PR #14267, opened 2018-06-11
[GLTFExporter] check for geometry.userData before trying to parse it

ngokevin:
Was erroring could not convert undefined to Object.


takahirox:
`BufferGeometry` has `userData` object as default now.

https://github.com/mrdoob/three.js/blob/dev/src/core/BufferGeometry.js#L41

How did you face that error?

takahirox:
I guess you're using an older Three.js revision and the latest `GLTFExporter`? Try to use `GLTFExporter` in the same revision.

I think we can close.

PR #10895, opened 2017-02-27
Enable keyboard in OrbitControls

tledoux:
In laptop and without a touchscreen, it's very hard to interact
with OrbitControls.

So I added the arrow keys (when pan is disabled) to allow for rotating
(if the Ctrl key is also pressed the rotation is boosted).

Moreover, the + and - keys of the numeric keypad to zoom.
For laptops without keypad, the A and Z keys works also.

PR #14273, opened 2018-06-11
Constraints settings for DragControl mouse move

szabcsee:
Added constraint function to dragControls THREE.DragControls.
Added constraint check to onDocumentMouseMove.
Default value is 'xyz'. If it's overriden, the missing axis will be constrained.

For example the following setting only allows dragging object on the Y axis:

dragControls = new THREE.DragControls( objects, camera, renderer.domElement );
dragControls.constrains('y');

### DIFF
--- a/examples/js/controls/DragControls.js
+++ b/examples/js/controls/DragControls.js
@@ -26,6 +26,32 @@ THREE.DragControls = function ( _objects, _camera, _domElement ) {
 
 	var scope = this;
 
+	// Drag constrains (eg. move along x-axis only, y only, x and y, or default xyz)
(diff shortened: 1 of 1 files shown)


Mugen87:
In general we try to keep the controls as universal as possible and avoid use-case-specific code. Also see: https://github.com/mrdoob/three.js/pull/10895#issuecomment-282790907

szabcsee:
I think this does not interfere with the generic behaviour just adds an extra feature. You are right that this is only use-case-specific feature but I think it can be needed in quite many cases.  Never mind I can set up a separate repository for it if someone needs it.

PR #11376, opened 2017-05-22
[WIP] VR + Post Processing

takahirox:
Let me share my WIP PR trying to enable VR + post-processing.
Feedback is welcome because I'm feeling like there'd be better designs.

Video: https://twitter.com/superhoge/status/866687749385891841

Policy:
- keep using `EffectComposer` + `Pass` approach
- reusable existing shader pass
- enable different post-processing for each left/right view

Basic idea:
1. Render normally with `WebVRCamera`
2. Extract the left/right half of the result of 1 to half size renderTarget. 
3. Apply pass to each half
4. Combine the left/right results

Disadvantage:
- Not good Performance. Rendering# increases in five. (two for 2., two for 3., one for 4.)

How to use:

    // Glitch example
    camera = new THREE.WebVRCamera( vrdisplay, renderer );
    composer = new THREE.EffectComposer( renderer );
    composer.setSize( window.innerWidth, window.innerHeight );
    composer.addPass( new THREE.RenderPass( scene, camera ) );
    composer.addPass( new THREE.ShaderVRPass( vrdisplay, new THREE.GlitchPass() ) );
    composer.passes[ composer.passes.length - 1 ].renderToScreen = true;

Change:
- Add `ShaderVRPass` which acts as the basic idea I wrote above.
- Update `WebGLState` properties instead of `WebGLRenderer` properties for rendering with `ArrayCamera`(see #11051)
- Add a way to prevent `vrDisplay.submitFrame()` invoking from `WebVRCamera` because I wanna invoke it after post-processing.
- Extract update logic (ex: glitch calculation) from `pass.render()` to `pass.update()` method because I wanna call `pass.render()` two times but run the update logic once in an animation frame.

PR #14274, opened 2018-06-11
Fix EffectComposer for vr mode

dmarcos:
With the new VR API (without VRControls / VREffect) the frame is rendered and submitted within the renderer. The current EffectComposer API is not suitable for VR use cases. This PR relies on the `onAfterRender` callback enabling postprocessing passes to access the frame before is submitted to the headset. I admit this is a bit of a contortion. I put myself at the disposal of the THREE gods to incorporate any API changes you think are appropriate.

### DIFF
--- a/examples/js/postprocessing/EffectComposer.js
+++ b/examples/js/postprocessing/EffectComposer.js
@@ -6,6 +6,8 @@ THREE.EffectComposer = function ( renderer, renderTarget ) {
 
 	this.renderer = renderer;
 
+	window.addEventListener( 'vrdisplaypresentchange' , this.resize.bind(this) );
+
 	if ( renderTarget === undefined ) {
 
 		var parameters = {
@@ -29,6 +31,7 @@ THREE.EffectComposer = function ( renderer, renderTarget ) {
 	this.readBuffer = this.renderTarget2;
 
 	this.passes = [];
+	this.maskActive = false;
 
 	// dependencies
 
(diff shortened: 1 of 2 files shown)


examples/js/postprocessing/EffectComposer.js
```diff
@@ -73,51 +106,54 @@ Object.assign( THREE.EffectComposer.prototype, {
...
+			// If VR mode is enabled and rendering the whole scene is required.
+			// The pass renders the scene and and postprocessing is resumed before
+			// submitting the frame to the headset by using the onAfterRender callback.
+			if ( this.renderer.vr.enabled && pass.scene ) {
 
-					var context = this.renderer.context;
+				currentOnAfterRender = pass.scene.onAfterRender;
 
-					context.stencilFunc( context.NOTEQUAL, 1, 0xffffffff );
+				pass.scene.onAfterRender = function () {
```
pailhead: maybe not create this function here?

dmarcos: yeah, good point. The function will be allocated once per frame. I'll factor it out.

pailhead: As is it looks like it would be instantiated `this.passes.length` times? I thought about it but it looked kinda complicated to carry the `i` over.

dmarcos:
@fernandojsg What do you think?

nbriz:
i would LOVE to see this update! there's another PR on the subject here: https://github.com/mrdoob/three.js/pull/11376

dmarcos:
@nbriz Thanks. I didn't see the other PR. This one makes use of the `onAfterRender ` callback. Either way works for me.

Mugen87:
Closing in favor of #15840.

PR #14288, opened 2018-06-14
CurvePath: Add isLineCurve3 check

### LINKED ISSUE #14285: isLineCurve3 must to be added
https://github.com/mrdoob/three.js/blob/8a9a106f79e13e7f5b0e7fa1a57ed4a31f8ba74b/src/extras/core/CurvePath.js#L173

I think it would be nice to add isLineCurve3 check in CurvePath::getPoints() function.

git-mormot:
Fixes #14285


src/extras/core/CurvePath.js
```diff
@@ -171,8 +171,9 @@ CurvePath.prototype = Object.assign( Object.create( Curve.prototype ), {
 			var curve = curves[ i ];
 			var resolution = ( curve && curve.isEllipseCurve ) ? divisions * 2
 				: ( curve && curve.isLineCurve ) ? 1
```
Mugen87: Why not?
```js
: ( curve && ( curve.isLineCurve || curve.isLineCurve3 ) ) ? 1
```

Mugen87 (changes requested):

src/extras/core/CurvePath.js
git-mormot: Oh its better. I will change it again.

Mugen87 (approved):

mrdoob:
Thanks!

PR #14297, opened 2018-06-15
Add new function Object3D.applyMatrices

mqp:
Since `Object3D.applyMatrix` both multiplies by the new matrix and decomposes into a new position, rotation, and scale, it's inefficient to use in the (common) situation where you want to apply multiple matrices in series to an object's transform. However, having application code manually call `obj.matrix.multiplyMatrices(mat, obj.matrix)` followed by a single `obj.matrix.decompose(obj.position, obj.rotation, obj.scale)` is verbose and error-prone -- application code could easily forget to decompose at the end, causing some surprising bug elsewhere.

I propose adding `applyMatrices`, which will be both efficient and easy to use in that case.

### DIFF
--- a/docs/api/core/Object3D.html
+++ b/docs/api/core/Object3D.html
@@ -200,6 +200,9 @@
 		<h3>[method:null applyMatrix]( [param:Matrix4 matrix] )</h3>
 		<p>Applies the matrix transform to the object and updates the object's position, rotation and scale.</p>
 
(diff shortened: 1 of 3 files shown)


mquander:
On reflection, maybe it's reasonable to expect a caller to multiply their matrices in-place into one transform matrix, and then apply that using `applyMatrix`. I think that's a better solution, so I'm closing this.

PR #14316, opened 2018-06-18
Fix example in SkinnedMesh documentation

### LINKED ISSUE #14314: Incorrect example in SkinnedMesh documentation
The example in the [SkinnedMesh documentation page](https://threejs.org/docs/#api/objects/SkinnedMesh) creates a `CylinderBufferGeometry`, and later tries to set skin indices and weights like this:
```javascript
geometry.skinIndices.push( new THREE.Vector4( skinIndex, skinIndex + 1, 0, 0 ) );
geometry.skinWeights.push( new THREE.Vector4( 1 - skinWeight, skinWeight, 0, 0 ) );
```
However, this code doesn't work since `BufferGeometry` doesn't have the `skinIndices` and `skinWeights` properties - those are only present in `Geometry`.

The issue could be fixed by either changing `CylinderBufferGeometry` to `CylinderGeometry` or setting the skin weights and indices as attributes in the `BufferGeometry`
(NOTE: The embedded example on the page correctly uses a `CylinderGeometry`, only the code provided below it is incorrect)

CakeWithSteak:
Fixes #14314

### DIFF
--- a/docs/api/objects/SkinnedMesh.html
+++ b/docs/api/objects/SkinnedMesh.html
@@ -38,7 +38,7 @@
 		<h2>Example</h2>
 
 		<code>
-		var geometry = new THREE.CylinderBufferGeometry( 5, 5, 5, 5, 15, 5, 30 );
+		var geometry = new THREE.CylinderGeometry( 5, 5, 5, 5, 15, 5, 30 );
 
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #14330, opened 2018-06-19
NodeMaterial: Update perturbNormal2Arb back/double-sided math.

donmccurdy:
Update to match changes from https://github.com/mrdoob/three.js/pull/13916, per https://github.com/mrdoob/three.js/pull/14237#issuecomment-397820642.

I'm not sure this fixes bumpmaps in NodeMaterial or how to reproduce that difference, but the DamagedHelmet example looks correct.

/cc @WestLangley

### DIFF
--- a/examples/js/nodes/NodeLib.js
+++ b/examples/js/nodes/NodeLib.js
@@ -136,19 +136,19 @@ THREE.NodeLib.add( new THREE.ConstNode( "vec3 LUMA vec3(0.2125, 0.7154, 0.0721)"
 THREE.NodeLib.add( new THREE.FunctionNode( [
 // Per-Pixel Tangent Space Normal Mapping
 // http://hacksoflife.blogspot.ch/2009/11/per-pixel-tangent-space-normal-mapping.html
-	"vec3 perturbNormal2Arb( vec3 eye_pos, vec3 surf_norm, vec3 map, vec2 mUv, vec2 scale ) {",
+	"vec3 perturbNormal2Arb( vec3 eye_pos, vec3 surf_norm, vec3 map, vec2 mUv, vec2 normalScale ) {",
 	"	vec3 q0 = dFdx( eye_pos );",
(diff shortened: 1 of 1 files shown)


examples/js/nodes/NodeLib.js
```diff
@@ -136,19 +136,19 @@ THREE.NodeLib.add( new THREE.ConstNode( "vec3 LUMA vec3(0.2125, 0.7154, 0.0721)"
 THREE.NodeLib.add( new THREE.FunctionNode( [
 // Per-Pixel Tangent Space Normal Mapping
 // http://hacksoflife.blogspot.ch/2009/11/per-pixel-tangent-space-normal-mapping.html
-	"vec3 perturbNormal2Arb( vec3 eye_pos, vec3 surf_norm, vec3 map, vec2 mUv, vec2 scale ) {",
+	"vec3 perturbNormal2Arb( vec3 eye_pos, vec3 surf_norm, vec3 map, vec2 mUv, vec2 normalScale ) {",
 	"	vec3 q0 = dFdx( eye_pos );",
 	"	vec3 q1 = dFdy( eye_pos );",
```
WestLangley: Add workaround for Adreno 3XX dFd*( vec3 ) bug?

sunag:
Currently `bumpMap` is converted to `normalMap` in `NodeMaterial` but in PR I will going to release I am following exactly the same as in the core.

Just remembering that an extra attention is needed to rename function inputs:
`vec2 scale` to `vec2 normalScale`

It is parsed in runtime and can be used with others `nodes` in `functionCallNode.inputs.scale = ...`.
For example:
https://github.com/mrdoob/three.js/blob/a41ca87ac0ecc651432cec819b49a9c3c4ddc8ff/examples/webgl_materials_nodes.html#L1108-L1118

sunag:
more directly fix `bumpMap` too since `bumpMap` uses perturbNormal2Arb

donmccurdy:
Looks like @sunag got it, thanks!

Issue #10690, opened 2017-01-31
*Material.lightMap should allow for HDR Encodings


bhouston:
##### Description of the problem 

*Material.lightMap is supposed to represent incoming light onto a material which then interacts with the surface to give a resulting intensity.  RIght now lightMap does not support encodings, such as RGBM*/RGBE/etc, thus it is limited to LDR, rather than HDR.  In reality, the incoming light is more accurately represented by HDR images.  To rectify this, we should allow for lightMap to have an encoding  (lightMapEncoding) similar to map (mapEncoding) and emissiveMap (emissiveMapEncoding).

##### Three.js version

- [x] Dev
- [x] r82

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

dherben:
I was surprised to find out that ThreeJS does not support HDR encodings in the lightmaps, seems to be one of the principal places where you would want this.

I'm considering picking this up myself, if someone could give me some information on what it would take to make it work.

tomdelerue:
Yes, something I've been looking forward to for some time. Maybe I can help?

Mugen87:
Should be fixed via #18052.

PR #14333, opened 2018-06-19
NodeMaterial - r7

sunag:
**NodeMaterial** r7 is the biggest review since the first release, has a better memory utilization and performance, `NodeMaterial` was simplified and all `nodes` revised and converted to `ES6 Modules`.

### Features

Now **NodeMaterial** use ES6 Modules

- new `MeshStandardNodeMaterial` with backward-compatible and `NodeMaterial` extended syntax
- support to `void-function` with `BypassNode`
- support to `struct` with `StructNode`
- support to `#define` syntax in `ConstNode`
- support to conditional using `CondNode`
- support varying using `VarNode` with `nodeVar.input = node`
- support to `TextureCubeNode`, `envMap` uses `PMREM` like `webgl_materials_standard.html`
- `copy` function per node and too `copy` read deserialization from `NodeMaterialLoader`
- new nodes: `RTTNode`, `CheckerNode`, `ColorSpaceNode`, `MaxMIPLevelNode`, `TextureCubeNode`, `TextureCubeUVNode`, `PropertyNode`, `ExpressionNode`, `StructNode`, `NodePostProcessing`

### Examples
- new examples: `mesh-standard`, `void-function`, `render-to-texture`, `temporal-blur`, `conditional`, `node-position`, `node-normal`
- update `varying`, `bump`, `all examples used normalmap`
- new example `motion-blur` in post-processing
- allow input param to show example, e.g: `webgl_materials_nodes.html?e=displace`

**NodeMaterial**
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html

**NodePostProcessing** instance of `NodePass`
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_postprocessing_nodes.html?e=motion-blur

**Performance**
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_performance_nodes.html

**All Examples**
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/index.html?q=nodes

### Changes
- All `NodeMaterial` code to build shader transferred to `NodeBuilder` as base to build code
- Use `ExpressionNode` instead of `FunctionNode` to inline expression
- Simplification `NodeMaterial`
- Simplification `NodeMaterialLoader`
- rename `BumpNode` to `BumpMapNode` ( use `bumpMapNode.toNormals` to convert to normals )
- rename node type `fv1` to `f`
- rename node type `iv1` to `i`
- rename all `.coord` properties to `.uv`
- rename all `.transform` to `.position` (  `.position` is `local position` = `new PositionNode()` )
- rename `GLNode` to `Node` ( node base )
- rename `nodePass.value` to `nodePass.input`
- removed libs, keyworks and dependencies function is added in yours respectives nodes.
- `material.normalScale` is deprecated use `material.normal.scale` using `NormalMapNode`
- new serialization struct
- update shader nodes from threejs core shaders
- new folders filter struct

### Fixes
- use build() to not need compile in .onBeforeCompile
- post-processing `mosaic` mask example
- `BlurNode` extra bright
- rough metals are too reflective ( https://github.com/mrdoob/three.js/issues/14238 ), fixed with the new `MaxMIPLevelNode`
- fix `NormalNode` `DoubleSided`
- perturbNormal2Arb back/double-sided math, because threejs core update in NodeMaterial ( https://github.com/mrdoob/three.js/pull/14330 )
- others minors fixes


sunag:
Usage: **MeshStandardNodeMaterial**
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html

```javascript
// use MeshStandardNodeMaterial or MeshStandardMaterial?
var useNodeMaterial = true;

var mtl = useNodeMaterial ? new THREE.MeshStandardNodeMaterial() : new THREE.MeshStandardMaterial();

// default syntax ( backward-compatible )

mtl.map = getTexture( "brick" );

mtl.normalMap = getTexture( "grassNormal" );
mtl.normalScale = new THREE.Vector2( .3, -.3 );

mtl.envMap = cubemap;

mtl.roughness = oldMaterial ? oldMaterial.roughness : .5;
mtl.metalness = oldMaterial ? oldMaterial.metalness : .5;

// extended syntax ( only for NodeMaterial materials )

if ( useNodeMaterial ) {

	// simple desaturated map using nodes

	mtl.map = new THREE.ColorAdjustmentNode( 
		new THREE.TextureNode( mtl.map ), 
		new THREE.FloatNode( 1 ), 
		THREE.ColorAdjustmentNode.SATURATION 
	);
	
}
```

sunag:
Now decoding function is per texture. 
https://github.com/sunag/three.js/blob/bc3e2068b2e743377fd5af9fd90f59506927dd0b/examples/js/nodes/inputs/TextureNode.js#L52

This can help with issues like these: https://github.com/mrdoob/three.js/issues/10690

@bhouston I have adapted the `ENVMAP_TYPE_CUBE_UV` to the node material system but it is still a slight difference in brightness that I have to check. 

https://github.com/sunag/three.js/blob/bc3e2068b2e743377fd5af9fd90f59506927dd0b/examples/js/nodes/utils/TextureCubeUVNode.js

sunag:
@martinRenou add `void-functions` example using `BypassNode` and update `varying` example with new feature `input` in `VarNode`. https://github.com/mrdoob/three.js/issues/7522#issuecomment-245903051

mrdoob:
Looking goood!

examples/js/nodes/NodeMaterialSource.js
```diff
@@ -0,0 +1,97 @@
...
+		for(var i = 0; i < scripts.length; i++) {
+			
+			loadScript( path + scripts[i] );
+			
+		}
+
+	}
+	
+	loadScripts( "js/nodes/", [
+		// NodeLibrary
```
mrdoob: I think we may want to turn all the nodes into ES6 Modules and let the browser resolve the dependencies for us 🙂

looeee: > I think we may want to turn all the nodes into ES6 Modules and let the browser resolve the dependencies for us 

Loading large numbers of modules in the browser will be slower than bundling

renderer-main-thread-time-breakdown


From this article: https://developers.google.com/web/fundamentals/primers/modules#performance

sunag: @mrdoob What do you think of including `build/three.nodes.js` as extesion?

mrdoob: > Loading large numbers of modules in the browser will be slower than bundling

I think it's too early to think about bundling. Besides, currently we're including scripts separately and adding dependency logic to the code.

I'm suggesting porting it to ES6 Modules so it'll be easier to integrate in src/ later on.

Anyway, this can be done in a different PR.

mrdoob:
> **This is in dev!**

Do you mean that it's work in progress? Or is the PR ready to be merged?

examples/js/nodes/NodeMaterialSource.js
mrdoob: > Yes! I had already converted to ES6 but the code turned out to be outdated. You have some example that I can base?

https://github.com/mrdoob/three.js/blob/dev/examples/webgl2_sandbox.html uses ES6 Modules in the example directly 😊

sunag: > https://github.com/mrdoob/three.js/blob/dev/examples/webgl2_sandbox.html uses ES6 Modules in the example directly blush

Niice!

sunag:
> Do you mean that it's work in progress?

Yes! I think it will take a few more days.

takahirox:
From https://github.com/mrdoob/three.js/issues/7522#issuecomment-398537534

> I see replacement as the ultimate goal, is better both for performance and maintenance. 

How can node based material system improve the performance? Because shader code can be simpler?

bhouston:
> How can node based material system improve the performance? Because shader code can be simpler?

I do not believe there are significant performance improvements to be expected if we choose the same options.

But the shader graph will make it easier to make procedural textures that run on the GPU rather than having to somehow create them via JavaScript and upload them.  Thus there is performance increases in this specific scenario.

Generally, I view the replacement as primarily giving us more flexibility, which is a major limiting issue with Three.JS right now with regards specifically to glTF conformance.

pailhead:
Is it possible to add a summary for a PR of this size? Without the documentation i would like to get up to speed and understand the proposed system, but i'm not sure where to start. I imagine something like:

```
chang foo
add bar
remove baz
```

examples/webgl_materials_nodes.html
```diff
@@ -467,6 +480,82 @@
...
+					
+					var sataturation = new THREE.FloatNode( 1 );
+					
+					function updateMaterial( useNodeMaterial ) {
+					
+						var oldMaterial = mtl;
+					
+						if (oldMaterial) oldMaterial.dispose();
+					
+						mtl = useNodeMaterial ? new THREE.MeshStandardNodeMaterial() : new THREE.MeshStandardMaterial();
```
pailhead: I think this causes a stutter on my old cpu and 1080GTX.

pailhead:
@takahirox 
>How can node based material system improve the performance? Because shader code can be simpler?

Would it make sense to make a comparison between some properties of these different systems? Hand written GLSL vs generated, parse/build system etc.?

sunag:
> How can node based material system improve the performance? Because shader code can be simpler?

Maybe I mispronounced. This is more faster and easy maintenance instead of two material system in the core system.

If compared to the current material system `NodeMaterial` 2x at 4x more slower to initialize the materia(this is something to improve) in visual language parse and code generation.  The GPU question follows as @bhouston said.

WestLangley:
>Without the documentation i would like to get up to speed and understand the proposed system, but i'm not sure where to start

If you do not have the skills to do that by studying `webgl_materials_nodes.html` and reading the source, you should not be a developer on this project.

Please wait until the feature is added and the documentation is complete. You will then be in a position to provide feedback.

sunag:
> @mrdoob Would it be OK to merge this now?

Give me another week and I think it is all done.

pailhead:
This has my approval in perpetuity if it's needed :)

sunag:
I am revising all and converting to ES6. `NodeMaterial` is more clean and `NodeBuilder` is the base to build shader. Keyworks and includes is added in yours respectives nodes.

```javascript
import './js/nodes/THREE.Nodes.js';
```

https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html

pailhead:
This is so sweet, think you for the good work!

sunag:
@pailhead Thanks!

mrdoob:
Looking good indeed!
Let me know when you want me to merge 😊

sunag:
@mrdoob I think it is good for now. I update the description.  :+1:

mrdoob:
Awesome!

mrdoob:
Thanks!

bhouston:
Very impressive system @sunag

sunag:
@mrdoob @bhouston Thank you! I see that I am on the right track.
I still have a few things in mind for this release. I will adding in the next PRs.

PR #14351, opened 2018-06-22
ObjectLoader: Added support for EdgesGeometry and WireframeGeometry.

### LINKED ISSUE #14302: EdgesGeometry: Add serialization/ deserialization
All my scene objects is created with a **wireframe**. When I export objects from the scene it works ok with the code bellow:

``` javascript
var objList = $this.getSceneObjects()
for (var i in objList) {
  $this.savefile.scene.objects.push(objList[i].toJSON())
}
```

Then I put all objects together, store some more needed information and save in a whatever file. So my program will **open** this file, it's not an importation at all, it's an opening to keep editing objects... But when I try to do it, I get a lot of errors. That's my opening code:

``` javascript
var fileJson = JSON.parse(fs.readFileSync(file))
var document = new App.Document(file)
        
// Load Scene saved objects
var loader = new THREE.ObjectLoader();
for (var obj in fileJson.scene.objects) {
  var item = fileJson.scene.objects[obj]
  document.getScene().add(loader.parse(item))
}
```

Note in the image attached that none of the objects could be rendered with that **wireframe** I said before... =/

##### Three.js version

- [ ] Dev
- [x] r93
- [ ] ...

##### Browser

- [ ] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer
- [x] Chromium

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

captura de tela de 2018-06-14 20-54-25

Mugen87:
Fixes #14302

The PR assumes that `EdgesGeometry` and `WireframeGeometry` are considered as helper geometries (like mentioned in the docs). So their base geometry has to be part of the scene graph.

### DIFF
--- a/src/geometries/WireframeGeometry.js
+++ b/src/geometries/WireframeGeometry.js
@@ -186,7 +186,7 @@ WireframeGeometry.prototype.toJSON = function () {
 
 	if ( geometry.isGeometry ) {
 
-		geometry = geometry.toBufferGeometry();
+		geometry = new BufferGeometry().fromGeometry( geometry );
 
 	} else {
 
--- a/src/geometries/WireframeGeometry.js
+++ b/src/geometries/WireframeGeometry.js
@@ -13,6 +13,10 @@ function WireframeGeometry( geometry ) {
 
 	this.type = 'WireframeGeometry';
 
+	this.parameters = {
+		geometry: geometry
+	};
+
 	// buffer
 
 	var vertices = [];
@@ -174,6 +178,16 @@ function WireframeGeometry( geometry ) {
 
 WireframeGeometry.prototype = Object.create( BufferGeometry.prototype );
 WireframeGeometry.prototype.constructor = WireframeGeometry;
+
+WireframeGeometry.prototype.toJSON = function () {
+
+	var data = BufferGeometry.prototype.toJSON.call( this );
+
+	data.geometry = data.geometry.uuid;
+
+	return data;
+
+};
 
 
 export { WireframeGeometry };
--- a/src/geometries/WireframeGeometry.js
+++ b/src/geometries/WireframeGeometry.js
@@ -182,8 +182,19 @@ WireframeGeometry.prototype.constructor = WireframeGeometry;
 WireframeGeometry.prototype.toJSON = function () {
 
 	var data = BufferGeometry.prototype.toJSON.call( this );
+	var geometry = data.geometry;
 
-	data.geometry = data.geometry.uuid;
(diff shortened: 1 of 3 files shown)


WestLangley:
>So their base geometry has to be part of the scene graph.

I have not experienced that limitation. Can you explain where the limitation comes from?

Mugen87:
The statement was related to the serialization/deserialization process. I've meant the following: Let's say you have a basic scene with a `BoxGeometry`. Now you want to use `EdgesGeometry` in order to visualize the edges of the box. If you now call `scene.toJSON()`, the original mesh must be part of the scene graph otherwise `ObjectLoader` can't restore `EdgesGeometry` (because `EdgesGeometry` refers with the `uuid` to its base geometry when it is transferred to JSON).

WestLangley:
> the original mesh must be part of the scene graph 

`EdgesGeometry` takes a geometry as an argument. There is no need for an "original mesh" at all, right?

Mugen87:
`EdgesGeometry` and `WireframeGeometry` are considered as helper geometries. How would you add one of these entities in the editor without a base geometry?

WestLangley:
>EdgesGeometry and WireframeGeometry are considered as helper geometries.

I guess I would have to say I don't agree with that statement.
```javascript
var geo = new THREE.IcosahedronBufferGeometry( 5, 0 );
var geometry = new THREE.WireframeGeometry( geo ); // or EdgesGeometry
var material = new THREE.LineBasicMaterial();
var wireframe = new THREE.LineSegments( geometry, material );
```

Mugen87:
Okay I understand your point. Let's see if we can implement serialization/deserialization differently.

Mugen87:
Okay, the new approach should regard @WestLangley's feedback.

src/geometries/WireframeGeometry.js
```diff
@@ -175,5 +179,26 @@ function WireframeGeometry( geometry ) {
...
+	var data = BufferGeometry.prototype.toJSON.call( this );
+	var geometry = data.geometry;
+
+	if ( geometry.isGeometry ) {
+
+		geometry = new BufferGeometry().fromGeometry( geometry );
+
+	} else {
+
+		geometry = geometry.clone();
```
Mugen87: This ensures the serialization/deserialization of both geometries is self-contained.

Mugen87:
I'd like to start with a fresh PR :innocent:

PR #8493, opened 2016-03-31
Triangle: new method .closestPointToPoint

Mugen87:
This PR adds the method `Triangle .closestPointToPoint`. The method calculates the closest point on the triangle for a given point, which is a useful feature for computational geometry (e.g. users can now easily calculate the distance from a point to a triangle).

PR #14377, opened 2018-06-27
Make Triangle.closestPointToPoint much faster

mqp:
This is an interesting function to improve because it's a small [hotspot](https://github.com/donmccurdy/three-pathfinding/blob/fc54f75b27b104fb6d60e39aabc07045a2fe408f/src/index.js#L223) for the [three-pathfinding](https://github.com/donmccurdy/three-pathfinding) library's functionality for clamping positions onto a navigation mesh.

The algorithm I implemented is very well-explained in [Real-Time Collision Detection](http://realtimecollisiondetection.net/) by Christer Ericson. You may be able to preview the relevant part of the book on [Google Books, starting on page 136](https://books.google.com/books?id=WGpL6Sk9qNAC&lpg=PA155&pg=PA136) (I don't know exactly how Google Books decides who can see what parts of what books.)

How much faster this is depends on the inputs. In the benchmark I wrote, I tried to make a representative sample of points and triangles and observed approximately an 8x speedup on both desktop Chrome and Firefox, which seems plausible. If you really want, I can produce more performance numbers, but it should clearly be much faster in all situations.

For correctness testing, I tested it alongside the old version on all of the point/triangle pairs in the benchmark and verified that the two produced the same output modulo floating-point error.

### DIFF
--- a/src/math/Triangle.js
+++ b/src/math/Triangle.js
@@ -1,6 +1,4 @@
 import { Vector3 } from './Vector3.js';
-import { Line3 } from './Line3.js';
-import { Plane } from './Plane.js';
 
 /**
  * @author bhouston / http://clara.io
@@ -218,12 +216,14 @@ Object.assign( Triangle.prototype, {
 
 	closestPointToPoint: function () {
 
-		var plane = new Plane();
-		var edgeList = [ new Line3(), new Line3(), new Line3() ];
-		var projectedPoint = new Vector3();
-		var closestPoint = new Vector3();
+		var vab = new Vector3();
+		var vac = new Vector3();
+		var vbc = new Vector3();
+		var vap = new Vector3();
+		var vbp = new Vector3();
+		var vcp = new Vector3();
 
-		return function closestPointToPoint( point, target ) {
+		return function closestPointToPoint( p, target ) {
 
 			if ( target === undefined ) {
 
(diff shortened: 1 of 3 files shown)


mrdoob:
@WestLangley looks good?

Mugen87:
You might want to consider https://github.com/mrdoob/three.js/pull/8493#issuecomment-204545731 as well.

WestLangley:
>verified that the two produced the same output modulo floating-point error.

Thanks @mquander.

Let's /ping @bhouston for 2nd opinion.

bhouston:
Amazing improvement!

I am pretty sure @sxenos's code that @Mugen87 linked to even more efficient but probably no more than 2x faster that this optimization, so lets at least merge this and go with it but remember @sxenos' code if you need to get even faster.

WestLangley:
@mrdoob It's a go!

mrdoob:
Hmm... @mquander any chance you could do some benchmark comparison against @sxenos' code?

bhouston:
I think I was wrong, I believe this algorithm is actually the same as @sxenos's.

mquander:
It's clearly closely related. I personally strongly prefer using an algorithm where I can read something that tells me how it was derived, so I prefer this implementation (unless there is an explanation of sxenos's hanging around somewhere.) But if you really feel strongly, I'll benchmark the other one sometime soon.

WestLangley:
Agreed. I prefer an algorithm with a published reference. Edge cases and numerical stability can be tricky.

PR #14402, opened 2018-06-30
Set Texture.image immediately in TextureLoader

gkjohnson:
This PR changes `TextureLoader` to set `Texture.image` immediately instead of waiting for the `ImageLoader` callback to fire. This lets you register for the image's `load` event and rerender the scene once it loads.

The `onUpdate` callback for `TextureLoader.load` isn't sufficient because you don't always have access to that function, such as when loading a model with the `ColladaLoader`.

### DIFF
--- a/src/loaders/TextureLoader.js
+++ b/src/loaders/TextureLoader.js
@@ -26,23 +26,22 @@ Object.assign( TextureLoader.prototype, {
 		loader.setCrossOrigin( this.crossOrigin );
 		loader.setPath( this.path );
 
-		loader.load( url, function ( image ) {
+		texture.image =
+			loader.load( url, function () {
 
-			texture.image = image;
+				// JPEGs can't have an alpha channel, so memory can be saved by storing them as RGB.
+				var isJPEG = url.search( /\.(jpg|jpeg)$/ ) > 0 || url.search( /^data\:image\/jpeg/ ) === 0;
 
-			// JPEGs can't have an alpha channel, so memory can be saved by storing them as RGB.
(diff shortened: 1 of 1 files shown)


takahirox:
Seems like the current code would be @mrdoob 's intentional. Perhaps, he thinks to let `TextureLoader` be friendly to `ImageBitmapLoader`? (We add an option to switch `ImageLoader` with `ImageBitmapLoader` someday soon?) https://github.com/mrdoob/three.js/commit/c8d42cd17d238e4577c0327bcabaae742d6f0c49

gkjohnson:
I thought that might be the case. Maybe there's another way to address my use case then? Unfortunately the `Texture.onUpdate` function only gets called on next render but the goal is to only rerender once the texture has updated.

Mugen87:
>  This lets you register for the image's load event and rerender the scene once it loads.

Maybe I miss something but why don't you trigger the rerender within the `onLoad()` callback of `TextureLoader.load()`?

gkjohnson:
@Mugen87 

When using the ColladaLoader you aren't able to set that callback. It looks like the GLTFLoader avoids the problem by firing its on load callback only when all the textures have loaded, as well. I don't think we can do the same without changing the ColladaLoader API.

This isn't an urgent issue for me -- just something I noticed when doing some testing.

Mugen87:
> When using the ColladaLoader you aren't able to set that callback.

Yes, that's right. Many loaders have internal loader instances in order to load specific resources. There is still the possibility to use the `onLoad()` callback of `LoadingManager.load()`. Could this be helpful to you?

gkjohnson:
That works great! Thanks -- somehow I missed that that was an option. I'll close the PR in that case.

PR #14422, opened 2018-07-06
Quaternion: Added .angleTo()

Mugen87:
I need this method for a project and I thought it might be a nice addition to `Quaternion`. Counterpart to Unity's [Quaternion.Angle](https://docs.unity3d.com/ScriptReference/Quaternion.Angle.html).

Reference: https://math.stackexchange.com/a/167828

### DIFF
--- a/src/math/Quaternion.js
+++ b/src/math/Quaternion.js
@@ -393,21 +393,11 @@ Object.assign( Quaternion.prototype, {
 
 	}(),
 
-	angleTo: function () {
+	angleTo: function ( q ) {
 
-		var p = new Quaternion();
+		return 2 * Math.acos( Math.abs( this.dot( q ) ) );
 
-		return function angleTo( q ) {
-
-			p.copy( this ).inverse();
-
-			p.premultiply( q );
-
-			return 2 * Math.acos( p.w );
-
-		};
-
-	}(),
+	},
 
 	inverse: function () {
 
--- a/src/math/Quaternion.js
+++ b/src/math/Quaternion.js
@@ -399,9 +399,9 @@ Object.assign( Quaternion.prototype, {
 
 		return function angleTo( q ) {
 
-			p.copy( q ).inverse();
+			p.copy( this ).inverse();
 
-			p.premultiply( this );
+			p.premultiply( q );
 
 			return 2 * Math.acos( p.w );
 
--- a/src/math/Quaternion.js
+++ b/src/math/Quaternion.js
@@ -393,6 +393,22 @@ Object.assign( Quaternion.prototype, {
 
 	}(),
 
+	angleTo: function () {
+
+		var p = new Quaternion();
+
+		return function angleTo( q ) {
+
+			p.copy( q ).inverse();
+
+			p.premultiply( this );
+
+			return 2 * Math.acos( p.w );
+
+		};
+
+	}(),
+
 	inverse: function () {
 
 		// quaternion is assumed to have unit length
--- a/test/unit/src/math/Quaternion.tests.js
+++ b/test/unit/src/math/Quaternion.tests.js
@@ -391,6 +391,18 @@ export default QUnit.module( 'Maths', () => {
 
 		} );
 
+		QUnit.test( "angleTo", ( assert ) => {
+
+			var a = new Quaternion();
+			var b = new Quaternion().setFromEuler( new Euler( 0, Math.PI, 0 ) );
+			var c = new Quaternion().setFromEuler( new Euler( 0, Math.PI * 2, 0 ) );
+
+			assert.ok( a.angleTo( a ) <= eps, "Passed!" );
+			assert.ok( a.angleTo( b ) - Math.PI <= eps, "Passed!" );
+			assert.ok( a.angleTo( c ) - ( Math.PI * 2 ) <= eps, "Passed!" );
+
+		} );
+
 		QUnit.test( "inverse/conjugate", ( assert ) => {
 
 			var a = new Quaternion( x, y, z, w );
--- a/test/unit/src/math/Quaternion.tests.js
+++ b/test/unit/src/math/Quaternion.tests.js
@@ -396,10 +396,12 @@ export default QUnit.module( 'Maths', () => {
 			var a = new Quaternion();
 			var b = new Quaternion().setFromEuler( new Euler( 0, Math.PI, 0 ) );
 			var c = new Quaternion().setFromEuler( new Euler( 0, Math.PI * 2, 0 ) );
+			var d = new Quaternion().setFromEuler( new Euler( 1, 1, 1 ) );
 
 			assert.ok( a.angleTo( a ) <= eps, "Passed!" );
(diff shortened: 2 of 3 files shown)


test/unit/src/math/Quaternion.tests.js
```diff
@@ -391,6 +391,18 @@ export default QUnit.module( 'Maths', () => {
...
 
+		QUnit.test( "angleTo", ( assert ) => {
+
+			var a = new Quaternion();
+			var b = new Quaternion().setFromEuler( new Euler( 0, Math.PI, 0 ) );
+			var c = new Quaternion().setFromEuler( new Euler( 0, Math.PI * 2, 0 ) );
+
+			assert.ok( a.angleTo( a ) <= eps, "Passed!" );
+			assert.ok( ( a.angleTo( b ) - Math.PI ) <= eps, "Passed!" );
+			assert.ok( ( a.angleTo( c ) - ( Math.PI * 2 ) ) <= eps, "Passed!" );
```
WestLangley: Sorry,  but this test is not adequate.  You are only testing the zero quaternion. And your angles are multiples of PI.

Also, `a` and `c` represent the same orientation, but `a.angleTo( c )` is not zero.

src/math/Quaternion.js
```diff
@@ -393,6 +393,22 @@ Object.assign( Quaternion.prototype, {
...
 
+	angleTo: function () {
+
+		var p = new Quaternion();
+
+		return function angleTo( q ) {
+
+			p.copy( q ).inverse();
+
+			p.premultiply( this );
```
WestLangley:
```javascript
p.copy( this ).inverse();
p.premultiply( q );
```

WestLangley:
I made some suggestions, but TBH, I am skeptical of this method and the implementation.

What is your interpretation of the returned angle? For example, what if `a` represents a quarter-turn around the x-axis and `b` represents a quarter-turn around the y-axis. What does `a.angleTo( b )` represent?

Mugen87:
> What does a.angleTo( b ) represent?

I think the Unity doc is precise in this point:

 _Returns the angle in degrees between two rotations a and b. Example: Think of two GameObjects (A and B) moving around a third GameObject (C). Lines from C to A and C to B create a triangle which can change over time. The angle between CA and CB is the value Quaternion.Angle provides._

test/unit/src/math/Quaternion.tests.js
Mugen87: > Also, a and c represent the same orientation, but a.angleTo( c ) is not zero.

The value is 360 Degrees. I think there is no "correct" results since it's open to interpretation.

test/unit/src/math/Quaternion.tests.js
```diff
@@ -391,6 +391,20 @@ export default QUnit.module( 'Maths', () => {
...
+
+			var a = new Quaternion();
+			var b = new Quaternion().setFromEuler( new Euler( 0, Math.PI, 0 ) );
+			var c = new Quaternion().setFromEuler( new Euler( 0, Math.PI * 2, 0 ) );
+			var d = new Quaternion().setFromEuler( new Euler( 1, 1, 1 ) );
+
+			assert.ok( a.angleTo( a ) <= eps, "Passed!" );
+			assert.ok( ( a.angleTo( b ) - Math.PI ) <= eps, "Passed!" );
+			assert.ok( ( a.angleTo( c ) - ( Math.PI * 2 ) ) <= eps, "Passed!" );
+			assert.ok( ( a.angleTo( d ) - ( 1.939087528822506 ) ) <= eps, "Passed!" );
```
WestLangley: Where did `1.939087528822506` come from?

test/unit/src/math/Quaternion.tests.js
WestLangley: I would say that is a problem, then, because it means output of this method is not particularly useful as a metric.

test/unit/src/math/Quaternion.tests.js
Mugen87: Calculator 😊

WestLangley:
>I think the Unity doc is precise in this point:
>>Lines from C to A and C to B create a triangle which can change over time. The angle between CA and CB is the value Quaternion.Angle provides.

That seems to be describing the angle between two vectors, which would be based on the object's relative positions. I am not sure what that has to do with the object's quaternions.

Mugen87:
Okay, I'll do some comparisons with Unity and see what happens. Closing the PR for now.

WestLangley:
I expect the following is what you are trying to do. This code is not optimized.
```javascript
p.copy( this ).inverse();
p.premultiply( q );
var v = new THREE.Vector4().setAxisAngleFromQuaternion( p );
return v.w;
```
And I expect that `v.w` is what you mean by your definition

    Returns the angle between this quaternion and quaternion [page:Quaternion q] in radians.

I am not sure the "angle between two quaternions" is a commonly-used term. But if it is, I think you would have to return the _shortest angle_ in this case -- that is, for the metric to be useful.

How are you using this returned value?

WestLangley:
Well, I am not in favor of this method, TBH, but an improvement would be to return
```javascript
2 * Math.acos( Math.abs( p.w ) ); // added abs()
```
But the entire method can be computed more easily with a single line:
```javascript
2 * Math.acos( Math.abs( this.dot( q ) ) );
```
But even so, if the output is being used as a metric, I would avoid `Math.acos()` alltogether. You could return:
```javascript
1 - Math.abs( this.dot( q ) );
```
Or a quantity equal to ( 1 − cos( angle ) ) / 2:
```javascript
1 - Math.pow( this.dot( q ), 2 );
```

Mugen87:
I have one uncertainty: Let's say we calculate the difference between two quaternions with the following formula:

d = ba<sup>-1</sup>

`d` represents now the angular displacement from `a` to `b`. The dot product of `a` and `b` is equal to the `w` component of `d`, right?

WestLangley:
What may be useful is to provide a method that computes the delta quaternion and return that. Forget about the metric. That is use-case specific. The user is then free to compute any metric he wants with the returned quaternion.

```javascript
THREE.Quaternion.prototype.deltaQuaternion = function ( qa, qb ) {

	this.copy( qa ).inverse();

	return this.premultiply( qb );

};
```

Thing is, I'm not sure I'm liking `deltaQuaternion` as the method name...

The above method is computing the quaternion you would have to apply to your object to rotate it from the orientation given by `qa` to the orientation given by `qb`.

WestLangley:
No user, as far as I know, has asked for "the angle between quaternions". It is not even a common term.

Users have asked how to calculate the delta quaternion.

Users who what a metric to measure the deviation of one quaternion from the other are going to have to know some math. And the appropriate metric to use is going to be application-specific.

This is why I suggested, instead, a method that computes the delta quaternion itself and returns it.

Quite frankly, I do not think that method really needs to be added to the library either. There is not a common name for it.

So, I would not add `Quaternion.angleTo( quaternion )`, and I would probably not add the "delta" method, either, but I would not oppose adding it.

Mugen87:
Reopening the PR with a simplified version of the code. I think it can't be wrong to add similar methods like in widely used game engines like Unity.

PR #14426, opened 2018-07-07
better synchronized scrolling

greggman:
moving the canvas each frame inside rAF means the browser will scroll the canvas with
the rest of the page even if we don't get a chance to update it. This removes the bounce that was there before where our canvas update was behind the page update.

you can compare the difference below. Both of these examples are set to render at only 15fps to exaggerate the issue (though the issue comes up naturally often while three.js pages are initializing their assets etc....)

scroll the results up and down

old way:  http://jsfiddle.net/greggman/w8dv3r1h/

new way: http://jsfiddle.net/greggman/rtps2ygz/

### DIFF
--- a/examples/webgl_multiple_elements.html
+++ b/examples/webgl_multiple_elements.html
@@ -16,19 +16,19 @@
 				font-size:13px;
 
 				background-color: #fff;
-				margin: 0px;
+				margin: 0;
 			}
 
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #14427, opened 2018-07-08
NodeMaterial - r7 (2)

sunag:
Full description here ( https://github.com/mrdoob/three.js/pull/14333#issue-195983205 )


bhouston:
@sunag, maybe profile the code and figure out what is the slow parts and share it?  Or better yet, create an example scene that is all about material creation using the current and the new system so others can easily profile it.  Maybe webgl_performance_materials.html or something like that.

You mentioned it was 2x to 4x slower than the current material system earlier.  I think that people will be reluctant to adopt the new material system if it is really that much slower.  Usually when something is slow, it is one little part of it that isn't that hard to address.

sunag:
I can do some example soon... there were other issues as memory too fixed in this release. This slower is only about `CPU` is performace in material initialization but I did some more tests and apparently this is something to be fixed inside threejs core and not in NodeMaterial.

The slowness is specifically here:
https://github.com/mrdoob/three.js/blob/23d0e6c71c838e43169ecd2ac1c1956cdcd55ca9/src/renderers/WebGLRenderer.js#L1474

Certainly to confirm the origin of the slowness I will need more tests. Still my tests were only with `MeshStandardMaterial` vs `MeshStandardNodeMaterial`.

sunag:
I found the reason. I add `encodings_pars_fragment` in vertex shader causing an increased `gl.getProgramInfoLog( program )` 2x or more in processing with node example related, I think the reason for everything is the extra code added without use. I did not imagine that not used functions in shader code was influencing so much. @bhouston But it is very good for `NodeMaterial` once ready no doubt it initilize the material more faster than the current material system on since it will only include the code necessary.

sunag:
@bhouston Now I am getting 1/1 with NodeMaterial vs current material system in performance
I will commit this example later

yasmim

bhouston:
I think you've addressed the only serious objection that others to have to this code - it is now as fast as the old material system, but infinitely more flexible.  I love those types of changes. :)

sunag:
More examples:

See this in special (RTTNode) bump map in procedural textures
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html?e=rtt

Conditional
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html?e=conditional

Performance
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_performance_nodes.html

sunag:
`RTTNode` can be used in post-processing, motion blur for example, cache of texture converting rgb to sampler2D like `render-to-texture` example, gpgpu simulations.,. I create a temporal example using ping-pong technique.

**Temporal Blur** - need click in reset because texture load after example
https://rawgit.com/sunag/three.js/dev-nodematerial-r7/examples/webgl_materials_nodes.html?e=temporal-blur

donmccurdy:
@sunag Nice work here! Where would be the best place to put general feedback? New issues, or in your PRs? A couple things I've run into testing this (well, actually testing the checked-in code but I think these issues remain) —

- [ ] How should alphaTest be specified? Using `material.alphaTest = 0.25` with one of the NodeFooMaterial classes will result in a duplicate definition of the `ALPHATEST` constant. Would the developer write `material.define('ALPHATEST', '0.25')` instead, or is that bug?
- [ ] Consider removing optional parameters from MathXNodes — if `method` is omitted, it's probably a typo and shouldn't default to something like `sin()`.

sunag:
I prefer that feedback like these stay inside the PR. I will considerer these changes. Thanks!

WestLangley:
@sunag Consider using `textures/decal/decal-normal.jpg` as the normal map instead, so it is easier to see if the normal maps and normal scale are working properly.

sunag:
Ok, @mrdoob for me already done to merge.
`webgl_sprites_nodes.html` only work fixing `onBeforeCompile` share code ( https://github.com/mrdoob/three.js/pull/14442 )

WestLangley:
Thanks, @sunag.

In `mesh-standard`, decal normals look good. It looks cool setting normal scale to ( 1, - 1 ).   :-)

In `mesh-standard`, if I toggle colors off, and then on again, saturation no longer works.

Teapot should be double-sided because of spout.

sunag:
@WestLangley Please, try again now.

WestLangley:
Same problem. In `mesh-standard`, if I toggle colors off, and then on again, the saturation controls no longer has any effect.

sunag:
I think that now work. I noticed a diference in `DoubleSided` using `normalMap` I have to check.

mrdoob:
Thanks!

PR #14429, opened 2018-07-08
WebGLRenderer: Fix mipmap generation in .copyTextureToTexture()

### LINKED ISSUE #14423: copyTextureToTexture(): Wrong processing of mipmaps

There is something seriously wrong with the dataTexture and copyTextureToTexture functions. 
I was struggling for many days with this issue until I was convinced that this is a three.js bug.

When I upload an image to texture1 and then copy that texture to a datatexture texture2, I cannot use the minFilter and magFilters for mipmaps at all (LinearMipmapLinearFilter and LinearFilter correspondingly) for the destination datatexture, as the texture will not be rendered, so the only option is to enjoy wild aliasing artifacts when applied to a moving plane or other object.

In order to make sure that this is a three.js bug, I added the WEBGL equivalent functions on an example WEBGL code and I was able to use mipmaps and filtering on the destination data texture with no problems at all.

>You can see the unfiltered datatexture artifacts on stackoverflow:
https://stackoverflow.com/questions/51071390/copytexturetotexture-results-in-horrible-aliasing-artifacts-in-three-js

>Or better try to mod this jsfiddle. You could try to uncomment and include the lines 73-76 that enable mipmaps and filtering.
https://jsfiddle.net/Debanker/4gL0eosh/27/

>You can also see the properly working WEBGL version that I mentioned, where the same image is applied on a cube (instead of a plane), but the two required helper scripts (on the top of HTML) are blocked (otherwise it works offline -unless you can host those files elsewhere and/or you can make it work -I spent half a day to find a proper free host to no avail). You can change line 326 from rendering texture2(copied) to texture1(original).
https://jsfiddle.net/Debanker/sybtL7ed/24/

##### Three.js version

- [ ] Dev
- [x] r93
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
Fixes #14423

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -7,7 +7,6 @@ import {
 	TriangleFanDrawMode,
 	TriangleStripDrawMode,
 	TrianglesDrawMode,
-	NoColors,
 	LinearToneMapping,
 	BackSide
 } from '../constants.js';
@@ -576,7 +575,7 @@ function WebGLRenderer( parameters ) {
 
 	}
 
-	this.renderBufferImmediate = function ( object, program, material ) {
+	this.renderBufferImmediate = function ( object, program ) {
 
 		state.initAttributes();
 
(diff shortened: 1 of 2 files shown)


Mugen87:
Closing. I've realized a side effect of this change since the `level` parameter of `.copyTextureToTexture()` can be used anymore.

PR #14430, opened 2018-07-08
WebGLRenderer: Avoid unnecessary texture uploads

Mugen87:
This is an optimization since `copyFramebufferToTexture()` and `copyTextureToTexture()` should only bind a texture but never trigger an upload with `.setTexture2D()`.


Mugen87:
`textureProperties.__webglTexture` can be undefined if the user passes in a new `Texture` object. Damn. I guess `WebGLTextures` needs some care...

PR #14435, opened 2018-07-08
Quaternion: Improve result of angleTo()

Mugen87:
This PR ensures that `Quaternion.angleTo()` calculates a value not greater than `Math.PI`.

### DIFF
--- a/src/math/Quaternion.js
+++ b/src/math/Quaternion.js
@@ -1,3 +1,4 @@
+import { _Math } from './Math.js';
 import { Vector3 } from './Vector3.js';
 
 /**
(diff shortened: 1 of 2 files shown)


WestLangley (approved):

WestLangley:
This method is useful in this context:
```javascript
if ( mesh.quaternion.angleTo( camera.quaternion ) > Math.PI / 8  ) mesh.quaternion.copy( camera.quaternion );
```

That is, the returned value from this method is useful as a distance metric. You will see the mesh 'hop' as the camera is rotated.

For the record, however, this method does _not_ compute the angle between quaternions as the documentation claims -- and as the method name implies.

Mugen87:
> For the record, however, this method does not compute the angle between quaternions as the documentation claims -- and as the method name implies.

But this is also true for what Unity writes, right?

Mugen87:
Since I need this change in `dev`, I'll merge this for now.

Issue #13288, opened 2018-02-09
SkinnedMesh bug on mobile


OrtOfOn:
##### Description of the problem

I think I found a bug about your SkinnedMesh **on mobile only** (tested on iPhone 6).

If this is already reported, I'm sorry, I didn't find it in the issue list :(

It seems the gpu skinning is not working correctly on mobile.
When a SkinnedMesh is moving or moved at high value positions (ex: x:1000, y:0, z:0), the skinning is not accurate anymore and starts spider dance.

The scale of the mesh is affecting the bug. Bigger the scale is, lesser the bug.
It seems the skeleton bones values are not calculated correctly at each frame and the bonesTexture on the skinning shader is pushing vertices at wrong place.
This is just my feeling of course.

I ran many tests before posting this... looking for a clue in my animated exports but I found the bug is happening with any kind of formats (GLTF, FBX, Collada, JSON, ...)

I made a clear example with clean source to expose the problem. It's quite easy to verify it on a smart phone:

http://cornflex.org/webgl/skinning-bug.html  (view source)

Again, all is fine on desktop.

That's very annoying because that means we are unable to develop a simple runner game with an avatar running (avatar.position.z increasing then) without having this issue :((
I still don't know how I'll manage it as morphTargets is not an option :(

Hope you guys can help here.

Thank you very much,

Q.

##### Three.js version

- [x] Dev
- [x] r89
- [x] ...

##### Browser

- [ ] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer
- [ ] Safari

##### OS

- [ ] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ x ] iOS

##### Hardware Requirements (graphics card, VR Device, ...)

iPhone 6

Mugen87:
Um, your demo looks good on my Pixel.

donmccurdy:
Confirmed issues on my iPhone SE —

![img_6324](https://user-images.githubusercontent.com/1848368/36045790-ac5abd72-0d8b-11e8-9a2f-39e67688b3d4.PNG)

WebGL precision issues in iOS perhaps

Mugen87:
But the iOS devices actually use the float texture based code path of the skinning implementation, right? Can you verify this with the following conformance test?

https://www.khronos.org/registry/webgl/conformance-suites/1.0.3/conformance/extensions/oes-texture-float.html

donmccurdy:
[...]

PR #14443, opened 2018-07-10
(Issue #13288) SkinnedMesh and ColladaLoader to support local skinning

denbo-ft:
Fixes distortion bug on ios and other platforms.

snowtrooper dae/skinning example updated to reflect local skinning changes


Mugen87:
Thanks for this but I think a solution for #13288 should not involve changes to loaders.

Issue #6241, opened 2015-03-16
cyclic dependencies


ghost:
Hey everyone.

@kumavis and I have been hard at work trying to find an efficient way to move THREE.js over to a browserify architecture. We made good progress, even up to the point of having all of the files moved over to a browserify build system and being able to generate a three.min.js with gulp.

Unfortunately, the examples don't work, because unlike commonjs browserify cannot handle cyclic dependencies, of which there are many in THREE.js.

I have made an interactive graph depicting the dependency relationships <a target="_blank" href="http://jsbin.com/medezu/2/edit?html,js,output">here</a>.

Unless and until these get untangled, we will not be able to move THREE.js over to a browserify build.

I do not consider this a deficiency of browserify, but rather a problem with THREE.js. Circular dependencies are a bad thing to have in software in general, and lead to all sorts of problems.

kumavis:
thats quite a knot to untie
http://jsbin.com/medezu/2/edit?html,js,output
![image](https://cloud.githubusercontent.com/assets/1474978/6681386/e989d1e0-cc22-11e4-8c4a-88ff1c872a10.png)

kumavis:
@coballast can you post the code you used to generate the dependency json?

bhouston:
Just use the precompiled three.min.js file directly.  There is no need to break up Three.js into individual files within Browserfy, you are just making your life more difficult for no real benefit.

bhouston:
I speak from experience, in that we use the npm module of three.js and it works great.  We just package it up as a single file and wrap it in a CommonJS style module.  This approach would work for browserfy and a lot of people are already doing it I understand.

Untangling this knot isn't needed for this use case.

coballast:
@kumavis I just dumped the dependency structure. The following code then generated the graph:
```javascript
var fs = require('fs-extra');
var unique = require('uniq');
var util = require('util');
 
function getRequiredObjects(dependencies){
  var result = [];
  for(var i = 0; i < dependencies.usedObjects.length; i++){
    var object = dependencies.usedObjects[i];
    if(dependencies.definedObjects.indexOf(object) == -1)
      result.push(object);
  }
 
  return result;
};
 
var dependencies = JSON.parse(fs.readFileSync('./dependencies.json'));
 
var objects = [];
for(var f in dependencies){
  objects = objects.concat(dependencies[f].usedObjects);
  objects = objects.concat(dependencies[f].definedObjects);
}
 
objects = unique(objects);
 
 
[...]

PR #14466, opened 2018-07-12
Quaternion: Refactor .setFromUnitVectors()

Mugen87:
This PR slightly refactors `Quaternion.setFromUnitVectors()` in order to avoid the closure and the dependency to `Vector3` (keyword: circular dependency). The cross product is now inlined.

### DIFF
--- a/src/math/Quaternion.js
+++ b/src/math/Quaternion.js
@@ -1,5 +1,4 @@
 import { _Math } from './Math.js';
-import { Vector3 } from './Vector3.js';
 
 /**
  * @author mikael emtinger / http://gomo.se/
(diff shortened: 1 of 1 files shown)


WestLangley:
You are inlining the `.crossVectors()` method to avoid a circular dependency.

I do not know if the circular dependency is an issue or not. I will have to defer to @bhouston.

If the circular dependency is not an issue, then I suppose this PR is not necessary, especially since it obfuscates what is going on.

BTW, I expect the method should have included `.onChangeCallback()`.  But that is a separate issue.

bhouston:
> I do not know if the circular dependency is an issue or not. I will have to defer to @bhouston.

I personally have no problem with with circular intra-modular dependencies.   I am against intra-modular circular dependencies.  They should only go one way between larger modules.  For example I am completely okay with all circular dependences between the classes in the /math folder.  But I am very very against /math depending on stuff in /core or /webgl.

Inline this math in another fundamental math class is just making the library bigger, more verbose and harder to maintain.  Do we really thing someone will want to just include Quaternion and not Vector3?  That scenario seems very unlikely.  If that scenario is unlikely, then why do this change?  We are just making the code bigger and more verbose for no real benefit.

But I can see someone wanting to include /math but not bring in /webgl or /core.

That said, I am against this type of change, but it does't really impact anyone significantly.  It is more of a make work project.  I think that most open source projects under go random drift like this and it is best to just coast with it for the most part, as it makes more people feel like they are contributors.

bhouston:
I am a broken record on this: https://github.com/mrdoob/three.js/issues/6241#issuecomment-82525014

Mugen87:
Sounds reasonable.

Issue #11682, opened 2017-07-04
GLTFLoader: Extensibility


Jeremy-Gaillard:
Hi,

We are currently implement [b3dm](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel) (a format that encapsulates glTF) support in our project and have encountered a few problems with the current GLTFLoader implementation.

We need to support addition attribute (`batchid`, used to identify the different objects packed in the geometry), extensions ([CESIUM_RTC](https://github.com/KhronosGroup/glTF/blob/master/extensions/Vendor/CESIUM_RTC/README.md)) and potentially additional uniforms. However, there does not seem to be any way to add these new features without modifying the GLTFLoader file itself.

So my question is: how to support the extensible nature of glTF? Do we contribute directly to the source and add the attributes/extensions/uniforms that we need? Or does the loader need to be modified so extensions can be "plugged in" without changing the source?

Related question: when loading glTFs in a scene rendered using a logarithmic depth buffer, the model's shader are not modified to account for it, resulting in depth issues: http://jsfiddle.net/x6ufnz3y/. I implemented a [small hack](https://github.com/iTowns/itowns2/commit/d3e2456d36ca1799d72cc4393291c2228895ce09#diff-4c76f3570ec49d62d5e66e22e861662aR87) to fix the problem. Should this process be added to the loader natively? Or could Three provide a helper function to patch the materials?

donmccurdy:
> So my question is: how to support the extensible nature of glTF?

Hi @Jeremy-Gaillard — I like the idea of modifying the glTF loader to support user-defined extensions, rather than just the extensions we've implemented ourselves. This would make it much easier to support draft extensions like #11577. Definite +1 from me on that. Hopefully we could do this by having a common interface for all extensions with methods like (`onAfterTexture`, `onAfterNode`, ...).

However, I think we would be implementing that extensibility on GLTF2Loader, and GLTFLoader — are you able to upgrade to the newer loader?

> when loading glTFs in a scene rendered using a logarithmic depth buffer, the model's shader are not modified to account for it, resulting in depth issues...

This seems specific to glTF1, as glTF2 does not have custom shaders currently. If it's not too much trouble, it may be best to keep your hack as-is, and if it's still an issue when GLTF2Loader gets custom shader support later, we'll figure it out then. Most development work is on GLTF2Loader at this point.

[...]

PR #14470, opened 2018-07-14
GLTFLoader: MOZ_alt_materials extension

robertlong:
The MOZ_alt_materials extension can be used to conditionally load alternate materials at runtime. We use it in Mozilla Hubs to publish a single glTF asset that uses a PBR material on high-end platforms and unlit on low-end platforms.

More info on the extension can be found here:
https://github.com/KhronosGroup/glTF/pull/1350


donmccurdy:
I'm a bit hesitant about merging this for a couple reasons:

(1) So far it's a fairly application-specific extension. I can imagine it being useful to others, but it's early at this point, and there are other ways to approach "responsive" 3D models.
(2) This functionality can be bypassed by using separate unlit and PBR assets, which might have different poly counts etc., so lack of this extension isn't blocking user goals.
(3) This is very much the type of extension that I'd like for users to be able to implement themselves, without modifications to GLTFLoader. See https://github.com/mrdoob/three.js/issues/11682 for some previous discussion. For example, we could provide a `lazy` property to the loader to prevent downloading anything up front except the raw JSON, and then return a parser so users can request pieces of the glTF asset a la cart:

```
var loader = new THREE.GLTFLoader();
loader.lazy = true;
loader.load('foo.gltf', function ( gltf ) {
  json = gltf.json;
  var myExtension = gltf.json.extensions.MOZ_foo;
  // ... traverse the JSON doing stuff with custom extensions
  var material = gltf.parser.getDependency( 'material', myExtension.myMaterialIndex );
  var mesh = gltf.parser.getDependency( 'mesh', myExtension.myMeshIndex );
});
```

Our glTF parser is already set up to work that way, so functionally we're already pretty close.

robertlong:
I agree with those points and I like the idea of the lazy API. That option will help us implement a few other modifications we are looking for. I'll try it out and make another PR. Thanks!

PR #14479, opened 2018-07-15
Depth Displacement effect

juniorxsound:
Added a depth displacement effect example inspired by `Pixi.js` [displacement filter](https://www.arpatech.com/blog/3d-parallax-effect-2d-images-depth-map/)
- Added `effects/displacement` example
- Added `js/effects/DepthDisplacement.js`
- Added a color image (`textures/depth_effect_color.png`) and depth map  (`textures/depth_effect_depthmap.png`)


looeee:
[Live version](https://rawgit.com/juniorxsound/three.js/f1cb779687061dd5d670dad638def8e83fc3dba9/examples/webgl_effects_displacement.html)

looeee:
This is pretty cool, but I suspect it would look better with a picture that has more depth. Here is the Pixi.js example for contrast.

juniorxsound:
@looeee Thanks and I totally agree

I think having a depth-map with more planes (background/mid-ground/foreground) would work better (as the `Pixi.js` example shows), let me try and commit a new set of textures and see whether it makes the example more compelling

looeee:
[Updated live version](https://rawgit.com/juniorxsound/three.js/7e35920042fca48b021c113c10d504e790ed1899/examples/webgl_effects_displacement.html) 

BTW you can make live versions using rawgit.com

This looks better, although I think the depth map still needs to be cleaned up. There's lots of artefacts as you move to the extreme left or right. Or is that a limitation of this technique?

juniorxsound:
Thanks for the live versions 🙌 I'll make sure to update them in future commits

It is a result of the technique, however, the `DepthDisplacementEffect` has a method for setting the amount of displacement applied using `effect.setDisplacement( amount );`. Perhaps, it would make sense to slightly lower the number on the example so the artifacts are not *as* visible.

juniorxsound:
[Live version](https://rawgit.com/juniorxsound/three.js/depthDisplacementEffect/examples/webgl_effects_displacement.html)

examples/webgl_effects_displacement.html
```diff
@@ -0,0 +1,157 @@
...
+
+		<script>
+
+			if ( ! Detector.webgl ) Detector.addGetWebGLMessage();
+
+			var container, stats;
+
+			var camera, scene, renderer, effect;
+
+      init();
```
Mugen87: Your code needs some cleanup seems it looks like you mix hard with soft tabs. There are also some code style issues. Please have a look at: https://github.com/mrdoob/three.js/wiki/Mr.doob%27s-Code-Style%E2%84%A2

Mugen87 (changes requested):

examples/js/effects/DepthDisplacement.js
```diff
@@ -0,0 +1,139 @@
...
+		].join( "\n" )
+
+	} );
+
+	this.setSize = function ( width, height ) {
+
+    _material.uniforms.filterArea.value = new THREE.Vector4(width,
+                                                            height,
+                                                            window.innerWidth,
+                                                            window.innerHeight);
```
WestLangley: The effect should not have a reference to `window`.

examples/js/effects/DepthDisplacement.js
```diff
@@ -0,0 +1,139 @@
...
+    _material.uniforms.filterClamp.value = new THREE.Vector4(0,
+                                                             0,
+                                                             (window.innerWidth - 1) / width,
+                                                             (window.innerHeight - 1) / height);
+
+	};
+
+  this.onMouseMove = function ( e ){
+
+    _material.uniforms.scale.value = new THREE.Vector2((window.innerWidth / 2 - e.clientX) * _displacementAmount,
```
WestLangley: This method probably does not belong here.

Also, do not instantiate a new `Vector3` every time the mouse moves. Use `.set()`, instead.

examples/webgl_effects_displacement.html
juniorxsound: Thanks, followed the guidelines and committed changes

examples/js/effects/DepthDisplacement.js
juniorxsound: Thanks @WestLangley, I swapped the `new Vector3` to `.set()`

examples/js/effects/DepthDisplacement.js
juniorxsound: Fixed in new commit 🙌

examples/webgl_effects_displacement.html
Mugen87: Much better! 👍

Mugen87 (approved):

examples/js/effects/DepthDisplacement.js
```diff
@@ -0,0 +1,136 @@
...
+	var _aspectRatio = color.image.width / color.image.height;
+
+	var _displacementAmount = 0.005;
+
+	var _direction = - 1.0;
+
+	var _material = new THREE.ShaderMaterial( {
+		uniforms: {
+			'colorMap': {
+				type: 't',
```
WestLangley: `type` is not needed. 6 occurrences.

examples/js/effects/DepthDisplacement.js
```diff
@@ -0,0 +1,136 @@
...
+		].join( "\n" )
+
+	} );
+
+	this.setSize = function ( rendererWidth, rendererHeight, windowWidth, windowHeight ) {
+
+		_windowWidth = windowWidth;
+		_windowHeight = windowHeight;
+
+		_material.uniforms.filterArea.value = new THREE.Vector4( rendererWidth, rendererHeight, _windowWidth, _windowHeight );
```
WestLangley: As I said previously, use `set()`. In all occurrences.

examples/webgl_effects_displacement.html
```diff
@@ -0,0 +1,153 @@
...
+					var rendererSize = renderer.getSize();
+
+					effect = new THREE.DepthDisplacementEffect( renderer, colorTexture, depthTexture );
+					effect.setSize( rendererSize.width, rendererSize.height, window.innerWidth, window.innerHeight );
+					scene.add( effect.getObject3D() );
+
+					document.addEventListener( 'mousemove', onDocumentMouseMove, false );
+					window.addEventListener( 'resize', onWindowResize, false );
+
+					animate();
```
WestLangley: There is no need for an animation loop.. Call `render()`, instead.

examples/webgl_effects_displacement.html
```diff
@@ -0,0 +1,153 @@
...
+				renderer.setSize( window.innerWidth, window.innerHeight );
+
+				var rendererSize = renderer.getSize();
+				effect.setSize( rendererSize.width, rendererSize.height, window.innerWidth, window.innerHeight );
+
+			}
+
+			function onDocumentMouseMove( event ) {
+
+				effect.updateMousePosition( event );
```
WestLangley: Call `render()` here.

examples/webgl_effects_displacement.html
```diff
@@ -0,0 +1,133 @@
...
+			//
+
+			function onWindowResize() {
+
+				camera.aspect = window.innerWidth / window.innerHeight;
+				camera.updateProjectionMatrix();
+				renderer.setSize( window.innerWidth, window.innerHeight );
+
+				var rendererSize = renderer.getSize();
+				effect.setSize( rendererSize.width, rendererSize.height, window.innerWidth, window.innerHeight );
```
WestLangley: When there is no animation loop, you have to add:
```javascript
effect.render( scene, camera );
```

examples/webgl_effects_displacement.html
```diff
@@ -0,0 +1,133 @@
...
+
+				} );
+
+				loadingManager.onLoad = function () {
+
+					var rendererSize = renderer.getSize();
+
+					effect = new THREE.DepthDisplacementEffect( renderer, colorTexture, depthTexture );
+					effect.setSize( rendererSize.width, rendererSize.height, window.innerWidth, window.innerHeight );
+					scene.add( effect.getObject3D() );
```
WestLangley: @Mugen87 This PR has an unusual API, but I will leave that up to you.

juniorxsound: @WestLangley I am open to suggestions/examples for design patterns for such an effect, given it is applied on a `PlaneGeometry` I picked the obvious of just adding it to the scene, but perhaps there is a simpler way of applying it?

Mugen87: If possible, remove `updateMousePosition()`, `setDirection() and `setDisplacement()` from `DepthDisplacementEffect`.

WestLangley: It is an interesting effect, but maybe the reason it doesn't seem to fit in is because it does not require three.js. Maybe it is more appropriate as a [ShaderToy](https://www.shadertoy.com) example.

I guess you could render a three.js scene, and use the scene's depth map, instead... I'm not sure what the point of that would be, though...

juniorxsound: @WestLangley Thanks! I will look into sharing this on ShaderToy

@Mugen87 Thanks for the suggestions! pushed some changes -https://github.com/mrdoob/three.js/pull/14479/commits/cf4d487e4d7fa28ea2dbe97c056ec36928001cca

Mugen87:
@mrdoob Are you interested in such an effect? The [demo](https://codepen.io/fatik/pen/jLMeeB) from the mentioned article does this with Pixi.js. Do want a similar effect in `three.js`?

However, I also understand @WestLangley view (see https://github.com/mrdoob/three.js/pull/14479#discussion_r204213705). The code would also fit well in the forum's `Showcase` category.

I don't feel strong about both option. In any event, the PR would require some subsequent works (creating TS files, module migration) if it is going to be merged.

juniorxsound:
@Mugen87 Happy to put in the work if it is going to be merged 🙌🏽

Mugen87:
> Also note @WestLangley view (see #14479 (comment)). The code would fit well in the forum's Showcase category.

In the meanwhile, it's better to share this code as a showcase in the [forum](https://discourse.threejs.org/). Or at Shadertoy like suggested in https://github.com/mrdoob/three.js/pull/14479#discussion_r204213705. 

In any event, PixiJS is probably a better choice than `three.js` if you want to create this kind of effect (the mentioned demo uses a the built-in [PIXI.filters.DisplacementFilter](https://pixijs.download/dev/docs/PIXI.filters.DisplacementFilter.html).

PR #14492, opened 2018-07-17
Add lazy option to GLTFLoader.

robertlong:
This PR adds the ability to lazily load or load parts of a glTF model. This can also be used to modify the glTF before calling `parser.parse()`. We intend to implement the `MOZ_alt_materials` extension in this way (see #14470).

This was previously discussed in: #11682

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -182,6 +182,9 @@ THREE.GLTFLoader = ( function () {
 
 			if ( this.lazy ) {
 
+				// Mark the special nodes/meshes in json for efficient parse
+				this.markDefs();
+
 				onLoad( { parser: parser } );
 
 				return;
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -12,6 +12,7 @@ THREE.GLTFLoader = ( function () {
 
 		this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
 		this.dracoLoader = null;
+		this.lazy = false;
 
 	}
 
@@ -72,6 +73,13 @@ THREE.GLTFLoader = ( function () {
 		setDRACOLoader: function ( dracoLoader ) {
 
 			this.dracoLoader = dracoLoader;
+			return this;
+
+		},
+
+		setLazy: function ( lazy ) {
+
+			this.lazy = lazy;
 			return this;
 
 		},
@@ -171,6 +179,14 @@ THREE.GLTFLoader = ( function () {
 				manager: this.manager
 
 			} );
+
+			if ( this.lazy ) {
+
+				onLoad( { parser: parser } );
+
+				return;
+
+			}
 
 			parser.parse( function ( scene, scenes, cameras, animations, json ) {
 
--- a/docs/examples/loaders/GLTFLoader.html
+++ b/docs/examples/loaders/GLTFLoader.html
@@ -156,6 +156,34 @@
 		Refer to this [link:https://github.com/mrdoob/three.js/tree/dev/examples/js/libs/draco#readme readme] for the details of Draco and its decoder.
 		</p>
 
+		<h3>[method:null setLazy]( [param:Boolean value] )</h3>
+		<p>
+		[page:String value] — When true, [page:Function parse] and [page:Function load] will return an object with the <em>parser</em> property set.
+		</p>
+		<p>
+			When lazy loading is enabled and the provided url points to a <em>.gltf</em> file [page:Function load] will only download the <em>.gltf</em> file. [page:Function parse] and [page:Function load] will not call <em>parser.parse()</em> or <em>parser.getDependency()</em>. Examples:
+		</p>
+		<code>
+		var loader = new THREE.GLTFLoader();
+
+		loader.setLazy(true);
+
+		loader.load('foo.gltf', function ( gltf ) {
+
+			var parser = gltf.parser;
+
+			// Modify the glTF before calling parse
+			parser.json.node[ 3 ].extensions = { EXT_foo: { bufferView: 3 } };
+
+			// Load part of the gltf 
+			parser.getDependency( "bufferView", 1 ).then( createNavMesh );
+
+			// Load the entire glTF
+			parser.parse( function ( scene, scenes, cameras, animations, json ) { } );
+
+		} );
+		</code>
+
 		<h3>[method:null parse]( [param:ArrayBuffer data], [param:String path], [param:Function onLoad], [param:Function onError] )</h3>
 		<p>
 		[page:ArrayBuffer data] — glTF asset to parse, as an ArrayBuffer or <em>JSON</em> string.<br />


examples/js/loaders/GLTFLoader.js
```diff
@@ -172,6 +180,14 @@ THREE.GLTFLoader = ( function () {
 
 			} );
 
+			if ( this.lazy ) {
+
+				onLoad( { parser: parser } );
+
+				return;
+
+			}
```
donmccurdy: Would you want to also return the `json`  object here? I imagine users will want to use that to access `extensions` data, or to list the available materials.

examples/js/loaders/GLTFLoader.js
```diff
@@ -172,6 +180,14 @@ THREE.GLTFLoader = ( function () {
 
 			} );
 
+			if ( this.lazy ) {
```
donmccurdy: There is also some pre-processing in `parser.parse()` that still needs to happen here — specifically `parser.markDefs()`.

takahirox:
Doesn't this work for `MOZ_alt_materials` even without `lazy`?

    var fileLoader = new THREE.FileLoader();
    var gltfLoader = new THREE.GLTFLoader();
    var path = THREE.LoaderUtils.extractUrlBase( url );

    fileLoader.setResponseType( 'text' ).load( url, function ( json ) {

        var materialDefs = json.materials;
        var altMaterialDefs = [];

        for ( var i = 0, il = materialDefs.length; i < il; i ++ ) {

            var materialDef = materialDefs[ i ];
            var materialIndex = i;

            if ( preferredTechnique !== undefined &&
                materialDef.extensions &&
                materialDef.extensions[ 'MOZ_alt_materials' ] &&
                materialDef.extensions[ 'MOZ_alt_materials' ][ preferredTechnique ] !== undefined ) {

                materialIndex = materialDef.extensions[ 'MOZ_alt_materials' ][ preferredTechnique ];

            }

            altMaterialDefs.push( materialDefs[ materialIndex ] );

        }

        json.materials = altMaterialDefs;

        gltfLoader.parse( json, THREE.LoaderUtils.extractUrlBase( url ), function ( gltf ) {

            scene.add( gltf.scene );

        } );

    } );

donmccurdy:
This extension could perhaps be implemented without the `lazy` option, yes, but per https://github.com/mrdoob/three.js/issues/11682 I think this is a good change all the same. We don't want to add too many vendor extensions into the loader, and having some basic hooks like this makes it easier for others to do custom things.

examples/js/loaders/GLTFLoader.js
robertlong: The `json` object is available via `parser.json`. I think this lines up well with the non-lazy method.

When specifying `lazy`:
```
onLoad( { parser: parser } )
```

When using the default non-lazy method:
```
onLoad( {
  scene: scene,
  scenes: scenes,
  cameras: cameras,
  animations: animations,
  asset: json.asset,
  parser: parser,
  userData: {}
} )
```

robertlong:
I agree, I think the `lazy` option makes implementing this extension easier to implement with very minimal changes.

Another example of something we will use this for in Hubs:
https://github.com/mozilla/hubs/blob/master/src/vendor/GLTFLoader.js#L43

We need to rewrite URLs for our file proxy using an asynchronous API after downloading the glTF json but before calling `parse()`. We could use the method you specified above, but as you mentioned we would also need to provide methods to continue to support `glb` files.

takahirox:
First, I agree with having hook points for handling user-defined extensions. My concern for `.lazy` is I guess it'd be a bit tricky for users...? My suggestion would be more friendly with Three.js users. (Yeah, there's a room for discussion.)

Before we go forward for further discussion, I have one question. How do you expect user uses `parser` before `parser.parse()` calling except for overriding json. I meant, is there any usecases where `lazy` can do while my suggestion (+ some fixes) can't?

donmccurdy:
In addition to the case of implementing extensions, note that the glTF spec says:

> The glTF asset contains zero or more scenes ... When scene is undefined, runtime is not required to render anything at load time.
> 
> **Implementation Note:** This allows applications to use glTF assets as libraries of individual entities such as materials or meshes.

For users who want to do this, manually downloading the JSON file and then "parsing" the entire thing doesn't seem like the right API. Even if you already had the JSON downloaded, I think you'd want a `lazy` option — or something like it — to access dependencies one by one, preferably without needing to know much about the structure of a glTF file.

Or maybe the file contains multiple scenes, one for each zone in the world, and you only want to download each zone as the user reaches it. This is preferable to having separate `.gltf` files for each zone, because materials and textures will automatically be recycled by the parser.

Out of scope for this PR, but we could make this even more user-friendly with a simple helper method that does lookups by name instead of by index: `getDependencyByName( 'mesh', 'castle_tower' )`.

***

**tl;dr** — I do think having a convenient way of getting dependencies out of a glTF file one by one is a good thing. But if there are better APIs for it than a `.lazy` option, I'm totally open to that. Maybe something like:

- `loader.getParser( 'foo.gltf', ( parser ) => { ... } )`
- `parser = loader.getParser( json | arraybuffer )`

robertlong:
I agree with @donmccurdy, being able to get dependencies out of a glTF file is important and one of the nice things about this API.

I'd like to see the ability to modify the JSON after loading a GLTF/GLB's JSON. Ideally the loader handles both formats and hands me a JSON object so I don't have to check the response content type and magic integer and call `parseGLB`.

I'd also like to have the ability to pull images, bufferViews, specific nodes, etc out of the glTF asset. This will be useful for packing additional assets into a glTF file.

However, if there's a better API than `.parse` I'm also open to suggestions.

takahirox:
Agreed with providing APIs to access the dependencies. Wondering which one do custom-extension developers want before parsing entire glTF, dependent Three.js object or dependent json content. I suppose json content rather than Three.js object?

I think one of the major use cases where to handle custom-extension before parsing entire glTF is to prevent downloading/loading/parsing unnecessary assets. For example, material's custom extension overrides texture with the lower resolution one for mobile use. In that case with parser's `.getDependency`, you get material's custom extension with `var material = parser.getDependency( 'material', 0 ); material.userData.gltfExtensions...; ` but default texture are already loaded inside.

takahirox:
@robertlong 

>  I'd also like to have the ability to pull images, bufferViews, specific nodes, etc out of the glTF asset. This will be useful for packing additional assets into a glTF file.

Adding assets? Can you give me some examples?

donmccurdy:
> In that case with parser's .getDependency, you get material's custom extension with var material = parser.getDependency( 'material', 0 ); material.userData.gltfExtensions...; but default textures for the material are already loaded inside. 

That's OK I think — the material and its dependencies should be fully loaded, just as long as we're only loading the textures needed for that material and no more. There are a few places in GLTFLoader where we use [getDependencies](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L2317), and we should try to remove those, but that can be another PR.

takahirox:
> I think one of the major use cases where to handle custom-extension before parsing entire glTF is to prevent downloading/loading/parsing unnecessary assets. For example, material's custom extension chooses lower resolution one for mobile use. 

I meant, for example, a glTF material refers to large size textures and the material's custom extension refers to small size textures for mobile use. For mobile use, user doesn't want to download the large textures and instead wants to download the small textures. But `var material = parser.getDependency( 'material', materialIndex );` loads the large textures because `GLTFLoader` doesn't know the custom extension and returns material after solving the dependencies.

I think the purpose of providing a hook point to user before parsing entire glTF is providing a way to prevent unnecessary downloading/parsing data for handling custom extension. But I don't think `.getDependency/ies` satisfies that demand, so I suppose what custom extension developers wants before parsing entire glTF is dependent json contents.

donmccurdy:
I see two use cases here:

1.  Modify the JSON before it is parsed. Users already can do that by fetching the JSON first, as you mentioned earlier, modifying it, then calling `parse()`. 
2. Load _unstructured_ or _incremental_ data out of the asset, which isn't necessarily replacing anything. For example, a `.gltf` asset might contain just a list of materials. Or the application might load additional meshes and textures as they are needed during an experience.

Technically both cases work with `lazy`, because you can access the JSON via `parser.json`. I don't think (1) is as important, because it can already be accomplished by pre-loading JSON or by simply creating different (or dynamically generated) `.gltf` JSON files that reference some shared buffers and textures.

On the other hand, (2) is not possible without the `lazy` option, unless you do things like constructing dummy meshes to extract your materials from the "material library" asset, and that seems hacky. I think this lazy option is a useful thing to have, although it will need some additional refinement in the loader (to avoid fetching all available buffers at various times) to make it flexible.

takahirox:
`GLTFLoader.load/parse()` returns `parser` as argument of `onLoad` callback (it's as temporal workaround tho). Can't it solve (2)?

donmccurdy:
Except that all available scenes will be loaded eagerly. The content author would need to be sure nothing is referenced by any scene, or that there are no scenes in the asset at all. For most users I think that would involve hand edits to the file, so an API for lazy-loading does still seem simpler.

takahirox:
Now I have some ideas/thoughts. Let me share after I'm back to home next week.

takahirox:
Sorry I'm late.

I suggest to add `.getParser()` method to `GLTFLoader` (as experimental) and remove `parser` from `onLoad`'s parameter of `GLTFLoader`'s `.parse()`, rather than adding `.lazy` option.

I'll write the reason this weekend, time to sleep now.

takahirox:
>This use case is not temporary — users should be able to implement custom extensions without hacking the loader, at least for basic use cases.

I agree with providing a way to process their custom extension in user side without hacking (I specified it as 3. above) I meant, returning `parser` via `onLoad` is temporal workaround until we find better solutions.

> How do you suggest getParser working? Could you write an example of how someone would use it? A GLTFLoader creates a new parser with every .load() call, you need to have loaded the JSON portion of the file to even create a parser... something like this?

Yes, something like that. I meant `.getParser()` as in https://github.com/mrdoob/three.js/pull/14492#issuecomment-407976476

donmccurdy:
With #14779 merged I think it would be a reasonable time to revisit this, if you are still interested. I'm still not sure of the right API here though.

takahirox:
I still prefer separating `parser` return from `.load()`. Making `.getParser()` PR and I think I can share in this year.

robertlong:
I think we can close this PR I'm in favor in the `.createParser()` API instead.

PR #12572, opened 2017-11-03
Geometry: Add skin information to .fromBufferGeometry()

Mugen87:
Just a little enhancement to `Geometry .fromBufferGeometry()` so skin information won't get lost when performing the conversion.

Issue #10729, opened 2017-02-04
Suggested change for fromBufferGeometry to include Skin information


filharvey:
(*** This section is for bug reports and feature requests only. This is NOT a help site. Do not ask help questions here. If you need help, please use stackoverflow. ***)

##### Description of the problem 
Skin information is not included when calling fromBufferGeometry.

`

fromBufferGeometry: function ( geometry ) {

			var scope = this;

			var indices = geometry.index !== null ? geometry.index.array : undefined;
			var attributes = geometry.attributes;

			var positions = attributes.position.array;
			var normals = attributes.normal !== undefined ? attributes.normal.array : undefined;
			var colors = attributes.color !== undefined ? attributes.color.array : undefined;
			var uvs = attributes.uv !== undefined ? attributes.uv.array : undefined;
			var uvs2 = attributes.uv2 !== undefined ? attributes.uv2.array : undefined;
			var skinIndex = attributes.skinIndex !== undefined ? attributes.skinIndex.array : undefined;
			var skinWeight = attributes.skinWeight !== undefined ? attributes.skinWeight.array : undefined;

			if ( uvs2 !== undefined ) this.faceVertexUvs[ 1 ] = [];

			var tempNormals = [];
			var tempUVs = [];
			var tempUVs2 = [];

			for ( var i = 0, j = 0, k = 0; i < positions.length; i += 3, j += 2, k += 4 ) {

				scope.vertices.push( new Vector3( positions[ i ], positions[ i + 1 ], positions[ i + 2 ] ) );

				if (skinIndex !== undefined)
				{
					scope.skinIndices.push( new Vector4( skinIndex[ k ], skinIndex[ k + 1 ], skinIndex[ k + 2 ], skinIndex[ k + 3 ] ) );
				}

				if (skinWeight !== undefined)
				{
					scope.skinWeights.push( new Vector4( skinWeight[ k ], skinWeight[ k + 1 ], skinWeight[ k + 2 ], skinWeight[ k + 3 ] ) );
				}

				if ( normals !== undefined ) {

					tempNormals.push( new Vector3( normals[ i ], normals[ i + 1 ], normals[ i + 2 ] ) );

				}

				if ( colors !== undefined ) {

					scope.colors.push( new Color( colors[ i ], colors[ i + 1 ], colors[ i + 2 ] ) );

				}

				if ( uvs !== undefined ) {

					tempUVs.push( new Vector2( uvs[ j ], uvs[ j + 1 ] ) );

				}

[...]

PR #14493, opened 2018-07-18
Geometry.fromBufferGeometry -> copy weights and morphs data

damienmortini:
Copy weights and morphs data to Geometry from source BufferGeometry to keep animations.

### DIFF
--- a/src/core/Geometry.js
+++ b/src/core/Geometry.js
@@ -263,16 +263,17 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 			}
 
-		}
-		if ( skinIndices !== undefined ) {
+			if ( skinIndices !== undefined ) {
 
-			scope.skinIndices.push( new Vector4( skinIndices[ k ], skinIndices[ k + 1 ], skinIndices[ k + 2 ], skinIndices[ k + 3 ] ) );
+				scope.skinIndices.push( new Vector4( skinIndices[ k ], skinIndices[ k + 1 ], skinIndices[ k + 2 ], skinIndices[ k + 3 ] ) );
 
-		}
+			}
 
-		if ( skinWeights !== undefined ) {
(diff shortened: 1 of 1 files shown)


Mugen87:
We already had two PRs (#12572 and #10729), none of them were accepted. I feel this is a `Won't fix`.

BTW: What is your use case for converting a `Geometry` to `BufferGeometry`?

src/core/Geometry.js
```diff
@@ -260,6 +263,65 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+			scope.skinWeights.push( new Vector4( skinWeights[ k ], skinWeights[ k + 1 ], skinWeights[ k + 2 ], skinWeights[ k + 3 ] ) );
+
+		}
+
+		var morphAttributes = geometry.morphAttributes;
+
+		if ( morphAttributes.position ) {
+
+			for ( const morphTargetAttribute of morphAttributes.position ) {
```
Mugen87: Please don't use `const` and `let`. Also avoid `for of` loops.

Mugen87 (changes requested):

src/core/Geometry.js
```diff
@@ -260,6 +263,65 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 			}
 
+		}
```
Mugen87: add a blank line here

damienmortini:
Just fixed these, thanks!

damienmortini (commented):
Updated requested changes

src/core/Geometry.js
damienmortini: All done thanks

src/core/Geometry.js
damienmortini: Ok

Mugen87 (approved):

Issue #14496, opened 2018-07-18
Make lookAt work with the scenegraph?


greggman:
As it is `lookAt` says it doesn't work if the object you're calling it for has parents etc... I was wondering what you all think about removing that restriction. My guess is most projects only call `lookAt` at most once a frame to aim the camera so making it a little heavier is not likely to affect performance since it's not the kind of function to be called 1000s of times a frame.

On the other hand, making it more flexible makes it more useful for things like pointing a character's head at an object of interest. Pointing a security camera what what it's tracking. Pointing a gun turret at it's target. 

Thoughts?

Here's a sample

https://jsfiddle.net/greggman/7gdux2ch/

and a preliminary pull request below

Mugen87:
I vote for this change. Unity also respects the parent transformation when using `lookAt()`. It also useful for #14444

Mugen87:
BTW: Cool fiddle 😊

pailhead:
>Thoughts?


```javascript
const myVec = new THREE.Vector3()
const myNode = new THREE.Object3D()

myNode.updateMatrixWorld()
myNode.localToWorld(myVec)

thing.lookAt(myVec)
```

With the `localToWorld` this limitation is not so far detached from the scene graph. Personally, i would maybe avoid overloading `lookAt` and do `lookAtNode`, i believe this conversation was had in the past.

greggman:
I'm just going by pretty much every other 3D engine except three.js.  There's matrix lookAt which is of course the standard function. Then there's scene graph lookAt which in 3DSMax, Maya, Unity, Unreal, and most other engines do take the position in the scene graph into account.  It really makes zero sense for scene graph functions not to work with the scene graph. In other words. `THREE.Matrix4.lookAt` deals with matrices. `THREE.Object3D.lookAt` deals with the scene graph. So you already have 2 functions, one for each case.


All of them require the camera to be part of the scene so that you can put cameras anywhere (like [the fiddle](https://jsfiddle.net/greggman/7gdux2ch/)).

WestLangley:
@greggman Don't worry, this will be fixed. It is going to have to be implemented in conjunction with some additional changes to ensure matrix updates during graph traversal are efficient.

greggman:
[...]

Issue #3845, opened 2013-09-05
Matrix Decomposition is Not Always a Valid Operation


WestLangley:
In the three.js paradigm, the user specifies an object's `position`, `quaternion` (or `rotation`), and `scale`, and from that, the object `matrix` is computed.

Sometimes, however, we preform the inverse calculation, and _infer_ a `position`, `quaternion`, and `scale` from an object's `matrix` or `matrixWorld`.

This decomposition occurs in the following functions

    Object3D.applyMatrix()
    Matrix4.decompose()
    SceneUtils.attach()/detach()
    Gyroscope.updateMatrixWorld()

Unfortunately, such a decomposition is not always possible, but we are doing it anyway.

The problems occur when the scale is `non-uniform`.

Here is a fairly simple explanation of what is happening.

If `R` is a pure rotation matrix, and `S` is a scale matrix, and `T` is a translation matrix, then an object's matrix is of the form:

    object.matrix = T * R * S.

When applied to a vector, the order in which these transforms occur is: scale first, then rotation, then translation -- i.e., from right to left. This order is important.

If the object has a parent, and the parent's `matrixWorld` is of the same form,

    parent.matrixWorld = T2 * R2 * S2.

then the object's world matrix is

    object.matrixWorld = T2 * R2 * S2 * T * R * S.

Now, as a special case, suppose that the object's `matrix` is a pure rotation matrix (i.e., `T` and `S` are the identity).

    object.matrix = R

Also suppose that the parent's `matrixWorld` is a pure scale matrix (i.e., `T` and `R` are the identity).

    parent.matrixWorld = S

Then, the object's `matrixWorld` is

    object.matrixWorld = S * R

Notice that the order is reversed -- `S` is on the left.

This means that when the matrix is applied to a vector, the rotation is applied first, and the scale is applied second.

When the scale is _non-uniform_, a transform of this form cannot, in general, be decomposed into a scale followed by a pure rotation.

What this means is that __parents with non-uniform scale need to be avoided__. It is OK if (childless) children have non-uniform scale.

The same problem occurs in `Object3D.applyMatrix()`.

Here is a fiddle showing the issue: http://jsfiddle.net/PwSmH/

This is all a consequence of the decision to represent an object's transform with a  `position`, `quaternion/rotation`, and `scale` -- applied in the order S-R-T.

The only solution, unfortunately, is to change the paradigm. A fallback position is to just say that three.js does not support non-uniform scaling of parent objects.

[...]

PR #14497, opened 2018-07-18
Make Object3D.lookAt take scenegraph into account

greggman:
#14496 

note I don't know three very well so this is just an idea. Not sure what might be wrong with this code.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -303,10 +303,10 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	lookAt: function () {
 
-		// This method does not support objects with rotated and/or translated parent(s)
-
 		var m1 = new Matrix4();
+		var m2 = new Matrix4();
 		var vector = new Vector3();
+		var worldPosition = new Vector3();
 
 		return function lookAt( x, y, z ) {
 
@@ -320,13 +320,24 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 			}
 
+			var position = this.parent ? this.getWorldPosition( worldPosition ) : this.position;
+
 			if ( this.isCamera ) {
 
-				m1.lookAt( this.position, vector, this.up );
+				m1.lookAt( position, vector, this.up );
 
 			} else {
 
-				m1.lookAt( vector, this.position, this.up );
+				m1.lookAt( vector, position, this.up );
+
+			}
+
+			if ( this.parent ) {
+
+				this.parent.updateMatrixWorld();
+				m2.getInverse( this.parent.matrixWorld );
+				m2.extractRotation( m2 );
+				m1.premultiply( m2 );
 
 			}
 


src/core/Object3D.js
```diff
@@ -303,10 +303,10 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 	lookAt: function () {
 
-		// This method does not support objects with rotated and/or translated parent(s)
-
-		var m1 = new Matrix4();
-		var vector = new Vector3();
+		const m1 = new THREE.Matrix4();
```
Mugen87: Please avoid the usage of the `THREE` namespace in the source. Besides, use `var` and not `const` or `let`.

src/core/Object3D.js
```diff
@@ -320,13 +320,23 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 			}
 
+			const position = this.parent ? this.position : this.getWorldPosition( worldPosition );
```
Mugen87: The order seems incorrect. If `this.parent` is not `null`, we want to use `Object3D.getWorldPosition()`. Otherwise `this.position` is already in world space.

Mugen87 (changes requested):

src/core/Object3D.js
Mugen87: Oh, the code in your fiddle is actually correct^^.

src/core/Object3D.js
```diff
@@ -303,10 +303,10 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 	lookAt: function () {
 
-		// This method does not support objects with rotated and/or translated parent(s)
-
-		var m1 = new Matrix4();
-		var vector = new Vector3();
+		var m1 = new THREE.Matrix4();
```
Mugen87: You are still using the `THREE` namespace...

Mugen87 (changes requested):

src/core/Object3D.js
greggman: Doh! I thought I fixed that before checking in 😅

src/core/Object3D.js
```diff
@@ -320,13 +320,23 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
-				m1.lookAt( vector, this.position, this.up );
+				m1.lookAt( vector, position, this.up );
+
+			}
+
+			if ( this.parent ) {
+
+				this.parent.updateMatrixWorld();
+				m2.getInverse( this.parent.matrixWorld );
+				m1.premultiply( m2 );
```
Mugen87: Is it actually safe to perform this calculation? `m2` is not necessarily a pure rotation matrix. Using `Matrix4.extractRotation()` should make the code more robust...

Demo: https://jsfiddle.net/7gdux2ch/41/

Uncomment the usage of the method to see the difference.

src/core/Object3D.js
greggman: I'm a little curious why no `const`. THREE is using ES6 modules and the build already translates `const` into `var`. Is this just a project preference? Is it an outdated preference? There's no perf difference and it's considered best practice by nearly every major tech company's style guide. It also prevents lots of errors and scoping issues. Is there a link to some rational for the policy?

Mugen87: > Is this just a project preference?

Yes. At least for now. 

/cc @mrdoob

looeee:
> I'm a little curious why no const. THREE is using ES6 modules and the build already translates const into var. Is this just a project preference? Is it an outdated preference? 

The last time this came up, there were still some issues - I think not so much with `const` as with `let`. However, it's been a while and it might be worth evaluating those again now to if things have changed in the meantime.

src/core/Object3D.js
greggman: Added that in.

Mugen87: Non-Uniform scaling is problematic in many aspects, see #3845.

Mugen87 (approved):

WestLangley:
In your implementation, you call `getWorldPosition()`, which calls `updateMatrixWorld()`, and later, you call `updateMatrixWorld()` again.

Also, as I said [here](https://github.com/mrdoob/three.js/issues/14496#issuecomment-406766433), you are relying on `updateMatrixWorld()` to work correctly when called on a scene node, but it doesn't. It only works correctly when called on the scene itself. Obviously, this is not your fault. :-)

I think your suggestion is a good one, and is worthy of getting right.

greggman:
Ah, my bad for not knowing three wasn't already doing it right 😅. Thanks
for doing it correctly.

PR #14505, opened 2018-07-19
GLTFLoader remove deprecated texture format, internalFormat, and type.

robertlong:
The glTF 2.0 spec no longer contains the `format`, `internalFormat`, and `type` properties ([texture spec](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/texture.schema.json)).

This PR removes these properties form being processed by the glTF loader. As a result JPEG textures can now use the RGB vs RGBA format. This also in turn makes the images exported from GLTFExporter the correct file types.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -2021,22 +2021,6 @@ THREE.GLTFLoader = ( function () {
 
 			if ( textureDef.name !== undefined ) texture.name = textureDef.name;
 
-			// .format of dds texture is set in DDSLoader
-			if ( ! textureExtensions[ EXTENSIONS.MSFT_TEXTURE_DDS ] ) {
-
-				texture.format = textureDef.format !== undefined ? WEBGL_TEXTURE_FORMATS[ textureDef.format ] : THREE.RGBAFormat;
-
-			}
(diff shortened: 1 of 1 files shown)


examples/js/loaders/GLTFLoader.js
```diff
@@ -2021,22 +2021,6 @@ THREE.GLTFLoader = ( function () {
...
-			}
-
-			if ( textureDef.internalFormat !== undefined && texture.format !== WEBGL_TEXTURE_FORMATS[ textureDef.internalFormat ] ) {
-
-				console.warn( 'THREE.GLTFLoader: Three.js does not support texture internalFormat which is different from texture format. ' +
-											'internalFormat will be forced to be the same value as format.' );
-
-			}
-
-			texture.type = textureDef.type !== undefined ? WEBGL_TEXTURE_DATATYPES[ textureDef.type ] : THREE.UnsignedByteType;
```
donmccurdy: Guess we no longer need `WEBGL_TEXTURE_DATATYPES` or `WEBGL_TEXTURE_FORMATS` definitions either?

takahirox: I think so.

PR #14508, opened 2018-07-19
Add boxselection demo

HypnosNova:
This feacture is very common used in BIM system.
It can help people to drag a box on the screen and then calculate a frustum in 3D space to check which objects are in frustum and these objects are selected.

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -59,6 +59,7 @@ var files = {
 		"webgl_interactive_lines",
 		"webgl_interactive_points",
 		"webgl_interactive_raycasting_points",
+		"webgl_interactive_boxselection",
 		"webgl_interactive_voxelpainter",
 		"webgl_kinect",
 		"webgl_lensflares",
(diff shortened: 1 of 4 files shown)


06wj:
[live demo](https://rawgit.com/HypnosNova/three.js/BoxSelection/examples/webgl_interactive_boxselection.html)

11zouzouzou:
good function

HypnosNova:
I found a small bug and fixed it.

mrdoob:
Thanks!

WestLangley:
@Mugen87 I think this PR missed a review.

Mugen87:
Yeah, we have missed that. Apart from code style issues, I see a lot of unnecessary object creation in `SelectionBox'. Especially in `SelectionBox.select()` and `SelectionBox.createFrustum()` which are frequently called in the `mousemove` event listener of the example.

@HypnosNova You should try to avoid object instantiation via `new` and `clone()` methods.You might want to wrap `SelectionBox` into a closure and define reusable variables in the new respective scope.

HypnosNova:
@Megun87 OK，i will update code tomorrow

HypnosNova:
OK，i will update code tomorrow@Mugen87

WestLangley:
Thanks, @Mugen87.

Issue #14419, opened 2018-07-05
SEA3D vs GLTF


RemusMar:
This is the current (july 2018) GLTF status.
Test case: the same skinned mesh exported from 3DS Max 2018.
Standard material: Diffuse + Specular + Normal

1) SEA3D exporter + SEA3D importer:
http://necromanthus.com/Test/html5/Lara.html 
SEA file size: 658 KB
Result: close to perfect

2) Babylon3D GLTF exporter + GLTF importer:
http://necromanthus.com/Test/html5/Lara_gltf.html 
GLB file size: 1,850 KB
Result: messed up materials

I've also tested the FBX2GLTF utility (by Facebook): the same wrong results

Important note: there is nothing wrong with THREE.js and PBR materials:
http://necromanthus.com/Test/html5/Lara_PBR.html 

In any case, PBR was a bad choice for GLTF and also, all the current converters are collection of bugs.

Mugen87:
I don't think github is the right place for this post. It's neither a feature request, nor a bug. So my question is: What are you trying to accomplish with this issue? Bashing `glTF`?

If you encounter problems with an exporter or converter, I suggest you open an issue at the respective github repo.

> In any case, PBR was a bad choice for GLTF and also, all the current converters are collection of bugs.

I generally reject such Trump-like statements. They have a provocative nature and are not objective at all.

RemusMar:
> I don't think this repo is the right place for this post.

It's the best place for sure and it shows the current GLTF status.

> I generally reject such Trump-like statements.

Really.
Here is a statement from Trump: Google already failed with UTF8.

looeee:
@RemusMar are you suggesting and particular actions we should take? If not I vote to close this issue.

RemusMar:
> RemusMar are you suggesting any particular actions we should take?

Three possible causes for the wrong GLTF results:
1. the Babylon3D exporter is buggy (they say it's not)
2. the FBX2GLTF converter is buggy (they say it's not)
3. the GLTF importer is buggy.

But you and Mugen87 want to close the topic because there is no issue and everytbody is happy ...

looeee:
> the GLTF importer is buggy

You mean the GLTFLoader? Can you identify what the bug is? It would be especially helpful if you can find a very simple model that demonstrates the bug.

>  (they say it's not)

If you've made bug reports on the Babylon3D exporter and FBX2GLTF can you link to them here?

RemusMar:
[...]

Issue #11337, opened 2017-05-15
Enhancement: add renderer.outputEncoding


WestLangley:
As discussed in #11332

1. `renderer.gammaInput` needs to be deprecated. We now support `texture.encoding` instead.

2. renderer.gammaOutput needs to be deprecated and replaced with `renderer.outputEncoding` when rendering to the screen. We already support `renderTarget.texture.encoding` when rendering to a render target.

3. WebGLPrograms.getProgramCode()  needs to be updated to respond to changes in `renderer.outputEncoding`.  Setting `material.needsUpdate = true` will then force the change.

4. Legacy warnings need to be added, and the examples need to be updated to accommodate this change.

WestLangley:
Actually, before we go down that path, perhaps we should decide on decoupling "encoding" into _color space_ and _encoding format_,  as [suggested](https://github.com/mrdoob/three.js/pull/11076#issuecomment-293570510) by @bhouston.

>If we split color space conversion from the encodings, we could do the above. It would be more flexible and clear. We could also support LUTs in the color space conversions as well.

>And then each texture and render target should have both a ColorSpace (Linear, sRGB, AdobeRGB, Gamma20, Gamma22) and a Encoding (None, RGBE, RGBM, RGBD, LogLUV, CMYK) parameter.

This is something I would support.

bhouston:
> Actually, before we go down that path, perhaps we should decide on decoupling "encoding" into color space and encoding format, as suggested by @bhouston.

Separating colorSpace and encoding is not that hard.  I would just copy the code used to inject functions for encoding into the glsl and modify it to inject colorSpace transforms.  And then split the enums into a color space list and an encoding list.  I guess I am saying just use the existing design for encodings, but duplicate it again for color spaces.

I think that is sort of the easy, obvious solution.

bhouston:
The injectors are pretty simple - they just create specially named functions that map onto the chosen encoding function: 

https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLProgram.js#L12

I use the same injection idea for tonemapping: getToneMappingFunctio

bhouston:
Hmm... I wonder if tone mapping could be combined with color spaces?  They are sort of the same thing theoretically -- a color transform.  And they sort of happen at the same time in the renderer pipeline (before the encoding.)

[...]

PR #14512, opened 2018-07-20
FBXLoader: Set sRGBEncoding for various texture types.

looeee:
As discussed in #14419, it would be best if loaders set texture encoding to `THREE.sRGBEncoding` by default, however only `GLTFLoader` currently does this.  Unfortunately, changing this now for other loaders would be a breaking change.

Instead I've added a flag to the `FBXLoader`. Defaults to false (previous behaviour), while setting `fbxLoader.sRGBTextureEncoding = true` will now set ` THREE.sRGBEncoding` on the diffuse map, emmissive map, and specular map if they are present. 

Enabling this flag will also require setting `renderer.gammaOutput = true;` for correct results. 

Unlike glTF, FBX format does not specify the texture encoding that should be used, so it's probably best that this is set by a flag rather than be default behaviour anyway. 

@donmccurdy I've followed the behaviour of `GLTFLoader` and only set the encoding for `diffuseMap`, `emissiveMap` and `specularMap`. Is this correct?


donmccurdy:
Thanks for looking into this! `envMap` should be included as well; glTF does not include environment maps at this point. Note that the Mixamo example will change noticeably with gammaOutput enabled, because it doesn't use textures at all.

I do like the idea of exposing an option for correct colorspace and lighting. But I wouldn't use the word "encoding" here, per https://github.com/mrdoob/three.js/issues/11337, we hope to rename that property on textures soon. Perhaps:

- loader.sRGBTextureColorSpace = true
- loader.textureColorSpace = THREE.sRGBEncoding
- loader.physicallyCorrectLighting = true

A couple references on the (lack of) colorspace clarity in FBX:

- [Vertex colors: sRGB, linear, ???](https://forums.autodesk.com/t5/fbx-forum/vertex-colors-srgb-linear/td-p/7985264) (unanswered)
- [Color management in 3DS Max marked for "future consideration"](https://forums.autodesk.com/t5/3ds-max-ideas/implement-color-management-in-3ds-max-support-for-icc-luts-etc/idi-p/6790724)

looeee:
I like the name `physicallyCorrectLighting`. However it's likely to cause confusion with the [WebGLRenderer.physicallyCorrectLights](https://threejs.org/docs/#api/renderers/WebGLRenderer.physicallyCorrectLights), which as far as I can make out is related to making lights have a physically correct fallof in a real-world sized scene, rather than anything to do with color space. 

Let's go with `loader.textureColorSpace = THREE.sRGBEncoding`, it seems like the most flexible.

looeee:
Before (right ) and after (left) comparison. 

srgb_comparison

donmccurdy (approved):

looeee:
@mrdoob are there any issues with this or can it be merged?

looeee:
/ping @mrdoob  

Is there some issue with adding color management to the loader? If so let me know and I'll close this.

mrdoob:
I think I would avoid having the additional flag. I think it's okay for this to be a breaking change.

looeee:
> I think I would avoid having the additional flag.

Yeah, I prefer that too. I've removed the flag and this is now done automatically:

``` js
material.map.encoding = THREE.sRGBEncoding;
material.emissiveMap.encoding = THREE.sRGBEncoding;
material.envMap.encoding = THREE.sRGBEncoding;
material.specularMap.encoding = THREE.sRGBEncoding;
```

mrdoob:
Thanks!

Mugen87:
Updated the migration guide with this change:

https://github.com/mrdoob/three.js/wiki/Migration-Guide#r106--r107

PR #14094, opened 2018-05-18
A simple improvement to blending cache

oguzeroglu:
In this PR a simple improvement is made inside WebGLState to prevent redundant calls to WebGL blending functions:
ekran resmi 2018-05-13 15 26 43

The modification insures that no redundant call would be made when the blending mode is NoBlending. The tests resulted successful.

PR #14516, opened 2018-07-20
WebGLState: More efficient blending

WestLangley:
Sharing this as a WIP. It is an another approach to #14094, and it is just a proposal.

An opaque scene should set blending only once.

### DIFF
--- a/src/renderers/webgl/WebGLState.js
+++ b/src/renderers/webgl/WebGLState.js
@@ -2,7 +2,7 @@
  * @author mrdoob / http://mrdoob.com/
  */
 
-import { NotEqualDepth, GreaterDepth, GreaterEqualDepth, EqualDepth, LessEqualDepth, LessDepth, AlwaysDepth, NeverDepth, CullFaceFront, CullFaceBack, CullFaceNone, CustomBlending, MultiplyBlending, SubtractiveBlending, AdditiveBlending, NoBlending, NormalBlending, DoubleSide, BackSide } from '../../constants.js';
+import { NotEqualDepth, GreaterDepth, GreaterEqualDepth, EqualDepth, LessEqualDepth, LessDepth, AlwaysDepth, NeverDepth, CullFaceFront, CullFaceBack, CullFaceNone, CustomBlending, MultiplyBlending, SubtractiveBlending, AdditiveBlending, NoBlending, NormalBlending, AddEquation, DoubleSide, BackSide } from '../../constants.js';
 import { Vector4 } from '../../math/Vector4.js';
 
 function WebGLState( gl, extensions, utils ) {
@@ -323,6 +323,7 @@ function WebGLState( gl, extensions, utils ) {
 
 	var currentProgram = null;
 
+	var currentBlendingEnabled = null;
 	var currentBlending = null;
 	var currentBlendEquation = null;
 	var currentBlendSrc = null;
@@ -400,8 +401,7 @@ function WebGLState( gl, extensions, utils ) {
 	setCullFace( CullFaceBack );
 	enable( gl.CULL_FACE );
 
-	enable( gl.BLEND );
-	setBlending( NormalBlending );
+	setBlending( NoBlending );
 
 	//
 
@@ -525,13 +525,23 @@ function WebGLState( gl, extensions, utils ) {
 
 	function setBlending( blending, blendEquation, blendSrc, blendDst, blendEquationAlpha, blendSrcAlpha, blendDstAlpha, premultipliedAlpha ) {
 
-		if ( blending !== NoBlending ) {
+		if ( blending === NoBlending ) {
+
+			if ( currentBlendingEnabled ) {
+
+				disable( gl.BLEND );
+				currentBlendingEnabled = false;
+
+			}
+
+			return;
+
+		}
+
+		if ( ! currentBlendingEnabled ) {
 
 			enable( gl.BLEND );
-
-		} else {
-
-			disable( gl.BLEND );
+			currentBlendingEnabled = true;
 
 		}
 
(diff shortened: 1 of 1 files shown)


src/renderers/webgl/WebGLState.js
```diff
@@ -525,80 +525,123 @@ function WebGLState( gl, extensions, utils ) {
 
 	function setBlending( blending, blendEquation, blendSrc, blendDst, blendEquationAlpha, blendSrcAlpha, blendDstAlpha, premultipliedAlpha ) {
 
-		if ( blending !== NoBlending ) {
+		if ( blending === NoBlending ) {
```
Mugen87: I would write the first block like this. It's a bit easier to read since all operations related to `NoBlending` are in a single `if/else` block.
```js
if ( blending === NoBlending ) {

	if ( currentBlending !== NoBlending ) {

		disable( gl.BLEND );
		currentBlending = NoBlending;

	}

} else {

	if ( currentBlending === NoBlending ) {

		enable( gl.BLEND );

	}

}
```

src/renderers/webgl/WebGLState.js
```diff
@@ -525,80 +525,123 @@ function WebGLState( gl, extensions, utils ) {
...
-						} else {
+							break;
 
-							gl.blendEquation( gl.FUNC_ADD );
-							gl.blendFunc( gl.ZERO, gl.SRC_COLOR );
+						default:
 
-						}
-						break;
+							// error
```
Mugen87: I would add an error message at this place. Something like:
```js
console.error( 'THREE.WebGLState: Invalid blending: ', blending );
```

src/renderers/webgl/WebGLState.js
```diff
@@ -525,80 +525,123 @@ function WebGLState( gl, extensions, utils ) {
...
+
+						case MultiplyBlending:
+
+							gl.blendFunc( gl.ZERO, gl.SRC_COLOR );
+
+							break;
+
+						default:
+
+							// error
```
Mugen87: Same here...

src/renderers/webgl/WebGLState.js
```diff
@@ -525,80 +525,123 @@ function WebGLState( gl, extensions, utils ) {
...
 
-					case MultiplyBlending:
+				gl.blendEquation( gl.FUNC_ADD );
 
-						if ( premultipliedAlpha ) {
+				if ( premultipliedAlpha ) {
+
+					switch ( blending ) {
+
+						case AdditiveBlending:
```
Mugen87: Consider to write it a little bit more compact:
```js
switch ( blending ) {

   case AdditiveBlending:
      gl.blendFunc( gl.ONE, gl.ONE );
      break;

   case SubtractiveBlending:
      gl.blendFuncSeparate( gl.ZERO, gl.ZERO, gl.ONE_MINUS_SRC_COLOR, gl.ONE_MINUS_SRC_ALPHA );
      break;
```
Similar to:
https://github.com/mrdoob/three.js/blob/3585caf4f5779c5eb433edd2838257cd65113e33/src/renderers/webgl/WebGLProgram.js#L55-L63

Mugen87 (changes requested):

src/renderers/webgl/WebGLState.js
WestLangley: Done!

src/renderers/webgl/WebGLState.js
WestLangley: Done!

src/renderers/webgl/WebGLState.js
WestLangley: Done!

src/renderers/webgl/WebGLState.js
WestLangley: Done!

src/renderers/webgl/WebGLState.js
```diff
@@ -525,80 +525,107 @@ function WebGLState( gl, extensions, utils ) {
...
 		}
 
-		if ( blending !== CustomBlending ) {
+		if ( blending === NormalBlending ) {
 
-			if ( blending !== currentBlending || premultipliedAlpha !== currentPremultipledAlpha ) {
+			if ( currentBlending !== NormalBlending || premultipliedAlpha !== currentPremultipledAlpha ) {
 
-				switch ( blending ) {
+				gl.blendEquation( gl.FUNC_ADD ); // todo: only set if not
```
Mugen87: Could you use `currentBlendEquation` to check this case?

WestLangley:
Another attempt has been posted. This is significantly more efficient than the current implementation.

I find the existing deeply-nested `if-then-else` structures difficult to even think about, so I have removed them to the extent that I can.

The most important thing at this point is to get the logic right.

Mugen87:
If tested with some examples and the visual outcome seems to be correct. I also could not detect redundant blending state changes via WebGL Inspector.

Mugen87:
@mrdoob BTW: I think it might be a good idea to enable "Allow squash merging" for the repo. In this way, all commits of a PR are merged in one and this goes in the base branch. In this way, it easier to track changes of PRs with obsolete commits.

image

Mugen87:
#13717 is a good example where a single squashed commit would be easier to track than 41...

Mugen87:
@WestLangley Um, why is [capabilities](https://github.com/mrdoob/three.js/pull/14516/files#diff-b100b307fb56276583bcb5ae053dc06eR912) not [enabledCapabilities](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLState.js#L886) at this line?

https://github.com/mrdoob/three.js/pull/14516/files#diff-b100b307fb56276583bcb5ae053dc06eR912

WestLangley:
@Mugen87 Thank you for testing this. I appreciate it.

The concept here is that most scenes have opaque objects, and maybe transparent objects with Normal Blending. So rather than setting the blending params over and over again, we can set blending correctly when it first occurs, and then just disable blending for an opaque object. When we enable blending again later, the previous blending settings will still be in effect -- we do not need to reset them.

In scenes where there are multiple blending modes, the code is still efficient. Making sure it is correct is my focus.

WestLangley:
@Mugen87 Would you be willing to test some bizarre use cases with custom blending using your inspector approach?

WestLangley:
Great. I am hopeful this latest version is correct. See if you can demonstrate otherwise.

Note that in the case of `AdditiveBlending`, I keep things simple and set `currentBlend*` to null. This is for simplicity, and should be good enough for such edge cases.

Mugen87:
I've realized the following examples are ideal for testing the PR. There are no redundant state calls and the visual result is the same compared to production.

https://threejs.org/examples/webgl_materials_blending.html
https://threejs.org/examples/webgl_materials_blending_custom.html

> Note that in the case of AdditiveBlending, I keep things simple and set currentBlend* to null.

You refer to this code, right?

https://github.com/mrdoob/three.js/blob/2de8571f91d10dda5e5ce7fc579695fc43f67614/src/renderers/webgl/WebGLState.js#L626-L629

But this happens not only in case of `AdditiveBlending` but whenever you use not a custom blending approach, correct?

WestLangley:
>I've realized the following examples are ideal for testing the PR.

I studied the first one, but forgot the second one existed. :-)

>You refer to this code, right?

Yes.

>but whenever you use a pre-defined blending approach, correct?

Yes.

Also, my biggest concern is making sure the final setting is what the user wanted. Eliminating redundancy is important, but it is secondary to correctness.

Mugen87:
I suggest you clean up the PR and then we merge it. From my point of view it looks good. The sooner this lands in `dev`  the sooner we detect a potential problem.

src/renderers/webgl/WebGLState.js
WestLangley: Undone. It's all different now...

WestLangley:
Rebased... hopefully no damage done.

Ready to merge.

Mugen87 (approved):

mrdoob:
Lets see how it goes 😊

mrdoob:
Thanks!

PR #14517, opened 2018-07-21
Object3D.lookAt(): support rotated parents

WestLangley:
This PR implements the suggestion by @greggman in #14496, and is based on his work. Credit goes to @greggman for both the suggestion and the initial implementation of this feature.

The enhancements to `lookAt()` are now quaternion-based, avoiding a matrix inversion.

The correct implementation of this feature required a new method which I have named `.updateWorldMatrix()`, for lack of a better term. It is actually a correct implementation of what `.updateMatrixWorld()` should have been.

Rather than getting diverted by how to best modify `.updateMatrixWorld()`, I have added the new method instead. It is currently undocumented. We can address `.updateMatrixWorld()`, and how to improve its performance, separately.

Because almost all scene nodes have a parent, this implementation of `lookAt()` will be a bit slower than it was previously.  We could probably maintain similar performance if we added `Matrix4.isIdentity()`.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -303,34 +303,50 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	lookAt: function () {
 
-		// This method does not support objects with rotated and/or translated parent(s)
+		// This method does not support objects having non-uniformly-scaled parent(s)
 
+		var q1 = new Quaternion();
 		var m1 = new Matrix4();
-		var vector = new Vector3();
+		var target = new Vector3();
+		var position = new Vector3();
 
 		return function lookAt( x, y, z ) {
 
 			if ( x.isVector3 ) {
 
-				vector.copy( x );
+				target.copy( x );
 
 			} else {
 
-				vector.set( x, y, z );
+				target.set( x, y, z );
 
 			}
 
+			var parent = this.parent;
+
+			this.updateWorldMatrix( true, false );
+
+			position.setFromMatrixPosition( this.matrixWorld );
+
 			if ( this.isCamera ) {
 
-				m1.lookAt( this.position, vector, this.up );
+				m1.lookAt( position, target, this.up );
 
 			} else {
 
-				m1.lookAt( vector, this.position, this.up );
+				m1.lookAt( target, position, this.up );
 
 			}
 
 			this.quaternion.setFromRotationMatrix( m1 );
+
+			if ( parent ) {
+
+				m1.extractRotation( parent.matrixWorld );
+				q1.setFromRotationMatrix( m1 );
+				this.quaternion.premultiply( q1.inverse() );
+
+			}
 
 		};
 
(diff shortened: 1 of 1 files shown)


src/core/Object3D.js
```diff
@@ -615,6 +631,44 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 	},
 
+	updateWorldMatrix: function ( updateParents, updateChildren ) {
```
Mugen87: Um, it can be very confusing for users when they encounter this method in the source code. I think it would be better for now to turn `updateWorldMatrix()` into a function and put it in the closure of lookAt().

Mugen87 (changes requested):

src/core/Object3D.js
WestLangley: Every method in the library that calls `updateMatrixWorld()` expecting it to work properly needs to call the new method instead.

That would be in the next PR.

Mugen87: Um, I'm not sure we are heading the right way with introducing a new method. At least, we should not rush with this change and think about the consequences.

WestLangley: I have purposely not touched `updateMatrixWorld()` with this PR.

src/core/Object3D.js
```diff
@@ -303,35 +303,51 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
 			} else {
 
-				vector.set( x, y, z );
+				target.set( x, y, z );
 
 			}
 
+			var parent = this.parent;
+
+			this.updateWorldMatrix( true, false );
```
mbredif: no need to update the matrix world if there is no parent (simply use this.position)

src/core/Object3D.js
```diff
@@ -303,35 +303,51 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
 
 			}
 
 			this.quaternion.setFromRotationMatrix( m1 );
 
+			if ( parent ) {
+
+				m1.extractRotation( parent.matrixWorld );
+				q1.setFromRotationMatrix( m1 );
+				this.quaternion.premultiply( q1.inverse() );
```
mbredif: Would not it be more efficient to compute the world quaternion of the parent by simply multiplying the ancestors quaternions? that may lift the non-uniformly-scaled contraint (to positively scaled?)

src/core/Object3D.js
WestLangley: I think that's almost never the case.

src/core/Object3D.js
WestLangley: Interesting...

mrdoob:
Thanks!

src/core/Object3D.js
mrdoob: @WestLangley are you planning on taking a look at the suggestion?

WestLangley: @mrdoob I have put it on my todo list. :-)

mrdoob: Okay 😊

WestLangley: @mbredif Extracting the quaternion from the world matrix is the correct approach. An ancestor's quaternion is not necessarily set, or even used (i.e., when `matrixAutoUpdate` is `false`).

Issue #14511, opened 2018-07-20
Camera CenterOn to fit objects Perfectly according to camera orientation


moroine:
Hello,

I manage to make it works, you can find an example here: 
https://codepen.io/bilouwan/project/full/XpgGOp/ 

As it is quite complex to implements, no really the code but more the maths and it is quite a common stuff when playing with the camera I am wondering about making a PR.

What do you think about it? Could it be useful? Better to add it under examples or inside the core like Math or Camera method may be...

selimbek:
Hi, 

I'm sure that will be usefull for lot of people. 
I recommand to make PR with commented code, under example section. ( center.js as an external library, and the others files merged into an html file example)

You should pack your center.js into an nes THREE object. A user case could be: THREE.CameraCentering()

selimbek:
Hi,

I'm sure that will be usefull for lot of people.
I recommand to make PR with commented code, under example section. ( center.js as an external library, and the others files merged into an html file example)

You should pack your center.js into a new THREE object (THREE.CameraCentering ?). A user case could be: 
var cameraCentring = new THREE.CameraCentering();
cameraCentring.addCamera( myCamera1);
cameraCentring.addCamera( myCamera2);
...
cameraCentring.centeringTo( threeObect3DArray );

haroldiedema:
That is a very nifty feature.

If you make a PR, please make sure to not have any references to things in the global scope though.

moroine:
Actually, in my case, I am doing some animations with the Camera so I don't want to modify the camera directly.

I think I can do an API similar to Box3 with kind of:

```
const viewBox = new CameraViewBox();
viewBox.setViewFromCamera(camera);
viewBox.setFromObject(object);

viewBox.getCenter(controls.target);
viewBox.getObserverPosition(camera.position);
```

@haroldiedema  sure, global variable are here just to validate the logic, I still need to figure out what would be the API.

haroldiedema:
So if I understand correctly, all you need for this feature to work is a `Camera` and a set of one or more `Object3D` instances right?

If that's the case, I think you're on the right track in my opinion.

```javascript
const camera  = new THREE.PerspectiveCamera();
const viewBox = new THREE.CameraViewBox(camera);

viewBox.centerObjects(object1, object2, object3, ...);
```
I don't think it gets any easier than this.

[...]

PR #14526, opened 2018-07-23
Add CameraViewBox example to fit objects to camera

### LINKED ISSUE #6784: "Fit all" (show all) feature
Dear ThreeJS developers!

Please implement "fit all" ("show all") feature to the camera and/or OrbitControl and other controls.
This feature is very missing thing. 

To provide some help, here is implementation of same feature in X3dom:
https://github.com/x3dom/x3dom/blob/master/src/Viewarea.js#L1077

Also, there is "fit object" (e.g. show object) feature:
https://github.com/x3dom/x3dom/blob/master/src/Runtime.js#L631

Many thanks in advance.

Please do not refer me to other feature requests like https://github.com/mrdoob/three.js/issues/1095

If you deny to do this, please at least point me, where is it better to code it, e.g. in camera, or in OrbitControl/etc, to get them (camera and control) synchronized?

moroine:
Hi,

I made a pixel perfect zoom method which allows setting the Camera position and target keeping the camera orientation.

This is linked to https://github.com/mrdoob/three.js/issues/14511 and could resolve https://github.com/mrdoob/three.js/issues/6784


moroine:
[Live](https://rawgit.com/moroine/three.js/CameraViewBox/examples/misc_camera_center.html)

moroine:
You need to click on an object to update the selection, then it will become red wireframe and center on that selection

looeee:
Got it. It would be more obvious if it started with one object selected and fit ratio ~1.5, with a message "click to add objects to selection". Also, I'd just use a different color material rather than wireframe. Since the camera jumps when you click, together with the wireframe material to me it feels like the object I just clicked on disappeared. 

This is a different approach than what I had in mind for camera fitting. It's more complex, but it may be a good solution. I'll let others give opinions on whether we should add this. 

However, your `CameraViewBox` is basically a `Box3` with a couple of extra methods. It would be better to either use a `Box3` internally, or extend the `Box3` object. 

Take a look at the [Group](https://github.com/mrdoob/three.js/blob/dev/src/objects/Group.js) source to see how we do inheritance.

greggman:
not sure it matters but is it compatible with `
PerspectiveCamera.setViewOffset`?

moroine:
No it's not compatible, but i don't this this would be very complicated

moroine:
Edit the example according to @looeee feedback

moroine:
Hello,
This has been a long time, @looeee what's your opinion about this PR?

looeee:
It's not up to me whether this gets added or not, but as I mentioned earlier, I do think that your example is more complex than it needs to be. However, it's been a while since I looked at your code so perhaps I'm missing something.

For very simple camera centering see @donmcmurdys code here: 

https://github.com/donmccurdy/three-gltf-viewer/blob/67bdd2f4bf5ce8ee512805a0b5252d8269b69ca6/src/viewer.js#L213

Fitting the camera to the object there is done in just a couple of lines. Can you explain why your example is better / accomplishes more than this?

moroine:
I didn't understood everything in the given code but I think @donmccurdy's code make sur that the object will fit into the camera view but doesn't get the optimal camera distance.

My implementation is far more complex, this is indisputable. That's the reason why I proposed it as an example rather in the three.js core.

Despite the complexity, my implementation will keep the current camera orientation and just adjust camera position to make the selection fit perfectly on the screen (you can try putting the fitRatio at 1 in the live example).

All code I've seen so far got some of the following limitations:

- Change the camera orientation
- Zoom is not consistent when the object is not a box axis aligned (see image)
img_20181025_100618

telramund:
Can this be merged, please? It serves a common need, but implementing it off the guidance available online is well beyond the skill of an average THREE.js user. I can attest to that after nearly two days of trying before I discovered this example.

looeee:
@telramund does @donmccurdy's solution that I linked above work for you? If not, can you explain why, and why you need this solution instead?

moroine:
The main difference is that @donmccurdy's solution is using the bounding box, which makes it great in term of optimization but it's not consistent when you have spread objects (such as the example in live version).

On this solution, it considers the optimal view, to make sure to perfectly adjust camera zoom and center to have all objects in the view while keeping the zoom as minimum as possible.

MH4CxhWSv5
> Red objects are selected objects

looeee:
I've done some testing of both methods. It was easy to extend @donmccurdy's method to work with multiple objects. However it doesn't take into account the aspect ratio, so while it works well in landscape or square views, it's not great in portrait. Also, it's a bit greedy, and selecting multiple objects makes the view expand too much. Aside from that, it doesn't take into account the camera's orientation, resetting it with every update..

@moroine's method does work well in all aspect ratios, and preserves camera orientation, so as a general solution it is superior (@donmccurdy's is tuned for his viewer, I guess, and works well in that situation). However, I still think it's much more complicated than necessary. 

I've made a hybrid of @moroine and @donmccurdy's approaches that takes this into account. It's much simpler than this PR and seems to work well for all aspect ratios. Here's code:

``` js
function zoomCameraToSelection( camera, controls, selection, fitOffset = 1.2 ) {
  
  const box = new THREE.Box3();
  
  for( const object of selection ) box.expandByObject( object );
  
  const size = box.getSize( new THREE.Vector3() );
  const center = box.getCenter( new THREE.Vector3() );
  
  const maxSize = Math.max( size.x, size.y, size.z );
  const fitHeightDistance = maxSize / ( 2 * Math.atan( Math.PI * camera.fov / 360 ) );
  const fitWidthDistance = fitHeightDistance / camera.aspect;
  const distance = fitOffset * Math.max( fitHeightDistance, fitWidthDistance );
  
  const direction = controls.target.clone()
    .sub( camera.position )
    .normalize()
    .multiplyScalar( distance );

  controls.maxDistance = distance * 10;
  controls.target.copy( center );
  
  camera.near = distance / 100;
  camera.far = distance * 100;

  camera.position.copy( controls.target ).sub(direction);
  
  controls.update();
  
}
```

### [Codepen](https://codepen.io/looeee/full/vwVeZB)

Notes: I haven't tested the `controls.maxDistance`, or the `camera.near` and `.far` to see whether they are optimal. The values adapted from @donmccurdy's code seem to work well though. 
Also, the `fitOffset = 1.2` is a minimum value that works. Since the view is set to the bound bound, when the rotation of the frustum doesn't fit closely to the box, objects close to the edge will be clipped without this.

Mugen87:
Centering an object based on a bounding box and deriving related camera parameters is easy to implement and provides usable results in many use cases. I don't think we should add more complex code to the repository. Instead I suggest to add the PR's example code as a showcase in the [forum](https://discourse.threejs.org/). You can then describe in the topic why you came up with this solution and how it works.

looeee:
>  I don't think we should add more complex code to the repository

I previously stated my opinion on this when we discussed it on the forum:

https://discourse.threejs.org/t/my-personal-threejs-scenes-gallery-portfolio-made-with-react/7564/9

I think that this is a conversation that needs to be had on a more formal level, and until we've done that and come a consensus as a community, or until @mrdoob reaches a decision, I don't think the right approach is to suddenly start telling people that their contributions are no longer welcome because we don't want more code added to the repo.

In this case, the  function that I just demonstrated, which accomplishes the same thing as @moroine's code, is not complex - certainly less complex than many existing examples demonstrating similar common use cases of the library. As @telramund [says above](https://github.com/mrdoob/three.js/pull/14526#issuecomment-497119067), what's obvious and easy to set up for you as an expert may be non-obvious and complex for other people. Zooming the camera to fit objects is a common question on the forum and I think it would be good to have a (simple) example demonstrating how to do it. This example demonstrates a solution to a far more common problem than [many](https://threejs.org/examples/#webgl_materials_grass) of [the](https://threejs.org/examples/#webgl_interactive_voxelpainter) [current](https://threejs.org/examples/#webgl_effects_ascii) [examples](https://threejs.org/examples/#webgl_effects_peppersghost). 

In the end, whether or not an example of how to do this gets added to the repo is fine by me, but we need to discuss it based on the the same merits that we've judged previous additions by. Or, if the repo has truly become too bloated to maintain, we need to do something about that. 

Joining in the discussion on @takahirox's issue #16471 is a better way of going about this than telling people that we no longer want their contributions, IMO. Or, open a new issue to discuss what we should keep/add to this repo and what we should remove or move to a separate repo. We have a new organisation at [github.com/threejs](https://github.com/threejs/) where we could move a lot of the example stuff if desired.

Mugen87:
> I don't think the right approach is to suddenly start telling people that their contributions are no longer welcome because we don't want more code added to the repo.

A contribution like @moroine's code is of course welcome, just at another place.

telramund:
> @telramund does @donmccurdy's solution that I linked above work for you? If not, can you explain why, and why you need this solution instead?

Sorry for the late response. I do not recall trying this particular solution, but I tried others based on bounding box and they were too imprecise. Also, it appears to change the camera orientation (lookAt) which is also an issue for me. My use case is a CAD app where you look at an assembly (the machine you're building) from a certain position and angle. You want to be able to select a few parts (or the whole) of that assembly and zoom in so they just fit and and without changing your viewing angle. The assemblies an be of very funny shapes (elongated, Г-shapedm etc), so the bounding box approach doesn't always work well. I was only able to accomplish what I want with the CameraViewBox and nothing else I could find.

Mugen87:
@looeee How about putting your code from https://github.com/mrdoob/three.js/pull/14526#issuecomment-497254491 into a new method of `SceneUtils` along with a new example? I think that would be a good compromise and also fixes #6784.

WestLangley:
zoom-to-scene is the solution to an optimization problem that can have any number of app-specific constraints.

A three.js method or utility requires that you are able to parameterize the problem statement somehow, and provide a sufficiently-general solution to it.  A sufficiently-general solution is not appropriate for three.js.

I, too, think the forum showcase is best.

Mugen87:
Agreed. Closing then.

PR #14527, opened 2018-07-23
fix intersectsBox

washstar:
computing time is reduced

### DIFF
--- a/src/objects/Mesh.js
+++ b/src/objects/Mesh.js
@@ -245,7 +245,7 @@ Mesh.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 			// Check boundingBox before continuing
 
-			if ( geometry.boundingBox === null ) geometry.computeBoundingBox();
+			if ( geometry.boundingBox === null ) return;
 			
 			if ( ray.intersectsBox( geometry.boundingBox ) === false ) return;
 
--- a/src/objects/Mesh.js
+++ b/src/objects/Mesh.js
@@ -162,7 +162,7 @@ Mesh.prototype = Object.assign( Object.create( Object3D.prototype ), {
 		function checkIntersection( object, material, raycaster, ray, pA, pB, pC, point ) {
 
 			var intersect;
-
+			if(!material) return null;
 			if ( material.side === BackSide ) {
 
 				intersect = ray.intersectTriangle( pC, pB, pA, true, point );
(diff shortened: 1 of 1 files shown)


pailhead:
You have some lint issues. 

What exactly does this do? Before if there were no bounding box this would do nothing, now you force the bounding box to be computed. What if someone doesnt want that to happen in a middle of a session, when clicking or hovering over stuff?

washstar:
thanks,I fixed this problem

src/objects/Mesh.js
```diff
@@ -162,7 +162,7 @@ Mesh.prototype = Object.assign( Object.create( Object3D.prototype ), {
 		function checkIntersection( object, material, raycaster, ray, pA, pB, pC, point ) {
 
 			var intersect;
-
+			if(!material) return null;
```
Mugen87: I'm not sure this line makes sense. Is `checkIntersection()` ever be called without a material instance?

src/objects/Mesh.js
```diff
@@ -245,11 +245,9 @@ Mesh.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 			// Check boundingBox before continuing
 
-			if ( geometry.boundingBox !== null ) {
-
-				if ( ray.intersectsBox( geometry.boundingBox ) === false ) return;
-
-			}
+			if ( geometry.boundingBox === null ) return;
```
Mugen87: You now leave the method if no bounding box is defined. This is clearly wrong. The usage of `boundingBox` in `Mesh.raycast()` is optional. If no bounding box is defined, you still want to perform the raycast.

Mugen87 (changes requested):

src/objects/Mesh.js
washstar: It happens when mesh.group.materialIndex=-1,

src/objects/Mesh.js
washstar: If the method does not return when boundingBox is null, it will have additional calculations.I want to reduce the amount of calculation by judging whether the boundingBox intersects.

src/objects/Mesh.js
Mugen87: This case should never happen. In general we do not check such special cases in the library.

src/objects/Mesh.js
Mugen87: > If the method does not return when boundingBox is null, it will have additional calculations.

The library should perform an intersection test even if no `boundingBox` is defined.

src/objects/Mesh.js
washstar: I did experience this during the development process

src/objects/Mesh.js
washstar: you are right

PR #14531, opened 2018-07-23
Remove Animation auto optimization and validation.

takahirox:
This PR removes animation auto optimization and validation. See #14336


donmccurdy:
I'm OK with this PR but there are other changes in https://github.com/mrdoob/three.js/pull/14385, including these changes, that I think would also be worth having.

takahirox:
I just didn't think you add "removal auto optimization/validation" in #14385. I'll close this PR now.

PR #14536, opened 2018-07-24
rad / deg mismatch in fbx loader

kubafran:
I was having a problem with some of my models animations (sorry I cannot share them). What I found is that if there is no angle defined for one of the axie in an animation key frame then fbx loader uses a value from initial state but this value is in degrees while the values in animation keys ar already converted to radians.

### DIFF
--- a/examples/js/loaders/FBXLoader.js
+++ b/examples/js/loaders/FBXLoader.js
@@ -2643,8 +2643,9 @@
 
 		}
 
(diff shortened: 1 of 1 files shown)


looeee:
I think this should be fixed in #14523 since I've updated the way animation transforms are applied. 
Can you try that and see if it fixes your issue?

kubafran:
@looeee Yes, on your branch my animations are working as expected, thanks. I will wait for the merge and close my PR.

PR #14537, opened 2018-07-24
Examples: more  Geometry -> BufferGeometry

WestLangley:
The minor visual differences between the new and updated versions is unavoidable, and is due to vertex indices in `IcosahedronBufferGeometry`(204 vertices) differing from those in `THREE.IcosahedronGeometry` (42 vertices).

### DIFF
--- a/examples/webgl_geometry_colors.html
+++ b/examples/webgl_geometry_colors.html
@@ -108,54 +108,47 @@
 				shadowMesh.rotation.x = - Math.PI / 2;
 				scene.add( shadowMesh );
 
-				var faceIndices = [ 'a', 'b', 'c' ];
+				var radius = 200;
 
-				var color, f, f2, f3, p, vertexIndex,
+				var geometry1 = new THREE.IcosahedronBufferGeometry( radius, 1 );
 
-					radius = 200,
+				var count = geometry1.attributes.position.count;
+				geometry1.addAttribute( 'color', new THREE.BufferAttribute( new Float32Array( count * 3 ), 3 ) );
 
-					geometry  = new THREE.IcosahedronGeometry( radius, 1 ),
-					geometry2 = new THREE.IcosahedronGeometry( radius, 1 ),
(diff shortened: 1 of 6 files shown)


WestLangley:
To clarify, the only difference is the middle sphere.

In the 'Before' example, `Geometry` vertices are shared, and face vertex colors are set by the vertex index in the vertices array:
```javascript
color.setHSL( 0.125 * vertexIndex / geometry.vertices.length, 1.0, 0.5 );
```
In the 'After' example, the `BufferGeometry` is triangle soup, and the color attribute for vertex` i` is set like so:
```javascript
color.setHSL( 0.125 * i / count, 1.0, 0.5 );
```
Quite frankly, the colors are not relevant to the primary feature being demonstrated in any of the six examples.

Mugen87:
I would choose different colors for the second sphere so the transition between the red and orange color in the center is not so steep.

image

WestLangley:
I coud do this:
```javascript
color.setRGB( 1, 1, ( positions3.getY( i ) / radius + 1 ) / 2 );
```
screen shot 2018-07-25 at 11 27 12 am

Mugen87:
Yeah, why not :+1:

mrdoob:
Thanks!

mrdoob:
Just updated the examples to match a bit more the previous version:

https://rawgit.com/mrdoob/three.js/dev/examples/webgl_geometry_colors.html
https://rawgit.com/mrdoob/three.js/dev/examples/webgl_multiple_canvases_circle.html
https://rawgit.com/mrdoob/three.js/dev/examples/webgl_multiple_canvases_complex.html
https://rawgit.com/mrdoob/three.js/dev/examples/webgl_multiple_canvases_grid.html
https://rawgit.com/mrdoob/three.js/dev/examples/webgl_multiple_renderers.html
https://rawgit.com/mrdoob/three.js/dev/examples/webgl_multiple_views.html

PR #14543, opened 2018-07-25
[WIP] Dirty Flag for updateMatrix optimization

takahirox:
See #14138 for the discussion and performance gain.

I still need to update some tests to pass, so still WIP. But you can start to evaluate the performance if you're interested in.

And feedback is welcome.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -31,20 +31,24 @@ function Object3D() {
 
 	this.up = Object3D.DefaultUp.clone();
 
-	var position = new Vector3();
+	var position = new Vector3WithDirtyFlag();
 	var rotation = new Euler();
 	var quaternion = new Quaternion();
-	var scale = new Vector3( 1, 1, 1 );
+	var scale = new Vector3WithDirtyFlag( 1, 1, 1 );
+
+	quaternion._dirty = false;
 
 	function onRotationChange() {
 
 		quaternion.setFromEuler( rotation, false );
+		quaternion._dirty = true;
 
 	}
 
 	function onQuaternionChange() {
 
 		rotation.setFromQuaternion( quaternion, undefined, false );
+		quaternion._dirty = true;
 
 	}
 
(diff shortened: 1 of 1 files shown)


src/core/Object3D.js
```diff
@@ -835,5 +849,79 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 } );
 
+function Vector3WithDirtyFlag( x, y, z ) {
```
takahirox: Is there any other good places where to define this new object, for example `src/math/Vector3WithDirtyFlag.js`?

src/core/Object3D.js
```diff
@@ -835,5 +849,79 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+	this._x = 0;
+	this._y = 0;
+	this._z = 0;
+
+	Vector3.call( this, x, y, z );
+
+	this._dirty = false;
```
takahirox: I don't wanna expose the existence of dirty flag to users then I added prefix `_` for `._dirty`.

src/core/Object3D.js
gkjohnson: I mentioned this in the other PR but there's the option of encapsulating this logic entirely in the constructor of `Object3D` so we don't have another vector type. The use of the this type might be pretty specific to the `Object3D` class.

src/core/Object3D.js
```diff
@@ -835,5 +849,79 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+			this._z = value;
+			this._dirty = true;
+
+		}
+
+	}
+
+} );
+
```
takahirox: I don't make `QuaternionWithDirtyFlag` yet in this PR because not sure why but somehow using the extension of `Quaternion` seems to lead the bad performance on my Windows Chrome...

src/core/Object3D.js
```diff
@@ -31,20 +31,24 @@ function Object3D() {
 
 	this.up = Object3D.DefaultUp.clone();
 
-	var position = new Vector3();
+	var position = new Vector3WithDirtyFlag();
 	var rotation = new Euler();
 	var quaternion = new Quaternion();
```
takahirox: I don't make `QuaternionWithDirtyFlag` yet in this PR because not sure why but somehow using the extension of `Quaternion` seems to lead the bad performance on my Windows Chrome...

src/core/Object3D.js
takahirox: No performance overhead?

gkjohnson: That's a good point -- I'm sure it does incur a bit more overhead on instantiation. I'm not sure how much, though.

takahirox:
Realized the dirty flag approach has the same issue mentioned in https://github.com/mrdoob/three.js/pull/15706#issuecomment-464450166, that is

```javascript
b.updateMatrix(); // clean
b.matrix.copy( a.matrix ); // manually update local matrix
b.updateMatrix(); // should be updated but no update because they are clean
```

Unless we make local matrix private, we also need dirty flag in local matrix but there is no way to observe array values change. `Proxy` may be only the way but IE doesn't support...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

Mugen87:
see https://github.com/mrdoob/three.js/pull/14138#issuecomment-532177576

PR #14544, opened 2018-07-25
Prevent navigation if modifiers pressed.

greggman:
In Windows Ctrl-Click opens a link in a new tab. In MacOS it's Cmd-Click.
But the code here was not respecting the user's wishes. A new tab
would open but the page itself would also navigate to the same link
which is not what a user is requesting with ctrl-click or cmd-click

Interestingly the demo page did not navigate when pressing Cmd-Click
on MacOS but the URL did change. Now the URL will correctly not
change and the link will open in a new window.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -93,8 +93,12 @@
 				link.setAttribute( 'target', 'viewer' );
 				link.addEventListener( 'click', function ( event ) {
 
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

Issue #13373, opened 2018-02-20
Add ability to set/get light "direction" explicitly 


bhouston:
##### Description of the problem

I'd like to add to the Directional Light and Spot light the ability to set/get the light direction in an explicit fashion.  Preferrably light.direction = ... or var direction = light.direction.  Right now both the DirectionalLight and SpotLight are really hard to figure out where they are pointing, it isn't one line of code but a bunch.  I would like to a generic accessor that hides this complexity.

##### Three.js version

- [x] Dev

##### Browser

- [x] All of them

##### OS

- [x] All of them

WestLangley:
>Right now both the DirectionalLight and SpotLight are really hard to figure out where they are pointing

They are pointing at `light.target` -- but you must add `light.target` to the scene, or call:
```javascript
light.target.updateMatrixWorld();
```

If you want to be able to _set_ the direction, then I expect we would have to remove `light.target`.

looeee:
I made a similar request a couple of months ago, that we switch to setting the direction of the light via the rotation / quaternion. 

At the time I was told that the problem with this was that the light shadow requires a target to work, and that this is the reason the lights have targets. 

If this is indeed the case, could we add `onRotationChange` type method to the lights that updates the target position when they are rotated, and possibly make the target invisible to the user?

Usnul:
@looee can you link the issue?

I have an adapter in my code specifically to address this issue as well.

WestLangley:
>At the time I was told that the problem with this was that the light shadow requires a target to work, and that this is the reason the lights have targets.

Someone may have said that, but my recollection is the reason lights have targets is so the user can set the target to be an object in the scene, and the light will follow it. That is why the target is an `Object3D`, and not a `Vector3`.

>If this is indeed the case, could we add onRotationChange type method to the lights that updates the target position when they are rotated, and possibly make the target invisible to the user?

Consequently, no.

looeee:
My apologies, it seems I remembered wrongly

https://github.com/mrdoob/three.js/issues/10220#issuecomment-263016723

The comment from @WestLangley was that lights require a `position` for shadows, which makes more sense.

Usnul:
@WestLangley 
Is light a meaningful primitive by itself in three.js? if so, it should be usable independently, and usable fully. Do you agree? As it stands - directional light can not be utilized fully without introducing  a target.

[...]

PR #14557, opened 2018-07-26
Allow setting direction of DirectionalLight by angles

EliasHasle:
I appreciate the Three.js philosophy of keeping things clean, simple and modular. Yet, I often find myself in situations where I have to create and manipulate a Spherical to modify the direction of DirectionalLights in a sensible way. Unit3D, by comparison (and I like Three.js much better for sure), has angles as the default way of configuring the direction. So I thought I could add it to DirectionalLight as an option. The implementation conserves the radius.

### DIFF
--- a/src/lights/DirectionalLight.js
+++ b/src/lights/DirectionalLight.js
@@ -1,7 +1,7 @@
 import { Light } from './Light.js';
 import { DirectionalLightShadow } from './DirectionalLightShadow.js';
 import { Object3D } from '../core/Object3D.js';
-import { _Math } from './Math.js';
+import { _Math } from '../Math/Math.js';
 
 /**
  * @author mrdoob / http://mrdoob.com/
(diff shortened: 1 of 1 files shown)


EliasHasle:
I have not tested it, though. Don't know how to build from source yet. So I have just tried my best at making it correct. Even if it is not correct, it serves as a starting point for discussion.

looeee:
This looks like it should work, however I would rather add a second directional light type without a target, then we can match the functionality from other apps such as 3DS Max (targeted and untargeted directional lights supported), Unity (untargeted only), Blender (untargeted only as Sun Lamp), Maya (untargeted only) etc. 

The current light is a `TargetDirectionalLight`, we should add a second `RotationalDirectionalLight`.

Mugen87:
Related: #13373

EliasHasle:
Oh, I have been ignoring the target issue. I sincerely thought that DirectionalLight direction was specified only by the position, always pointing toward the origin.

WestLangley:
```javascript
light.position.setFromSpherical( spherical ).add( light.target.position );
```

EliasHasle:
@WestLangley I think so, yes, if target is in same space and the "radius" is computed as distance to target instead of distance to origin.

WestLangley:
I think my suggestion is all that you need and this PR can be closed.

Mugen87:
> I think my suggestion is all that you need and this PR can be closed.

👍

PR #14571, opened 2018-07-27
Quaternion: Added .lookAt()

Mugen87:
This method is a neat shortcut since it allows to directly create a quaternion that orients an object to face towards a specified target. I've often found it annoying to create a temporary matrix object in order ot use `Matrix4.lookAt`. This is now done internally in `Quaternion.lookAt()`.

BTW: That makes the code in `Object3D.lookAt()` a bit nicer :blush: .

Similar to methods in other projects like:

https://github.com/juj/MathGeoLib/blob/master/src/Math/Quat.h#L234-L260 or
https://docs.unity3d.com/ScriptReference/Quaternion.LookRotation.html

### DIFF
--- a/docs/api/math/Matrix4.html
+++ b/docs/api/math/Matrix4.html
@@ -205,9 +205,9 @@ zAxis = (c, g, k)
 		<h3>[method:this identity]()</h3>
 		<p>Resets this matrix to the [link:https://en.wikipedia.org/wiki/Identity_matrix identity matrix].</p>
 
-		<h3>[method:this lookAt]( [param:Vector3 eye], [param:Vector3 center], [param:Vector3 up], )</h3>
+		<h3>[method:this lookAt]( [param:Vector3 eye], [param:Vector3 target], [param:Vector3 up], )</h3>
 		<p>
-			Constructs a rotation matrix, looking from [page:Vector3 eye] towards [page:Vector3 center]
+			Constructs a rotation matrix, looking from [page:Vector3 eye] towards [page:Vector3 target]
 			oriented by the [page:Vector3 up] vector.
 		</p>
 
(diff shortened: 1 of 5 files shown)


WestLangley:
`Quaternion.lookAt()` in this PR does not work for non-camera objects.

As implemented in this PR, you would need to tell the user to swap the order of the arguments to `Quaternion.lookAt()` if the quaternion is going to be applied to an object other than a camera.

`Matrix4.lookAt()` is a method for internal use. It is one of the oldest methods in the library, and was written to be used when orienting a camera.

We already have a look-at method, just as the other libraries do. We just have a slightly-different paradigm.

In three.js, we define our `lookAt()` method at the `Object3D` level, and at the `Geometry` level. This hides implementation details from the user, and it  "just works".

WestLangley:
Also, in the unit tests, it is best to avoid using the library to test against itself. Instead, perform the calculation by hand and test against a known value.

Trivial tests are not particularly useful. You need to test all relevant code paths, which in the math library, can include edge cases.

Mugen87:
The unit test for `Quaternion.lookAt` should just ensure that the same result is produced if you use the "old way" via `Matrix.lookAt()`  and `Quaternion.setFromRotationMatrix()`. Against this backdrop, the test should be sufficient.

I think it's important to understand that `Quaternion.lookAt()` is intended for developers who directly work with quaternions. Like I said before, it's somewhat annoying if you always have to work with a rotation matrix. `Quaternion.lookAt()` hides this effort so you directly have the representation you need for your code.

> Quaternion.lookAt() in this PR does not work for non-camera objects.

As you said before, users need to pass in parameters differently (like in `Object3D.lookAt()`). I've not changed the signature in order to ensure `Matrix4.lookAt()` and `Quaternion.lookAt()` behave the same. The signature of `Matrix4.lookAt()` is somewhat unfortune. The following approach seems better to me:

https://github.com/juj/MathGeoLib/blob/7d94b714ff17b954b1302ff29d040b76b9b40476/src/Math/float3x3.cpp#L616-L677

I could address your concerns if `Quaternion.lookAt()` gets a different signature and uses a new `Matrix3.lookAt()`.

Issue #14476, opened 2018-07-14
Add `DrawGeometry` function to WebGLRenderer


gkjohnson:
Hello! 
THREE provides a really nice, quick approach for setting up a scene but there are times when the default render path isn't ideal. I want to propose adding a function like `DrawGeometry` to the WebGLRenderer API.

### WebGLRenderer.DrawGeometry(<Geometry|BufferGeometry\>, <Material\>, <Matrix4\>, <Camera\>, <Scene\>, <RenderTarget\>, <Options\>)
Draws the geometry to the render target immediately.

#### Geometry|BufferGeometry
The geometry to render.

#### Material
The material to render with.

#### Matrix4
The world matrix to transform the geometry.

#### Camera
The camera to render with.

#### Scene
The scene to render in the context of (for lighting, fog, etc).

#### RenderTarget
The target to render to.

#### Options
A set of draw options including things like `frustumCulled`.

### Use Cases
At a high level this would afford more flexibility to create custom render paths to suit different types of optimizations and scene complexity.

- In cases where you're dealing with tens or hundreds of thousands of meshes in a scene it can be beneficial to save transform information in a minimal representation to save memory (an array of matrices, for example). This can't be done all the easily at the moment because THREE requires Object3Ds and Meshes to render.
- If a set of visible geometry is already known (through a custom octree frustum culling implementation or precomputed visible sets) there's no reason to have THREE iterate over all geometry in the scene and it can be much slower to do so. This lets you control what gets drawn and iterated over.
- Some effects are just convenient to implement with a custom draw and don't need an object in the scene hierarchy.
- You may want to draw for only X milliseconds during a frame -- this would let you measure how long you've spent drawing so far. 

In a couple of these cases the solution at the moment is to remove everything (or almost everything) from the scene and rebuild it specifically so THREE doesn't doesn't spend time iterating over everything, which can take time in itself.

Thoughts?

Mugen87:
Why not using [WebGLRenderer.renderBufferDirect](https://threejs.org/docs/#api/renderers/WebGLRenderer.renderBufferDirect) for this?

Mugen87:
> How are light uniforms set on the material in this case? Do they need to be updated manually?

No, that happens within `renderBufferDirect()`. But you need to call at least `.compile()` so lights and shadows are correctly prepared. Normally you use `renderBufferDirect()` for special objects rendered along the normal workflow. Lensflares are a good example for this.

Mugen87:
[...]

PR #14578, opened 2018-07-30
Add WebGLRenderer.DrawGeometry and associated functions

gkjohnson:
Hey! This PR adds three functions to `WebGLRenderer` to address https://github.com/mrdoob/three.js/issues/14476 (and probably others) and allow you to have more control over materials and how objects get drawn outside of the `render` function because `renderBufferDirect` wasn't sufficient in all cases. There's a lot to learn and I'm pretty new to poking around the renderer code, so this PR might be more for discussion and feedback:

#### SetLightingState( scene, camera )
Rolls up the lighting information in the scene and caches it in the WebGLRenderState for that scene / camera combination, much like the beginning of `render`. Calling this with `null` resets the state.

#### SetMaterial( material )
Sets the material to be used when `DrawGeometry` is used next and invalidates the currently set material so the material uniforms are guaranteed to be updated on the next draw. This allows for the same material to be reused on multiple objects and modified between draws if desired.

#### DrawGeometry(  camera, fog, geometry, object, group )
Draws the geometry with the last set material using the world matrix on `object` and ensures that the geometry is read by calling `objects.update`. Note `object` _must_ be of a drawable type like `Mesh`.

---

These functions allow you to draw a scene (or draw more after rendering) like so:
```js
var geometry, object, material, camera, scene;

// ...

renderer.clear();
renderer.setLightingState( scene, camera );
for ( var i = 0; i < 100; i ++ ) {
      
   object.position.set( Math.random(), Math.random(), Math.random() );
   object.updateMatrixWorld( true );
   material.color.set( Math.random(), Math.random(), Math.random() );
   
   renderer.setMaterial( material );
   renderer.drawGeometry( camera, null, geometry, object, null );

}

// increment the frame so geometry will get updated if needed. This
// can be called between draws to modify and draw the same geometry multiple
// times in a single frame.
renderer.info.reset();

```

`render()` does _not_ need to be called unless you want shadows to be rendered (which I suppose you could render to their maps directly using this?)

---

#### Comments & Questions
- With these functions, `compile` and `renderBufferDirect` may not need to be in the public API.
- Ultimately what I'd like to see is enough of the functionality of `WebGLRenderer` be cleanly exposed so you can forego using the `render()` function entirely allow for a really customize-able render path. I know I've seen this discussed in issues in the past but I'm having trouble finding them now.
- I'm less familiar with using VR with THREE so I don't know exactly how this will work in a VR case. Presumably you'd need to call `renderer.vr.submitFrame()` manually?
- ~~I've noticed that updating individual vertices in a mesh isn't working when not using `render()` so I'll have to track that down. Geometry may need it's own type of cache-invalidating function.~~
- And lastly (this may be related to some of the lighting-hash issues I've seen): is there a reason that `WebGLRenderStates` are cached even though they're reinitialized and recreated on every render? Why not remove the cache entirely?

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -677,10 +677,18 @@ function WebGLRenderer( parameters ) {
 
 	};
 
+	var currentMaterial = null;
 	var uploadMesh = new Mesh();
 	var emptyState = new WebGLRenderState();
 	emptyState.init();
-	this.drawGeometry = function ( camera, fog, geometry, material, object, group ) {
+	this.setMaterial = function ( material ) {
+
+		_currentMaterialId = - 1;
+		currentMaterial = material;
+
+	};
+
+	this.drawGeometry = function ( camera, fog, geometry, object, group ) {
 
 		if ( vr.enabled ) {
 
@@ -693,8 +701,6 @@ function WebGLRenderer( parameters ) {
 			currentRenderState = emptyState;
 
 		}
-
-		_currentMaterialId = - 1;
 
 		if ( camera !== _currentCamera ) {
 
@@ -713,7 +719,7 @@ function WebGLRenderer( parameters ) {
 
 		object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
 		object.normalMatrix.getNormalMatrix( object.modelViewMatrix );
-		this.renderBufferDirect( camera, fog, geometry, material, object, group );
+		this.renderBufferDirect( camera, fog, geometry, currentMaterial, object, group );
 		state.setPolygonOffset( false );
 
 		if ( currentRenderState === emptyState ) {
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -34,7 +34,7 @@ import { WebGLObjects } from './webgl/WebGLObjects.js';
 import { WebGLPrograms } from './webgl/WebGLPrograms.js';
 import { WebGLProperties } from './webgl/WebGLProperties.js';
 import { WebGLRenderLists } from './webgl/WebGLRenderLists.js';
-import { WebGLRenderStates } from './webgl/WebGLRenderStates.js';
+import { WebGLRenderStates, WebGLRenderState } from './webgl/WebGLRenderStates.js';
 import { WebGLShadowMap } from './webgl/WebGLShadowMap.js';
 import { WebGLState } from './webgl/WebGLState.js';
 import { WebGLTextures } from './webgl/WebGLTextures.js';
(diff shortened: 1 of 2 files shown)


takahirox:
PR shouldn't include `build/three.js`.

gkjohnson:
I was able to take a look at this again and found that `renderer.info.reset()` needs to be called in order for geometry to be properly updated if it's modified.

@mrdoob @Mugen87 Any thoughts on this PR or similar new functions? Is there anything I can do to help move it forward?

Thanks!

cnstoba:
I was looking for a solution to this problem. Your's works really well for me, thanks! I think it would be nice to be able to pass an array of geometries/meshes and maybe a THREE.Group/THREE.Group[ ] to the drawGeometries function.

Mugen87:
Sorry for the late feedback but honestly I was never a fan of the proposed API. I think signatures like `setLightingState( scene, camera );` and `renderer.drawGeometry( camera, null, geometry, object, null );` are not really user-friendly. In the meanwhile, I personally interpret `renderBufferDirect()` as no part of the API which is intended for users. We use it in the examples only once (not even in the HTML but in the `Lensflare` implementation).

I think `three.js` should focus on the idea to render a _scene graph_ with a _camera_. If devs need more control, I think modifying `WebGLRenderer` is the better apporach.

gkjohnson:
Thank you for the feedback @Mugen87!

This PR (and associated issue) was made quite awhile ago and now that I've spent more time contributing to the project maybe I can explain what I want more succinctly and propose something smaller to addressing my needs.

> I think three.js should focus on the idea to render a scene graph with a given camera. If devs need more control (for whatever exotic reasons), I think advising to modify WebGLRenderer is the better approach. 

To be blunt asking anyone to fork the renderer and modify it is a bit absurd and something I don't feel is reasonable for a production project that depends on updates to three.js. Ultimately original intent was to make it easier for people to write their own renderer to suit their needs. Three.js has some really nice abstractions and if WebGLRenderer were broken out into some more bite sized pieces then it could make writing your own renderer more simple and be extremely powerful. The way I see it the renderer should be made up of some publicly available atomic functions such as `setLights`, `setMaterial`, `drawGeometry`, `updateUniform`, etc that could be composed into a custom renderer to suit more "exotic" use cases.

The issue that's really biting me right now is the inability to control the draw order of a set of meshes in a scene. I've tried `Group.renderOrder` to control the render order of specific sets of objects but it's extremely painful and doesn't work as expected (which is elaborated on in other issues). Rendering another scene (or multiple) with the objects that I need to render in a specific order might work but there's no way to simply use the lighting state from another scene. If I want to draw an object or scene after my main scene has rendered I have to copy all lights to the new scene and keep them in sync.

Maybe some simpler API change with less reaching changes might look like so?

```js
WebGLRenderer.render(
    scene : Scene | Array<Object3D>,
    camera : Camera,
    lights? : Array<Lights>
);
```

This would let me render a scene or set of objects more cleanly (by passing an array of objects to render) and you can specify an array of lights you'd like to use.

Mugen87:
It was always the goal of the project to keep the API of `WebGLRenderer` lean. Providing more atomic render functions will blow up the API which is not something I would vote for. Keeping the API simple makes it also easier to abstract WebGL more effectively. Something like `updateUniform()` is  clearly counterproductive in that regard. This can become beneficial if we want switch to other renderers e.g. `WebGPURenderer`  in the future.

TBH, I did not have the chance to dive into #18176 yet. Can you maybe demonstrate the issue in a live example? I guess that will make it easier for me and @mrdoob to understand the current restrictions.

gkjohnson:
@Mugen87  Sorry for the delay! Hope you had a good holiday. I've followed up on the other PR like you've suggested.

> Providing more atomic render functions will blow up the API which is not something I would vote for.

I understand but I'm trying to find an approach that will allow an easy API for newcomers while providing options for advanced projects to make a more customized renderer without the suggestion being to fork source and modify the source. There must be a middle ground here, right?

> Keeping the API simple makes it also easier to abstract WebGL more effectively. Something like updateUniform() is clearly counterproductive in that regard. This can become beneficial if we want switch to other renderers e.g. WebGPURenderer in the future.

I agree with the idea here but I disagree that the functions abd ideas I've suggested are counterproductive to that. They rely on core constructs like geometry, lights, materials, and uniforms (which are needed for ShaderMaterial). I wouldn't expect any of these fundamental objects to go away with WebGPURenderer -- and if they do it means a massive change for the library as a whole anyway.

I've also suggested a simpler change of adding an override array of lights to the `render` function signature so that multiple draws with the same lighting state can be used across multiple scenes. Do you have any thoughts on that? Or maybe something like `Scene.overrideLights` like `overrideMaterial` is more agreeable?

Issue #14576, opened 2018-07-29
WebGLLights: Inefficient string concatenations


oguzeroglu:
https://github.com/mrdoob/three.js/blob/8c6d24b0876f0f7d23aa60be55d6570f4ddb4d6b/src/renderers/webgl/WebGLLights.js#L319

Related PR: https://github.com/mrdoob/three.js/pull/14568

This line is executed 60 times per second thus produces garbage as seen on Chrome's Allocation Profiler: 

![ekran resmi 2018-07-29 12 28 20](https://user-images.githubusercontent.com/15003836/43365615-14a75d68-9338-11e8-828b-34df72122b1c.png)

The hash value of WebGLLights should be an object instead of concatanated string:

```
hash: {
stateID: -1,
directionalLength: -1,
pointLength: -1,
spotLength: -1,
rectAreaLength: -1,
hemiLength: -1,
shadowsLength: -1
}
```

In the WebGLRenderer.js the values should be compared as:

```
} else if ( materialProperties.lightsHash.stateID !== lights.state.hash.stateID ||
        materialProperties.lightsHash.directionalLength !== lights.state.hash.directionalLength ||
	materialProperties.lightsHash.pointLength !== lights.state.hash.pointLength ||
        materialProperties.lightsHash.spotLength !== lights.state.hash.spotLength ||
	materialProperties.lightsHash.rectAreaLength !== lights.state.hash.rectAreaLength ||
	materialProperties.lightsHash.hemiLength !== lights.state.hash.hemiLength ||
	materialProperties.lightsHash.shadowsLength !== lights.state.hash.shadowsLength ) {
```

Since JS assigns objects by reference materialProperties.lightsHash should be updated like this to avoid potenial bugs caused by string to object switch:

```
materialProperties.lightsHash.stateID = lights.state.hash.stateID;
materialProperties.lightsHash.directionalLength = lights.state.hash.directionalLength;
materialProperties.lightsHash.pointLength = lights.state.hash.pointLength;
materialProperties.lightsHash.spotLength = lights.state.hash.spotLength;
materialProperties.lightsHash.rectAreaLength = lights.state.hash.rectAreaLength;
materialProperties.lightsHash.hemiLength = lights.state.hash.hemiLength;
materialProperties.lightsHash.shadowsLength = lights.state.hash.shadowsLength;
```
instead of:

`materialProperties.lightsHash = lights.state.hash`

Any ideas/opinions?

Mugen87:
[...]

PR #14588, opened 2018-07-30
Removed inefficient string concats from WebGLLights

oguzeroglu:
This PR fixes the issue mentionned in #14576

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1548,7 +1548,7 @@ function WebGLRenderer( parameters ) {
 		// store the light setup it was created for
 		if ( materialProperties.lightsHash === undefined ) {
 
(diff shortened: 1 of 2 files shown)


haroldiedema (approved):
Looks fine to me. Although if it isn't too much trouble, I would like to see a performance comparison between the current dev branch and this, just to see how much difference it really makes.

oguzeroglu:
@haroldiedema 
I don't think it is a huge improvement but still most of the 3D apps should be careful about the memory management as the the GC activity may drop the FPS from 60 to around 57/58 from time to time. This is no big deal but why not do our best for solid 60 FPS?

Like in the PR #14025 it is not a huge improvement for the performance that you can easily measure but still, it's an improvement.

mrdoob:
Thanks!

PR #14596, opened 2018-07-31
Examples: Removed webgl_objects_update.html

### LINKED ISSUE #14583: webgl_objects_update.html example
What three.js feature(s) do we want this demo to demonstrate?

What should be added to, or removed from, this demo?

https://threejs.org/examples/webgl_objects_update.html

(Note that It is not disposing of geometries properly.)

Mugen87:
Fixed #14583


mrdoob:
Thanks!

PR #14597, opened 2018-07-31
UnrealBloomPass memory optimization

oguzeroglu:
In this PR a simple modification is made inside the UnrealBloomPass to reduce possible GC activity caused by this line:

https://github.com/mrdoob/three.js/blob/5ce33263bf9c08ec981f71f02d1be15a416fb1e1/examples/js/postprocessing/UnrealBloomPass.js#L192

A new THREE.Color instance should not be created inside the render loop (60 times a second), it should be created once and be reused later in my opinion.

### DIFF
--- a/examples/js/postprocessing/UnrealBloomPass.js
+++ b/examples/js/postprocessing/UnrealBloomPass.js
@@ -1,6 +1,6 @@
 /**
  * @author spidersharma / http://eduperiment.com/
- * 
+ *
  * Inspired from Unreal Engine
  * https://docs.unrealengine.com/latest/INT/Engine/Rendering/PostProcessEffects/Bloom/
  */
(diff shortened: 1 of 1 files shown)


examples/js/postprocessing/UnrealBloomPass.js
```diff
@@ -13,6 +13,9 @@ THREE.UnrealBloomPass = function ( resolution, strength, radius, threshold ) {
 	this.threshold = threshold;
 	this.resolution = ( resolution !== undefined ) ? new THREE.Vector2( resolution.x, resolution.y ) : new THREE.Vector2( 256, 256 );
 
+	// create color only once here, reuse it later inside the render function
+	this.reusableColor = new THREE.Color( 0, 0, 0 );
```
Mugen87: I would call it just `clearColor`.

Mugen87 (changes requested):

examples/js/postprocessing/UnrealBloomPass.js
oguzeroglu: Done

Mugen87 (approved):

mrdoob:
Thanks!

PR #14224, opened 2018-06-05
Introduce usage of Vertex Array Objects (VAO)

Mugen87:
This PR introduces the usage of [Vertex Array Objects](https://blog.tojicode.com/2012/10/oesvertexarrayobject-extension.html) (VAO), a performance primitive of WebGL (see https://github.com/mrdoob/three.js/issues/8705). A new class `WebGLBindingStates` manages now the buffer bindings and pointers for a single entity. VAO can only be used via an extension, the support though is really high (96% according to http://webglstats.com/webgl/extension/OES_vertex_array_object). Besides VAOs are part of the core of WebGL2.

If the extension is not available, `three.js` renders the old way. It also uses the old render path if morph targets are present, since it's pretty complicated to manage the flexible state of morph target attributes with VAOs. Maybe we find a solution for this in the future.

The amount of changes is actually manageable. Most changes are caused by refactoring, since certain code like `setupVertexAttributes()` moved from `WebGLRenderer` to `WebGLBindingStates`. I've tested most examples and they still work, although intensive testing is important for such a change. 

Also see https://github.com/mrdoob/three.js/issues/8705#issuecomment-296177105

PR #14600, opened 2018-07-31
VAO support

takahirox:
(I made a new PR because I've broken #14590 by my misoperation)

This PR introduces Vertex Array Object (VAO) into Three.js.

- Benefit
  - Performance improvement. We can skip `setupVertexAttributes()`, except for initial call, which is one of the CPU hot-spot functions.

- Changes
  - Add `WebGLBindingStates` which manages VAO and vertex attributes.
  - Add `.version` to `BufferGeometry`. Counting up when its attribute is added or removed. `WebGLBindingStates` sees geometry's version and decides whether to update vertex attributes.
  - Move the following functions to `WebGLBindingStates`. Now managing vertex attributes per VAO.
    - `initAttributes()`, `disableUnusedAttributes()`, `enableAttribute()`, and `enableAttributeAndDivisor()` of `WebGLBindingStates`.
    - `setupVertexAttributes()` of `WebGLRenderer`
  - (Update) Move index buffer update into `WebGLBindingStates`
  - (Update) Add/Remove morph attributes only when necessary in `WebGLMorphtargets`
  - (Update) Dispose VAOs in `renderer.dispose()`
  - (Update) Release VAO when geometry/program using it is released

- Discussion
  - ~How to dispose VAO. Now Disposing when `renderer.dispose()` is called. Should we dispose in `onGeometry/MaterialDispose()`?~
  - Do we need fine-grain attribute update detection, like attribute's `.stride`, `.itemSize`, `.offset`?
  - How should the data structure of cache keeping VAOs be? I create VAO per set of `geometry`, `program`, and `material.wireframe`. Then I do like `bindingStates[ geometry.id ][ program.id ][ material.wireframe === true ] = { object: vao, ... }`.

Feedbacks and bug reports are welcome. Let me know if this change breaks existing examples/apps.

Inspired by #14224


src/renderers/webgl/WebGLBindingStates.js
```diff
@@ -0,0 +1,393 @@
...
+		var state = getBindingState( geometry, program, material );
+
+		if ( currentState !== state ) {
+
+			currentState = state;
+			bindVertexArrayObject( currentState.object );
+
+		}
+
+		var updateBuffers = hasMorphTarget;
```
takahirox: Update vertex attribute if morph targets are present so far because current morph targets rendering path requires attributes update. We could find solution in the future.

src/renderers/webgl/WebGLBindingStates.js
```diff
@@ -0,0 +1,393 @@
...
+				currentState.program = program.id;
+				currentState.wireframe = wireframe;
+
+				updateBuffers = true;
+
+			}
+
+		}
+
+		if ( updateBuffers ) {
```
takahirox: Following the discussion https://github.com/mrdoob/three.js/pull/14590#pullrequestreview-141652383

> What happens if the user change a attribute data? We have to update the VAO. It would be a bad usage of the library maybe.

Regarding `BufferGeometry`'s `stride/offset/count`, I don't wanna fine-grain update check for the performance and simple code. So I'm thinking of a workaround that is asking users to explicitly call `geometry.version ++;` if users update them on the fly.

@DavidPeicho @Mugen87

Mugen87: An other more strict option is to think of these properties as immutable. If the user wants to change them, it's necessary to create a new attribute and add it via `BufferGeometry.addAttribute()`. In this case, `geometry.version` would be updated automatically. 

In any event, the library would lose flexibility. Especially since we assume `.addAttribute()` is actually used. Right now, you can change `BufferGeometry.attributes` directly without breaking anything.

src/renderers/WebGLRenderer.js
```diff
@@ -1038,9 +883,7 @@ function WebGLRenderer( parameters ) {
 
 		// reset caching for this frame
 
-		_currentGeometryProgram.geometry = null;
-		_currentGeometryProgram.program = null;
-		_currentGeometryProgram.wireframe = false;
+		bindingStates.reset();
```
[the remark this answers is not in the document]
takahirox: I moved indexBuffer update in VAO (`WebGLBindingStates`) and seems working, thanks.

By moving it, it seems ok not to unbind VAO in the beginning of `.render()` now.

takahirox:
This is the change in core. I wanna test this change more. I'd be pleased if you check that this change doesn't break your apps.

takahirox:
Let me share my VAO performance benchmark tests.

https://rawgit.com/takahirox/three.js/VAOSupportTest/examples/webgl_vao_test.html (With VAO)
https://rawgit.com/takahirox/three.js/VAOSupportTest/examples/webgl_non_vao_test.html (Without VAO)

They render meshes (as default 100 meshes) which don't share geometry. Available vertex attributes per geometry are three as `posiition`, `normal`, and `uv`. And plus, index buffer.

You can controls the number of meshes by adding `?num` into the URL, like

https://rawgit.com/takahirox/three.js/VAOSupportTest/examples/webgl_vao_test.html?10

You can compare the fpses and CPU time number between the test. And also you can profile with browser dev tools. My impression is `gl.bindVertexArray()` is slower than I expected... I'll look into more.

Mugen87:
I don't see any differences in the framerate with low geometry count. But if I define 10.000 unique meshes with unique geometries, I have the following FPS numbers on my iMac (4 GHz Intel Core i7) with Chrome:

| Mesh Count   | FPS (VAO)       | FPS (Non VAO)  |
| ------------- |:-------------:| -----:|
| 2.500      | 60 | 60 |
| 5.000      | 48  | 41 |
| 7.500      | 31  | 27 |
| 10.000 | 24  |20 |

takahirox:
Cool. How is Elapsed CPU time on the small number geometries? Do you see the performance gain?

Mugen87:
A sample with 1000 geometries (both examples produce 60 FPS):

With VAO:
image

Without VAO:
image

Mugen87:
Here you can see the performance data of my Pixel 1 with Chrome:

| Mesh Count   | FPS (VAO)       | FPS (Non VAO)  |
| ------------- |:-------------:| -----:|
| 1000      | 60 | 51|
| 2.500      | 29 |23 |
| 5.000      | 15  | 12|
| 7.500      | 10  |<10 |
| 10.000 | <10  |<10|

takahirox:
Thanks for the test. I think we can say we'll see the performance benefit from VAO. (At least no degradation).

One concern is memory consumption. Currently in this PR I create an object + VAO per a set of geometry + program + material.wireframe. I see 1.8MB larger heap memory consumption (19.1MB with VAO and 17.3MB without VAO) with 2000 geometries on my Windows + Chrome.

Any ideas for reducing memory consumption?

BTW, no limit to number of VAO creation?

simon-paris:
Hi, is it possible to allow some user control for this? I don't think it'll perform better in every case, I can imagine wanting to disable it for some geometries.

And secondly, can users be allowed to share vaos between similar geometries? I've been able to improve performance by about 10% overall by sharing vaos between all my static geometries.

Mugen87:
I think the library should always use VAO if possible and not allow user control for this. I would also not expose VAOs to the user in any way in order to keep things simple. AFAIK, other engines like Babylon.js do this as well.

looeee:
>  I don't think it'll perform better in every case

@simon-paris can you provide any examples in which performance is worse? If not then I'm with @Mugen87 on this, we should aim to keep things as simple as possible from the user's perspective.

simon-paris:
I can think of a few:

 * Objects with dynamic geometry
 * Objects rendered for one frame
 * Objects where the vertexAttribPointer calls would be no-ops
 * Objects with morph targets

An option on BufferGeometry would be pretty nice.

simon-paris:
> Any ideas for reducing memory consumption?

These 3 Uint8Arrays are 184 bytes each: https://github.com/mrdoob/three.js/pull/14600/files#diff-d1d3e9b3d0928ce5d609d6e4f2cb51a5R149

takahirox:
Is that the number from heap memory snapshot?

takahirox:
About control option, I think it very depends on how big penalty(performance and memory) is. Basically I'd like to try to make the penalty small and not to expose control option. So evaluation and optimization are first.

takahirox:
About sharing VAOs, I'm wondering if we can automatically do inside...  As I mentioned we need some ideas to reduce memory consumption.

simon-paris:
> Is that the number from heap memory snapshot?
Yup

> About sharing VAOs, I'm wondering if we can automatically do inside
It'd need to automatically detect geometries with identical attributes. That seems like a bit much though.

mrdoob:
> It'd need to automatically detect geometries with identical attributes. That seems like a bit much though.

Shouldn't we consider that a user error? We currently do that for materials but we always recommend reusing materials for saving memory.

simon-paris:
Yeah, to share automatically it'd need to do something similar to what it does for materials. Imo, that'd be too complex. What I've been doing internally is just exposing the key used to lookup the vao [here](https://github.com/mrdoob/three.js/pull/14600/files#diff-d1d3e9b3d0928ce5d609d6e4f2cb51a5R109) like `geometryMap = bindingStates[ geometry.vaoKey || geometry.id ]; `.


> Basically I'd like to try to make the penalty small as much as possible

I measured the penalty by making it always call setupVertexAttributes, like it does the first time an object is rendered or if the object has morph targets.

 * VAOs enabled: 12% faster
 * VAOs enabled, but always call setupVertexAttributes: 15% slower
 * VAO shared between all geometries: 22% faster

takahirox:
Hm, 15% performance degradation isn't good...

mrdoob:
> I wanna keep trying small the penalty but it may take time. 

No rush. I prefer to keep experimenting rather than adding experiments to the renderer.

Mugen87:
> How to dispose VAO. Now Disposing when renderer.dispose() is called. Should we dispose in onGeometry/MaterialDispose()?

Yes, I think we should try to delete VAOs in `onGeometryDispose()`. Although I'm not super familiar with the internals of babylon.js, I've noticed that the engine deletes VAOs when `dispose()` is called on a geometry object. Deleting VAOs only when `renderer.dispose()` is called seems not sufficient.

src/renderers/webgl/WebGLBindingStates.js
Mugen87: In the meanwhile, I think it's okay to accept such a change. I personally think the performance gain due to VAO usage is more important than the mentioned flexibility with buffer attributes.

Mugen87:
> BTW, no limit to number of VAO creation?

AFAIK, there is no limit.

> I see 1.8MB larger heap memory consumption (19.1MB with VAO and 17.3MB without VAO) with 2000 geometries on my Windows + Chrome.

I think a slightly higher memory footprint is acceptable in order to achieve the mentioned performance gain.

takahirox:
I'm gonna start to work on VAO again soon. Right now my main concern is performance penalty for object with morph targets...

Mugen87:
@takahirox If you need support in any form (e.g. testing), I'm glad to help out 😊. I really would like to see VAO in the library and we already made some presentable progress.

src/renderers/webgl/WebGLBindingStates.js
takahirox: I realized I can remove `hasMorphTarget` because geometry version increments in `WebGLMorphTargets` when morph attributes are added/removed.

And I optimized morph attributes addition/removal in `WebGLMorphTargets`. They're added/removed only when necessary. So I accomplished to narrow the case where VAO needs to be updated even with morph targets.

src/renderers/webgl/WebGLMorphtargets.js
```diff
@@ -2,6 +2,12 @@
  * @author mrdoob / http://mrdoob.com/
  */
 
+function numericalSort( a, b ) {
```
takahirox: More appropriate function name idea?

src/renderers/webgl/WebGLMorphtargets.js
```diff
@@ -13,6 +19,14 @@ function WebGLMorphtargets( gl ) {
 	var influencesList = {};
 	var morphInfluences = new Float32Array( 8 );
 
+	var workInfluences = [];
```
takahirox: Any better name and algorithm ideas?

takahirox:
Updated. Previously `setupVertexAttributes` was always called for geometry with morph targets. Then I optimized `WebGLMorphtargets`. Now morph attributes (attributes whose names are morphTarget i or morphNormal i) are added/removed only when necessary, and only when they're added/removed `setupVertexAttributes` is called. So I accomplished to narrow the case where `setupVertexAttributes` is called.

src/renderers/webgl/WebGLBindingStates.js
```diff
@@ -0,0 +1,479 @@
...
+
+		return state;
+
+	}
+
+	function createBindingState( vao ) {
+
+		var newAttributes = [];
+		var enabledAttributes = [];
+		var attributeDivisors = [];
```
takahirox: I replaced three `Uint8Array`s with regular arrays for less memory consumption because there seems being no big reason why `Uint8Array`s are used.

takahirox:
I added two changes
- Release VAO when `geometry`/`material`referring to it is released
- Replacing `Uint8Array` with regular arrays for less memory consumption

I think this PR is good to go now (unless there's any bugs).

takahirox:
One concern left is this change looses some flexibilities.

1. Users can't change attribute data on the fly
2. Users can't directly add/remove `attribute` to/from `geometry.attributes` like `geometry.attributes[ something ] = attribute;` or `delete geometry.attributes[ something ]`

But I think it may be ok because

For 1. there is workaround, we may ask users to explicitly call `geometry.version ++;`. Another option is we may think of attribute is immutable. If users wanna change the data, we could recommend them to create a new one and replace with `geometry.add/removeAttribute()`.

For 2. I think using `geometry.add/removeAttribute()` is the right API. We can encourage using the `.add/removeAttribute()` instead of directly touching `geometry.attributes`.

See https://github.com/mrdoob/three.js/pull/14600/files#r206937767

takahirox:
Confirmed the examples keep working fine. I've removed "[WIP]" from the title of this PR.

takahirox:
I realized I didn't take account of dynamic attribute. I'm looking into...

takahirox:
Sorry my bad, seems no problem. Actually `webgl_geometry_dynamic.html` looks working fine.

takahirox:
Any plans when we will merge this? Or is something blocking this PR?

Mugen87:
What about https://github.com/mrdoob/three.js/pull/14600#discussion_r206937767

I think there is still no agreement on this point. What happens when users modifies the attribute and index configuration without using `addAttribute()`, `removeAttribute()` and `setIndex()`?

takahirox:
I think it may be good to make `geometry.attributes` private and ask users to use accessing method not only for VAO but also clear API. Currently we have two ways to access `attributes/index`, directly accessing `geometry.attributes/index` and using access methods. I think only one way is clearer API. But I know this API change breaks a lot of codes. So I'd like to open an new issue to discuss.

Let me know if anyone has better solutions without breaking the compatibility.

Mugen87:
Making a separate issue is a good idea. There is also another issue that might be related to VAOs, see https://github.com/mrdoob/three.js/issues/14730. Does the current VAO implementation works if you resize an attribute via `BufferAttribute.setArray()`? If not, that's another reason to remove the method an consider `BufferAttribute` as immutable in their structure.

takahirox:
That works because array is updated to GPU in `projectObject()` which `bindingState` doesn't skip.

https://github.com/mrdoob/three.js/blob/r103/src/renderers/WebGLRenderer.js#L1313
https://github.com/mrdoob/three.js/blob/r103/src/renderers/webgl/WebGLAttributes.js#L66-L96

And count is used when drawing which `bindingState` also doesn't skip.

https://github.com/mrdoob/three.js/blob/r103/src/renderers/WebGLRenderer.js#L773-L781

takahirox:
These are the properties runtime change of which are ignored with the current VAO implementation, ever I'm aware of. I don't think they can be changed runtime in general use case. So I guess it may be ok to make them immutable. Let me know if there's any use cases where they can be changed runtime.

`BufferAttribute`: `.normalized`, `.itemSize`, `.type`,
`InterleavedBufferAttribute`: `.normalized`, `.itemSize`, `.offset`, `.data.stride`
`InstancedBufferAttribute/InstancedInterleavedBuffer`: `.meshPerAttribute`.

takahirox:
Opened an issue #16132 to discuss `.attributes/index` properties access scope.

takahirox:
Testing the performance again. Even without `geometry.version`, caching attribute to detect update may be good enough.

```javascript

updateBuffer = needsUpdate( geometry );

if ( updateBuffer ) {

    saveCache( geometry );

    setupVertexAttributes( ... );

    ....

}

function needsUpdate( geometry ) {

    var cache = currentState.attributes;
    var attributes = geometry.attributes;

    if ( Object.keys( cache ).length !== Object.keys( attributes ).length ) return true;

    for ( var key in attributes ) {

        if ( cache[ key ] !== attributes[ key ] ) return true;

    }

    return false;

}

function saveCache( geometry ) {

    var cache = {};
    var attributes = geometry.attributes;

    for ( var key in attributes ) {

        cache[ key ] = attributes[ key ];

    }

    currentState.attributes = cache;

}
```

Test I posted above on my Chrome.

VAO + geometry.version: 42fps
VAO + attribute cache: 40fps:
No VAO: 36fps:

`geometry.version` is faster but the performance gain of attribute cache looks also good. Attribute cache doesn't require big `geometry.attribute` API change. Thinking to go with attribute cache, and to start to think `geometry.version` if we need further better performance.

takahirox:
Updated. Reflecting https://github.com/mrdoob/three.js/pull/14600#issuecomment-482799338 and now this PR requires only https://github.com/mrdoob/three.js/pull/14600#issuecomment-478768913. No longer private `geometry.attributes/index` is necessary (so far).

takahirox:
Made draft PR #16264 to show how the change for https://github.com/mrdoob/three.js/pull/14600#issuecomment-478768913 looks like

takahirox:
Closing for now because I made VAO without any breaking existing APIs PR #16287

PR #14605, opened 2018-08-01
Color: Added .toString()

trusktr:
A standard JavaScript method.

Useful with other APIs that serialize/deserialize using `toString`.

Example based on [SkateJS](http://skatejs.netlify.com) props:

```js
export default
function createPropType(Type, override = {}) {
    return {
        attribute: { source: true },
        coerce: val => val instanceof Type ? val : new Type(val),
        default: new Type,
        deserialize: val => new Type(val),
        serialize: val => val.toString(),
        ...override,
    }
}
```

```js
import * as THREE from 'three'
import createPropType from './createPropType'

skateTHREEColorProp = createPropType(THREE.Color)
```

## TODO

- [ ] Doc update

### DIFF
--- a/src/math/Color.js
+++ b/src/math/Color.js
@@ -452,6 +452,12 @@ Object.assign( Color.prototype, {
 		return 'rgb(' + ( ( this.r * 255 ) | 0 ) + ',' + ( ( this.g * 255 ) | 0 ) + ',' + ( ( this.b * 255 ) | 0 ) + ')';
 
 	},
+	
+	toString: function () {
+
+		return '#' + this.getHexString();
+
+	},
 
 	offsetHSL: function () {
 


WestLangley:
Yet, depending on the use case, it could be
```javascript
'0x' + color.getHexString();
```

Mugen87:
Um, but values like `0xffffff` are numbers, right? You normally don't work with them as strings. Against this backdrop, I think the PR is okay like that.

src/math/Color.js
```diff
@@ -452,6 +452,12 @@ Object.assign( Color.prototype, {
 		return 'rgb(' + ( ( this.r * 255 ) | 0 ) + ',' + ( ( this.g * 255 ) | 0 ) + ',' + ( ( this.b * 255 ) | 0 ) + ')';
 
 	},
+
```
Mugen87: Please remove these trailing spaces in line 455.

Mugen87 (changes requested):

WestLangley:
I'm not liking the proposed method name. I do not think it is sufficiently descriptive.

IMO, `.getHexString()` should be modified to return proper hexadecimal syntax:  `#0000ff`.

Optionally, `.getStyle()` could be renamed to `.getRGBString()`, although I understand why it was named the way it was.

The method names `.getHexString()` and `.getRGBString()` would be an improvement, IMO.

Mugen87:
Closing. Users can still monkey path `Color` with a custom `toString()` method.

Issue #11340, opened 2017-05-15
Feature request: a method to disable TransformControl axes


brmscheiner:
I wish there was an easy way to prevent users from moving an object about a particular axis. For example, it would be nice to say 

```
var controls =  new THREE.TransformControls(camera, domElement);
controls.disableAxis('x');
```

Then the transformControls would have draggable arrows pointing in the Y and Z directions, but not the X direction. Thanks for your consideration!

makc:
TransformControls are not maintained in this repo - last time I asked to fix the bug in it, I was told by the author that he can't do that, because his original TransformControls evolved far beyond three.js version: #9312

brmscheiner:
It does look like @arodic has done a lot with the transform controls for ree.js, but also contributors here have made relatively recent contributions to the transform controls (including @mrdoob [late last year](https://github.com/mrdoob/three.js/commits/dev/examples/js/controls/TransformControls.js)), so I don't think it's fair to say that the transform controls aren't maintained in this repo. Also note that the issue you created is still open - my guess is that if it was not related to this project it would have been closed already.

If people think this feature is a good idea, I would be happy to take a stab at it myself, although I am a little bit intimidated by the sheer size of the file...

tentone:
@brmscheiner I have a split version of the TransformControl in my nunuStudio repository and i have worked on them a bit already, i might be able to take a look at this on the weekend, it should be fairly easy to do this 👍 

I used to have similar ones that i wrote that are very simple and easy to change  (https://github.com/tentone/nunuStudio/tree/0fb60a9cc7a0c3d07d90bf7cb644054831b79c0f/editor/tools) but the ones made by arodic are a lot better.

pailhead:
> although I am a little bit intimidated by the sheer size of the file...

I believe this is what may have deterred the original author to keep working on this.

brmscheiner:
@tentone nunuStudio is a very cool project! If you do get the chance to take a look I would be massively appreciative, if not just let me know.

brmscheiner:
Hi @tentone, any chance to look into this?

tentone:
Hi @brmscheiner sorry i still haven't got a change to take a look at this but as soon as i can i will work on it (i also need this for my project).

Sorry for the delay!

brmscheiner:
Dope, no problem!

brmscheiner:
Hey @tentone just wanted to circle back again :)

brean:
[...]

PR #14619, opened 2018-08-02
Transform controls v2 stopPropagation patch

arodic:
@trusktr @but0n @brettkromkamp @AustinDuvin reported issues with new TransformControls not playing nice with OrbitControls and it appears to be due to removed `event.stopPropagation()` on mousedown and mousemove.

This patch adds **dragging-changed** event that lets developers configure other controls appropriately. See discussion for details.

### DIFF
--- a/examples/js/controls/TransformControls.js
+++ b/examples/js/controls/TransformControls.js
@@ -104,11 +104,9 @@ THREE.TransformControls = function ( camera, domElement ) {
 		domElement.addEventListener( "touchstart", onPointerDown, false );
 		domElement.addEventListener( "mousemove", onPointerHover, false );
 		domElement.addEventListener( "touchmove", onPointerHover, false );
-		domElement.addEventListener( "mousemove", onPointerMove, false );
+		document.addEventListener( "mousemove", onPointerMove, false );
 		domElement.addEventListener( "touchmove", onPointerMove, false );
-		domElement.addEventListener( "mouseup", onPointerUp, false );
-		domElement.addEventListener( "mouseleave", onPointerUp, false );
-		domElement.addEventListener( "mouseout", onPointerUp, false );
+		document.addEventListener( "mouseup", onPointerUp, false );
 		domElement.addEventListener( "touchend", onPointerUp, false );
 		domElement.addEventListener( "touchcancel", onPointerUp, false );
 		domElement.addEventListener( "touchleave", onPointerUp, false );
@@ -122,11 +120,9 @@ THREE.TransformControls = function ( camera, domElement ) {
 		domElement.removeEventListener( "touchstart", onPointerDown );
 		domElement.removeEventListener( "mousemove", onPointerHover );
 		domElement.removeEventListener( "touchmove", onPointerHover );
-		domElement.removeEventListener( "mousemove", onPointerMove );
+		document.removeEventListener( "mousemove", onPointerMove );
 		domElement.removeEventListener( "touchmove", onPointerMove );
-		domElement.removeEventListener( "mouseup", onPointerUp );
-		domElement.removeEventListener( "mouseleave", onPointerUp );
-		domElement.removeEventListener( "mouseout", onPointerUp );
+		document.removeEventListener( "mouseup", onPointerUp );
 		domElement.removeEventListener( "touchend", onPointerUp );
 		domElement.removeEventListener( "touchcancel", onPointerUp );
 		domElement.removeEventListener( "touchleave", onPointerUp );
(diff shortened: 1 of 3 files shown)


but0n:
Looks good to me, thanks!

but0n (changes requested):

but0n:
Yeah, I've noticed it too, the editor is using `THREE.EditorControls()`.

ieskudero:
This patch fixes the issue of letting the THREE.OrbitControls() work, but if the object is dragging it shouldn't let move. Maybe it´s a better idea to put the stopPropagation() function after the checking line inside the pointerMove() function:

if ( object === undefined || axis === null || this.dragging === false || ( pointer.button !== undefined && pointer.button !== 0 ) ) return;

pointer.e.stopPropagation(); //assuming we pass the event int the pointer data structure

This way the camera doesn't move when we are dragging the object.

arodic:
> This patch fixes the issue of letting the THREE.OrbitControls() work, but if the object is dragging it shouldn't let move.

@ieskudero are you still experiencing OrbitControls moving camera while TransformControls is being dragged with this patch? Im getting correct behavior in webgl_geometry_spline_editor, misc_controls_transform and editor.

It appears to me that this somewhat depends on how your app handles multiple controls.

Personally, I think the best solution to remove `stopPropagation()` entirely and let application listen transformControl state changes and configure other controls appropriately. (OrbitControls or whatever other tool). The downside is that this approach requires additional code on application side.

Other approach (which worked OK before) is to use `stopPropagation()` to hack both controls to work simultaneously out of the box. This lets developers to simply attach both TransformControls and OrbitControls and they simply work which is great. The downside of this approach is that it is kind of a hack and might have other side-effects.

arodic:
`if ( object === undefined || axis === null || this.dragging === false || ( pointer.button !== undefined && pointer.button !== 0 ) ) return;`

`object === undefined` is redundant because this case always causes axis to be falsey.

`this.dragging === false` this makes sense, I'l add it to pointerMove as sanity check.

`pointer.button !== undefined && pointer.button !== 0` this also seems redundant since dragging will be false unless button is pressed.

arodic:
@mrdoob do you have any thoughts on this?

Personally I would remove `stopPropagation()` hack alltogether and attach mousemove listeners to document so drag works outside document borders. However, this would force developers to manually enable/disable multi-control setups in similar fashion like the editor in Viewport.js

ieskudero:
I agree that **stopPropagation()**  should be removed in order to keep the control clean (it´s dangerous to have a control stopping events), but I would set a comment or something that tells the user to make use of the **dragging** variable to know when the control is dragging.

Maybe it would be good to have a mainControl class to manage situations where there are multiple controls in the scene.

By the way, and out of topic, another feature that would be great would be the posibility of hiding gizmo's axis separately,  having boolean variables like **showPosX = false** or **showRotZ = false** to do it.

arodic:
> Maybe it would be good to have a mainControl class to manage situations where there are multiple controls in the scene.

I agree maybe it can be called `ControlManager` or something like that. I can imagine situations where you might want to have more than two controls operating simultaneously in 3D modeling tools for example. Also, I proposed this many times before, it would be useful to standardize all controls by extending a single `Control` class. I am quite interested in this topic (interactive tools and controls) so I'll probably create a separate repository just for that.

> By the way, and out of topic, another feature that would be great would be the posibility of hiding gizmo's axis separately, having boolean variables like showPosX = false or showRotZ = false to do it.

What is the intended use for this? Similar issue #11340 was closed now that axes hide automatically when facing camera. I thought the use case was to have better UX in orthographic side views for example. Your suggestion sounds like you want to prevent users from moving in particular axis regardless of camera view, right? Perhaps we can reopen #11340 and discuss it there.

arodic:
OK, so I removed the ugly `stopPropagation()` hack. This means that multi-control apps need to listen to **dragging-changed** event on TransformControls and enable/disable other controls. For example:

```javascript
transformControl.addEventListener( 'dragging-changed', function ( event ) {
    orbitControls.enabled = !event.value
} );
```

examples using TransformControls have also been updated to reflect this change. Editor (Viewport.js) is unaffected by this change.

ieskudero:
I didn't know of #11340, and I definitely would love to have axis disabled. A work project I'm working on needs to move some objects alongside X axis, restricting the other axis. I can stop the movement in some axis, but it would be great if the control showed only the axis that the object could move.

An event before showGizmo or something similar where I could change the visibility of the axis, alongside the booleans variables that I commented before should covert almost every posibility.

I think it´s a feature that many people could use. Anyway, #11340 should be the place to discuss it.

mrdoob:
@arodic could you fix the conflicts?

arodic:
@mrdoob done

mrdoob:
Thanks!

PR #14182, opened 2018-05-31
TrackballControls: Fixed bad zoom performance

boulabiar:
This is only a fix for chrome #14013

PR #14622, opened 2018-08-02
[Examples] Fix trackball controls "lag" on mobile.

XanderLuciano:
This fixes: https://github.com/mrdoob/three.js/pull/14182, https://github.com/mrdoob/three.js/issues/14013, 

The lag was incorrectly blamed on event handlers previously, when it's actually just incorrectly styled HTML causing a small overflow that interferes with gesture detection. 

Here we have a 1920 pixel wide screen, but this div is 1930px due to it having a 5px border on all sides, and not using `box-sizing: border-box` property. Thus it will always overflow the body by 10px.

image

There is also some vertical mis-sizing on mobile, but we can eliminate that by forcing both the HTML and Body elements to be the same size as the screen,

With these modifications, chrome will not "try" to scroll the screen, and the framerate is steady. (or at least steadier than before).

Test link for this PR:   
https://cdn.rawgit.com/XanderLuciano/three.js/fix-trackball-html/examples/misc_controls_trackball.html


XanderLuciano:
Requesting confirmation that this also works on iOS devices with safari. (I'm android only)

( and for reference, there are multiple different ways of solving this problem on stack overflow: https://stackoverflow.com/questions/14270084/overflow-xhidden-doesnt-prevent-content-from-overflowing-in-mobile-browsers/14271049#14271049 )

boulabiar:
I'm still having the lag with the link you provided.

XanderLuciano:
Try this link: https://rawgit.com/XanderLuciano/three.js/fix-trackball-html/examples/misc_controls_trackball.html

I'm getting a constant 60 fps now, now weird lag spikes when it was trying to zoom before, no delay when moving my find perpendicular to the edges. 

Also what version of Chrome are you running on your phone? Try downloading Chrome Canary and testing in Canary.

XanderLuciano:
Here's a video comparing the current version with the this PR: https://youtu.be/QqQ-qKrXLoI

Maintains a steady 60 fps no problem.

boulabiar:
The fps indicator keeps a good number, but the scene still have the lag.
I'll post a video about it later.

boulabiar:
https://www.youtube.com/watch?v=qQov0Eei_uU

FaiyazMdAbdul:
This issue still persists with latest three.js release . I am not using trackball but OribalControls. It seems to happens without either of them
chrome 72.03626.105 | Android

Mugen87:
Try to use the CSS `touchAction` property like mentioned in #14182

PR #14653, opened 2018-08-06
FBXLoader: improved texture checks

looeee:
Improved check for TGALoader, and added check for PSD textures since I've found that lots of models downloaded from SketchFab have references to PSD files. 

~~I've also switched to creating an empty `Texture` object when the loader can't find a texture.~~ 
EDIT: removed this as it's redundant, the TextureLoader and TGALoader both already return an empty `Texture` if the image file isn't found.

Sample console output with these checks:
```
FBXLoader: TGALoader not found, creating empty placeholder texture for LM_Final.tga

PSD textures are not supported, creating empty placeholder texture for props_alpha.psd
```

### DIFF
--- a/examples/js/loaders/FBXLoader.js
+++ b/examples/js/loaders/FBXLoader.js
@@ -389,13 +389,33 @@ THREE.FBXLoader = ( function () {
 
 			var texture;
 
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

PR #14656, opened 2018-08-07
Box3: revert applyMatrix4

### LINKED ISSUE #14655: Box3#applyMatrix4() no longer supports projection transformation matrices

#13993 introduced an optimization to `Box3#applyMatrix4()` that only supports affine transforms.

[screen shot 2018-08-06 at 9 24 21 pm](https://codepen.io/razh/full/mjGBVV/)

[**CodePen**](https://codepen.io/razh/full/mjGBVV/)

- **`geometry.applyMatrix()`** - The geometry is transformed with `applyMatrix()`.
- **`r93 Box3#applyMatrix4()`** & **`r92 Box3#applyMatrix4()`** - A `Box3` instance is constructed from the geometry. A non-affine matrix is then applied.

For `BoxGeometry`, it is expected is that both operations result in equivalent `Box3` instances.

The `r92` version transformed each `Box3` vertex, then used `setFromPoints()` to resize to fit.

- **r92**: https://github.com/mrdoob/three.js/blob/r92/src/math/Box3.js#L562-L596
- **r95**: https://github.com/mrdoob/three.js/blob/r95/src/math/Box3.js#L562-L585

See https://github.com/mrdoob/three.js/issues/6039 for a similar issue with a `Vector3#applyMatrix4()` optimization.

##### Three.js version

- [x] Dev
- [x] r95
- [x] r94
- [x] r93

##### Browser

- [x] All of them

##### OS

- [x] All of them

WestLangley:
Fixes #14655.

### DIFF
--- a/src/math/Box3.js
+++ b/src/math/Box3.js
@@ -559,30 +559,41 @@ Object.assign( Box3.prototype, {
 
 	},
 
-	applyMatrix4: function ( matrix ) {
+	applyMatrix4: function () {
 
-		// transform of empty box is an empty box.
-		if ( this.isEmpty( ) ) return this;
+		var points = [
+			new Vector3(),
+			new Vector3(),
+			new Vector3(),
+			new Vector3(),
+			new Vector3(),
+			new Vector3(),
+			new Vector3(),
+			new Vector3()
+		];
 
-		var m = matrix.elements;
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #14663, opened 2018-08-08
SpotLightShadow.cameraAutoUpdate / Light.shadowAutoUpdate

mbredif:
Preventing the Autoupdating of the `SpotLight.shadow.camera` was previously performed through the hack of [overwriting](https://github.com/mrdoob/three.js/compare/dev...mbredif:SpotLightCameraUpdateOptOut?expand=1#diff-3a405b9c95eb5dd79f92e5ba070fafa9L115) the SpotLightShadow with a LightShadow. 

Instead, this PR proposes a new property ~`SpotLightShadow.cameraAutoUpdate`~`Light.shadowAutoUpdate` that makes optional the fitting of the shadow camera frustum to the square cone defined by light.angle/light.distance.

(this PR is the item 2 of #14658, extracted here for an easier review)

== Edit ==
Instead of overwriting the shadow, an alternative hack to get the same result would be to set `this.shadow.isSpotLightShadow = false`. As that was the sole use of `isSpotLightShadow`, this PR is essentially moving and renaming `Light.shadow.isSpotLightShadow` to `Light.shadowAutoUpdate`.

### DIFF
--- a/docs/api/lights/SpotLight.html
+++ b/docs/api/lights/SpotLight.html
@@ -175,6 +175,13 @@ light.target = targetObject;
 			The spotlight will now track the target object.
 		</p>
 
+		<h3>[property:Boolean shadowAutoUpdate]</h3>
+		<p>
+			If set to *true*, the shadow camera will be automatically updated each frame to match the
+			geometry of the spotLight (a square frustum tightly fitting the cone of the spot light defined
+			by the [page:.angle angle] and [page:.distance distance] properties).
+			The default is *true*.
+		</p>
 
 		<h2>Methods</h2>
 
--- a/docs/api/lights/SpotLight.html
+++ b/docs/api/lights/SpotLight.html
@@ -177,9 +177,7 @@ light.target = targetObject;
 
 		<h3>[property:Boolean shadowAutoUpdate]</h3>
 		<p>
-			If set to *true*, the shadow camera will be automatically updated each frame to match the
-			geometry of the spotLight (a square frustum tightly fitting the cone of the spot light defined
-			by the [page:.angle angle] and [page:.distance distance] properties).
+			If set to *true*, the shadow camera frustum will be automatically sized to match the cone of the spotLight.
 			The default is *true*.
 		</p>
 
--- a/docs/api/lights/shadows/LightShadow.html
+++ b/docs/api/lights/shadows/LightShadow.html
@@ -134,6 +134,11 @@ scene.add( helper );
 		Serialize this LightShadow.
 		</p>
 
+		<h3>[method:void update]( [param:Light light] )</h3>
+		<p>
+		Updates this LightShadow (e.g. [page:SpotLightShadow.update SpotLightShadow.update]).
+		</p>
+
 		<h2>Source</h2>
 
 		[link:https://github.com/mrdoob/three.js/blob/master/src/lights/[name].js src/lights/[name].js]
--- a/docs/api/lights/shadows/LightShadow.html
+++ b/docs/api/lights/shadows/LightShadow.html
@@ -137,6 +137,7 @@ scene.add( helper );
 		<h3>[method:void update]( [param:Light light] )</h3>
 		<p>
 		Updates this LightShadow (e.g. [page:SpotLightShadow.update SpotLightShadow.update]).
+		The default [page:.update update] method does nothing.
 		</p>
 
 		<h2>Source</h2>
--- a/src/lights/SpotLightShadow.js
+++ b/src/lights/SpotLightShadow.js
@@ -9,15 +9,12 @@ import { PerspectiveCamera } from '../cameras/PerspectiveCamera.js';
 function SpotLightShadow() {
 
 	LightShadow.call( this, new PerspectiveCamera( 50, 1, 0.5, 500 ) );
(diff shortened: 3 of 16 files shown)


src/lights/SpotLightShadow.js
```diff
@@ -20,6 +21,8 @@ SpotLightShadow.prototype = Object.assign( Object.create( LightShadow.prototype
 
 	update: function ( light ) {
 
+		if ( ! this.cameraAutoUpdate ) return;
```
WestLangley: Instead, I would do the following in the calling method:
```javascript
if ( shadow.isSpotLightShadow && shadow.cameraAutoUpdate ) {

	shadow.update( light );

}
```

WestLangley:
I definitely support this.

I also support removal of the need for this pattern:
```javascript
light.shadow = new THREE.LightShadow( ... );
```

Forcing the shadow frustum to auto-size to the cone can have the side-effect of causing pixelated shadows.

src/lights/SpotLightShadow.js
mbredif: done

mbredif:
2 more commits : 
- I found 2 more usages of the `light.shadow = new THREE.LightShadow( ... );` pattern in the examples.
- I moved the `cameraAutoUpdate` check to the calling site, as suggested in the review.

mbredif:
I am not sure about 2da8234 though, at it breaks http://localhost:8000/examples/webgl_lights_spotlight.html due to https://github.com/mrdoob/three.js/blob/2cbd28958be86cb5f34ee5532a3ee18a23467b94/examples/webgl_lights_spotlight.html#L142 which is now ungarded (it is required to have the spotlight and camera helpers in sync)

mbredif:
@WestLangley what do you think of this new commit 64b191c ?
I moved `SpotLightShadow.cameraAutoUpdate` to `Light.shadowAutoUpdate`, updated the docs, and added the (conditional) light shadow update to the SpotLight helper. I also remove the now unused `isSpotLightShadow`.

docs/api/lights/SpotLight.html
```diff
@@ -175,6 +175,13 @@ <h3>[property:Object3D target]</h3>
 			The spotlight will now track the target object.
 		</p>
 
+		<h3>[property:Boolean shadowAutoUpdate]</h3>
+		<p>
+			If set to *true*, the shadow camera will be automatically updated each frame to match the
+			geometry of the spotLight (a square frustum tightly fitting the cone of the spot light defined
```
WestLangley: If set to *true*, the shadow camera frustum will be automatically sized to match the cone of the spotLight

docs/api/lights/SpotLight.html
```diff
@@ -175,6 +175,13 @@ <h3>[property:Object3D target]</h3>
 			The spotlight will now track the target object.
 		</p>
 
+		<h3>[property:Boolean shadowAutoUpdate]</h3>
```
WestLangley: [property:Boolean autoSizeShadowFrustum] ?

docs/api/lights/shadows/LightShadow.html
```diff
@@ -134,6 +134,11 @@ <h3>[method:Object toJSON]()</h3>
 		Serialize this LightShadow.
 		</p>
 
+		<h3>[method:void update]( [param:Light light] )</h3>
+		<p>
+		Updates this LightShadow (e.g. [page:SpotLightShadow.update SpotLightShadow.update]).
```
WestLangley: This method is akin to a virtual base class and does not do anything. Define what it does where it is overloaded.

WestLangley:
Actually, I think some others need to chime in regarding both the feature and the implementation details.

You are not "auto-updating" the shadow camera frustum, you are "auto-sizing" it. At least that is how I see it.

mbredif:
The naming shadowAutoUpdate refers to what it does for all lights (it guards a call to shadow.update, similarly to matrixAutoUpdate) rather than the previous  naming cameraAutoUpdate or autoSizeShadowFrustum which refers to what it does for spotlights. It is thus more generic (could be used unifiedly for using a custom shadow update function) and make isSpotLightShadow unnecessary.

docs/api/lights/SpotLight.html
mbredif: done

mbredif:
Up !

There is no new functionality here. The goal is to provide a better API for the hack of overwriting the SpotLightShadow with a LightShadow to prevent the lightshadow update.

The alternatives are : 
A- renaming `LightShadow.isSpotLightShadow` to `LightShadow.autoUpdate`
B- moving `LightShadow.isSpotLightShadow` to `Light.shadowAutoUpdate`
C- use a more specific naming like `autoSizeShadowFrustum` or `cameraAutoUpdate`, but that opens the door for users to use the LightShadow.update function as a generic custom callback, guarded by a flag with a specific naming.

(Complementarily the `autoUpdate` approach, a `needsUpdate` approach could be implemented as well for a 1-time updating dirty flag.)

This PR implements B without the `needsUpdate` approach. What API would you prefer @WestLangley @mrdoob ?

Mugen87:
Implemented via #19715.

PR #14674, opened 2018-08-09
FileLoader: Add listener for 'abort' event

### LINKED ISSUE #14652: FileLoader cannot handle request.abort()
The change, that let's FileLoader cache duplicate requests by default, introduced a problem, when aborting a request before it's loaded.

XMLHttpRequest.abort() does not tirgger 'load' or 'error' events but instead an 'abort' event.

Because that event is not handled, the request will be stuck in the FileLoaders 'loading' object, that is used to determine, if a file is already loading.

[Demo jsfiddle](https://jsfiddle.net/42brk186/) 

By adding a listener for 'abort' this could be prevented:

    request.addEventListener( 'abort', function ( event ) {
        ... handle the same way an error would be handled
    }, false );

I'm not sure, if handling it as an error is the way to go, because aborting has to be called on purpose, but this is how i do it for now.

##### Three.js version

- [x] r88+

##### Browser

- [x] All of them

##### OS

- [x] All of them

okitu:
Fixes #14652

### DIFF
--- a/src/loaders/FileLoader.js
+++ b/src/loaders/FileLoader.js
@@ -239,6 +239,24 @@ Object.assign( FileLoader.prototype, {
 
 			}, false );
 
+			request.addEventListener( 'abort', function ( event ) {
+
+				var callbacks = loading[ url ];
+
+				delete loading[ url ];
+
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

Issue #14676, opened 2018-08-09
JSPM package contains unnecessary files


LordBenjamin:
Including three.js using JSPM (`"three": "github:mrdoob/three.js@r89",`) results in a 400MB dependency - the package includes all source, documentation and examples.

For an application to depend on three.js, only the contents of `build/` are required. Would it be possible to ship a package that only contains these files?

I would expect that if I wanted to work with the docs / examples / etc, I could simply clone or export the Github repo.

![image](https://user-images.githubusercontent.com/13396705/43892214-45b4fcec-9bc3-11e8-907c-02f689e4aa8b.png)

Mugen87:
I don't think there is much what we can do on our side. The `npm` package of `three.js` is small:

https://bundlephobia.com/result?p=three@0.95.0

LordBenjamin:
PR incoming - let me know if there's any issue with adding the JSPM config to `package.json`.

kevinlandsberg:
> PR incoming - let me know if there's any issue with adding the JSPM config to `package.json`.

Hello,

why is the folder examples not included into the jspm files?
I am installing the library via jspm and i am missing the examples folder with all the javascripts inside...
(/examples/js/*). Can you please add it???

PR #14678, opened 2018-08-09
Add JSPM overrides to reduce the package size.

LordBenjamin:
Pull request relating to #14676.

Overriding the files / directories that JSPM pulls in reduces the download from ~400MB to ~2.5MB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
   "jspm": {
     "files": [
       "package.json",
-      "bower.json",
       "LICENSE",
       "README.md",
       "build/three.js",
--- a/package.json
+++ b/package.json
@@ -62,5 +62,18 @@
     "rollup-watch": "^4.3.1",
     "serve": "6.5.8",
     "uglify-js": "^3.3.28"
+  },
+  "jspm": {
+    "files": [
+      "package.json",
+      "bower.json",
+      "LICENSE",
+      "README.md",
+      "build/three.js",
+      "build/three.min.js",
+      "build/three.module.js"
+    ],
+    "directories": {
+    }
   }
 }
--- a/package.json
+++ b/package.json
@@ -61,5 +61,18 @@
     "rollup-watch": "^4.3.1",
     "serve": "6.5.8",
     "uglify-js": "^3.3.28"
+  },
+  "jspm": {
+    "files": [
+      "package.json",
+      "bower.json",
+      "LICENSE",
+      "README.md",
+      "build/three.js",
+      "build/three.min.js",
+      "build/three.module.js"
+    ],
+    "directories": {
+    }
   }
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "module": "build/three.module.js",
   "files": [
     "package.json",
-    "bower.json",
     "LICENSE",
     "README.md",
     "build/three.js",


package.json
```diff
@@ -62,5 +62,18 @@
     "rollup-watch": "^4.3.1",
     "serve": "6.5.8",
     "uglify-js": "^3.3.28"
+  },
+  "jspm": {
+    "files": [
+      "package.json",
+      "bower.json",
```
Mugen87: There is no `bower.json`.

LordBenjamin: Fair comment - I copy/pasted from the existing `files` section in `package.json`, is that also wrong?

Mugen87: Oh, yeah. The entry is actually obsolete.

Mugen87: Solved 4688a5fb211b338224d6f47b99684fcee138320e

LordBenjamin: OK. I'll remove both entries in my branch.

Mugen87: Maybe it's better if you merge the commit into your branch.

LordBenjamin: Didn't see your commit / comment (I'm using my phone), will do.

Mugen87:
Do you now why JSMP does not use the `files` definition of `npm`? Adding this additional config is somewhat redundant...

LordBenjamin:
Agreed - I don't know why JSPM ignores the base `files` definition.

package.json
```diff
@@ -65,7 +65,6 @@
   "jspm": {
     "files": [
       "package.json",
-      "bower.json",
```
Mugen87: Um, not sure why this line is still in...

LordBenjamin: The diff is showing this line being removed? Unless I'm missing something? (I've really not used Github much so that's entirely possible!)

image

The diff for the entire branch doesn't show this line being present either:

image

The line isn't present in my working copy and there is nothing to commit / push.

image

LordBenjamin: Ah, I see what you mean.

LordBenjamin: This is what I did to pull in your change and apply it to my branch. Am I holding it wrong?

```
git fetch upstream
git checkout dev
git merge upstream/dev

git checkout jspm-package-size
git rebase dev
```

mrdoob:
I bet jspm people are going to ask for code in examples at some point. But until then...

mrdoob:
Thanks!

kevinlandsberg:
Hello,

why is the folder examples not included into the jspm files?
I am installing the library via jspm and i am missing the examples folder with all the javascripts inside...
(/examples/js/*). Can you please add it???

Mugen87:
> I bet jspm people are going to ask for code in examples at some point. 

😆

mrdoob:
@LordBenjamin it's up to you 🙃

kevinlandsberg:
Thank you for your answer @LordBenjamin .
I am using three.js in a frontend project of mine. I am including the CanvasRender from 'three.js/examples/js/renderers/CanvasRenderer' and the Projector from 'three.js/examples/js/renderers/Projector'. I really need those functions. How can i still use them in my javscript without having the examples folder installed via jspm?

Regards.

LordBenjamin:
IMHO it isn't a good idea to reference something in the `examples` folder from your codebase.

I wouldn't expect that people will be careful to maintain API compatibility when updating the examples, as the assumption would be that they are just sample code. This would mean that your dependent code may break without warning.

If these files represent stable and functional components that are used by third-party production systems, I'd recommend moving them into the core three.js distribution or creating separate packages for them.

My personal approach towards these files would be to treat them as a starting point and copy them into your own codebase (maintaining copyright headers etc in compliance with the three.js licence).

mrdoob:
@kevinlandsberg what do you need `CanvasRenderer` and `Projector` for?

kevinlandsberg:
@mrdoob , i am using those classes for a project to create an animation. I really need them. Can you help me?

mrdoob:
@kevineschmidt can't you just download the repo and copy the files you're using?

mrdoob:
Also, just so you know, `CanvasRenderer` has been removed from the repo as it was a renderer that was only useful when WebGL was yet adopted by all browsers.

kevinoe:
@mrdoob  - Think you mentioned the wrong person here.

mrdoob:
@kevinoe You're right. Sorry about that.

PR #14679, opened 2018-08-09
Sprite: Fix raycast for offscreen sprites

### LINKED ISSUE #14624: Sprite: Raycasting with offscreen sprites fails

A user in the forum reported a bug in `Sprite.raycast()`.

https://discourse.threejs.org/t/problematic-raycasting-of-offscreen-three-sprite-object/3644

The method depends on `Object3D.modelViewMatrix`. But this matrix is only updated if the sprite is actually rendered. If not, `modelViewMatrix` does not regard the actual object's transformation. This means that `Sprite.raycast()` detects intersections of sprites outside the view frustum.

https://github.com/mrdoob/three.js/blob/1ffaa41a4b5f053ad62da1682844b53b041e63ca/src/objects/Sprite.js#L99

##### Three.js version

- [x] Dev
- [x] r95
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
Fixes #14624

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1240,6 +1240,10 @@ function WebGLRenderer( parameters ) {
 
 			} else if ( object.isSprite ) {
 
+				// Sprite.raycast() needs modelViewMatrix, even when the sprite is not rendered
+
+				object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
+
 				if ( ! object.frustumCulled || _frustum.intersectsSprite( object ) ) {
 
 					if ( sortObjects ) {
@@ -1389,7 +1393,7 @@ function WebGLRenderer( parameters ) {
 		object.onBeforeRender( _this, scene, camera, geometry, material, group );
 		currentRenderState = renderStates.get( scene, _currentArrayCamera || camera );
 
-		object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
+		if ( ! object.isSprite ) object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
 		object.normalMatrix.getNormalMatrix( object.modelViewMatrix );
 
 		if ( object.isImmediateRenderObject ) {


src/renderers/WebGLRenderer.js
```diff
@@ -1389,7 +1393,7 @@ function WebGLRenderer( parameters ) {
 		object.onBeforeRender( _this, scene, camera, geometry, material, group );
 		currentRenderState = renderStates.get( scene, _currentArrayCamera || camera );
 
-		object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
+		if ( ! object.isSprite ) object.modelViewMatrix.multiplyMatrices( camera.matrixWorldInverse, object.matrixWorld );
```
WestLangley: Can code in `sprite.onBeforeRender()` require this calculation?

Mugen87: If `matrixWorld` would be manipulated in some way, then yes. However, there would be the same problem than before. The new `matrixWorld` would not be calculated for offscreen sprites although the value is required by `Sprite.raycast()`.

Um, tricky. An other approach to solve the problem is to introduce a third parameter `camera` to `Sprite.raycast()` and calculated the `modelViewMatrix` right there.

WestLangley: Can `sprite.matrixWorld` be modified in `sprite.onBeforeRender()`?

Mugen87: Well, there might be a problem if `object.matrixWorld` is changed in `onBeforeRender()`. But in this case, there is the same problem. `Sprite.raycast()` needs the `modelViewMatrix` for all sprites. Um, tricky...

Mugen87: Sry, I've added the comment too late^^. I'm not sure if modifying the transformation of an `Object3D` is valid. Something like this would actually invalidate render lists.

WestLangley: I'd leave the line as-is, especially if you are not sure.

WestLangley: >I'm sure changing Object3D.matrixWorld in onBeforeRender() is no valid approach.

I just did a 30-second test and it works fine.
```javascript
sprite.onBeforeRender = function() { this.position.y += 0.01; this.updateMatrixWorld(); };
```
 I would not modify this line as you are doing.

Mugen87: What do you mean with "works fine"? You now produce inconsistent data in your app since the transformation of the sprites is different than `projectObject()` worked with.

WestLangley:
Raycasting for sprites is now done in camera space, which is why we have this problem.

It would be best if the raycast method could be modified, instead of the renderer, but that may not be doable. 

So this is a workaround -- maybe a permanent one.

PR #14685, opened 2018-08-10
Camera: added projectionMatrixInverse property

WestLangley:
As suggested in https://github.com/mrdoob/three.js/pull/14628#issuecomment-411926123.

### DIFF
--- a/docs/api/cameras/Camera.html
+++ b/docs/api/cameras/Camera.html
@@ -55,6 +55,9 @@
 		<h3>[property:Matrix4 projectionMatrix]</h3>
 		<p>This is the matrix which contains the projection.</p>
 
+		<h3>[property:Matrix4 projectionMatrixInverse]</h3>
+		<p>The inverse of projectionMatrix.</p>
+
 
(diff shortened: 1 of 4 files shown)


WestLangley:
We will  just need to make sure the projection matrix and its inverse remain in sync.

Users who set the projection matrix manually will have to set the inverse, too.

WestLangley:
@mrdoob Search for `projectionMatrix.copy` in the code and examples, and you will find a number instances. So, we will need to copy the inverse too -- or do something else, such as `camera.copy()`.

How do you feel about that?

WestLangley:
@mrdoob I will further explain:

We want to use `projectionMatrixInverse` instead of computing the inverse over and over again.

Thing is, `projectionMatrixInverse` is computed in `updateProjectionMatrix()` only; it is not set elsewhere. For example, it is not set whenever `projectionMatrix.copy()` is used. This will cause the matrix and its inverse to get out-of-sync.

We need to come up with an approach that will ensure they are synchronized if we want to make use of `projectionMatrixInverse`.

Issue #14035, opened 2018-05-10
Object3D.toJSON() not working for Points


mamun4122:
##### Description of the problem

I have created some points using the following code.
```
    let geometry = new THREE.Geometry();
    geometry.vertices.push(new THREE.Vector3( 5, 5, 0));
    geometry.vertices.push(new THREE.Vector3( 10, 5, 0));
    geometry.vertices.push(new THREE.Vector3( 15, 5, 0));

    geometry.colors.push(new THREE.Color(1, 0, 0));
    geometry.colors.push(new THREE.Color(0, 1, 0));
    geometry.colors.push(new THREE.Color(0, 0, 1));

    let material = new THREE.PointsMaterial({size: 2,  alphaTest: 0.5, vertexColors: THREE.VertexColors});
    let mesh =  new THREE.Points(geometry, material);
``` 
Now when i want to convert it using `mesh.toJSON()` the resulting json doesn't contain the colors for the vertices.

Here is the working [jsfiddle](http://jsfiddle.net/hp5bhf3y/3/).

##### Three.js version

- [ ] Dev
- [x] r92
- [ ] ...

Mugen87:
It seems that `Geometry.toJSON()` does not respect `Geometry.colors` so far.

But you can use `BufferGeometry` as a workaround: http://jsfiddle.net/hp5bhf3y/4/

gooderist:
Hello, hope you don't mind I put in a PR for this. It seemed like an easy change.

Mugen87:
Closing. Duplicate of #7149.

PR #14686, opened 2018-08-10
add colors to Geometry.toJSON()

gooderist:
Add colors to Geometry.toJSON() output

See #7149 and #14035


Mugen87:
Just adding `Geometry.toJSON()` does not entirely solve the problem. It's also important that `JSONLoader`/`ObjectLoader` are able to parse these data correctly.

Mugen87:
Closing since serializing and loading `Geometry` is not recommended anymore. Please use `BufferGeometry`.

PR #14687, opened 2018-08-10
setLength micro optimization

mbredif:
this micro optim saves a few function calls and multiplications...

### DIFF
--- a/src/math/Vector2.js
+++ b/src/math/Vector2.js
@@ -416,7 +416,7 @@ Object.assign( Vector2.prototype, {
 
 	setLength: function ( length ) {
 
-		return this.normalize().multiplyScalar( length );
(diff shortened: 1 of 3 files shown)


WestLangley:
You are inlining an existing method.

The code is written for clarity. So far, we do not inline code or otherwise micro-optimize. I expect a compelling reason would be required to do so, such as a demonstrated performance issue.

Mugen87:
> I expect a compelling reason would be required to do so, such as a demonstrated performance issue.

Even if the performance would be slightly better, I would prefer the current style which is easier to read.

WestLangley:
@mbredif Yes, we have had this discussion before, and have opted for clarity and reusability. But please continue to make suggestions! :-)

mbredif:
Ok, no problem. I am closing it then. I doubt that this would be a bottleneck in any case :)

PR #14695, opened 2018-08-11
Remove IIFEs, to loosen cyclical dependencies

Rich-Harris:
Following @Mugen87's suggestion in https://github.com/mrdoob/three.js/pull/14654#issuecomment-412291063:

A number of classes have methods that are returned from immediately-invoked function expressions (IIFEs), meaning that helper instances are created eagerly. This creates cyclical dependencies that make it tricky to convert the entire codebase to classes, and impede tree-shaking.

Because the codebase is composed of modules, we can simply hoist the helper variable declarations to the top of the modules that use them, and instantiate them lazily when the methods are invoked.

### DIFF
--- a/build/three.js
+++ b/build/three.js
@@ -61,31 +61,27 @@
 		// Missing in IE
 		// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
 
-		( function () {
+		Object.assign = function ( target ) {
 
-			Object.assign = function ( target ) {
+			if ( target === undefined || target === null ) {
 
-				if ( target === undefined || target === null ) {
+				throw new TypeError( 'Cannot convert undefined or null to object' );
 
-					throw new TypeError( 'Cannot convert undefined or null to object' );
+			}
 
-				}
+			var output = Object( target );
 
-				var output = Object( target );
+			for ( var index = 1; index < arguments.length; index ++ ) {
 
-				for ( var index = 1; index < arguments.length; index ++ ) {
+				var source = arguments[ index ];
 
-					var source = arguments[ index ];
+				if ( source !== undefined && source !== null ) {
 
-					if ( source !== undefined && source !== null ) {
+					for ( var nextKey in source ) {
 
-						for ( var nextKey in source ) {
+						if ( Object.prototype.hasOwnProperty.call( source, nextKey ) ) {
 
-							if ( Object.prototype.hasOwnProperty.call( source, nextKey ) ) {
-
-								output[ nextKey ] = source[ nextKey ];
-
-							}
+							output[ nextKey ] = source[ nextKey ];
 
 						}
 
@@ -93,11 +89,11 @@
 
 				}
 
-				return output;
+			}
 
-			};
+			return output;
 
-		} )();
+		};
 
 	}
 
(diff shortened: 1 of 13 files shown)


takahirox:
You don't need to include `build/*` in PR.

Mugen87:
What about the methods listed below. They have the following style (note the missing pair of brackets):
```js
distanceToPoint: function () {

    var v1 = new Vector2();

    return function distanceToPoint( point ) {

        var clampedPoint = v1.copy( point ).clamp( this.min, this.max );
        return clampedPoint.sub( point ).length();

    };

}(),
```
`Vector3.applyEuler()`, `Vector3.applyAxisAngle()`, `Vector3.clampScalar()`, `Vector3.unproject()`, `Vector3.projectOnPlane()`, `Vector3.reflect()`

`Object3D.rotateOnAxis()`, `Object3D.rotateOnWorldAxis()`, `Object3D.rotateX()`, `Object3D.rotateY()`, `Object3D.rotateZ()`, `Object3D.translateOnAxis()`, `Object3D.translateX()`, `Object3D.translateY()`, `Object3D.translateZ()`, `Object3D.worldToLocal()`, `Object3D.lookAt()`, `Object3D. getWorldQuaternion()`, `Object3D. getWorldScale()`

`PropertyBinding.sanitizeNodeName()`

`Camera.getWorldDirection()`

`BufferGeometry.rotateX()`, `BufferGeometry.rotateY()`, `BufferGeometry.rotateZ()`, `BufferGeometry.translate()`, `BufferGeometry.scale()`, `BufferGeometry.lookAt()`, `BufferGeometry.center()`, `BufferGeometry.normalizeNormals()`, `BufferGeometry.computeBoundingSphere()`

`Geometry.rotateX()`, `Geometry.rotateY()`, `Geometry.rotateZ()`, `Geometry.translate()`, `Geometry.scale()`, `Geometry.lookAt()`, `Geometry.center()`

`Box2.setFromCenterAndSize()`, `Box2.distanceToPoint()`

`Box3.setFromCenterAndSize()`, `Box3.expandByObject()`, `Box3.distanceToPoint()`

`Color.setHSL()`, `Color.copySRGBToLinear()`, `Color.copyLinearToSRGB()`,  `Color.offsetHSL()`

`Euler.setFromQuaternion()`,  `Euler.reorder()`

`Frustum.intersectsObject()`, `Frustum.intersectsSprite()`, `Frustum.intersectsBox()`

`Line3.closestPointToPointParameter()`

`Matrix3.applyToBufferAttribute()`

`Matrix4.extractRotation()`, `Matrix4.makeRotationFromQuaternion()`, `Matrix4.lookAt()`, `Matrix4.applyToBufferAttribute(), `Matrix4.decompose()`

`Plane.intersectLine()`, `Plane.setFromCoplanarPoints()`, `Plane.applyMatrix4()`

`Quaternion.setFromUnitVectors()`

`Ray.recast()`, `Ray.distanceSqToPoint()`, `Ray.distanceSqToSegment()`, `Ray.intersectSphere()`, `Ray.intersectTriangle()`

`Sphere.setFromPoints()`

`Triangle.getNormal()`, `Triangle.getBarycoord()`, `Triangle.containsPoint()`, `Triangle.closestPointToPoint()`

`Vector2.clampScalar()`

`Vector3.applyEuler()`, `Vector3.applyAxisAngle()`, `Vector3.project()`, `Vector3.unproject()`, `Vector3.clampScalar()`, `Vector3.projectOnPlane()`, `Vector3.reflect()`

`Vector4.clampScalar()`

`LOD.update()`

Rich-Harris:
@takahirox understood — removed.

@Mugen87 argh, I was searching for `)()`, completely missed `}()`. Yes, they also need to be converted. There's enough of them that I might try and automate it... 

In a situation like [this one](https://github.com/mrdoob/three.js/blob/dev/src/core/BufferGeometry.js#L173-L207)...

```js
	rotateX: function () {

		// rotate geometry around world x-axis

		var m1 = new Matrix4();

		return function rotateX( angle ) {

			m1.makeRotationX( angle );

			this.applyMatrix( m1 );

			return this;

		};

	}(),

	rotateY: function () {

		// rotate geometry around world y-axis

		var m1 = new Matrix4();

		return function rotateY( angle ) {

			m1.makeRotationY( angle );

			this.applyMatrix( m1 );

			return this;

		};

	}(),
```

...does each method need its own instance? Or can they all use a shared `m1`?

Mugen87:
> ...does each method need its own instance? Or can they all use a shared m1?

A method might call another method of the same class that both use the same type of temporary helper object. If we would now have a single instance, intermediate results get overwritten. An example for this is `Triangle.containsPoint()` which calls `Triangle.getBarycoord()`. Both have a `v1` vector object in their closure scope.

https://github.com/mrdoob/three.js/blob/a8e50626bbd8387059ebaeca607db3ca2895642f/src/math/Triangle.js#L50-L109

BTW: These two methods are something special since they are directly assigned to the function constructor (not the prototype). In context of classes, these methods should ideally use the `static` keyword so they become static class methods.

Mugen87:
@Rich-Harris I think we could also change the mentioned methods by hand. If you like, I can support you in this task.

Rich-Harris:
I've made good progress on the automated conversion but haven't had a chance to finish it off this week — I was thinking that a manual conversion in those cases probably makes more sense, yes (otherwise we'd have a bunch of auto-generated `v1_1`, `v1_2` etc). I'll update the PR soon

bhouston:
I think you should use some type of prefix notation to denote these module scope variables from local function variables, otherwise people will make mistakes.  Could be as simple as _position or m_position (flashbacks to [MFC](https://en.wikipedia.org/wiki/Microsoft_Foundation_Class_Library)) or [functionName]_position would also be nice to reduce the scope of each variable to a specific function.

marcofugaro:
What happened to this awesome PR? This is the first step to make the tree.js bundle tree-shakable

Mugen87:
Closing. All IIFEs are now removed from the code base.

PR #13587, opened 2018-03-14
ExtrudeGeometry: Remove option "frames"

Mugen87:
I'd like to suggest another simplification of `ExtrudeGeometry`: The removal of `options.frames` or the possibility to apply custom Frenet–Serret frames to the constructor. 

Analogous to `TubeGeometry` frames are now always calculated by the given curve via [Curve.computeFrenetFrames](https://threejs.org/docs/index.html#api/extras/core/Curve.computeFrenetFrames).

https://github.com/mrdoob/three.js/blob/0c6c0050982cf3697005bb7ca419b4dbd6a9279a/src/geometries/TubeGeometry.js#L74

`option.frames` was never used in the examples so there is no need for modifications.

PR #14696, opened 2018-08-12
fix the bug which lost tangent, binormal, and normal in ExtrudeGeometry

kuixu:
When tangent, binormal, and normal are provided by users, the previous code will lost them in ExtrudeGeometry, and now I fixed the bug.

### DIFF
--- a/src/geometries/ExtrudeGeometry.js
+++ b/src/geometries/ExtrudeGeometry.js
@@ -140,9 +140,13 @@ function ExtrudeBufferGeometry( shapes, options ) {
 
 			// TODO1 - have a .isClosed in spline?
 			if ( options.frames === undefined ) {
+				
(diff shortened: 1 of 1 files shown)


Mugen87:
This is not a bug. The option `frames` was removed on purpose, see #13587.

barrykui:
But if I draw some self-designed geometry, the frames are computed incorrectly, which make the geometry structure with a wrong style.

Mugen87:
What do you mean with "self-designed geometry"? Do you mean custom Frenet–Serret frames? As mentioned in the linked PR, this workflow is not supported anymore. Use a 3D modeling tool to design the geometry and then load in into your app. Best with `glTF`.

Mugen87:
Closing. Please read the discussion in #13587 in order to understand why the option `frames` was removed. You should probably use a modeling software and then import your geometry data via one of `three.js`s loaders.

PR #14710, opened 2018-08-14
Removed dead code (handleEvent) from controls.

arodic:
Found this dead piece of code across examples/js/controls.

Perhaps it was once used but now it serves no purpose and it is not referenced anywhere else in examples/ or editor/ It appears that it was added to FlyControls long time ago by @alteredq, then copy/pasted into TrackbalConrols by @egraether


Mugen87 (approved):

mrdoob:
Thanks!

PR #14720, opened 2018-08-16
Fix bounding box calculations on BufferGeometries with indices/groups

simon-paris:
FYI, the /wiki/How-to-contribute-to-three.js page is not loading for me, sorry if I do something wrong.

This PR makes the Box3.setFromBufferAttribute, Box3.expandByObject, and BufferGeometry.computeBoundingSphere methods consider geometry's indices and groups. This makes them behave more like BufferGeometry.computeVertexNormals.

Previously, if you combined multiple geometry's buffers together, calculating the bounding box would scan the entire vertices buffer, giving incorrect results and being very slow.

### DIFF
--- a/src/core/BufferGeometry.js
+++ b/src/core/BufferGeometry.js
@@ -634,7 +634,13 @@ BufferGeometry.prototype = Object.assign( Object.create( EventDispatcher.prototy
 
 				var center = this.boundingSphere.center;
 
-				box.setFromBufferAttribute( position );
+				if ( this.index && this.groups && this.groups.length === 0 ) {
+
+					this.addGroup( 0, this.index.array.length );
+
+				}
+
+				box.setFromBufferAttribute( position, this.index || null, this.groups || null );
 				box.getCenter( center );
 
 				// hoping to find a boundingSphere with a radius smaller than the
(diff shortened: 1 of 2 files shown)


WestLangley:
Are you aware that both indexed and non-indexed `BufferGeometry` support `groups`?

If we are going to implement these changes, then we must identify all the methods that will need to honor groups, but currently do not. Is your list complete?

Should we honor `groups` in these methods, but not `drawRange`?

Groups are designed to support multi-materials. The fact that groups can limit the number of vertices rendered is a side-effect.

I am not sure we should go down this path...

Mugen87:
I do not vote to merge this PR. When calculating an AABB of bounding sphere, it's sufficient to focus on the vertices. The bounding volume will enclose the resulting shape, no matter how the faces are actually structured. Same for a convex hull.

From my point of view, @simon-paris is faced with an issue that needs to be solved on application level.  If users merge multiple geometries into a single one, they also need a special approach to calculate the desired bounding volumes.

simon-paris:
Alright, I am able to monkey patch those functions. But in that case, computeVertexNormals should probably ignore groups too.

WestLangley:
>computeVertexNormals should probably ignore groups too.

I am inclined to agree with that.

PR #13253, opened 2018-02-05
Example: Basic OffscreenCanvas example

Mugen87:
see discussion in https://github.com/mrdoob/three.js/issues/13235#issuecomment-362798479

This is a very basic example that shows how use `three.js` in a worker in combination with `OffscreenCanvas`. The demo produces asynchronously frames for the canvas element in the main thread. 

It was actually not necessary to change the core or apply any polyfills to get things working 🙌 . Because `OffscreenCanvas` is an experimental API, you have to activate a respective flag in Chrome (Experimental Canvas Features).

image

PR #14721, opened 2018-08-16
Improve offscreencanvas example for cross-browser inter operability

devnook:
In the recent Chrome implementation of OffscreenCanvas, there is no commit() call and instead DedicatedWorker.requestAnimationFrame is used (https://www.chromestatus.com/features/5681560598609920). This PR updates the OffscreenCanvas demo to work in current Chrome and Firefox browsers.

It also allows drawing un-textured cube if loader.load fails. It happens currently in FF because of  a wrong function signature (Chrome allows it while FF rejects, https://bugzilla.mozilla.org/show_bug.cgi?id=1335594).

### DIFF
--- a/examples/js/workers/OffscreenCanvas.js
+++ b/examples/js/workers/OffscreenCanvas.js
@@ -65,9 +65,13 @@ function animate() {
 	renderer.render( scene, camera );
 
 	if ( self.requestAnimationFrame ) {
+
 		self.requestAnimationFrame( animate );
+
 	} else if ( renderer.context.commit ) {
+
 		renderer.context.commit().then( animate );
+
 	}
 
 }
--- a/examples/js/workers/OffscreenCanvas.js
+++ b/examples/js/workers/OffscreenCanvas.js
@@ -9,6 +9,12 @@ self.onmessage = function ( message ) {
 
 var camera, scene, renderer, mesh, clock;
 
+function createBoxMesh( size, material ) {
+	var material = new THREE.MeshBasicMaterial(material);
+	var geometry = new THREE.BoxBufferGeometry(size, size, size);
+	return new THREE.Mesh(geometry, material);
+}
+
 function init( offscreen, width, height, pixelRatio ) {
 
 	camera = new THREE.PerspectiveCamera( 70, width / height, 1, 1000 );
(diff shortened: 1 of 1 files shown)


examples/js/workers/OffscreenCanvas.js
```diff
@@ -52,7 +58,10 @@ function animate() {
 	mesh.rotation.y += delta * 0.5;
 
 	renderer.render( scene, camera );
-
-	renderer.context.commit().then( animate );
+	if (self.requestAnimationFrame) {
```
Mugen87: Please respect the code style. E.g. instead of your code do this:
```
if ( self.requestAnimationFrame ) {

    self.requestAnimationFrame( animate );

} else ( renderer.context.commit ) {

    renderer.context.commit().then( animate );

}
```

examples/js/workers/OffscreenCanvas.js
```diff
@@ -25,23 +31,23 @@ function init( offscreen, width, height, pixelRatio ) {
...
 
 	} );
 
-	//
-
-	renderer = new THREE.WebGLRenderer( { antialias: true, canvas: offscreen } );
-	renderer.setPixelRatio( pixelRatio );
-	renderer.setSize( width, height, false );
-
+	renderer = new THREE.WebGLRenderer({ antialias: true, canvas: offscreen });
```
Mugen87: Code style.

Mugen87:
https://rawgit.com/devnook/three.js/offscreencanvas/examples/webgl_worker_offscreencanvas.html

Mugen87 (changes requested):

Mugen87:
> It happens currently in FF because of a wrong function signature (Chrome allows it while FF rejects).

Do you refer to a FF bug here?

devnook:
Yes, the bug is this one:  https://bugzilla.mozilla.org/show_bug.cgi?id=1335594 (now added to the description too).

Mugen87:
Um, are you sure this bug is related to the example? We are not setting any options via `ImageBitmapLoader.setOptions()`. So when `createImageBitmap()` is called, `undefined` is passed as a second parameter. I guess the problem is related to the issue linked below. Especially the last link with titel `Crash in mozilla::WebGLContext::FromDomElem when uploading canvas as WebGL texture` is relevant here.

https://github.com/mrdoob/three.js/pull/13253#issuecomment-363438920

https://bugzilla.mozilla.org/show_bug.cgi?id=1436036
https://bugzilla.mozilla.org/show_bug.cgi?id=1342694

Because of this, I suggest you undo the texture related changes in the example. A worker demo without `ImageBitmapLoader` and thus textures seems to meaningful to me.

devnook:
I'm rather sure it's about the createImageBitmap signature bug. Even if the options parameter is undefined, FF will still complains about the signature:

screen shot 2018-08-17 at 13 27 05

Without the texture related changes in the example, the demo will not work in FF (it does not crash, it just stays blank and the error can be logged to console):

screen shot 2018-08-17 at 13 19 20

This happens on current FF (61.0.2 (64-bit)).

I think it would be nice if the demo run out of the box on FF, despite this bug, as the bug is not related to OffscreenCanvas itself. Do you think we can add this workaround?

Mugen87:
> Do you think we can add this workaround?

@mrdoob should decide this. I remember cases where we did not add workarounds and refer to the existing browser bugs instead (e.g. #7802).

mrdoob:
Thanks!

PR #14729, opened 2018-08-17
Log the image in the warning about resizing power of two textures.

ngokevin:
Lets developer know what image needs resizing.


mrdoob:
Apparently, doing this causes memory leaks... #14483

ngokevin:
Will re-open just logging the src

PR #14732, opened 2018-08-17
Corrected an error in OrbitControls documentation

matheowis:
there was an error with controls.mouseButtons keys. I changed them from [LEFT, MIDDLE, RIGHT] to [ORBIT, ZOOM, PAN]


Mugen87:
When looking at the latest code, it seems that the documentation is already correct:

https://github.com/mrdoob/three.js/blob/af878f446dc4d5ee6153000cee88dd32e902eb91/examples/js/controls/OrbitControls.js#L78

PR #14738, opened 2018-08-18
spotLightHelper: fix duplicate rotation

but0n:
When I rotate the parent of spot light, the cone of light helper will rotate twice.

image

### Scene tree
- scene
  - spotLight group (transform control attach here)
    - spot light
    - target of spot light

[Example](https://gist.github.com/but0n/29939ee781ce75902b431208816e9570)

`this.matrix = light.matrixWorld;` 

 - The only thing this object need is the world position of the spot light, but this matrix here include world position and world rotation (not direction) of the light, so the helper will apply the world rotation matrix of the light.

`this.cone.lookAt( vector2.sub( vector ) );`
 -  this will also locally rotate the cone to the correct direction.

### DIFF
--- a/src/helpers/SpotLightHelper.js
+++ b/src/helpers/SpotLightHelper.js
@@ -18,9 +18,6 @@ function SpotLightHelper( light, color ) {
 	this.light = light;
 	this.light.updateMatrixWorld();
 
-	this.matrix = light.matrixWorld;
-	this.matrixAutoUpdate = false;
-
 	this.color = color;
 
 	var geometry = new BufferGeometry();
@@ -52,7 +49,7 @@ function SpotLightHelper( light, color ) {
 	this.cone = new LineSegments( geometry, material );
 	this.add( this.cone );
 
-	this.update();
+		this.update();
 
 }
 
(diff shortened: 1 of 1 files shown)


src/helpers/SpotLightHelper.js
```diff
@@ -83,6 +80,7 @@ SpotLightHelper.prototype.update = function () {
 		vector.setFromMatrixPosition( this.light.matrixWorld );
 		vector2.setFromMatrixPosition( this.light.target.matrixWorld );
 
+		this.position.setFromMatrixPosition( this.light.matrixWorld );
```
WestLangley: Do this, instead:
```javascript
this.position.copy( vector );
```

WestLangley:
`SpotLightHelper` is limited in that it only supports lights without rotated parents.

An advantage of this change is it eliminates this pattern:
```javascript
this.matrix = light.matrixWorld;
this.matrixAutoUpdate = false;
```

If #14658 is merged, then this PR would have to be revisited.

/ping @mbredif

mbredif:
> /ping @mbredif 

This is a quick and simple fix, I think it can be merged before the discussion on #14658  settles.

As a side note you could save a temporary Vector3 variable and a Vector3.copy call by replacing
```
var vector = new Vector3();
var vector2 = new Vector3();
[...]
vector.setFromMatrixPosition( this.light.matrixWorld );
vector2.setFromMatrixPosition( this.light.target.matrixWorld );
this.position.copy( vector );
this.cone.lookAt( vector2.sub( vector ) );
```
with
```
var target = new Vector3();
[...]
this.position.setFromMatrixPosition( this.light.matrixWorld );
target.setFromMatrixPosition( this.light.target.matrixWorld );
this.cone.lookAt( target.sub( this.position ) );
```

WestLangley:
Thanks!

mrdoob:
Do you still have the issue after #14738?

but0n:
Nope, thanks!

PR #14742, opened 2018-08-19
use rAF in OffscreenCanvas worker

trusktr:
Looks like `commit()` is the old way to do it. https://ebidel.github.io/demos/offscreencanvas.html


trusktr:
Oh, nvm, it's already in dev, and better. It defaulted to showing me `master`, so I didn't realize.

PR #14752, opened 2018-08-21
Make distance fog calculation overflow less in mediump

### LINKED ISSUE #14736: Mobile Fog Bug

greggman:
Fixes #14736

### DIFF
--- a/src/renderers/shaders/ShaderChunk/fog_fragment.glsl
+++ b/src/renderers/shaders/ShaderChunk/fog_fragment.glsl
@@ -1,6 +1,9 @@
 #ifdef USE_FOG
 
-	float fogDepth = length( vFogPosition );
+	// divide then multiply by max component to try to avoid saturating mediump precision
+	vec3 fogPositionAbs = abs( vFogPosition );
+	float fogMaxComponent = max( fogPositionAbs.x, max( fogPositionAbs.y, fogPositionAbs.z ) );
+	float fogDepth = length( vFogPosition / fogMaxComponent ) * fogMaxComponent;
 
 	#ifdef FOG_EXP2
 


greggman:
If this works do you want to do the same for specular lighting calculations? Or conditionally do the same thing if in mediump? Or does no one even notice their highlights are in the wrong place and/or missing on mediump?

Note: running the webgl_animation_cloth examples in mediump and then using my finger the roll the camera around the net appears and disappears from time to time. Probably a similar issue

titansoftime:
I want to be clear that as noted in #14736, my device does not render fog at all. Nothing.

It works just fine in the master build.

mrdoob:
@titansoftime can you try this fiddle?

https://jsfiddle.net/bsp7h5uk/

mrdoob:
> Or does no one even notice their highlights are in the wrong place and/or missing on mediump?

Wouldn't be surprised. Just yesterday I realised light were not correct in VR... https://github.com/mrdoob/three.js/commit/c2a469a9c7fa40fd604b341dab6f83547833cde5

mrdoob:
I can reproduce the issue on a Galaxy S8 and this PR fixes the issue for me.

titansoftime:
Works for me =]

mrdoob:
Can also reproduce (and confirm the fix works) on a Huawei Y5 lite 2017 🙌

mrdoob:
Thanks!

mrdoob:
> If this works do you want to do the same for specular lighting calculations? Or conditionally do the same thing if in mediump? Or does no one even notice their highlights are in the wrong place and/or missing on mediump?

Currently `MeshStandardMaterial` doesn't render properly in `mediump` I guess this would be a good first step?

@WestLangley @bhouston

bhouston:
I am all for making mediump work, but adding divides on a GPU is very slow.  I guess if you are adding divides I would prefer to make this a separate code path, but I also hate maintaining separate code paths.  Maybe you can make this a macro?

titansoftime:
Doesn't it seem a little odd to be doing more complex calculations when one chooses mediump?

On the other hand, I am not sure how much the extra calculations will affect performance.

greggman:
another option is you could add `THREE.FogDistance` and
`THREE.FogDistamceExp2` or some other option for you to opt in to distance
based or depth based fog

```
fog = new THREE.Fog(...)
fog.type = THREE.DepthFog;
```

bhouston:
One could also use fog in the z direction rather than as a distance from camera?  Then it would become something like this:
```
float fogFactor = ( gl_FragDepth - gl_DepthRange.near ) / gl_DepthRange.diff;

OR

// using fog limits not liked to the camera frustum
float fogFactor = ( gl_FragDepth - fogNear ) / ( fogFar - fogNear );
```
???

greggman:
That was what it used to be and is wrong and is why it was switched to distance fog.

With depth based fog the fog changes as you rotate the camera. Stuff that was in the fog when directly in front of the camera comes out of the fog is it moves toward the edge of the screen even though it's the same distance from the viewer.

titansoftime:
```
fog = new THREE.Fog(...)
fog.type = THREE.DepthFog;
```

I mean if I had a vote =]

bhouston:
https://docs.microsoft.com/en-us/windows/desktop/direct3d9/fog-formulas

From the above, it seems that there are these general options:

- Range (distance from camera) versus Depth (distance along forward axis.)
- Linear, Exp and Exp2.

It seems that Range is preferred, thus we can drop support for Depth.  I only suggested depth because it was simpler math.

And then you have Linear, Exp and Exp2 type transforms on the fog.  I am unsure if all three are equally useful or not.

WestLangley:
>Doesn't it seem a little odd to be doing more complex calculations when one chooses mediump?

See https://github.com/mrdoob/three.js/pull/14740#issuecomment-414520872

looeee:
> And then you have Linear, Exp and Exp2  type transforms on the fog. I am unsure if all three are equally useful or not.

Regarding Exp2, @spite mentioned a use here yesterday

https://twitter.com/thespite/status/1032041367168593926

mrdoob:
> I'd suggest something like this so that the precision safe function can be reused.  I suspect there may be more precision safe functions like this:
>
> ```glsl
> #ifdef HIGH_PRECISION
> 	float precisionSafeLength( vec3 v ) {
> 		return length( v );
> 	}
> #else
> 	float max3( vec3 v ) {
> 		return max( max( v.x, v.y ), v.z );
> 	}
> 
> 	float precisionSafeLength( vec3 v ) {
> 		float maxComponent = max3( abs( v ) );
> 		return length( v / maxComponent ) * maxComponent;
> 	}
> #endif  
> ```

I like that ideas. Here's a list of all the other instances of  `length()`:

https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/bsdfs.glsl#L141
https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/lights_pars_begin.glsl#L66
https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/lights_pars_begin.glsl#L102
https://github.com/mrdoob/three.js/blob/dev/src/renderers/shaders/ShaderChunk/shadowmap_pars_fragment.glsl#L213

Do you guys think it is worth adding the code?

WestLangley:
In the first link, I don't think overflow is actually an issue, but consistency has its merits...

titansoftime:
There is is still an issue.

The fog is "disappearing" oddly at some points on mobile. I'll work on making a fiddle and posting a new issue.

bug

mrdoob:
Oh noes! Thanks for the fiddle in advance!

Issue #14737, opened 2018-08-18
JSON models need to be replaced


WestLangley:
There are 30+ examples that render legacy `Geometry` because `JSONLoader` returns `Geometry`.

We will need to use a different loader in these examples -- one that returns `BufferGeometry`. That means the following models either need to be converted to another format or replaced with a different model.

models/json:

* [x] (https://github.com/mrdoob/three.js/pull/14747) ~models/json/male02/Male02_dds.json~
* [x] (https://github.com/mrdoob/three.js/pull/14747) ~models/json/female02/Female02_slim.json~
* [x] (#14765) ~models/json/cubecolors/cubecolors.json~
* [x] (#14774) ~models/json/WaltHeadLo.json~
* [X] (#14769) ~models/json/suzanne_geometry.json~
* [x] (#14806) ~models/json/leeperrysmith/LeePerrySmith.json~
* [x] (#14924) ~models/json/lightmap/lightmap.json~
* [x] (#14788) ~models/json/QRCode.json~

models/animated:

* [x] (https://github.com/mrdoob/three.js/pull/14747) ~models/animated/horse.js~
* [x] (https://github.com/mrdoob/three.js/pull/14747) ~models/animated/flamingo.js~
* [x] (https://github.com/mrdoob/three.js/pull/14747) ~models/animated/sittingBox.js~
* [x] (https://github.com/mrdoob/three.js/pull/14747) models/animated/monster/monster.js (moved)

models/skinned:

* [x] (#14781) ~models/skinned/simple/simple.js~
* [x] (#15132) ~models/skinned/knight.js~

Suggestions?

> **EDIT:** Edited by @donmccurdy and @Mugen87 to track progress.

Mugen87:
I vote to convert the models to `glTF`. We could import them in the editor and then use the `Export GLB` option.

takahirox:
What's the easiest way to make Draco compressed .glb now?

donmccurdy:
I'm glad to work on converting or replacing some of these. I'll update the OP with ones I'm working on, in case others want to work in parallel. A few (horse, flamingo, monster) are already available in glTF format I think. The lightmap example would need to use another format though — glTF does not currently support lightmaps, just AO.

Maybe the lightmap model, and perhaps some others, should be updated to the newer JSON format, since having the ability to serialize objects to/from JSON is still valuable to maintain.

> What's the easiest way to make Draco compressed .glb now?

[...]

PR #14781, opened 2018-08-25
Examples: Moved webgl_skinning_simple to gltf

Mugen87:
See #14737

The example uses now `SimpleSkinning.glb`. Besides, the scene looks a little bit nicer now.

I've not deleted the JSON model since it's used by the unit tests.


takahirox:
BTW I'm wondering what this example is for... There're a lot of examples showing skinning animation. Do we really need this example?

Mugen87:
The armature of the model is very simple. There is also corresponding blend file in the repo. I guess it was initially added to illustrate a basic workflow from Blender to `three.js`. Users can check out the blend file in order to see how things are setup.

takahirox:
If that's the purpose of this example, I think we need to add a note or something, for example the link to `simple.blend` file on the example page. It's unclear now.

Mugen87:
Done.

WestLangley:
The example looks much better! It would be even better if the skeleton-helper toggle were present, IMHO.

Unfortunately, the point of the example is lost as the model file is now not human readable. I am not sure an ascii .gltf file would be much better. I am also not sure what this example adds now. Are you sure blender workflow was the point of this example?

mrdoob:
Maybe we can convert it to .gltf (embeded) instead?

Mugen87:
> Maybe we can convert it to .gltf (embeded) instead?

Done!

mrdoob:
Thanks!

mrdoob:
Btw, this example was very handy when I refactored the `WebGLRenderer` to use `BufferGeometry` internally years ago 😊

takahirox:
Don't we add a skeleton-helper in this example? Agreed with @WestLangley

Mugen87:
I've checked this but because of the simple skeleton the visualization is not really meaningful.

takahirox:
Yeah, absolutely.

image

PR #14783, opened 2018-08-25
FBXLoader little endian performance improvement

akoww:
Since the byteorder of FBX files is little endian, we can simply copy whole arrays in a singleshot on little endian platforms. This saves lot of execution time on big FBX files.

### DIFF
--- a/examples/js/loaders/FBXLoader.js
+++ b/examples/js/loaders/FBXLoader.js
@@ -3292,7 +3292,7 @@ THREE.FBXLoader = ( function () {
 
 				var value = subNode.propertyList[ 0 ];
 
-				if ( Array.isArray( value ) ) {
+				if ( Array.isArray( value ) || ArrayBuffer.isView( value ) ) {
 
 					node[ subNode.name ] = subNode;
 
@@ -3587,6 +3587,19 @@ THREE.FBXLoader = ( function () {
 
 		getInt32Array: function ( size ) {
 
+			if ( this.littleEndian ) {
+
+				return this.getInt32ArrayLE( size );
+
+			} else {
+
+				return this.getInt32ArrayBE( size );
+
+			}
+		},
+
+		getInt32ArrayBE: function ( size ) {
+
 			var a = [];
 
 			for ( var i = 0; i < size; i ++ ) {
@@ -3594,6 +3607,17 @@ THREE.FBXLoader = ( function () {
 				a.push( this.getInt32() );
 
 			}
+
+			return a;
+
+		},
+
+		getInt32ArrayLE: function ( size ) {
+
+			var buffer = this.dv.buffer.slice( this.offset, this.offset + size * 4 );
+			var a = new Int32Array(buffer);
+
+			this.offset += 4 * size;
 
 			return a;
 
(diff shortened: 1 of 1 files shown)


looeee:
This is cool, I seem to be getting an improvement of ~10% on buffer parsing time for most models. 

However I've also found at least one model that no longer loads. Give me a bit of time to test this and I'll see if I can figure out why.

looeee:
Unfortunately I don't think I can, that one was shared with me privately and I don't wanna step on anyone's toes! Sorry.

I'll test again tomorrow though and see if I can find any other models that cause problems.

looeee:
This is the problem:

```js
getFloat32ArrayLE: function ( size ) {

        var a = new Float32Array( this.dv.buffer.slice( this.offset, this.offset + size * 4 ) );
	this.offset += 4 * size;

	return a;

},
```

The method names are actually a little confusing, it should be called something like `getArrayOfFloat32`. What you want is this, to return a normal `Array`:

```js
var a = new Array( this.dv.buffer.slice( this.offset, this.offset + size * 4 ) );
```

akoww:
`var a = new Array( this.dv.buffer.slice( this.offset, this.offset + size * 4 ) );`
This wont work out of the box. This will create an `Array` with `size * 4` elements of type `byte/interger`
However we need only `size` elements of type `float32`. Thats the reson I used `TypedArrays` to represent the whole sliced chunck of bytes as a array of `size` floats.

Some metadata like version/exporter/... about the FBX model may be enough to reconstruct the problem.

looeee:
> This wont work out of the box. This will create an Array with size * 4 elements of type byte/interger

Oops, you are correct. 

The problem is that your `getXXXArrayLE` methods return typed arrays, while your `getXXXArrayBE` methods return normal arrays. They each need to return the same thing - this could be typed arrays, but then we'll need to do some refactoring to make sure that we are only using methods available to typed arrays.

The actual error in this case was caused by the `inject()` helper function since `Float32Array().concat()` doesn't exist. 

Here's a variation of your method that returns a normal array, however it's about 20% slower than the current method. 

```js
getFloat32ArrayLE: function ( size ) {

	var buffer = this.dv.buffer.slice( this.offset, this.offset + size * 4 )

	this.offset += 4 * size;

	return Array.prototype.slice.call( new Float32Array( buffer ));

},
```

akoww:
I nether like the fact, that I used mixed return types to get it work, but it seemed to be the esiest way to get it work. For my use case and the used FBX models with limited animation it seems to work propper but as you mentioned for not for all. 

Now we got two options. 1) discard this hack since its broken or 2) let me rewrite the code with TypedArrays.

The second option would take a while, since my free time is limited and my JS skills not the best. So if you think rewriting/refactoring is a good idea then let me know it.

looeee:
@akoww sorry for the delay in getting back to you on this. 

> Now we got two options. 1) discard this hack since its broken or 2) let me rewrite the code with TypedArrays.

What the loader is currently doing for binary FBX files (focusing on the parts that deal with large arrays): 

1. Parsing typed data from the binary file and converting to it to ordinary, untyped JavaScript data 
2. Operating on that data 
3. Converting it back into typed data in some cases (geometry), or leaving it as untyped data (animation) 

However, in step 3 the typed arrays are not necessarily of the same type as the original data. Also, this is ignoring ASCII files which contain untyped data to start with, although I suspect most people use those just for testing so performance is not so important there.  

You method is currently getting around 10% speedup on LE systems for step 1 (parsing) only - I think that it's likely that, especially due to the lack of a `concat` method for typed arrays, we would lose most or all of this 10% in in step 2 while processing the data, and this would also likely end up making the loader slower on BE systems. 

However, I could be wrong and the only way to know for sure would be to at least partially refactor the loader and do lots of testing. If you want to try this then go ahead, otherwise I would be inclined to close this PR.

looeee:
Closing this for now. If @akoww or someone else wants to refactor the loader as described above and provide benchmarks demonstrating that there is a performance improvement across multiple systems and browsers then we can reconsider it.

PR #14791, opened 2018-08-27
Example Variable Clean up

gkjohnson:
I'm looking into what it would take to get the example files running using es6 imports and found that some of the code can't be run in a `modules` context, so I thought I'd make a PR to fix some of those problems.

For the most part the changes address:

- assumption of shared scope between script tags
- using variables before they're defined
- assuming the global `this` is `window`

### DIFF
--- a/examples/webgl_physics_terrain.html
+++ b/examples/webgl_physics_terrain.html
@@ -204,7 +204,7 @@
 
 						// Create the terrain body
 
-						var groundShape = createTerrainShape( heightData );
+						var groundShape = createTerrainShape();
 						var groundTransform = new Ammo.btTransform();
 						groundTransform.setIdentity();
 						// Shifts the terrain, since bullet re-centers it on its bounding box.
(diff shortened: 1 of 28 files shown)


examples/webgl_physics_terrain.html
```diff
@@ -204,7 +204,7 @@
 
 						// Create the terrain body
 
-						var groundShape = this.createTerrainShape( heightData );
+						var groundShape = createTerrainShape( heightData );
```
Mugen87: The parameter of `createTerrainShape()` can be removed.

Mugen87 (changes requested):

examples/webgl_physics_terrain.html
gkjohnson: Removed! Thanks!

Mugen87 (approved):

mrdoob:
Thanks!

PR #14795, opened 2018-08-27
Move EffectComposer "Pass" into its own file

gkjohnson:
#### What Changed

- Moves `THREE.Pass` definition from `postprocessing/EffectComposer.js` to `postprocessing/Pass.js`
- Add warning to EffectComposer if `THREE.Pass` isn't defined to notify users of the change
- Update all examples that use the EffectComposer to include the new file

#### Why

I'm looking into making a script to convert all the example scripts into es6 module files and this file was proving troublesome because it creates cyclic dependencies:

`EffectComposer (EffectComposer.js)` depends on `ShaderPass` which depends on `Pass (EffectComposer.js)`.

So EffectComposer.js imports ShaderPass before it has defined Pass, yet, which causes ShaderPass to throw an undefined variable error. Moving Pass out the EffectComposer file removes the cyclic dependency.

### DIFF
--- a/examples/js/postprocessing/EffectComposer.js
+++ b/examples/js/postprocessing/EffectComposer.js
@@ -32,6 +32,12 @@ THREE.EffectComposer = function ( renderer, renderTarget ) {
 
 	// dependencies
 
+	if ( THREE.Pass === undefined ) {
+
+		console.error( 'THREE.Pass has been moved into a separate file, "Pass.js". Include it, as well.' );
+
+	}
+
 	if ( THREE.CopyShader === undefined ) {
 
 		console.error( 'THREE.EffectComposer relies on THREE.CopyShader' );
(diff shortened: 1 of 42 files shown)


WestLangley:
Is there a way to inline the code if it is not defined so existing fiddles will continue to work?

Can you swap the order?
```html
<script src="js/postprocessing/EffectComposer.js"></script>

<script src="js/postprocessing/Pass.js"></script>
<script src="js/postprocessing/ShaderPass.js"></script>
<script src="js/postprocessing/MaskPass.js"></script>
```

gkjohnson:
> Is there a way to inline the code if it is not defined so existing fiddles will continue to work?

I could do that by copy / pasting the code from `Pass` into `EffectComposer` and then printing a warning if it's been inlined that it should actually be included separately. I suppose another approach would be to fetch and `eval` the Pass script if it doesn't exist but that feels dirty. It would keep the definitions in sync, though.

> Can you swap the order?

You _can_ swap the order and everything will work but the warning will print if Pass is included after EffectComposer. I tried putting the warning inside of the EffectComposer constructor but other code would fail from the missing Pass definition before the EffectComposer was created in the examples so the warning would never get printed, making kinda useless.

Is there a better way to handle the printed warning?

gkjohnson:
> Is there a better way to handle the printed warning?

I suppose I could wrap the log in a `requestAnimationFrame` so the the other scripts have had a chance to load before checking if Pass exists. That way it won't matter what order the scripts are included in.

gkjohnson:
/ping @mrdoob @WestLangley 

Any further thoughts on what I should do here? Should I copy the `Pass` source into `EffectComposer` and wrap the warning in `requestAnimationFrame`?

Thanks!

WestLangley:
Sorry, I am not sure of the preferred way to handle this...

PR #14796, opened 2018-08-28
Fix error handling for PCDLoader parsing

### LINKED ISSUE #14776: PCDLoader does not handle errors correctly

The PCDLoader doesn't handle errors well.  If there's an error when parsing, the error handler passed in is not called and the resulting exception is only caught by a global process handler.  This may apply to all loaders as if there's an error in the `setTimeout` method used for loading there's no way to catch that exception.  

```js
        var url = 'https://threejs.org/examples/textures/sprites/disc.png';
        var loader = new THREE.PCDLoader().load(
          url, 
          () => alert('something loaded'),
          () => {},
          () => alert('error handler - never gets called')
        );
```

##### Three.js version

-  r91

##### Browser

- [x] All of them
- [x]  Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [x] macOS
- [x] Linux
- [ ] Android
- [ ] iOS

any

elcommie:
This fixes an issue with the PCD Loader so it will use the `onError` parameter (when there is one) if an issue occurs in the `parse` method. 

Note: This fixes one loader.  There are many others that have a similar pattern of passing the `onError` to the file loader, and parsing afterword without catching any errors that may occur.  Might be worth moving this catch logic to the `FileLoader`.

Fixes #14776

### DIFF
--- a/examples/js/loaders/PCDLoader.js
+++ b/examples/js/loaders/PCDLoader.js
@@ -28,7 +28,23 @@ THREE.PCDLoader.prototype = {
 		loader.setResponseType( 'arraybuffer' );
 		loader.load( url, function ( data ) {
 
-			onLoad( scope.parse( data, url ) );
+			try {
+
+				onLoad( scope.parse( data, url ) );
+
+			} catch ( e ) {
+
+				if ( onError ) {
+
+					onError( e );
+
+				} else {
+
+					throw e;
+
+				}
+
+			}
 
 		}, onProgress, onError );
 


Mugen87 (approved):

mrdoob:
Thanks!

gkjohnson:
> Note: This fixes one loader. There are many others that have a similar pattern of passing the onError to the file loader, and parsing afterword without catching any errors that may occur. Might be worth moving this catch logic to the FileLoader.

@elcommie Is there an existing issue to track the other loaders' error handling?

elcommie:
@gkjohnson There is not, but my particular issue was the PCD Loader.  I haven't used the other loaders (and by other loaders, I mean the loaders in example/js/loaders), so I can't speak to how well they handle errors other than what I saw in the code.  I suppose we could test something out and create an issue if it does have a similar problem

gkjohnson:
My assumption is that they're all similar and don't handle the error when parsing an incompatible format. I can take a look and make the issue when I get a chance.

PR #14750, opened 2018-08-20
Docs: Added localisation support.

mrdoob:

PR #14820, opened 2018-09-01
Fix broken links to source

### LINKED ISSUE #14819: Documentation: Broken link in InstancedBufferGeometry doc page

The bottom of the doc page of [InstancedBufferGeometry](https://threejs.org/docs/index.html#api/en/core/InstancedBufferGeometry) has a link to the source [src/en/core/InstancedBufferGeometry.js](https://github.com/mrdoob/three.js/blob/master/src/en/core/InstancedBufferGeometry.js) which is broken.

##### Three.js version

- [x] r96

##### Browser

- [x] Opera
- [x] Chrome
- [x] and maybe all others

##### OS

- [x] All of them

untitled

»(◕‸◕)«

fordhurley:
Fixes #14819 

This was noted in https://github.com/mrdoob/three.js/pull/14750#issuecomment-417635324.

The problem was that the `[path]` for the docs changed to support localization, but the source code didn't. For example, `/docs/#api/objects/Mesh` became `/docs/#api/en/objects/Mesh`.

The doc links follow the pattern:

    [link:https://github.com/mrdoob/three.js/blob/master/src/[path].js src/[path].js]

where `[path]` is, for our example, `en/objects/Mesh`.

I introduced a new variable, `[nonLocalizedPath]`, and changed the links to:

    [link:https://github.com/mrdoob/three.js/blob/master/src/[nonLocalizedPath].js src/[nonLocalizedPath].js]

Not an elegant solution, but it gets the job done. Extremely open to other approaches.

### DIFF
--- a/docs/api/en/Template.html
+++ b/docs/api/en/Template.html
@@ -43,6 +43,6 @@
 
 		<h2>Source</h2>
 
-		[link:https://github.com/mrdoob/three.js/blob/master/src/[path].js src/[path].js]
+		[link:https://github.com/mrdoob/three.js/blob/master/src/[nonLocalizedPath].js src/[nonLocalizedPath].js]
 	</body>
 </html>
--- a/docs/api/en/animation/AnimationAction.html
+++ b/docs/api/en/animation/AnimationAction.html
@@ -357,6 +357,6 @@
 
 		<h2>Source</h2>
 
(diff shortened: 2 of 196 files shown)


fordhurley:
Ok, great! I was afraid there would be a reason to keep the full `[path]` around, but this is dramatically simpler. I just opened a new PR with that solution: #14846

PR #14828, opened 2018-09-03
Added MeshMatcapMaterial

WestLangley:
`MeshMatcapMaterial`  supports maps, bump maps, normal maps, object-space normal maps, displacement maps, alpha maps, flat-shading, alpha test, skinning,  morphs, and local and global clipping.

`MeshMatcapMaterial` does not respond to lights. (The matcap image file encodes baked lighting). It will cast a shadow onto an object that receives shadows (and shadow clipping works), but it will not self-shadow or receive shadows.

The included example supports drag-and-drop, so the user can easily experiment with alternate matcap image files.


WestLangley:
If preferred, the example can be modified to use the glTF LeePerrySmith model once #14806 is merged.

WestLangley:
Bubblegum Knight

bubblegum_knight

mrdoob:
Thanks!

src/materials/MeshMatcapMaterial.js
```diff
@@ -0,0 +1,129 @@
...
+
+	this.lights = false;
+
+	this.setValues( parameters );
+
+	// a matcap is required
+
+	if ( this.matcap === null ) {
+
+		var canvas = document.createElement( 'canvas' );
```
WestLangley: @mrdoob Is referencing `document` here going to be a problem?

Is there another way we can have a default matcap?

This is actually not necessary. If the matcap is null, the model will render black, I guess. But for drag and drop support, ensuring the matcap was a valid texture to begin with makes things easy.

mrdoob: This is how it's done in `WebGLTextures`:
https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLTextures.js#L31

mrdoob: You can probably create a `DataTexture` instead though.

WestLangley: >document.createElementNS

That presupposes a document, too. Is that OK? I'm not sure of the advantage of `createElementNS` here...

>You can probably create a DataTexture instead though.

I tried a `DataTexture` first, but there were errors when the user replaced the default matcap with a `Texture`, or when drag-and-drop was used. It is possible I did something wrong, of course...

mrdoob: > I tried a DataTexture first, but there were errors when the user replaced the default matcap with a Texture, or when drag-and-drop was used. It is possible I did something wrong, of course...

Oh, sounds like a bug. I'll have to investigate.

WestLangley: >I'll have to investigate.

Sweet. I appreciate that!

Here is my original attempt:
```javascript
if ( this.matcap === null ) {

	this.matcap = new THREE.DataTexture( new Uint8Array( [ 255, 255, 255 ] ), 1, 1, THREE.RGBFormat, THREE.UnsignedByteType ); // white

	this.matcap.needsUpdate = true;

}
```

WestLangley: >Oh, sounds like a bug. I'll have to investigate.

/bump

/ping @Mugen87 Should it be this, instead? Or doesn't it matter?

```
var canvas = document.createElementNS( 'http://www.w3.org/1999/xhtml', 'canvas' );
```

Mugen87: Yes, that's correct. Otherwise we assume a HTML document context. Explicitly defining the HTML namespace is more robust.

mrdoob: >> Oh, sounds like a bug. I'll have to investigate.
>
> /bump

#15313

Issue #13651, opened 2018-03-21
Deprecate BinaryLoader


Mugen87:
This was discussed before but i'd like to make a separate issue so it's easier to track the progress. `BinaryLoader` is still used in three examples. However, there are already some PRs by @looeee

- [x] `webgl_materials_cubemap` → #14843 
- [x] `webgl_materials_cars` → #14843
- [x] `webgl_effects_parallaxbarrier` → #13084

There is also a dependency to `BinaryLoader` in `examples/js/Car.js`.

Looking at the PRs, i wonder if it's not more expedient to just focus on the removal of `BinaryLoader` and don't perform any `Car.js` related refactoring or simplification. I would see these changes in different PRs. In this way, i think we can easier make progress. 

What do you guys think?

looeee:
> don't perform any Car.js related refactoring or simplification

That was my original intention, however it turned out that Car.js is quite strictly tied in to the Binary format model being used, which means that just replacing the model with one in a different format doesn't work. 

It makes sense to me that if we have something as specific as car controls in the repo, then it should be possible to  use it with any generic car model, consisting of a body, 4 wheels and an optional steering wheel. I haven't tested my update with that many models, but it's at least a step in the right direction. 

You could try exporting the current model using the GLTFExporter and make minimal changes that way, but then we still have a not very useful car controls file tied to a specifically designed model. However since I seem to have hit a dead end here, if you have more success with that approach then I won't complain 😛

Mugen87:
Um, let's start with: `webgl_effects_parallaxbarrier`

I would just copy `webgl_effects_anaglyph` and replace `AnaglyphEffect` with `ParallaxBarrierEffect`.

Mugen87:
@looeee Would you like to perform the change (in #13084) or should I make a fresh PR?

looeee:
If you don't mind waiting a day or so I can do it, otherwise feel free to make a fresh PR 👍

Mugen87:
Let's reuse your PR 😊

Mugen87:
@looeee Okay, the first example `webgl_effects_parallaxbarrier` is done. Next, I suggest to remove `webgl_materials_cubemap_dynamic` and rename `webgl_materials_cubemap_dynamic2` to `webgl_materials_cubemap_dynamic`. I would use `Car.js` in `webgl_materials_cars` instead.

looeee:
>  I would use Car.js in webgl_materials_cars instead

[...]

PR #14843, opened 2018-09-04
Examples: update webgl_materials_cars

looeee:
[Live version](https://rawgit.com/looeee/three.js/examples_car_controls/examples/webgl_materials_cars.html) 

webgl_materials_car

As discussed in #13651: 

* combined `webgl_materials_cubemap_dynamic` and `webgl_materials_cars` so that there is now one place that demonstrates car controls with car.js and dynamic material updating. 
* removed dependency on `BinaryLoader` from `Car.js` so that it now accepts a generic car model in any format
* Replaced the old car models with [this](https://sketchfab.com/models/57bf6cc56931426e87494f554df1dab6) one, converted to glTF and with draco compression it comes in at 1.6mb. 
* Renamed `webgl_materials_cubemap_dynamic2` -> `webgl_materials_cubemap_dynamic` and removed the old example. 

After this PR there should be nowhere that uses the `BinaryLoader`, actually deprecating it and tidying old files is left to a seperate PR.


examples/webgl_materials_cars.html
```diff
@@ -66,541 +61,338 @@
...
 
-					$( button_name( car, i ) ).counter = i;
-					$( button_name( car, i ) ).addEventListener( 'click', function() {
+				carModel.getWorldPosition( cameraTarget );
+				cameraTarget.y = 2.5;
+        cameraTarget.z += distance;
 
-						for ( var j = 0; j < material_indices.length; j ++ ) {
+				camera.position.lerp( cameraTarget, delta * damping );
+        camera.lookAt( carModel.position );
```
Mugen87: There are some minor format issue at the end of the file.

Mugen87 (changes requested):

examples/webgl_materials_cars.html
```diff
@@ -66,541 +61,334 @@
...
+      }
 
-						}
-
-					}, false );
-
-				}
-
-			}
+      function resetPosition() {
```
Mugen87: Um, this looks still not correct. BTW: You can easily detect wrong formatting by enabling invisible characters in your editor.

looeee: Thanks, should all be fixed now.

Mugen87 (approved):

mrdoob:
Thanks!

donmccurdy:
Looks great! 🎉

PR #14859, opened 2018-09-05
add computeAngleVertexNormals: calculation of normals taking into account an angle to neighboring polygons 

Ni55aN:
Example: https://codepen.io/Ni55aN/pen/zROmoe

### DIFF
--- a/src/core/Geometry.js
+++ b/src/core/Geometry.js
@@ -549,27 +549,33 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 		function weightedNormal( normals, vector ) {
 
-		  var normal = new THREE.Vector3();
+			var normal = new THREE.Vector3();
 
-			for ( var i = 0; i < normals.length; i ++ ) {
+			for ( var i = 0, l = normals.length; i < l; i ++ ) {
 
 				if ( normals[ i ].angleTo( vector ) < angle ) {
 
-			  		normal.add( normals[ i ] );
+					normal.add( normals[ i ] );
 
 				}
 
 			}
 
-		  return normal.normalize();
+			return normal.normalize();
 
 		}
 
 		this.computeFaceNormals();
 
-		var vertexNormals = this.vertices.map( () => [] );
+		var vertexNormals = [];
 
-		for ( var i = 0; i < this.faces.length; i ++ ) {
+		for ( var i = 0, l = this.vertices.length; i < l; i ++ ) {
+
+			vertexNormals[ i ] = [];
+
+		}
+
+		for ( var i = 0, fl = this.faces.length; i < fl; i ++ ) {
 
 			var face = this.faces[ i ];
 
@@ -579,7 +585,7 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 		}
 
-		for ( var i = 0; i < this.faces.length; i ++ ) {
+		for ( var i = 0, fl = this.faces.length; i < fl; i ++ ) {
 
 			var face = this.faces[ i ];
 
--- a/src/core/Geometry.js
+++ b/src/core/Geometry.js
@@ -545,6 +545,58 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	},
 
+	computeAngleVertexNormals: function ( angle ) {
+
+		function weightedNormal( normals, vector ) {
+
+		  var normal = new THREE.Vector3();
+
+			for ( var i = 0; i < normals.length; i ++ ) {
+
+				if ( normals[ i ].angleTo( vector ) < angle ) {
+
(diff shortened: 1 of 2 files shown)


src/core/Geometry.js
```diff
@@ -545,6 +545,58 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 	},
 
+	computeAngleVertexNormals: function ( angle ) {
+
+		function weightedNormal( normals, vector ) {
+
+		  var normal = new THREE.Vector3();
+
+			for ( var i = 0; i < normals.length; i ++ ) {
```
Glinkis: Reduce the indentation of this loop by one step.

Ni55aN:
@Galiaf47 yes, sure

src/core/Geometry.js
```diff
@@ -545,6 +545,58 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+			}
+
+		  return normal.normalize();
+
+		}
+
+		this.computeFaceNormals();
+
+		var vertexNormals = this.vertices.map( () => [] );
```
Mugen87: Please avoid the usage of an array function.

src/core/Geometry.js
```diff
@@ -545,6 +545,58 @@ Geometry.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+		for ( var i = 0; i < this.faces.length; i ++ ) {
```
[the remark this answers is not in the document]
Glinkis: Why are the loops written like this?
In all modern browsers performance is pretty much the same if if you don't cache the length.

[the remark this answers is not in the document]
Mugen87: > I understand if it's a stylistic choice

Um, I remember benchmarks that showed that this approach is a bit faster. Anyway, we are not using this style of `for` loop at all places in the code base. I've made this suggestion so the code of `computeAngleVertexNormals()` is more consistent to the other `compute*Normals()` methods.

src/core/Geometry.js
Ni55aN: hm.. three.js not welcome with es6?

Ni55aN:
@Glinkis implementation for `BufferGeometry` looks more difficult (due to non-indexed geometry). I suggest to leave this for later, but for now we can use `fromBufferGeometry` / `fromGeometry`

WestLangley:
This is a thing-of-beauty on extruded `TextGeometry`: smooths the faceted edges only. Likewise for `ExtrudeGeometry`.

screen shot 2018-09-06 at 6 28 36 pm

I'd prefer the following API, instead, however. It would be a significant improvement to the existing method, which is limited in this regard.
```javascript
computeVertexNormals( areaWeighted, creaseAngle ); // angle in radians; undefined same as Math.PI
```

It is unfortunate it is for `Geometry` only, not `BufferGeometry`, but I'll take it.

Ni55aN:
@WestLangley the present `computeVertexNormals()` contains many of logic. I don't think it is good idea to add some code to this method. Or maybe it will only call the method if there is a second parameter
```js
computeVertexNormals( areaWeighted, creaseAngle ){
   if(creaseAngle) return computeAngleVertexNormals(creaseAngle)
}
```
but then the first parameter is lost

WestLangley:
I think it would be OK to always use area-weighted:
```javascript
computeVertexNormals( creaseAngle ); // creaseAngle optional
```
(We know how to change the API in a backwards-compatible way.)

I do realize that @mrdoob does not want to make any more changes to `Geometry`, so this entire PR may be moot.

Ni55aN:
Moving to 'utils' it would be good solution

Glinkis:
My colleague has now written a utility function that does exactly this, but for `BufferGeometry`,
so should I issue a separate pull request to put this function in `BufferGeometryUtils` or something?

Mugen87:
I vote for `BufferGeometry`.

Glinkis:
Yes, me too. But I suppose it's ultimately up to @mrdoob ?

WestLangley:
`BufferGeometry.computeVertexNormals( creaseAngle )` will only support `creaseAngle` for non-indexed. If indexed, users will have to convert their geometry to non-indexed first.

WestLangley:
>Would it be all right to issue a separate pull request for inserting said function in either BufferGeometry, or BufferGeometryUtils?

Yes, but I expect it will require a restructuring of the data if the geometry is indexed. It may be better as a utility, in that case. I'd propose it as a utility at first. It can always be promoted.

Glinkis:
Yes, it needs to split points for hard edges, and offset eventual geometry groups accordingly.

Mugen87:
Closing, since `Geometry` will not be enhanced with such methods anymore. It's better to make a new PR with an enhancement for `BufferGeometryUtils`. I think this will have more chance to get merged...

PR #14870, opened 2018-09-06
Added layers example

aardgoose:
There are no examples demonstrating layers apart from some use in some of the VR demos which don't really highlight their utility (layers seen to be overlooked).

Also added a flat background texture which also has no examples.


examples/webgl_layers.html
```diff
@@ -0,0 +1,156 @@
...
+				info.style.width = '100%';
+				info.style.textAlign = 'center';
+				info.innerHTML = '<a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> webgl - layers';
+				container.appendChild( info );
+
+				var bgTexture = new THREE.TextureLoader().load( 'textures/brick_diffuse.jpg' );
+				var bgColor = new THREE.Color( 0xf0f0f0 );
+
+				camera = new THREE.PerspectiveCamera( 70, window.innerWidth / window.innerHeight, 1, 10000 );
+				camera.layers.enable( 0 );
```
Mugen87: Maybe you could add a comment that says channel zero is enabled by default?

examples/webgl_layers.html
```diff
@@ -0,0 +1,156 @@
...
+				container = document.createElement( 'div' );
+				document.body.appendChild( container );
+
+				var info = document.createElement( 'div' );
+				info.style.position = 'absolute';
+				info.style.top = '10px';
+				info.style.width = '100%';
+				info.style.textAlign = 'center';
+				info.innerHTML = '<a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> webgl - layers';
+				container.appendChild( info );
```
Mugen87: Please write this code as HTML markup.

examples/webgl_layers.html
```diff
@@ -0,0 +1,156 @@
...
+				var layers = { red: true, green: true, blue: true, background: false };
+
+				//
+				// Init gui
+				var gui = new dat.GUI();
+				gui.add( layers, 'red' ).onChange( function () { camera.layers.toggle( 0 ); } );
+				gui.add( layers, 'green' ).onChange( function () { camera.layers.toggle( 1 ); } );
+				gui.add( layers, 'blue' ).onChange( function () { camera.layers.toggle( 2 ); } );
+
+				gui.add( layers, 'background' ).onChange( function ( value ) { scene.background = value ? bgTexture : bgColor; } );
```
Mugen87: I would remove the background since it feels out of place in this example. Let's just focus on layers.

Mugen87 (changes requested):

examples/webgl_layers.html
Mugen87: Maybe this one:
https://github.com/mrdoob/three.js/blob/master/examples/webgl_loader_fbx.html

Mugen87 (approved):

Mugen87:
For testing: https://rawgit.com/aardgoose/three.js/layers-example/examples/webgl_layers.html

examples/webgl_layers.html
```diff
@@ -0,0 +1,156 @@
...
+			function render() {
+
+				theta += 0.1;
+
+				camera.position.x = radius * Math.sin( THREE.Math.degToRad( theta ) );
+				camera.position.y = radius * Math.sin( THREE.Math.degToRad( theta ) );
+				camera.position.z = radius * Math.cos( THREE.Math.degToRad( theta ) );
+				camera.lookAt( scene.position );
+
+				camera.updateMatrixWorld();
```
WestLangley: Can you explain why are you calling `camera.updateMatrixWorld()` here?

examples/webgl_layers.html
```diff
@@ -0,0 +1,156 @@
...
+
+				camera = new THREE.PerspectiveCamera( 70, window.innerWidth / window.innerHeight, 1, 10000 );
+				camera.layers.enable( 0 ); // enabled by default
+				camera.layers.enable( 1 );
+				camera.layers.enable( 2 );
+
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color( 0xf0f0f0 );
+
+				var light = new THREE.DirectionalLight( 0xffffff, 1 );
```
WestLangley: How about
```javascript
var light = new THREE.PointLight( 0xffffff, 1 );
scene.add( camera );
camera add( light );
```

WestLangley:
>Also added a flat background texture which also has no examples.

There is no texture, but I don't think it is necessary in this example.

aardgoose:
Yes, I removed it from the first version,

examples/webgl_layers.html
aardgoose: cut and paste from another example - I'll review it.

examples/webgl_layers.html
aardgoose: lighting fixed as you suggested.

examples/webgl_layers.html
```diff
@@ -0,0 +1,155 @@
...
+				camera = new THREE.PerspectiveCamera( 70, window.innerWidth / window.innerHeight, 1, 10000 );
+				camera.layers.enable( 0 ); // enabled by default
+				camera.layers.enable( 1 );
+				camera.layers.enable( 2 );
+
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color( 0xf0f0f0 );
+
+				var light = new THREE.PointLight( 0xffffff, 1 );
+				light.position.set( 1, 1, 1 ).normalize();
```
WestLangley: Why are you normalizing the position?

Why isn't ( 0, 0, 0 ), the default, adequate?

aardgoose: junk left in by mistake. apologies.

mrdoob:
Thanks!

PR #14888, opened 2018-09-09
Added currentCamera property to Scene and scene to Object3D.

arodic:
I'd like to propose this very small change to updateMatrixWorld method which I found extremely useful when creating viewport helpers and other objects that need camera information for their matrix updates.

One example would be a scene helper that always looks towards the current camera. This can easily be implemented by extending the `Object3D.updateMatrixWorld` with custom matrix update logic and reference to current camera's `matrixWorld`. This PR enables that.

If this idea is welcome, I can add add an example to this PR.

Alternative solutions exist. One is to manually update `.camera` property on all objects which need it but that can add a lot of additional complexity to application logic. Especially when multiple cameras and multiple viewports are used.

Another solution is to transform geometry in the shader (where camera information is obtainable). I believe points currently do this. This on the other hand, requires custom shaders which is not ideal.

### DIFF
--- a/docs/api/en/core/Object3D.html
+++ b/docs/api/en/core/Object3D.html
@@ -112,6 +112,9 @@
 		<p>Object's parent in the [link:https://en.wikipedia.org/wiki/Scene_graph scene graph]. An object can have at most
 		one parent.</p>
 
+		<h3>[property:Scene scene]</h3>
+		<p>Object's top-most ancestor ([page:Scene]) if object is added to a scene.</p>
+
 		<h3>[property:Vector3 position]</h3>
 		<p>A [page:Vector3] representing the object's local position. Default is (0, 0, 0).</p>
 
--- a/docs/api/en/scenes/Scene.html
+++ b/docs/api/en/scenes/Scene.html
@@ -44,6 +44,9 @@
 		If not null, sets the background used when rendering the scene, and is always rendered first. Can be set to a [page:Color] which sets the clear color, a [page:Texture] covering the canvas, or a [page:CubeTexture]. Default is null.
 		</p>
 
+		<h3>[property:Camera currentCamera]</h3>
+		<p>Reference to the most recent camera the scene has been (or is about to be) be rendered with.</p>
+
 		<h2>Methods</h2>
 
 		<h3>[method:JSON toJSON]</h3>
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -27,6 +27,7 @@ function Object3D() {
 	this.type = 'Object3D';
 
 	this.parent = null;
+	this.scene = null;
 	this.children = [];
 
 	this.up = Object3D.DefaultUp.clone();
@@ -382,6 +383,19 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 			}
 
 			object.parent = this;
+
+			var scene = this.scene;
+
+			if ( object.scene !== scene ) {
+
+				object.traverse( function ( child ) {
+
+					child.scene = scene;
+
+				} );
+
+			}
+
 			object.dispatchEvent( { type: 'added' } );
 
 			this.children.push( object );
@@ -415,6 +429,12 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 		if ( index !== - 1 ) {
 
 			object.parent = null;
+
+			object.traverse( function ( child ) {
+
+				child.scene = null;
+
+			} );
 
 			object.dispatchEvent( { type: 'removed' } );
 
--- a/src/renderers/WebGL2Renderer.js
+++ b/src/renderers/WebGL2Renderer.js
@@ -149,6 +149,8 @@ function WebGL2Renderer( parameters ) {
 		var background = scene.background;
 		var forceClear = false;
 
+		scene.currentCamera = camera;
+
 		if ( background === null ) {
 
 			state.buffers.color.setClear( _clearColor.r, _clearColor.g, _clearColor.b, _clearAlpha, _premultipliedAlpha );
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1037,6 +1037,8 @@ function WebGLRenderer( parameters ) {
 		_currentMaterialId = - 1;
 		_currentCamera = null;
 
+		scene.currentCamera = camera;
+
 		// update scene graph
 
 		if ( scene.autoUpdate === true ) scene.updateMatrixWorld();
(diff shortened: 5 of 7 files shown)


Mugen87:
Although this change is convenient in certain situations, I'm not happy with the new signature of `.updateMatrixWorld()`. TBH, it feels a bit...hacky. I think it would be better if your helper has a method `setCamera()` that allows to inject the camera on app level.

arodic:
For example `Object3D` already has `scene` reference. It would be possible to set `scene.activeCamera = camera;` every time `render(scene, camera)` is called. This would provide activeCamera reference to all Object3D instances and feels much less hacky than doing it in `updateMatrixWorld` path. What do you think?

Mugen87:
> For example Object3D already has scene reference.

Um, can you please explain what you mean? Something like `Object3D.scene` does not exist.

> It would be possible to set scene.activeCamera = camera; every time render(scene, camera) is called. This would provide activeCamera reference to all Object3D instances and feels much less hacky than doing it in updateMatrixWorld path. 

Indeed, this approach sounds better than changing the signature of `updateMatrixWorld()`.

arodic:
> Um, can you please explain what you mean? Something like Object3D.scene does not exist.

Basically, you can walk down the scenegraph tree with `Object3D.parent` until you find `isScene === true`.  It's not optimal but it is there.

Ideally, there should be a `.scene` property which gets set/unset inside `add/remove()` methods exactly like `.parent` property, except it would have to be applied to all descendants of added object.

haroldiedema:
I welcome this change as it removes a lot of traversal though the scene graph.

Being able to just do `obj.scene.activeCamera` can have quite a positive impact on performance :+1:

arodic:
@haroldiedema I think `updateMatrixWorld` approach was too hacky and this approach is more elegant and pretty much free.

I changed the property name to `currentCamera` since the same nomenclature is used in WebGLRenderer internally but I do have slight preference for `activeCamera`

Mugen87:
@arodic Can you please also update the docs for [Object3D](https://threejs.org/docs/index.html#api/en/core/Object3D) and [Scene](https://threejs.org/docs/index.html#api/en/scenes/Scene)?

I think this change is a useful addition to the library and should be merged.

> I changed the property name to currentCamera since the same nomenclature is used in WebGLRenderer internally but I do have slight preference for activeCamera

`currentCamera` is a nomenclature which is also used in other engines like Unity. I think it's okay to stay with it.

https://docs.unity3d.com/ScriptReference/Camera-current.html

takahirox:
I have a concern of memory leak.

1. `scene` has a reference to `camera`. Even if user throws away `camera` it won't be released until next `renderer.render()` call. 

2. `scene` and underlying `Object3D`s form a circular reference. To release a scene graph, all the `Object3D`s should be removed from the scene graph.

arodic:
> Can you please also update the docs for Object3D and Scene?

@Mugen87 will do. I can also add relevant tests.

Regarding memory leak concerns by @takahirox

> scene has a reference to camera. Even if user throws away camera it won't be released until next renderer.render() call.

I wouldn't be concerned about this one since (as you said it yourself) the camera reference will be cleared as soon as you render the scene with another camera.

> scene and underlying Object3Ds form a circular reference. To release a scene graph, user needs to remove all the Object3Ds from the scene graph.

This is already the case with `parent` property. The way `scene` property is implemented right now will remove scene reference from all ancestors of removed object. I'll make sure this is covered by the test.

arodic:
@Mugen87 I added documentation and unit tests for both `.parent` and `.scene` property.

I noticed that there are no unit tests for `WebGLRenderer` where test for `scene.currentCamera` should be so I wasn't sure if I should add it there. Let me know.

I also realized that in current implementation scenes have `.scene` property pointing to themselves - which makes the implementation super simple. However, it also assumes that users never add scenes as children to other objects or scenes, which is technically possible. If that happens, the `.scene` property would corrupt on the scene itself (being null or reference to another scene). How should this hypothetical situation be handled?

1. Do nothing since no sane person should ever add scene as a child. (no fix)

2. Define `.parent` property on scene as a getter which **always** returns itself and throws a warning if something tries to set it. (partial fix and a warning)

3. Improve current implementation to support nested scenes and other edge cases. (fully fixed but slightly more complex implementation)

Mugen87:
TBH, I've never seen that somebody is doing this:
```js
object.add( scene );
```
Hence, I personally tend to option one.

arodic:
The only reason I can think of to do `scene.add( scene1 )` would be multi-scene rendering. For example in Unity editor you can load multiple scenes at once and each scene is managed separately while objects are sorted and rendered collectively in the editor. With three.js this can be achieved by adding scenes as children to master(editor) scene. I know scenes are not intended to be used like that, but since `Scene` is also `Object3D` it actually works just fine. However, there are probably other ways to sort and render multiple scenes simultaneously.

takahirox:
> I wouldn't be concerned about this one since (as you said it yourself) the camera reference will be cleared as soon as you render the scene with another camera.

I don't wanna say it's ok. I've seen similar behaviors, that is a reference to certain objects is made in `.render()` and the reference won't be cleared until next `.render()`. They ended up being fixed because they caused memory leak issues in some cases. 

> This is already the case with parent property. The way scene property is implemented right now will remove scene reference from all ancestors of removed object. I'll make sure this is covered by the test.

Ah, I've missed `parent` property... Off topic but it means scene graph won't be released unless all underlying objects are removed from the graph even without this PR? I've never tested, I think I need to check...

BTW can't this be done in app level because added `scene` and `camera` referenced seem to be used by only user code? (I haven't caught on the discussion but just read through the diff.) These properties remind me of `Bone`'s `.skin` property referencing to `SkinnedMesh`. It has been removed because it wasn't used in core, no user seemed to use, and it just made references complex.

    parent.add( child );
    child.userData.scene = parent.userData.scene;

    scene.userData.currentCamera = camera;

donmccurdy:
Note that a few loaders may return a model's content as a `Scene`:

- BabylonLoader
- ColladaLoader
- GLTFLoader
- VRMLLoader
- LegacyGLTFLoader

We could change that, of course, but as-is it would be a common source of nested Scene objects.

arodic:
@takahirox as you probably noticed in #14946 currentCamera is used in the core by the renderer itself.

Making it available through scene would have positive effect on app complexity. This might not be very obvious if you use threejs mostly for nuclear single scene/camera sketches like most of the examples and demos out there in the wild. However, if you attempt to build a more complex application, such as a game engine editor, 3D modeling tool or such, the benefits of currentCamera reference becomes quite obvious and imho justifies it's place in the memory.

arodic:
Closing this pull request in favor of #14953 and another pull request which will follow up shortly.

Issue #14900, opened 2018-09-12
FeatureRequest: Is it possible to add type everywhere ?


Itee:
This is a feature request.

Currently many class in threejs have their own implementation of type = "ClassName"

Is it possible to consider this as a global pattern for all threejs class ?

Mugen87:
AFAIK, `type` is only considered for serialization/deserialization. If a class is no part of the serialization/deserialization process, the `type` property is superfluous.

Itee:
About this [comment](https://github.com/mrdoob/three.js/pull/14901#issuecomment-420632851) the implementation of type property is (i think) missing in lot of serialisable class.

Should it be interesting to inherit an "serializable interface" with type, toJson and fromJson ?

Mugen87:
Related #11266

Itee:
I'm aware of it, thanks.

Mugen87:
How about closing this issue and proceed the discussion in #11266?

PR #14901, opened 2018-09-12
Global type

Itee:
PR about #14900

It define on top level class the type property and remove useless redefinition in subclass.

/!\ I really don't know why but this PR should be unable to be merge automatically.

### DIFF
--- a/examples/js/MorphAnimMesh.js
+++ b/examples/js/MorphAnimMesh.js
@@ -6,8 +6,6 @@ THREE.MorphAnimMesh = function ( geometry, material ) {
 
 	THREE.Mesh.call( this, geometry, material );
 
-	this.type = 'MorphAnimMesh';
-
 	this.mixer = new THREE.AnimationMixer( this );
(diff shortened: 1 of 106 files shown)


haroldiedema:
Just a quick observation, maybe I'm missing some information, in which case just ignore the following, but...

1. Minified builds will have the value of `this.constructor.name` change to a generated variable name such as `da`.
2. Because of (1), deserialization of, for example exported JSON objects, will no longer work on minified builds.

I think the hard coded "type name" is important for use cases like this?

Mugen87:
Besides, there is no need to create a commit for each class. Try to put your changes into a single one.

Itee:
@haroldiedema Nice remark about minification/serialization, you're totally right. And i can't see how use inheritable/minimifiable property for this purpose... damn !

So may be i will need to add type for all serialisable class that doesn't contain it... (exactly the inverse of this PR 😣)

@Mugen87 yes i do for next PR like this

PR #14910, opened 2018-09-14
Replace bi-directional syncing with lazy Euler syncing

WestLangley:
three.js is quaternion-based, and the Euler rotation is provided only for user convenience.

However, in the `Object3D` class, we currently do this:

1. update the quaternion whenever the Euler rotation changes, and
2. update the Euler rotation whenever the quaternion changes.

This PR implements the minimal change to effect the following, more efficient, strategy:

1. update the quaternion whenever the Euler rotation changes, and
2. lazily update the Euler rotation only when it is accessed.

In addition, a simple caching system is used to further prevent unnecessary computations.

This means that the `Quaternion` class no longer needs to update anybody else.

If there is agreement that this approach works, we can proceed to remove all the `onChange` callbacks from the `Quaternion` class (yay!). In addition, as proposed in #11289, we can eliminate some unnecessary closures. But for now, I'd like to keep this PR simple.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -36,20 +36,27 @@ function Object3D() {
 	var quaternion = new Quaternion();
 	var scale = new Vector3( 1, 1, 1 );
 
+	var sync = new Quaternion();
+
 	function onRotationChange() {
 
 		quaternion.setFromEuler( rotation, false );
+		sync.copy( quaternion );
 
 	}
 
-	function onQuaternionChange() {
+	function onRotationAccess() {
 
-		rotation.setFromQuaternion( quaternion, undefined, false );
+		if ( ! sync.equals( quaternion ) ) {
+
+			rotation.setFromQuaternion( quaternion, undefined, false );
+			sync.copy( quaternion );
+
+		}
 
 	}
 
 	rotation.onChange( onRotationChange );
-	quaternion.onChange( onQuaternionChange );
 
 	Object.defineProperties( this, {
 		position: {
@@ -58,7 +65,12 @@ function Object3D() {
 		},
 		rotation: {
 			enumerable: true,
-			value: rotation
+			get: function () {
+
+				onRotationAccess();
+				return rotation;
+
+			}
 		},
 		quaternion: {
 			enumerable: true,


WestLangley:
Any interest in this?

@takahirox ? @Mugen87 ?

takahirox:
I like this idea. But about removing `onChange`, I still want it because of dirty flag introduction #14543. Let's keep discussing in another thread.

Mugen87:
I know this sounds insane...but are there any chances to transform `Object3D.rotation` to type `Quaternion` and remove `Object.quaternion`?

I've always thought that `Object3D.rotation` being of type `Euler` is somewhat a design flaw. One the one hand, it makes it easy for users to change the orientation of the object. But on the other hand since `three.js` internally works with quaternions, we always have to perform this inconvenient sync (in any form whatsoever).

 It's of course okay to have an `Euler` class in the library but I think it would be best to just have `Quaternion.setFromEuler()`. This is similar to what Unity does:

https://docs.unity3d.com/Manual/QuaternionAndEulerRotationsInUnity.html

_When dealing with handling rotations in your scripts, you should use the Quaternion class and its functions to create and modify rotational values._

WestLangley:
This PR is 10 lines. It saves considerable computation, provided it works correctly. That is why I have been waiting for contributors to test it to confirm.

three.js is now quaternion-based. It was not originally. When the change was made, the Euler rotation property was retained for user convenience.

This PR retains the current API, but improves the synchronization. In a future PR, we could remove the rotation property altogether, and ask users to use this pattern instead:
```javascript

var euler = new THREE.Euler( 1, 1, 1, 'YXZ' ); // global
...
euler.x += 0.01;
euler.y += 0.01;
mesh.quaternion.setFromEuler( euler ); 
```
If the user had many meshes, presumably the pattern would be to add `userData.euler` to each mesh.

titansoftime:
Regarding removing Euler rotations from Object3D. This seems to violate one of the core principles of this library.

image

Mugen87:
Working with the [Quaternion](https://docs.unity3d.com/ScriptReference/Quaternion.html) class in Unity seemed to hard to me...

WestLangley:
Closing due to lack of interest.   :-(

takahirox:
Ahhh, sorry I'm still interested in. I was gonna dive into after we've done update matrices optimization.

WestLangley:
@takahirox OK, I'll reopen for now. Thank you for your interest. If you are willing to apply your performance testbed to this PR separately, I would be grateful.

WestLangley:
You will also want to test animations where only quaternions are used. In those cases, the Euler angles will never be computed.

takahirox:
Yes, that's the plan. I'm gonna edit performance example so and evaluate the performance gain.

WestLangley:
>I was gonna dive into after we will have done matrices update optimization.

@takahirox Since your matrix optimization experiment has been put on hold for now, would you be willing to consider this PR?

takahirox:
Yup that's the plan

takahirox:
I made performance test based on `webgl_performance` but somehow I didn't get good result.

With Dev branch `three.js`

https://raw.githack.com/takahirox/webgfx-tests/threejs-pr14910/examples/tests/threejs-pr14910/webgl_performance.html

With `three.js` including this PR

https://raw.githack.com/takahirox/webgfx-tests/threejs-pr14910/examples/tests/threejs-pr14910/webgl_performance_pr14910.html

You can control the objects num by adding `?num=xxx` to URL.

Note that in `webgl_performance_pr14910` objects aren't rotated.

```javascript
// render loop in webgl_performance.html
for ( var i = 0, il = objects.length; i < il; i ++ ) {

	objects[ i ].rotation.x += 0.01;
	objects[ i ].rotation.y += 0.02;

}
```

```javascript
// render loop in webgl_performance_pr14910.html
for ( var i = 0, il = objects.length; i < il; i ++ ) {

	//objects[ i ].rotation.x += 0.01;
	//objects[ i ].rotation.y += 0.02;

}
```

FPS numbers on my Windows10 laptop, bigger is better.

| Browser | `webgl_performance` | `webgl_performance_pr14910` |
| ---- | ---- | ---- |
| Chrome | 32.61 | 29.66 |
| FireFox | 23.91 | 16.68 |

`webgl_performance_pr14910` is slower even without rotating objects. 

I suppose this may be from the existence of getter method defined via `Object.defineProperties()`.

https://github.com/mrdoob/three.js/blob/774dc255774d9ead911f30456ce20e2e84b0c5af/src/core/Object3D.js#L68-L73

I guess accessing any properties may become slower if any getter method exists? By replacing getter method with `value: rotation`, the numbers became 39.26 for Chrome and 25.65 for FireFox respectively.

WestLangley:
Thanks for studying this. Can you please also try an animation example that uses quaternions only, such as `webgl_loader_fbx.html`?

takahirox:
OK, tomorrow.

takahirox:
I ran `webgl_loader_fbx` but I don't see clear difference. I think I need more quaternion update intensive example.

takahirox:
I made another test based on `webgl_animation_skinning_morph`. In this test A`quaternion` is updated via mixer and no one touches `rotation`.

Dev branch Three.js

https://raw.githack.com/takahirox/webgfx-tests/threejs-pr14910/examples/tests/threejs-pr14910/webgl_animation_gltf.html

Dev branch Three.js + this PR

https://raw.githack.com/takahirox/webgfx-tests/threejs-pr14910/examples/tests/threejs-pr14910/webgl_animation_gltf_pr14910.html

`webgl_animation_gltf_pr14910.html` is slower on my Windows10 laptop.

FPS numbers where `num=30`

| Browser | `webgl_animation_gltf` | `webgl_animation_gltf_pr14910` |
| ---- | ---- | ---- |
| Chrome | 42.12 | 39.70 |
| FireFox | 26.30 | 25.20 |

And cpu elapsed time increases by 7% both on Chrome and FireFox.

Perhaps we need to fix `Object.defineProperties( this, ... )` performance issue I've mentioned above if we go with this change.

WestLangley:
Thanks, @takahirox

Avoiding updating Euler angles when they are not needed seemed to me to be a good idea. But if we can't demonstrate a useful benefit to doing so, then I doubt @mrdoob will be interested in pursuing this further.

If someone can suggest a way to fix the performance issue, that would be great; otherwise, I will close this.

takahirox:
> Avoiding updating Euler angles when they are not needed seemed to me to be a good idea.

Yeah, the idea itself is very nice to me. I hope we can find something solutions.

takahirox:
Based on #11298 idea, I moved rotation getter to `Object3D.prototype` from `this` like

```javascript
function Object3D() {

    ....

    this._rotation = rotation;
    this._sync = sync;

    ....

}

Object.defineProperties( Object3D.prototype, {

    rotation: {
        configurable: true,
        enumerable: true,
        get: function () {

            if ( ! this._sync.equals( this.quaternion ) ) {

                this._rotation.setFromQuaternion( this.quaternion, undefined, false );
                this._sync.copy( this.quaternion );

            }

            return this._rotation;

        }
    }

} );
```

Test

https://raw.githack.com/takahirox/webgfx-tests/threejs-pr14910/examples/tests/threejs-pr14910/webgl_animation_gltf_pr14910_2.html

It seems the performance problem has been solved although we need to add `_rotation` and `_sync` properties to `Object3D`.

FPS numbers where `num=30`

| Browser | `webgl_animation_gltf` | `webgl_animation_gltf_pr14910` | `webgl_animation_gltf_pr14910_2` |
| ---- | ---- | ---- | ---- |
| Chrome | 41.83 | 39.98 | 41.93 |
| FireFox | 26.31 | 25.30 | 26.80 |

In this test, FPS number gain is only a little bit and cpu elapsed time decreases by just 1%.

So we might not see good performance improvement with this PR but still might be worth as simpler code because maybe we can remove onChange callback from `Quaternion`.

WestLangley:
Thanks, @takahirox

It seems improvements in performance will have to be part of a larger solution.

I'll close this, then.

takahirox:
Let's revisit if we find any use cases where sync callback is performance bottleneck.

PR #14912, opened 2018-09-14
Update rollup config

HypnosNova:
I add some more rollup plugins and update some configs.

1. Babel plugins: From now on, Three.js could be written in ES6 gramma and babel plugins can compile code into ES5 code. And code in ES6 would be much easier to write and read.
2. ESLint plugins: I put eslint loader so when developers build Three.js, it can check the code style without using "npm run lint". In that way, developers must follow the code style rules, otherwise there would be many eslint errors when build Three.js
3. I remove the glsl loader function from config.js and publish the loader on npm: https://www.npmjs.com/package/rollup-plugin-three-glsl
4. Minify the code, I create the rollup.uglify.config.js, that would be much easier to use.


package.json
```diff
@@ -50,14 +51,23 @@
...
+    "babel-core": "^6.26.3",
+    "babel-plugin-external-helpers": "^6.22.0",
+    "babel-preset-env": "^1.7.0",
     "concurrently": "^4.0.1",
     "electron": "^2.0.8",
     "eslint": "^5.5.0",
     "eslint-config-mdcs": "^4.2.3",
     "google-closure-compiler": "20180805.0.0",
     "http-server": "^0.11.1",
-    "qunit": "^2.6.2",
```
Mugen87: Why do you downgrade `qunit` and `rollup` version?

package.json
```diff
@@ -50,14 +51,23 @@
...
-    "rollup": "^0.65.0"
+    "qunit": "^2.6.1",
+    "rollup": "^0.63.5",
+    "rollup-watch": "^4.3.1",
+    "rollup-plugin-babel": "^3.0.7",
+    "rollup-plugin-eslint": "^5.0.0",
+    "rollup-plugin-three-glsl": "^1.0.1",
+    "rollup-plugin-uglify": "^5.0.2",
+    "uglify-js": "^3.3.28"
```
Mugen87: We recently removed `uglify` (#14871). I don't think we should add it again.

Mugen87:
> Three.js could be written in ES6 gramma and babel plugins can compile code into ES5 code. And code in ES6 would be much easier to write and read.

I think PRs like #14654 and #14695 should be merged first before changing the build configuration.

package.json
HypnosNova: Actually, the qunit  and rollup is latest version in this PR now. I made a mistake at first.( I created the branch based on master ) and after I PR, I found the conflicts and fixed it. So they are latest version now.

arodic:
I would love to see ES6 widely adopted in threejs. I know there has been a lot of discussion about performance and reverse compatibility issues in #11552 and #14654 (and few other issues I cant find right now).

> babel plugins can compile code into ES5 code

If this works and performs more or less same as current code, that would be a very encouraging!

Also, it might be a god idea to squash those 15 commits considering this is a very small PR.

.gitignore
```diff
@@ -8,3 +8,4 @@ npm-debug.log
 .jshintrc
 .vs/
 test/unit/three.*.unit.js
+.package-lock.json
```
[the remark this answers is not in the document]
donmccurdy: Note it's only moved in this PR. Maybe worth filing a separate issue? I don't have a strong preference here but threejs has only devDependencies; there are no prod dependencies to resolve.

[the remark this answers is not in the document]
HypnosNova: I ignore this file because in many times, developers might use npm mirror. In my company, we have a mirror so we install form mirror instead of npm website because it will be faster. And I know many companies have their own mirror like Alibaba and Baidu. If we use mirror, the package-lock file is total different. And we know that three.js doen't depend on any third party library and we use npm just install develop tools like babel, eslint, rollup etc. In that way, I think ignore package-lock file is a better choice. If we commit that file, there might be a lot of conflicts in package-lock

Mugen87:
Closing, see https://github.com/mrdoob/three.js/pull/14912#issuecomment-421541793

PR #14917, opened 2018-09-15
Fixed Raycaster for OrthographicCamera. Fixes #14916

### LINKED ISSUE #14916: Raycaster.setFromCamera is incorrect for back half of OrthographicCamera frustum.
It appears to be failing [here](https://github.com/mrdoob/three.js/blame/dev/src/core/Raycaster.js#L85)

Correct calculation, to start the ray from back of the frustum I think should be:

```javascript
this.ray.origin.set( coords.x, coords.y, camera.near / ( camera.near - camera.far ) ).unproject( camera ); 
```

Notice the sphere position is not updated in back half of the frustum in live example:

* [jsfiddle](https://jsfiddle.net/wmbyhgxr/) (dev branch)

##### Three.js version

- [x] Dev

##### Browser

- [x] All of them

##### OS

- [x] All of them

arodic:


arodic:
Although this PR fixes the broken [example](https://jsfiddle.net/wmbyhgxr/) from #14916, the test appears to catch unexpected result. as far as I can tell, the test is correct. Will investigate more.

src/core/Raycaster.js
```diff
@@ -82,8 +82,8 @@ Object.assign( Raycaster.prototype, {
 
 		} else if ( ( camera && camera.isOrthographicCamera ) ) {
 
-			this.ray.origin.set( coords.x, coords.y, ( camera.near + camera.far ) / ( camera.near - camera.far ) ).unproject( camera ); // set origin in plane of camera
-			this.ray.direction.set( 0, 0, - 1 ).transformDirection( camera.matrixWorld );
+			this.ray.origin.set( coords.x, coords.y, camera.near / ( camera.near - camera.far ) ).unproject( camera ); // set origin in plane of camera
+			this.ray.direction.set( 0, 0, 1 ).transformDirection( camera.matrixWorld );
```
Mugen87: FYI: The default forward vector of a camera is `( 0, 0, - 1 )` and not `( 0, 0, 1 )` in `three.js`.

PR #14929, opened 2018-09-18
Prevent ShapePath#toShapes() from throwing exception

ghost:
For certain inputs (like some converted fonts), `ShapePath#toShapes()` throws an exception since `tmpPoints` happens to be empty and thus `tmpPoints[0]` will be `undefined`. This change simply ignores `tmpPoints` if it is empty.

CC @MichaelBuerge

### DIFF
--- a/src/extras/core/ShapePath.js
+++ b/src/extras/core/ShapePath.js
@@ -188,6 +188,15 @@ Object.assign( ShapePath.prototype, {
 
 			} else {
 
+				if ( ! tmpPoints.length ) {
+
+					// If tmpPoints is empty, the point ('p'-property) in the object
+					// added to newShapeHoles will be undefined.
+					// This causes an error in inPointInsidePolygon() further down.
+					continue;
+
+				}
+
 				newShapeHoles[ mainIdx ].push( { h: tmpPath, p: tmpPoints[ 0 ] } );
 
 				//console.log('ccw', i);


ffdybuster:
No, we didn't investigate this any further. The shapes were produced by http://gero3.github.io/facetype.js/; the resulting shapes happen to have empty paths in some of the letters (mostly umlauts).

WestLangley:
@ffdybuster It is best to try to fix the problem. Maybe file a new issue at https://github.com/gero3/facetype.js/issues and include an example.

ffdybuster:
The root problem seems to be that the shape description starts with two move operations: `"m 3 0 m 540 38 q ..."`. While this can be fixed in gero3/facetype.js, I think it is also a bug in three.js. Calling `ShapePath#moveTo()` twice should not result in a crash when calling `ShapePath#toShapes()` later on. This should either be documented as an invalid operation, or there should be some code preventing the crash. (The simplest code change is the one proposed in this PR; handling it earlier involves more complicated logic.)

Is there a preference on how this should be fixed (add note to doc vs. fix in code), and if a fix in code is preferred, where do you want the fix?

Mugen87:
I would prefer a note in the docs.

WestLangley:
>I prefer a note in the docs.

Please, no. We can't be documenting things that will not work. That would be a long list.

You need to study the code. If a sub-path has zero length, there are problems earlier when computing the winding order of an empty shape.

You can either handle empty shapes silently in this class, or fix `facetype.js` so it does not return empty shapes in the first place. I prefer the latter if possible, and the former only if necessary.

ffdybuster:
I looked again at the code. Making sure that no empty paths are inserted is pretty complicated (if you don't want new book-keeping logic). It's much easier to modify `toShapes` to simply ignore empty-length paths (as this PR currently does). But what one could do instead of what this PR does is to check earlier, like this:
``` .js
		for ( var i = 0, l = subPaths.length; i < l; i ++ ) {

			tmpPath = subPaths[ i ];

			if ( !tmpPath.curves.length ) {

				continue;

			}

			tmpPoints = tmpPath.getPoints();
			solid = isClockWise( tmpPoints );
			solid = isCCW ? ! solid : solid;
```

ffdybuster:
I created an issue at facetype.js. Feel free to close this if you think this is enough.

Mugen87:
It seems https://github.com/gero3/facetype.js/issues/18 is solved so the error should not be triggered anymore by converted fonts.

> Calling ShapePath#moveTo() twice should not result in a crash when calling ShapePath#toShapes() later on. 

I've tried to reproduce this issue here but without success. Notice how the method is called twice in `transformSVGPath()`. The final call of `ShapePath.toShapes()` works.

https://jsfiddle.net/g2q8wb5a/1/

Mugen87:
Closing for now. Feel free to create a new issue if the runtime error still occurs.

PR #14932, opened 2018-09-18
ColladaLoader: Added .setResourcePath()

Mugen87:
See #12942

### DIFF
--- a/examples/js/loaders/ColladaLoader.js
+++ b/examples/js/loaders/ColladaLoader.js
@@ -19,7 +19,7 @@ THREE.ColladaLoader.prototype = {
 
 		var scope = this;
 
-		var path = scope.path === undefined ? THREE.LoaderUtils.extractUrlBase( url ) : scope.path;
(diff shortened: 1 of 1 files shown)


donmccurdy (approved):

examples/js/loaders/ColladaLoader.js
```diff
@@ -19,7 +19,7 @@ THREE.ColladaLoader.prototype = {
 
 		var scope = this;
 
-		var path = scope.path === undefined ? THREE.LoaderUtils.extractUrlBase( url ) : scope.path;
+		var path = ( scope.path === undefined ) ? THREE.LoaderUtils.extractUrlBase( url ) : scope.path;
 
 		var loader = new THREE.FileLoader( scope.manager );
 		loader.load( url, function ( text ) {
```
Mugen87: Um, wait. `setPath()` should affect the original file, right? Then we need something like this, correct?
```js
var fileUrl = ( scope.path === undefined ) ? url : ( scope.path + url );

loader.load( fileUrl, function ( text ) {
```

PR #14936, opened 2018-09-19
Fixed sprite raycast bug when sizeAttenuation is false

### LINKED ISSUE #14934: Raycast problem with SpriteMaterial.sizeAttenuation set to false
Hi Friends!
```
SpriteMaterial.sizeAttenuation = false
```
seems not working as expected with `rayCaster`.
I doubt the sprite size(dimension) estimation may be a little bit off, ray is very hard to hit them after apply `scale()`, especially scale down. 
I can upload a simplified code that regenerate this problem a bit later.

- [x] r96

##### Browser
- [x] All of them

Thanks!

06wj:
Fixed #14934

[demo](https://raw.githack.com/06wj/three.js/patch0_build/examples/webgl_raycast_sprite.html)

If using an orthographic camera the results is wrong.

### DIFF
--- a/examples/webgl_raycast_sprite.html
+++ b/examples/webgl_raycast_sprite.html
@@ -87,6 +87,26 @@
 			sprite.material.rotation = Math.PI / 3;
 			group2.add( sprite );
 
+
+			var num = 20;
+			var radius = 10;
+			while(num --) {
+
+				var x = Math.random() - 0.5;
+				var y = Math.random() - 0.5;
+				var z = Math.random() - 0.5;
+
+				var sprite = new THREE.Sprite( new THREE.SpriteMaterial( { color: '#96f' } ));
+				sprite.position.set( x, y, z );
+				sprite.position.normalize();
+				sprite.position.multiplyScalar( radius );
+				sprite.scale.set(0.05 * Math.random(), 0.05 * Math.random(), 0.05 * Math.random());
+				sprite.material.rotation = Math.random() * Math.PI * 2;
+				sprite.material.sizeAttenuation = false;
+				group.add( sprite );
+
+			}
+
 			window.addEventListener( 'resize', onWindowResize, false );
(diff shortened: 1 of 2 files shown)


06wj:
If the camera is orthographic, the ```sizeAttenuation``` has no effect on render result, we should not ```worldScale.multiplyScalar( - mvPosition.z );```.

WestLangley:
>I can't check if it is the camera is orthographic.

Right. That is what I was afraid of...

06wj:
Maybe we should add a note if the camera is orthographic do not set sizeAttenuation of the sprite

WestLangley:
The user could be using a custom `ShaderMaterial` or `SpriteMaterial`.
The material may, or may not, have a `sizeAttenuation` property (`undefined`, `true`, or `false`).
The user may, or may not, be using an orthographic camera when raycasting.
There are 2 x 3 x 2 = 12 cases.

06wj:
#14888 Maybe this is a solution.

WestLangley:
A raycast can emanate from any point, in any direction -- it does not have to emanate from a camera. 

Consequently, raycasting against a `Sprite` that is camera-dependent and sized in screen-space does not necessarily make sense.

06wj:
Sprite is a 3d object that is always facing the camera. We can do this by dynamically setting the vertices in js or using shader.Just because we chose shader for performance reasons makes raycast implementation difficult, I don't think Sprite raycast is meaningless.

WestLangley:
>Sprite is a 3d object that is always facing the camera

Sprite is a 2D object, actually.

>This is the main reason raycast relies on cameras

Raycasting does not require a camera. It is the sprite that is camera-dependent.

>I don't think Sprite raycast is meaningless

Raycasting against a sprite using a ray that does not emanate from a camera is pretty meaningless because there is no way to determine how big the sprite should be, or how it should be oriented.

06wj:
Fixed in #16423, so close it.

PR #14953, opened 2018-09-22
Added scene property to Object3D.

arodic:
This is a new feature for `Object3D` similar to `.parent` property except it points to top-most `Scene` ancestor. It is a first step towards #14888

### DIFF
--- a/docs/api/en/core/Object3D.html
+++ b/docs/api/en/core/Object3D.html
@@ -112,6 +112,9 @@
 		<p>Object's parent in the [link:https://en.wikipedia.org/wiki/Scene_graph scene graph]. An object can have at most
 		one parent.</p>
 
+		<h3>[property:Scene scene]</h3>
+		<p>Object's top-most ancestor ([page:Scene]) if object is added to a scene.</p>
+
 		<h3>[property:Vector3 position]</h3>
 		<p>A [page:Vector3] representing the object's local position. Default is (0, 0, 0).</p>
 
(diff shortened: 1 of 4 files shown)


WestLangley:
> first step towards #14888

That PR has been closed. Can you please explain your use case for this feature? Can you use `onBeforeRender()`, instead?

arodic:
I closed the PR in favor of two separate PRs

> Can you use onBeforeRender(), instead?

Now that you mention it, perhaps I could. I wasn't aware of  `onBeforeRender()` arguments since they are not documented in the code. Perhaps this is an argument towards using @jsdoc :) I will investigate this path.

arodic:
@WestLangley I was very excited to learn that `onBeforeRender()` has useful arguments (scene, camera etc...). It could have been a neat way to pass camera and scene information to objects. However, it appears that `.onBeforeRender()` does not get called on objects without materials (`Object3D`, `Group`, etc.) since they are not actually renderable.

Would it be possible to modify the renderer to call `.onBeforeRender()` on `Object3D`, `Group`, etc...?

WestLangley:
>Would it be possible to modify the renderer to call .onBeforeRender() on Object3D, Group, etc...?

That is not up to me, but I expect you will need to describe your use case.

arodic:
OK, Im' gonna close this PR as well as #14956 in favor of pursuing `onBeforeRender()` approach. I will describe the use case in a new issue.

PR #14956, opened 2018-09-22
Added currentCamera property to Scene.

arodic:
Following #14953 this pull request allows any object in a rendered scene to access current camera. 

For example, you can do `mesh.scene.currentCamera`.

This small change can have positive impact on application-side complexity. As previously discussed in 
#14888 it can be used, for example, to make helper objects with camera-dependent transformations.

The only concern that was brought up so far by @takahirox is theoretically possible memory leak because the most recent camera will be referenced on the scene until the scene gets rendered with a different camera. While I understand the potential memory leaks for lingering post-render references, I cant imagine a scenario where camera doesn't get released for GC eventually while the scene itself does.

### DIFF
--- a/docs/api/en/scenes/Scene.html
+++ b/docs/api/en/scenes/Scene.html
@@ -44,6 +44,9 @@
 		If not null, sets the background used when rendering the scene, and is always rendered first. Can be set to a [page:Color] which sets the clear color, a [page:Texture] covering the canvas, or a [page:CubeTexture]. Default is null.
 		</p>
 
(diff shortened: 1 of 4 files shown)


WestLangley:
>This small change can have positive impact on application-side complexity. As previously discussed in
#14888 it can be used, for example, to make helper objects with camera-dependent transformations.

That sounds application-specific to me. Can you handle this in your app by setting `.currentCamera` in `scene.onBeforeRender()`?

arodic:
Same comment as #14953 I will investigate.

PR #14957, opened 2018-09-24
Volume rendering example: refactor and clean up

WestLangley:
As proposed in https://github.com/mrdoob/three.js/pull/14847#issuecomment-423853738.

### DIFF
--- a/examples/js/shaders/VolumeShader.js
+++ b/examples/js/shaders/VolumeShader.js
@@ -8,10 +8,10 @@
 
 THREE.VolumeRenderShader1 = {
 	uniforms: {
-        "u_size": { value: [1, 1, 1] },
+        "u_size": { value: new THREE.Vector3( 1, 1, 1 ) },
         "u_renderstyle": { value: 0 },
         "u_renderthreshold": { value: 0.5 },
-        "u_clim": { value: [0.0, 1.0] },
+        "u_clim": { value: new THREE.Vector2( 1, 1 ) },
         "u_data": { value: null },
         "u_cmdata": { value: null }
     },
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

looeee:
> I had put renderer.setPixelRatio( window.devicePixelRatio ); in the onWindowResize handler

My understanding is that this is best practice because of multi-monitor setups where monitors can have different pixel ratios. 

However as far as I'm aware none of the examples currently do this.

almarklein:
> because of multi-monitor setups where monitors can have different pixel ratios.

Yes, and users hitting those zoom buttons inside the browser settings (or Ctrl+scroll) :)

looeee:
Doesn't the meta tag prevent this being an issue?

`<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">`

takahirox:
We don't put stats in the example? I think raycasting based Volume rendering is heavy, good to know fps.

almarklein:
Well, we moved to only drawing when needed (interaction, gui changes) so the fps stat becomes rather unreliable, so I removed it.

takahirox:
Ah, ok. I didn't check the implementation in the detail.

WestLangley:
@almarklein 

>I think this might be a typo: "u_clim": { value: new THREE.Vector2( 1, 1 ) }

Yes. It doesn't matter though. Could default to ( 0, 0 ) just as well.

>There is a bit of repetition with the setting of uniforms. Is there a reason why you don't simply call updateUniforms after instantiating the uniforms object?

What you say is true. However, `updateUniforms()` calls `render()`, which we were not ready for.

Actually, `render()` should be called after the mesh is added. That was an omission. You could call `updateUniforms()` there, instead, but that would be sort of confusing, I think.

>does Three.js apply throttling

Not to my knowledge.

I am also not familiar with the browser-zoom issue you refer too.

PR #14968, opened 2018-09-26
Docs: Reword light.distance explanation.

donmccurdy:
Per https://github.com/mrdoob/three.js/pull/14962#discussion_r220306665, /cc @WestLangley.


mrdoob:
Thanks!

PR #14974, opened 2018-09-26
GLTFLoader Extension Example: Fix Spotlight

takahirox:
This PR fixes spot light in glTF extensions example broken by #14962

Before
image

After
image

BTW, did anyone confirm lights in all the example keep working fine?

### DIFF
--- a/examples/webgl_loader_gltf_extensions.html
+++ b/examples/webgl_loader_gltf_extensions.html
@@ -211,6 +211,7 @@
 					spot1.position.set( 10, 20, 10 );
 					spot1.angle = 0.25;
 					spot1.penumbra = 0.75;
+					spot1.decay = 0.0001;
 
 					if ( sceneInfo.shadows ) {
 
--- a/examples/webgl_loader_gltf_extensions.html
+++ b/examples/webgl_loader_gltf_extensions.html
@@ -115,9 +115,9 @@
 					url: './models/gltf/Monster/%s/Monster.gltf',
 					author: '3drt.com',
 					authorURL: 'http://www.3drt.com/downloads.htm',
-					cameraPos: new THREE.Vector3( 30, 10, 70 ),
-					objectScale: new THREE.Vector3( 0.4, 0.4, 0.4 ),
-					objectPosition: new THREE.Vector3( 2, 1, 0 ),
+					cameraPos: new THREE.Vector3( 3, 1, 7 ),
+					objectScale: new THREE.Vector3( 0.04, 0.04, 0.04 ),
+					objectPosition: new THREE.Vector3( 0.2, 0.1, 0 ),
 					objectRotation: new THREE.Euler( 0, - 3 * Math.PI / 4, 0 ),
 					animationTime: 3,
 					addLights: true,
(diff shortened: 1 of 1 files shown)


examples/webgl_loader_gltf_extensions.html
```diff
@@ -211,6 +211,7 @@
 					spot1.position.set( 10, 20, 10 );
 					spot1.angle = 0.25;
 					spot1.penumbra = 0.75;
+					spot1.decay = 0.0001;
```
takahirox: Let me know if there is any other better way to fix.

takahirox:
@donmccurdy 

BTW glTF Monster + light extension with #14962 (without this PR)

image

Is this glTF file correct?

donmccurdy:
> BTW, did anyone confirm lights in all the example keep working fine?

I think the only example using physically-correct lights, other than the glTF examples, was webgl_lights_physical.html. I did check it. My change should not affect the default lighting mode. But maybe we should have some more examples with physical lighting..

donmccurdy:
I'm not sure about the glTF lights monster model. The extension is still unfinished, and we've marked it as experimental in three.js docs. Let's revisit that in a couple weeks, I think we will be able to have an updated model and compare visual results with some other engines.

examples/webgl_loader_gltf_extensions.html
mrdoob: Maybe changing the `distance` or the `intensity` makes more sense?

mrdoob:
> I think this bug fix should be in r97 unless it's too late.

Sure thing. We can merge into `master` and `gh-pages` as soon as we decide what the right fix is.

examples/webgl_loader_gltf_extensions.html
donmccurdy: `distance` won't help as the default is infinite, but is there an `intensity`  value that looks good for all models in this example? It would be nice to use that, as it's physically correct, but may take some tuning for each model.

takahirox: Yeah, trying `intensity` but seems we need tuning for each model...

donmccurdy: Maybe just swap the spotlight for a second directional light. 😇

takahirox:
> My change should not affect the default lighting mode

    if( decayExponent > 0.0 ) {

    #if defined ( PHYSICALLY_CORRECT_LIGHTS )

    ... snip ...

    #else

        return pow( saturate( -lightDistance / cutoffDistance + 1.0 ), decayExponent );

    #endif

    }

	return 1.0;

    #if defined ( PHYSICALLY_CORRECT_LIGHTS )

        ... snip ....

    #else

        if( cutoffDistance > 0.0 ) {

            return pow( saturate( -lightDistance / cutoffDistance + 1.0 ), decayExponent );

        }

        return 1.0;

    #endif

takahirox:
I think this PR is ok now? https://rawgit.com/takahirox/three.js/FixGLTFExampleLight/examples/webgl_loader_gltf_extensions.html

donmccurdy (approved):
This looks good except that `Monster/glTF-Lights` isn't showing up... I think that should be fixed by https://github.com/mrdoob/three.js/pull/14978.

takahirox:
> Monster/glTF-Lights isn't showing up... I think that should be fixed by #14978.

Oh, I wasn't aware of that. Good catch.

takahirox:
/ping @mrdoob Do you still have any concerns in this PR?

mrdoob:
Thanks!

PR #14981, opened 2018-09-27
EXAMPLES: added per-vertex ao (cpu) example

looeee:
per_vertex_ao

[Live](https://rawgit.com/looeee/three.js/per_vertex_ao/examples/webgl_per_vertex_ao.html)

Based on the technique by [Bunnel](https://developer.nvidia.com/gpugems/GPUGems2/gpugems2_chapter14.html) in  GPUGems 2.

I've implemented this as two CPU passes (see the linked page for details). 

**Pass one**: calculate surface element areas
**Pass two**: calculate the element to element occlusion and save as vertex color buffer

The advantages of this approach are:

* can be pre-calculated 
* Works with any three.js material type, so shadows etc. work as normal

Disadvantages:
* It's slow
* won't work for animated models

Most implementations of this that I've seen do **pass two** in a shader - then this can work for animated models. However, it's likely that would prove too slow for real time WebGL applications. 

**Possible improvements**

* **Pass two** can be performed multiple times to get a more accurate result. 
* The _bent normals_ are calculated but currently ignored since after a single pass they don't seem to give a good visual result. 
* This is currently O(n^2) on the number of vertices. Various methods for building up a hierarchy of the surface elements according to their distance from each other are possible which could reduce this to O(n log n) and would certainly be required if implementing this as a shader pass. 
* [Hoberock and Jia, GPU Gems 3](https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch12.html) describe a number of additional methods that can be used to reduce artefacts.


Mugen87:
Nice example!

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+
+				// ignoring bent normals for now since with a single pass they don't improve appearance of the current model
+				// geometry.attributes.normal.set( bentNormals );
+
+				var vertexColors = [];
+				for( var k = 0; k < colors.length; k ++ ) vertexColors.push( colors[ k ], colors[ k ], colors[ k ] );
+
+				geometry.addAttribute( 'color', new THREE.BufferAttribute( new Float32Array( vertexColors ), 3 ) );
+
+				}
```
Mugen87: Minor formatting issue. One tab too much.

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+					}
+					model.material = materials[ num ];
+					model.material.needsUpdate = true;
+
+				} )
+			}
+
+			function getAllIndexes(arr, val) {
+
+				var indexes = [];
```
Mugen87: I think I would use the term `indices`.

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+
+					if ( arr[i] === val ) indexes.push( i );
+
+				}
+
+				return indexes;
+
+			}
+
+			function saturate( num ) {
```
Mugen87: I would write it like this:
```js
function saturate( num ) {

    return THREE.Math.clamp( num, 0, 1 );

}
```

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+					model.material = materials[ num ];
+					model.material.needsUpdate = true;
+
+				} )
+			}
+
+			function getAllIndexes(arr, val) {
+
+				var indexes = [];
+				for( var i = 0; i < arr.length; i++) {
```
Mugen87:
```js
for ( var i = 0; i < arr.length; i ++ ) {

File: examples/webgl_per_vertex_ao.html [hunk 5]
```diff
@@ -0,0 +1,450 @@
...
+		</div>
+
+		<script src="../build/three.js"></script>
+		<script src="js/libs/dat.gui.min.js"></script>
+
+		<script src="js/controls/OrbitControls.js"></script>
+
+		<script src="js/loaders/GLTFLoader.js"></script>
+
+		<script src="js/Detector.js"></script>
```
Mugen87: `Detector` is deprecated. You definitely need to rebase your branch.

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+		<script src="js/libs/stats.min.js"></script>
+
+		<script>
+
+			if ( ! Detector.webgl ) Detector.addGetWebGLMessage();
+
+			var container, stats, controls;
+			var camera, scene, renderer, light, model, materials, envMap, shadowLight;
```
Mugen87: It seems `light` is not used.

examples/webgl_per_vertex_ao.html
```diff
@@ -0,0 +1,450 @@
...
+				]
+
+			}
+
+			function buildGUI() {
+
+				var gui = new dat.GUI( { width: 300 } );
+				gui.open();
+
+				params = {
```
Mugen87: `params` has no declaration.

looeee: Fixed!

examples/webgl_per_vertex_ao.html
looeee: Thanks, removed it.

examples/webgl_per_vertex_ao.html
looeee: Updated.

examples/webgl_per_vertex_ao.html
looeee: Yup, there were quite a few... 😓 Fixed now

examples/webgl_per_vertex_ao.html
looeee: Linter errors should be fixed now.

examples/webgl_per_vertex_ao.html
looeee: Thanks for reviewing 😄

donmccurdy:
In a real use case AO should not be multiplied against baseColor directly — two solutions here:

1. (3-4 LOC + dependency) Use NodeMaterial to enable per-vertex AO. Roughly:

```js
material.ao = new THREE.AttributeNode( 'ao', 'float' );
```

2. (20-30 LOC) Construct a 1x256 B/W gradient with Canvas API, assign as `material.aoMap`, and set vertex `uv2` based on its AO value.

If these are too complex for an otherwise concise example, I'd just include a comment above this line to the effect of:

> Ambient occlusion should not really be stored as a base color, or multiplied directly against color information. This is a hack to display the output of AO computation.

I may also have some sample models where you could retain base color along with the AO and show a nice effect, such as:

after

looeee:
> Use NodeMaterial to enable per-vertex AO

I'd rather avoid this for the moment, since the main point in implementing this in it's current form is to demonstrate that this technique can be used with the default, unmodified three.js materials. If `NodeMaterial` is moved into the core at some point then we can revisit this. 

> Construct a 1x256 B/W gradient with Canvas API, assign as material.aoMap, and set vertex uv2 based on its AO value.

I like this idea, I'll test it in the morning and see how it looks. 

> I may also have some sample models where you could retain base color along with the AO and show a nice effect

Sure, if @mrdoob is OK with adding another model to the repo 😄

donmccurdy:
> I like this idea, I'll test it in the morning and see how it looks.

As it happens I have code sitting around, so if this helps:

```js
const aoTexture = (() => {
  const canvas = document.createElement('canvas');
  canvas.width = 256;
  canvas.height = 1;
  const ctx = canvas.getContext('2d');
  const grd = ctx.createLinearGradient(0, 0, 255, 0);
  grd.addColorStop(0, '#000');
  grd.addColorStop(1, '#fff');
  ctx.fillStyle = grd;
  ctx.fillRect(0, 0, 256, 1);
  document.body.appendChild(canvas);
  return new THREE.CanvasTexture(canvas);
})();

// Write UV set and add AO map.
const numVertices = ao.length;
const uv2Data = new Float32Array(numVertices * 2);
for (let i = 0; i < numVertices; i++) {
  uv2Data[i * 2] = uv2Data[i * 2 + 1] = 1 - ao[i];
}
const uv2 = new THREE.BufferAttribute(uv2Data, 2);
mesh.geometry.addAttribute('uv2', uv2);
mesh.material.aoMap = aoTexture;
```

But in any case, very nice example! :)

WestLangley:
Unfortunately, this is not a good example to be used with `MeshStandardMatereial` which is physically-based. Ambient occlusion occludes light, it does not tint the albedo, as you are doing with your vertex colors. It also occludes only ambient light, not all light.

You either need to use your technique to create an AO map, and assign the map as a property of `MeshStandardMaterial`, or use `MeshPhongMaterial` or `MeshLambertMaterial` instead, where AO vertex colors are more acceptable as a hack.

The latter technique is already demonstrated in the [minecraft_ao] example(https://threejs.org/examples/webgl_geometry_minecraft_ao.html) example; the former approach should probably be implemented with modeling software.

In light of that, I'd suggest using a different model with a different material.

Perhaps it would be interesting to use your technique on the minecraft_ao example to see the difference it makes.

looeee:
@donmccurdy I've implemented this as an `aoMap` as per your suggestions, using a pre-generated gradient map, and the live example above is updated:

```js
// 1x256 B/W gradient
var aoMap = new THREE.TextureLoader().load( 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAQAAAAABCAYAAAAxWXB3AAAAIElEQVQ4T2NkYGD4z8TExAACo/RoOIymg5GTDxgZGRkA/oMD/vOwS0YAAAAASUVORK5CYII=' );
```

@WestLangley this is now being applied as an `aoMap` rather than multiplied with the albedo. Does this look better to you now? I'm not tied to the idea of using `MeshStandardMaterial` with this, rather I chose it because it's the default material for this model.

> the former approach should probably be implemented with modeling software. 

Agreed that generating an ao map in modelling software would give better results than the example as it is now. However, the purpose of this example is to demonstrate a simple skeleton that can be used to expand on this technique - in particular, the "surface element areas" generated in **pass one** can be used to create high quality GI + AO, where the AO is generated on the fly for animated models. That's something that can't be done in modelling software. 

There are quite a few examples of **pass 2** being implemented in  a shader around the web, but none that I found demonstrating how to do it on the CPU which I part of the reason that I chose to do it this way. If you think it would be better implemented as a shader, perhaps we can find a simple animated model that can be used to show realtime GI + AO using this approach.

looeee:
@WestLangley thanks for the suggestions, I've simplified the GUI. 

@donmccurdy I'd like to test with your model, can you share it here?

WestLangley:
@looeee Much better!

I do think your demo is better-served with a material other than Standard -- such as Lambert or Basic, for example.

looeee:
@donmccurdy unfortunately that model doesn't produce an indexed geometry. The `calculateAreas` function, as I've currently written it, requires indexes. It could be rewritten to convert to `Geometry` and use the `faces` instead, but that would probably be much slower.

WestLangley:
>The calculateAreas function, as I've currently written it, requires indexes.

That's unfortunate.

You can try your algorithm on other geometries:
```javascript
model.geometry = new THREE.TorusKnotBufferGeometry( .4, .15, 256, 32, 3, 7 );
```

screen shot 2018-10-06 at 3 26 26 pm

donmccurdy:
Here are two indexed versions of the same `monu7.glb` file:

[monu7.zip](https://github.com/mrdoob/three.js/files/2453640/monu7.zip)

Unoptimized voxel art turns out to be a perfect test case for glTF+Draco... the model goes from 13MB to 21KB, for 99.8% size savings. 😅

donmccurdy:
I'm curious to see results with those, but we unfortunately can't use them in an example at this point. The sample repo does not list a license, although I'm asking about it: https://github.com/ephtracy/voxel-model/issues/22

looeee:
> The calculateAreas function, as I've currently written it, requires indexes.

>> That's unfortunate.

@WestLangley I've rewritten the function to work with any geometry, using an intermediate `Geometry`. Surprisingly, it's actually _much_ faster (~130ms vs 1500ms for the helmet model on my laptop). 

@donmccurdy thanks for sharing the model. For voxel based models, calculating the areas is redundant since the area of each face is the same. Also, I haven't been able to get this work with that model so far, I guess the complexity is a bit much for the current algorithm. How about we stick with the helmet for now, and revisit this when we hear back about the license?

WestLangley:
>I've rewritten the function to work with any geometry

Does it produce better results than what you see in my screenshot above? That does not appear correct to me.

looeee:
> That does not appear correct to me.

This is partially to do with the line `if ( d2 < 4 * emitterArea ) {`, which is used ( in the absence of creating a distance based hierarchy of faces ) to prevent too much calculation. Basically, we assume that faces beyond a certain distance from the current face will have a very weak occlusion effect and ignore them. 

Since I'm doing a single run of **Pass 2** here, this technique creates a too dark occlusion effect. To account for this I had set the cutoff to a pretty low value, which worked well for the helmet but not for the torus knot. I've updated the line above to give a stronger occlusion effect and explicitly named the parameter: `if ( d2 * d2 < cutoffParam * emitterArea ) {`.

In the case of the torus knot, you can now see that the hole in the centre is occluded, at the expense of about 33% increase in computation time. 

torusknot

Aside from that, I guess the torus knot is a bad geometry for this technique since the vertex density along the curve boundaries is low, leading to the artefacts that you see in the image. Papers discussing this technique acknowledge that one of its main limitations is that it's often neccessary to manually increase vertex density in areas of the model that will experience heavy AO. 

Also, the paper by Hoberock and Jia in GPUGems 3 that I linked above describe a number of additional techniques that could be used to help reduce these artefacts (in particular see the section **Smoothing Out Discontinuities**). They are too complex too be used in this simple example though. Perhaps at a future date we could explore creating a more fully featured GPU based version of this.

looeee:
@WestLangley do you have any further objections to this?

WestLangley:
>@WestLangley do you have any further objections to this?

The decision is not up to me, I think it is best to leave the computation of AO maps to modeling tools. Your algorithm seems to have significant limitations.

So sorry not to be more supportive, @looeee. I see you have worked hard on this...

looeee:
> So sorry not to be more supportive

It's OK, I'd rather honesty. Thanks for taking the time to evaluate it. If this doesn't get accepted here I'll move it into a separate repo. 

> Your algorithm seems to have significant limitations

1. It's not my algorithm, it's the one described by Bunnel and linked to above. 
2. The limitations of the current implementation are clearly discussed in the two GPUGems papers linked above, along with possible solutions. However, implementing these solution would make the example considerably more complex than it currently is.

The example as it currently stands is supposed to be a _absolute minimal skeleton_ of the technique. I chose to implement it like this for two reason: 

a) there are no three.js per vertex ao examples on the web (that I could find)
b) there are literally no CPU based implementations of this technique, anywhere (that I could find)
c) this was the simplest possible implementation and my spare time is very short at the moment 😅 

If other people agree that it's not a useful example in it's current form then I'm happy to close this for now and improve it when I have more time.

Mugen87:
From an educational perspective, I think this example is very useful. Especially since you have linked to the respective article of Nvidia.

mrdoob:
@looeee Do you know how https://twitter.com/super_eggbert/status/1150418776019623942 differs from the approach in this PR?

looeee:
> Do you know how https://twitter.com/super_eggbert/status/1150418776019623942 differs from the approach in this PR?

No idea, although that is what reminded me to close this PR 😅  
I asked him on Twitter what technique he's using, I'll update here if he responds.

looeee:
> I asked him on Twitter what technique he's using

He's using an algorithm that he wrote himself, raycasting from vertices and then attenuating the AO based on the distance to the other vertices that the ray hits using `Math.min(1, (distance * distance) / ao_distance)`. Pretty simple and looks like it gives nice results, but it's presumably much slower than the algorithm I used here.

mrdoob:
I'm a way, it's also similar to the technique implemented here:
https://threejs.org/examples/#webgl_simple_gi

looeee:
Yes, GI and AO techniques are often pretty similar. GI is a bit more complex though because you need to consider the color.

Issue #14919, opened 2018-09-15
NodeMaterial: Instancing API


donmccurdy:
I believe NodeMaterial could support an expressive instancing API, without requiring users to modify builtin shaders. The [instancing • lambert](https://github.com/mrdoob/three.js/blob/master/examples/webgl_buffergeometry_instancing_lambert.html) example shows the simplest pre-existing approach, and is still complex enough that users will find it challenging to instance a 3D model created in a DCC tool.

Proposal:

```js
import * as THREE from 'three';
import {
  StandardNodeMaterial,
  InstanceTransformNode,
  AttributeNode,
  FloatNode,
  OperatorNode,
  TextureNode
} from 'three/js/nodes/THREE.Nodes';

var material = new StandardNodeMaterial();

// Configure shared material properties as usual.
material.metalness = new FloatNode( 0 );

// Configure material.instanceTransform to reference 1-3 InstancedBufferAttribute nodes.
material.instanceTransform = THREE.InstanceTransformNode(
  new AttributeNode( 'instancePosition', 'vec3' ),
  new AttributeNode( 'instanceQuaternion', 'vec4' ),
  new AttributeNode( 'instanceScale', 'vec3' )
);

// Configure per-instance material properties using node expressions and InstancedBufferAttribute.
material.roughness = new AttributeNode( 'instanceRoughness' );
material.color = new OperatorNode(
  new TextureNode( baseColorTexture ),
  new AttributeNode( 'instanceColor' ),
  OperatorNode.MUL
);
```

With this API, and as loaders enable the NodeMaterial system, instancing models created in DCC tools becomes much easier:

```js
var treePositions = new Float32Array(  [
  0, 0, 0,
  10, 0, 0,
  20, 0, 0
] );

var loader = new THREE.GLTFLoader();
loader
  .setUseNodes( true )
  .load( 'tree.glb', ( gltf ) => {
    var model = gltf.scene.children[ 0 ];

    model.geometry = new THREE.InstancedBufferGeometry().fromBufferGeometry( model.geometry );
    model.geometry.addAttribute( 'instancePosition', new THREE.InstancedBufferAttribute( treePositions, 3 );
    model.material.instanceTransfom = new THREE.InstanceTransformNode( THREE.AttributeNode( 'instancePosition' ) );

    scene.add( model );
  } );
```

[...]

PR #15000, opened 2018-10-02
Instance-ready materials

sweetenr:
- based on the instanced lambert example, I updated the standard three.js materials ( basic, lambert, phong, physical, standard, toon) to be instance friendly.

( https://github.com/mrdoob/three.js/blob/master/examples/webgl_buffergeometry_instancing_lambert.html )

- tested the updated shader materials with all relevant texture maps per material for both instance and single use, everything seems to check out

### DIFF
--- a/src/renderers/shaders/ShaderChunk.js
+++ b/src/renderers/shaders/ShaderChunk.js
@@ -94,6 +94,7 @@ import cube_frag from './ShaderLib/cube_frag.glsl';
 import cube_vert from './ShaderLib/cube_vert.glsl';
 import depth_frag from './ShaderLib/depth_frag.glsl';
 import depth_vert from './ShaderLib/depth_vert.glsl';
+import depthRGBA_vert from './ShaderLib/depthRGBA_vert.glsl';
 import distanceRGBA_frag from './ShaderLib/distanceRGBA_frag.glsl';
 import distanceRGBA_vert from './ShaderLib/distanceRGBA_vert.glsl';
 import equirect_frag from './ShaderLib/equirect_frag.glsl';
@@ -216,6 +217,7 @@ export var ShaderChunk = {
 	cube_vert: cube_vert,
 	depth_frag: depth_frag,
 	depth_vert: depth_vert,
+	depthRGBA_vert: depthRGBA_vert,
 	distanceRGBA_frag: distanceRGBA_frag,
 	distanceRGBA_vert: distanceRGBA_vert,
 	equirect_frag: equirect_frag,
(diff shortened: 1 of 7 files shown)


WestLangley:
Instancing is infinitely flexible. `webgl_buffergeometry_instancing_lambert.html` demonstrates an _example_ of instancing. You can't assume users will require the same instanced attributes as in the example.

rsweeten:
@WestLangley I didn't, this update was JUST adding ifdefs to the shaders for MeshLambert etc and adding a depthRGBA shader to the ShaderLib so users can use the 3js materials for either instance or single use without having to override them in their code.

rsweeten:
@WestLangley oh yea..  I get it, sorry.  I'll think about this some more :)

donmccurdy:
This is a clean and concise PR, thanks for taking the time to write it up.  ❤️

I do agree with @WestLangley that a general solution is quite hard here, and may require a certain amount of configurability. Another option is proposed in https://github.com/mrdoob/three.js/issues/14919.

Just curious, do you have a particular use case for instancing yourself? Or simply wanted to package the feature up better for other three.js users?

rsweeten:
Thanks @donmccurdy :) 

I totally agree w/ @WestLangley , I just hadn't thought it through! 

I have several use cases and also wanted to help other users interested in getting going with instancing. 

As for use cases, my partner and I have an art practice and are exploring WebGL. We've been iterating an app that generalizes how we create works with 3js for live and exhibition applications. It's basically an angular app that provides a UI for non-coders that generates a JSON file, and then another JS/Express app that translates the JSON to 3js and binds sensor/gamepad/midi inputs to scene elements. We also use this app for the occasional adjunct teaching gig.

Anyway, I recently built something that tried to add thousands of meshes to a scene had hit a hard limit, which led me to instancing. 

I've never used the node API, or really knew about it until you pointed it out. I don't see any docs on it, so I'll read through the code. Is the design intent of the node API similar to something you'd find in other GL applications, like nodes in houdini? 

btw, I [expletive] love 3js. I'm pretty new at it, and am just learning glsl, but the breadth to 3js is amazing

donmccurdy:
> I've never used the node API, or really knew about it until you pointed it out. I don't see any docs on it, so I'll read through the code. Is the design intent of the node API similar to something you'd find in other GL applications, like nodes in houdini?

It's still pretty experimental, but I hope we'll have documentation soon. The idea is similar to Houdini's, but strictly for generating shaders (and not geometry). Unity Shader Graph and Blender nodes are also good for comparison, among others.

PR #15003, opened 2018-10-02
Update OrbitControls.js

yaroslavkhudchenko:
cleaned a bit comments and code align


WestLangley:
Thank you for your contribution. In this case, however, the existing code is probably fine as-is.

Your changes to the comment lines are subjective, and the other changes violate the white-space rules.

You may find https://zz85.github.io/mrdoobapproves/ helpful.

PR #15013, opened 2018-10-05
Configurable Object3D.position/scale/rotation/quaternion properties

jbaicoianu:
Object3D's position, scale, rotation, and quaternion properties are currently set up via a call to Object.defineProperties(), which sets them up as immutable references to an instance of their underlying type.  This pull request allows the user to reconfigure those properties after the fact, if they know what they're doing.

This can be useful if you want to "bind" two objects together by pointing them at the same underlying position and/or orientation objects.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -54,18 +54,22 @@ function Object3D() {
 	Object.defineProperties( this, {
 		position: {
 			enumerable: true,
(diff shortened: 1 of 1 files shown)


WestLangley:
Duplicate of #11134.

Mugen87:
Closing in favor of #11134.

PR #15043, opened 2018-10-12
Add set() methods to perspective and orthographic camera.

XanderLuciano:
Proposal to add new functions to ortho and persepctive cameras that allow easily and quickly updating camera frustums.

When creating a camera, you must specify it's [viewing frustrum](https://en.wikipedia.org/wiki/Viewing_frustum). Once you've defined this viewing viewing frustrum, there is no easy and direct way to update the viewing frustrum.

Currently, if you want to update the viewing frustrum, you must manually update each variable individually like so:

```js
camera.top    =  screenHeight;
camera.right  =  screenWidth;
camera.bottom = -screenHeight;;
camera.left   = -screenWidth;

```

But this feel very verbose compared to how other objects are updated in Three.js , thus I propose a new function to improve this: `camera.setViewFrustrum()` which accepts the same arguments passed when creating a camera originally.

The above code could then be replaced with:

```js
camera.setViewFrustrum( -screenWidth, screenWidth, screenHeight, -screenHeight );
```

Thoughts? 

(I'll happily write up accompanying documentation if this proposal is considered)

### DIFF
--- a/src/cameras/OrthographicCamera.js
+++ b/src/cameras/OrthographicCamera.js
@@ -79,6 +79,20 @@ OrthographicCamera.prototype = Object.assign( Object.create( Camera.prototype ),
 
 	},
 
+	setViewFrustrum: function ( left, right, top, bottom, near, far ) {
+
+			this.left = left;
+			this.right = right;
+			this.top = top;
+			this.bottom = bottom;
+
+			this.near = ( near !== undefined ) ? near : 0.1;
+			this.far = ( far !== undefined ) ? far : 2000;
+
+			this.updateProjectionMatrix();
+
+	},
+
 	clearViewOffset: function () {
 
 		if ( this.view !== null ) {
--- a/src/cameras/OrthographicCamera.js
+++ b/src/cameras/OrthographicCamera.js
@@ -79,7 +79,7 @@ OrthographicCamera.prototype = Object.assign( Object.create( Camera.prototype ),
 
 	},
 
-	setViewFrustrum: function ( left, right, top, bottom, near, far ) {
+	set: function ( left, right, top, bottom, near, far ) {
 
 		this.left = ( left !== undefined ) ? left : this.left;
 		this.right = ( right !== undefined ) ? right : this.right;
--- a/src/cameras/OrthographicCamera.js
+++ b/src/cameras/OrthographicCamera.js
@@ -82,9 +82,9 @@ OrthographicCamera.prototype = Object.assign( Object.create( Camera.prototype ),
 	setViewFrustrum: function ( left, right, top, bottom, near, far ) {
 
 		this.left = ( left !== undefined ) ? left : this.left;
-		this.right = ( right !== undefined ) ? right: this.right;
+		this.right = ( right !== undefined ) ? right : this.right;
 		this.top = ( top !== undefined ) ? top : this.top;
-		this.bottom = ( bottom !== undefined ) ? bottom: this.bottom;
+		this.bottom = ( bottom !== undefined ) ? bottom : this.bottom;
 
 		this.near = ( near !== undefined ) ? near : this.near;
 		this.far = ( far !== undefined ) ? far : this.far;
--- a/src/cameras/OrthographicCamera.js
+++ b/src/cameras/OrthographicCamera.js
@@ -81,15 +81,15 @@ OrthographicCamera.prototype = Object.assign( Object.create( Camera.prototype ),
 
 	setViewFrustrum: function ( left, right, top, bottom, near, far ) {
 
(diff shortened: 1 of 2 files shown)


WestLangley:
>camera.setViewFrustrum( -screenWidth, screenWidth, screenHeight, -screenHeight );

To be precise, you are not really setting the frustum, you are setting camera properties from which a frustum can be inferred. A frustum is set like so:
```javascript
frustum.set( p0, p1, p2, p3, p4, p5 );
```

We could add two new methods:
```javascript
perspectiveCamera.set( fov, aspect, near, far );

orthographicCamera.set( left, right, top, bottom, near, far );
```
However, I am not aware of a use case in which all those properties are reset simultaneously. On window resizing, for example, only a subset of the constructor parameters are reset:
```javascript
perspectiveCamera.set( undefined, aspect );

orthographicCamera.set( left, right );
```
That API seems a bit odd to me...

What is your use case?

XanderLuciano:
@WestLangley Fair enough, and I quite like the short simpler syntax of your `cameraType.set()` method.

> We could add two new methods: `perspectiveCamera.set()`, and `orthographicCamera.set()`

I could update the PR accordingly.

> What is your use case?

Part of one my apps my apps has an adjustable "viewing window" which is a div that can adjusted by a user. This screen maintains can also maintain it's aspect ratio for example, and therefore requires updating all four corners. I also adjust the far plane distance at occasional intervals as a simple way to improve performance slightly. Additionally needed when toggling fullscreen.

Oh and to explain why I need to adjust all 4 corners when maintaining the same aspect ratio, that is because without adjusting the frustum, increasing the viewer would appear to be a "zooming" movement. However if we factor in a scale, then we can maintain a consistent screen size. 

Example:

image


vs. large

image


Increasing the window size here actually reveals more of what is in view.

It's not a function I imagine seeing with lots of usage, but when used is more convenient and automatically calls `updateProjectionMatrix()` as well, making a window resize event very easy to handle given the above use case.

WestLangley:
That seems like a reasonable use case to me. I would change the method names as I suggested.

This method may be even better for your use case:
```javascript
orthographicCamera.setViewOffset( fullWidth, fullHeight, x, y, width, height );
```

XanderLuciano:
Updated accordingly.

Looking at `setViewOffset` though, you may be correct. I had only thought of that function in terms of using it for a multi-camera setup. Looking at how the the projection matrix is updated when using a view offset, I could achieve a similar result (and modifying the far plane by separately). 

`setViewOffset` isn't a perfect _solution_, but it's a lot better than what I have currently, so cheers for that tip and a little thinking outside the 'frustrum' haha.

Just a few other points in favor of this PR:   
This PR would allow for simpler switching between camera 'settings', on one project I work on we are ingesting data from a game's replay files and using player camera settings to directly setup the camera view for that player. Players can cycle through these 'player views' and would see the game nearly the same way that player did. In other words, 'hot swapping' a camera view could be a single function call.  Also not all arguments need to be defined in order to update the camera view. All undefined arguments would remain unchanged. 

Thanks for help! Should I add some short documentation on the `orthographic` and `perspective` camera's pages into the PR as well?

WestLangley:
>Should I add some short documentation

Yes.

The final decision is up to @mrdoob.

Mugen87:
After looking at the source code, I wonder why other camera properties like `zoom` or `view` are not set by both methods. Normally, `set()` methods allow to set all public properties of an object (see the math classes for reference). Hence, it would be more consistent to me if all properties are respected which are also processed in `.copy()`.

The signatures would look like so then:
```js
set: function ( left, right, top, bottom, near, far, zoom, view ) {} // ortho cam
set: function ( fov, aspect, near, far, zoom, view, focus, filmGauge, filmOffset ) {} // persp cam
```

Mugen87:
Closing for now. We can still add these methods if more devs request them.

mrdoob:
Hmm, actually... I think these methods look good, but I would get rid of the `undefined` check.

ANDIKKY:
But this feel very verbose compared to how other objects are updated in Three.js , thus I propose a new function to improve this: camera.setViewFrustrum() which accepts the same arguments passed when creating a camera originally.....?

Mugen87:
@XanderLuciano In this case feel free to reopen the issue and apply @mrdoob's feedback 😅

WestLangley:
Agreed. We can't set some of the parameters and leave `zoom` at its current, arbitrary value. That would be confusing.

There has been no other demand for this method, so maybe we don't need it...

Mugen87:
Um, it seems @XanderLuciano is currently not active at github. Should I make the change?

Mugen87:
Replacement: #17792

PR #15046, opened 2018-10-13
[Examples] added scene comparison example

looeee:
Added a simple example that shows how to use scissor test to compare two scenes using a slider, similar to how many image comparison tools work. 

download

[Live example](https://htmlpreview.github.io/?https://github.com/looeee/three.js/blob/examples_add_scene_comparison/examples/webgl_multiple_scenes_comparison.html)


looeee:
@Mugen87 good catch, should be fixed now!

Mugen87 (approved):

mrdoob:
Thanks!

PR #15063, opened 2018-10-16
BoxHelper: added copy and clone methods

### LINKED ISSUE #7824: Cannot clone BoxHelper
In version r71 I could clone BoxHelper meshes. When I upgrade to r72 or r73, anytime I try to clone a BoxHelper or any object containing a BoxHelper I get the following exception:

Uncaught TypeError: object.updateMatrixWorld is not a function
You can replicate this by making the following change to the example code in "webgl_helpers.html"

Change line 73 from:

```
scene.add( new THREE.BoxHelper( mesh ) );
```

To:

```
var helper = new THREE.BoxHelper( mesh );
var helper1 = helper.clone();
scene.add( helper1 );
```

Is there a workaround?

WestLangley:
With #15062, this PR fixes #7824.

### DIFF
--- a/src/helpers/BoxHelper.js
+++ b/src/helpers/BoxHelper.js
@@ -101,5 +101,20 @@ BoxHelper.prototype.setFromObject = function ( object ) {
 
 };
 
+BoxHelper.prototype.copy = function ( source ) {
+
+	LineSegments.prototype.copy.call( this, source );
+
+	this.object = source.object;
+
+	return this;
+
+};
+
+BoxHelper.prototype.clone = function () {
+
+	return new this.constructor().copy( this );
+
+};
 
 export { BoxHelper };


src/helpers/BoxHelper.js
```diff
@@ -101,5 +101,20 @@ BoxHelper.prototype.setFromObject = function ( object ) {
...
+
+	LineSegments.prototype.copy.call( this, source );
+
+	this.object = source.object;
+
+	return this;
+
+};
+
+BoxHelper.prototype.clone = function () {
```
WestLangley: Since the `clone()` method is inherited, it does not need to be redefined here. However, it doesn't hurt, and I think it makes the code easier to follow. I do not object to removing it though.

mrdoob:
Thanks!

PR #15074, opened 2018-10-17
remove material.program

gero3:
This removes the link between material and program. And this makes sure the program is used through properties


src/renderers/WebGLRenderer.js
```diff
@@ -277,7 +277,7 @@ function WebGLRenderer( parameters ) {
 		morphtargets = new WebGLMorphtargets( _gl );
 		programCache = new WebGLPrograms( _this, extensions, capabilities );
 		renderLists = new WebGLRenderLists();
-		renderStates = new WebGLRenderStates();
+		renderStates = new WebGLRenderStates( properties );
```
Mugen87: Why have you added this line? The ctor WebGLRenderStates has no arguments.

gero3: This was indeed wrong

Mugen87 (approved):

PR #15105, opened 2018-10-22
Docs: Toon Material Browser

mghini:
Added material-browser for MeshToonMaterial documentation page

### DIFF
--- a/docs/api/en/materials/MeshToonMaterial.html
+++ b/docs/api/en/materials/MeshToonMaterial.html
@@ -14,7 +14,7 @@
 
 		<div class="desc">An extension of the [page:MeshPhongMaterial] with toon shading.</div>
 
-		<!-- <iframe id="scene" src="scenes/material-browser.html#MeshStandardMaterial"></iframe>
+		<iframe id="scene" src="scenes/material-browser.html#MeshToonMaterial"></iframe>
 
 		<script>
 
@@ -30,7 +30,7 @@
 
 		}
 
-		</script> -->
+		</script>
 
 		<h2>Examples</h2>
 		[example:webgl_materials_variations_toon materials / variations / toon]<br />
--- a/docs/scenes/js/material.js
+++ b/docs/scenes/js/material.js
@@ -440,6 +440,18 @@ function guiMeshPhongMaterial( gui, mesh, material, geometry ) {
 
 }
 
+function guiMeshToonMaterial( gui, mesh, material, geometry ) {
+
+	var data = {
+		gradientMap: textureMapKeys
+	};
+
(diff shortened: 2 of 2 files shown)


Mugen87:
For testing: https://raw.githack.com/mghini/three.js/ToonMaterialBrowser/docs/index.html#api/en/materials/MeshToonMaterial

Mugen87:
Maybe we should add a note that you often want to use `MeshToonMaterial` with `OutlineEffect` in order to produce the typical visual appearance of a toon shader. AFAIK, `MeshToonMaterial` is always used with `OutlineEffect` in the examples.

WestLangley:
That is not what that material is supposed to look like. One also doesn't use a grass map for a gradient.

Either the example here has issues, or there is a problem with the material.

takahirox:
Gradient map should be like "Gradient Mask_B" in http://technical-eden.blogspot.com/2012/01/gradient-mapping-awesome-way-to-get.html

WestLangley:
>The grass map is also used for other map types in context of other materials which is not ideal:

I think it would be a worthwhile effort to further improve the docs.

IMHO, In the material examples, you don't need all the menu choices from the base classes to be repeated every time.

@takahirox I have never seen a toon shader that supports hotspots, environment maps, or even maps for that matter... Or multiple lights. (Maybe I am missing something?)

A toon shader can be written in a few lines of code, and IMHO, should not have extended `Phong`, which is way too flexible. I would have implemented it like `MeshMatcapMaterial`.

Mugen87:
> A toon shader can be written in a few lines of code, and IMHO, should not have extended Phong, which is way too flexible. I would have implemented it like MeshMatcapMaterial.

I agree. I think we should consider to refactor `MeshToonMaterial` at some point and also `OutlineEffect` so it only supports the toon material.

WestLangley:
>@takahirox I have never seen a toon shader that supports hotspots, environment maps, or even maps for that matter... Or multiple lights. (Maybe I am missing something?)

>A toon shader can be written in a few lines of code, and IMHO, should not have extended Phong, which is way too flexible. I would have implemented it like MeshMatcapMaterial.

/ping @takahirox

takahirox:
OK, let me revisit `MeshToonMaterial`. Not sure if `OutlineEffect` should be specific to toon material tho.

WestLangley:
/bump

I think a refactoring of `MeshToonMaterial` along the lines of ` MeshMatcapMaterial` is worthwhile.

Mugen87:
@WestLangley Let's create a separate issue for this. The PR can be closed now because of #16255

PR #15120, opened 2018-10-23
Fix water reflection sampling coordinates

fats:
Hello!

This is a fix for the slightly wrong texture coordinates used for sampling the reflection texture in the current water shader example (ocean).

Especially at lower angles, reflection looks warped, and seams can appear at various angles.

reflection_bug

(Water distortion is disabled in the image to showcase the issue)


mrdoob:
Nice catch!

mrdoob:
Thanks!

PR #15137, opened 2018-10-25
Docs: Fix two tags

moraxy:
In `core/BufferGeometry` there's a `[param:.boundingBox]` instead of `[page:.boundingBox]`:
https://github.com/mrdoob/three.js/blob/12450bedbef53649992cf55f10d7c1b9a253570f/docs/api/en/core/BufferGeometry.html#L235

And in `materials/ShaderMaterial` a `[link:..]` has a `)` instead of a `]` at the end:
https://github.com/mrdoob/three.js/blob/12450bedbef53649992cf55f10d7c1b9a253570f/docs/api/en/materials/ShaderMaterial.html#L447

### DIFF
--- a/docs/api/en/core/BufferGeometry.html
+++ b/docs/api/en/core/BufferGeometry.html
@@ -232,7 +232,7 @@
 
 		<h3>[method:null computeBoundingBox]()</h3>
 		<p>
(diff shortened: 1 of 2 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #15147, opened 2018-10-27
Docs - add links to base class

rexdk:
I've added links in the documentation to the base class properties and methods. This is already in some existing doc pages, eg Line.html. I think it is helpful for experienced users so they can quickly get to the base class, and inexperienced users who thus become aware of the inheritance from the base class.  If agreed, I'd propose to do the same for other doc pages, to improve uniformity.


looeee (approved):

looeee (approved):

Mugen87:
see #17027

PR #15150, opened 2018-10-27
Vignette and chromatic pass

but0n:
image

### DIFF
--- a/examples/js/postprocessing/ChromaticPass.js
+++ b/examples/js/postprocessing/ChromaticPass.js
@@ -1,0 +1,86 @@
+/**
+ * @author but0n / https://github.com/but0n
+ */
+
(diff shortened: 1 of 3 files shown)


but0n:
Considering add some blur effects on the corner.

Mugen87:
There is already a implementation of a vignette shader so I don't think we need two of them:

https://github.com/mrdoob/three.js/blob/dev/examples/js/shaders/VignetteShader.js

Besides, there is already [RGBShiftShader](https://github.com/mrdoob/three.js/blob/dev/examples/js/shaders/RGBShiftShader.js). It seems to me that your chromatic shader does essentially the same thing.

but0n:
@Mugen87 Ok, I didn’t noticed them

Issue #15154, opened 2018-10-27
GLTFLoader - TypeError: Cannot read property 'response' of undefined


Jonybang:
##### Description of the problem
```
const loader = new GLTFLoader();
loader.setCrossOrigin('anonymous');
loader.load('model-assets/untitled.gltf', function (gltf) {
   // ...
});
```
The code above throws the error:
```
three.module.js:32183 Uncaught TypeError: Cannot read property 'response' of undefined
    at three.module.js:32183
    at XMLHttpRequest.wrapped (index.js:3134)
```
Maybe because i used webpack for build js files into one.

##### Three.js version

- [ ] Dev
- [x] r97

##### Browser

- [] All of them
- [x] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [] All of them
- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
Closing, see https://github.com/mrdoob/three.js/pull/15155#issuecomment-464200925

PR #15155, opened 2018-10-27
replace this by event.target in request load callback

MicrowaveDev:
Fixed the issue: https://github.com/mrdoob/three.js/issues/15154

### DIFF
--- a/src/loaders/FileLoader.js
+++ b/src/loaders/FileLoader.js
@@ -168,7 +168,8 @@ Object.assign( FileLoader.prototype, {
 
 			request.addEventListener( 'load', function ( event ) {
 
-				var response = event.target.response;
+				var target = event.target;
+				var response = target.response;
 
 				Cache.add( url, response );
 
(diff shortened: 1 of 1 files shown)


src/loaders/FileLoader.js
```diff
@@ -168,20 +168,20 @@ Object.assign( FileLoader.prototype, {
 
 			request.addEventListener( 'load', function ( event ) {
 
-				var response = this.response;
+				var response = event.target.response;
```
WestLangley: If what you propose is correct, how about this instead:
```javascript
var target = event.target;
```
...and then make the remaining necessary changes.

Jonybang: done

src/loaders/FileLoader.js
```diff
@@ -168,20 +168,21 @@ Object.assign( FileLoader.prototype, {
 
 			request.addEventListener( 'load', function ( event ) {
 
-				var response = this.response;
+				var target = event.target;
```
Mugen87: I think the name `target` is a bit confusing. It would be more clear to highlight that `target` represents the `XMLHttpRequest` object in this context.

Mugen87 (changes requested):

donmccurdy:
> Do you know the reasons why using this in the onLoad() event listener is problematic for your use case?

+1, it would be helpful to understand the conditions where you're seeing https://github.com/mrdoob/three.js/issues/15154... the use of `this` is a [documented part of the XMLHttpRequest API](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onload#Values). The bit about `XMLHttpRequest.wrapped` is interesting... are you possibly using some other library (RavenJS, AngularJS) that might "wrap" the XMLHttpRequest object?

Jonybang:
@donmccurdy i used VueJS, it may be a problem?

donmccurdy:
Maybe... I'm not sure sorry. Any chance you can share a demo?

Mugen87:
@Jonybang This PR gets never merged without more investigation. If you can't reproduce the reference error for testing purposes, I suggest to close the PR and the respective issue.

Jonybang:
ok, i will try to create an example

Mugen87:
@Jonybang Any news on this? 😉

Mugen87:
Closing. We can reopen when there is a test case for reproduction.

PR #15157, opened 2018-10-28
Edit button in docs localization bug fix #15136

### LINKED ISSUE #15136: Edit button in docs broken after r96
#14846 stripped the localization from the `[path]` placeholder to get the links to the source files working again but the edit button in the top right needed that to point to the localized docs file.

https://github.com/mrdoob/three.js/blob/12450bedbef53649992cf55f10d7c1b9a253570f/docs/page.js#L102-L106

First noticed in https://github.com/mrdoob/three.js/pull/14750#issuecomment-414792240 as far as I can tell. A previous hotfix(hiding the button) was applied in 0c58464ab1b37451d6440025bc5d8c0095fec8da, but got removed with the release of r96.

maru0301:
\docs\index.html
edit button localization link bug fix
[https://github.com/mrdoob/three.js/issues/15136](#15136)

### DIFF
--- a/docs/page.js
+++ b/docs/page.js
@@ -30,10 +30,10 @@ function onDocumentLoad( event ) {
 	var pathname = window.location.pathname;
 	var section = /\/(manual|api|examples)\//.exec( pathname )[ 1 ].toString().split( '.html' )[ 0 ];
 	var name = /[\-A-z0-9]+\.html/.exec( pathname ).toString().split( '.html' )[ 0 ];
-	var languageReg = new RegExp( "/" + section + "/[A-z0-9\-]+/[A-z0-9\-]+/" );
+	var languageReg = new RegExp( '/' + section + '/[A-z0-9\-]+/[A-z0-9\-]+/' );
 	var language = languageReg.exec( pathname );
-	languageReg = new RegExp( "/"+section+"/[A-z0-9\-]+/" );
-	language = ( language !== null ) ? languageReg.exec( language ).toString().replace( "/" + section + "/", "" ) : "";
+	languageReg = new RegExp( '/' + section + '/[A-z0-9\-]+/' );
+	language = ( language !== null ) ? languageReg.exec( language ).toString().replace( '/' + section + '/', '' ) : '';
 
 	switch ( section ) {
 
--- a/docs/page.js
+++ b/docs/page.js
@@ -30,6 +30,10 @@ function onDocumentLoad( event ) {
 	var pathname = window.location.pathname;
 	var section = /\/(manual|api|examples)\//.exec( pathname )[ 1 ].toString().split( '.html' )[ 0 ];
 	var name = /[\-A-z0-9]+\.html/.exec( pathname ).toString().split( '.html' )[ 0 ];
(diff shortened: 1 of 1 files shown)


moraxy:
That would break the `Source` links at the bottom of the API pages (e.g. [live link](https://raw.githack.com/maru0301/three.js/dev/docs/index.html#api/en/cameras/ArrayCamera))

Mugen87:
Correct. This PR will not solve the problem. One solution is presented in https://github.com/mrdoob/three.js/issues/15136#issuecomment-433650832. But there might be a smarter approach...

maru0301:
It seems best to think that it is best to pass the language setting on the right page when selecting the language and receive it.
Other than that
· Extract language setting for edit button
· Do not include language setting on source link Set to URL

Is not it?
I would like to try it with any of the above

Mugen87:
I think `page.js` needs a bit of refactoring in order so make things more nice. But I don't think it's necessary to do this now. I vote for a quick/dirty fix just to ensure the `Edit Button` works again. In the next step, it's okay to think about a different handling of the language setting.

Mugen87:
For testing: https://raw.githack.com/maru0301/three.js/dev/docs/index.html

Mugen87:
Your approach seems to work fine 👍

docs/page.js
```diff
@@ -30,15 +30,15 @@ function onDocumentLoad( event ) {
 	var pathname = window.location.pathname;
 	var section = /\/(manual|api|examples)\//.exec( pathname )[ 1 ].toString().split( '.html' )[ 0 ];
 	var name = /[\-A-z0-9]+\.html/.exec( pathname ).toString().split( '.html' )[ 0 ];
+	var languageReg = new RegExp( "/" + section + "/[A-z0-9\-]+/[A-z0-9\-]+/" );
```
Mugen87: Just some style feedback: Please use ' instead of ".

maru0301:
Yes, thank you!

docs/page.js
```diff
@@ -30,15 +30,15 @@ function onDocumentLoad( event ) {
...
+	var language = languageReg.exec( pathname );
+	languageReg = new RegExp( "/"+section+"/[A-z0-9\-]+/" );
```
Mugen87: Please write it like so (notice the additional spaces)
```js
languageReg = new RegExp( '/' + section + '/[A-z0-9\-]+/' );
```

Mugen87 (changes requested):

maru0301:
I applied feedback.

Mugen87 (approved):

Mugen87:
Let's see what @mrdoob thinks about the fix 😊

maru0301:
Yes
Thank you so much today!

Mugen87:
Funny, your name "Maru" remembers me at a certain cat 😁 

https://www.youtube.com/watch?v=hPzNl6NKAG0

maru0301:
Hahaha,
maru is my real nickname 😊

Mugen87:
@mrdoob Do you think this fix is appropriate? Or do we need a different approach?

Mugen87:
Solved via #15399

@maru0301 In any event, thanks for your work ❤️

PR #15159, opened 2018-10-28
Support for offset (non-centered) frustum in CSS3DRenderer

### LINKED ISSUE #15143: Changing frustum planes of Orthographic camera doesn't affect projection in CSS3DRenderer

Hi,

So I recently discovered a potential bug:
I use an orthographic camera with CSS3DRenderer, and I implemented a simple pan functionality. When I adjust the position of the camera (either with `camera.position.setX(camera.position.x + delta.x)`, or `camera.translateX(delta.x)`, or similar), it works perfectly, as it should. But when I try to change the frustum planes of the camera (eg.: `camera.left += delta.x`), it doesn't have any visual effect.

So in my opinion, in terms of visual differences, these 4 lines:
```
camera.left += delta.x;
camera.right += delta.x;
camera.top += delta.y;
camera.bottom += delta.y;
```

Should give similar results to the effect of these 2 lines:
```
camera.translateX(delta.x);
camera.translateY(delta.y);
```

And with WebGLRenderer, they do. But with CSS3DRenderer, they don't. In fact, they don't seem to do anything. I also played around with the frustum plane values in the resize function, but it didn't have an effect there, either. So I think CSS3DRenderer doesn't take the frustum planes into account, only the initial ones..?

I know it's probably not the best practice to change the frustum planes in this case (and I should change the position instead), but I believe it should work this way too, as it does with WebGLRenderer.

I've made a simple demo to represent the problem. Check from line ~80 in `index.html` in the zip, or line ~60 in the JS section here: https://codepen.io/soadzoor/pen/vVbzyE

Thank you!
[CSS3DRenderer.zip](https://github.com/mrdoob/three.js/files/2520398/CSS3DRenderer.zip)

##### Three.js version
- [x] r97

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

soadzoor:
Fixes #15143

### DIFF
--- a/examples/js/renderers/CSS3DRenderer.js
+++ b/examples/js/renderers/CSS3DRenderer.js
@@ -291,8 +291,15 @@ THREE.CSS3DRenderer = function () {
 
 		if ( camera.parent === null ) camera.updateMatrixWorld();
 
+		if ( camera.isOrthographicCamera ) {
+
+			var tx = - ( camera.right + camera.left ) / 2;
+			var ty = ( camera.top + camera.bottom ) / 2;
+
+		}
+
 		var cameraCSSMatrix = camera.isOrthographicCamera ?
-			'scale(' + fov + ')' + getCameraCSSMatrix( camera.matrixWorldInverse ) :
+			'scale(' + fov + ')' + 'translate(' + epsilon( tx ) + 'px,' + epsilon( ty ) + 'px)' + getCameraCSSMatrix( camera.matrixWorldInverse ) :
 			'translateZ(' + fov + 'px)' + getCameraCSSMatrix( camera.matrixWorldInverse );
 
 		var style = cameraCSSMatrix +


WestLangley:
/ping @yomotsu Can you have a look?

examples/js/renderers/CSS3DRenderer.js
```diff
@@ -95,28 +95,68 @@ THREE.CSS3DRenderer = function () {
...
+				epsilon( elements[ 7 ] ) + ',' +
+				epsilon( sx * elements[ 8 ] + tx * elements[ 11 ] ) + ',' +
+				epsilon( - sy * elements[ 9 ] + ty * elements[ 11 ] ) + ',' +
+				epsilon( sz * elements[ 10 ] + tz * elements[ 11 ] ) + ',' +
+				epsilon( elements[ 11 ] ) + ',' +
+				epsilon( sx * elements[ 12 ] + tx * elements[ 15 ] ) + ',' +
+				epsilon( - sy * elements[ 13 ] + ty * elements[ 15 ] ) + ',' +
+				epsilon( sz * elements[ 14 ] + tz * elements[ 15 ] ) + ',' +
+				epsilon( elements[ 15 ] ) +
+			')';
```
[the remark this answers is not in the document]
soadzoor: You're absolutely right, I pushed an update with this

WestLangley:
`camera.near` can be zero, so you can't divide by it.

Can you explain why the z-component would be affected?

Try

```javascript
var sz = 0
var tz = 1;
```

examples/js/renderers/CSS3DRenderer.js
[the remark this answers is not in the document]
soadzoor: > `camera.near` can be zero, so you can't divide by it.
> 
> Can you explain why the z-component would be affected?
> 
> Try
> 
> ```js
> var sz = 0
> var tz = 1;
> ```

You're right again. I picked those values from the [wiki](https://en.wikipedia.org/wiki/Orthographic_projection)

But I made a typo. `camera.far / camera.near` should be `camera.far - camera.near` based on this.


I believe you mean:
```js
var sz = 1;
var tz = 0;
```

I tried it, and it doesn't seem to mess up anything, so I simplified the matrix again with this idea. Thanks.

soadzoor:
> `camera.near` can be zero, so you can't divide by it.
> 
> Can you explain why the z-component would be affected?
> 
> Try
> 
> ```js
> var sz = 0
> var tz = 1;
> ```

You're right again. I picked those values from the [wiki](https://en.wikipedia.org/wiki/Orthographic_projection)

But I made a typo. `camera.far / camera.near` should be `camera.far - camera.near` based on this.


I believe you mean:
```js
var sz = 1;
var tz = 0;
```

I tried it, and it doesn't seem to mess up anything, so I simplified the matrix again with this idea. Thanks.

WestLangley:
>I tried it, and it doesn't seem to mess up anything, so I simplified the matrix again with this idea

We can't be using trial and error. Can you please explain the basis for your original formulas?

examples/js/renderers/CSS3DRenderer.js
```diff
@@ -95,28 +95,68 @@ THREE.CSS3DRenderer = function () {
...
+			//  0  0 sz  tz
+			//  0  0  0   1
+			//
+			// sz === 1, tz === 0
+
+			return 'matrix3d(' +
+				epsilon( sx * elements[ 0 ] ) + ',' +
+				epsilon( - sy * elements[ 1 ] ) + ',' +
+				epsilon( elements[ 2 ] ) + ',' +
+				'0,' +
```
WestLangley: Please do not hardwire values here.
```javascript
epsilon( elements[ 3 ] ) + ',' +
```

soadzoor: Why? You told me it's always 0, and it can be simplified.

soadzoor:
> > I tried it, and it doesn't seem to mess up anything, so I simplified the matrix again with this idea
> 
> We can't be using trial and error. Can you please explain the basis for your original formulas?

My original intention was to multiply the worldmatrix with the projectionmatrix, and I used the [wiki](https://en.wikipedia.org/wiki/Orthographic_projection) to create the projectionmatrix (though I've made a typo first).

I'm not sure why it has those values for the Z axis, while Z's value doesn't really matter in an orthogonal projection.

examples/js/renderers/CSS3DRenderer.js
WestLangley: That's a separate optimization. If it were to be implemented in this case, it should be implemented consistently throughout the file.

WestLangley:
Via experimentation, it appears that your PR works with the changes I suggested. But I have not tried to derive the formulas. If they are not clear to you, we will have to hold off a bit.

Maybe someone else will have an idea.

soadzoor:
Okay, so here's what I think: when it comes to an orthographic projection, the z-values decide only the clipping, and the z-order. They don't affect anything else.

Now, since we're talking about CSS, clipping is not involved. And the z-order is managed by the `preserve-3d` rule (remember: the z-order was absolutely fine before these changes as well).

So that's why we can use 1 for `sz`, and 0 for `tz` here, because they won't affect the existing (working) z-order.

WestLangley:
What I am asking is: how did you derive your formulas?

soadzoor:
I just copied it from the wiki, but I'll try to derive it myself if you'd like. I'd need a couple of days though, I'm busy with other projects

soadzoor:
Okay, here they are. So the basic idea is to write a system of equations, like shown below, then calculate the values (`Sx`, and `Tx`, then `Sy`, `Ty`).

I assume we are using NDC system, so we go from `-1` to `1` in each direction. `(0, 0)` is the center of the viewport.

The first expression must be `-1` because we want to have the vertex at `left` value to be at the left edge of our viewport. For the same reason, the second expression must be `1` (we want to have the vertex at `right` value to be at the right edge of our viewport).

The same exact thing applies when we calculate with `y`, `top` and `bottom`.
img_20181105_121614
img_20181105_121655
img_20181105_121727
img_20181105_121935
img_20181105_122413

The only difference here, is that we have to multiply the `ty` value with `-1`, because in `CSS3D`, the y axis is upside down.

soadzoor:
@WestLangley any thoughts?

yomotsu:
@WestLangley 
Sorry, I just realized your ping.
Looks good to me. It doesn't affect existing examples at least.

Please take a look at his formulas?

soadzoor:
````tx = - ( camera.right + camera.left ) / 2;
ty = ( camera.top + camera.bottom ) / 2;
```
These values don't seem to be correct to me. How did you come up with them?

WestLangley:
As I said, make ONLY the change I [posted](https://github.com/mrdoob/three.js/pull/15159#issuecomment-437124863) above. Do not change any other parts of the file.

soadzoor:
> As I said, make ONLY the change I [posted](https://github.com/mrdoob/three.js/pull/15159#issuecomment-437124863) above. Do not change any other parts of the file.

In that case there would be a syntax error, because `camera` is not defined at `if ( camera.isOrthographicCamera ) {`.

So that parameter change is needed anyways. I see your point, to keep the `scale` part here:
`'scale(' + fov + ')' + getCameraCSSMatrix( camera )`

But I don't agree with it. `fov` doesn't event make much sense when it comes to orthographic projection. At least we shouldn't call it `fov`. So I'd suggest we stick to my [suggestion](https://github.com/mrdoob/three.js/pull/15159#issuecomment-437128789) for simplifying the function. Why don't you like it? :(

soadzoor:
Alright, I applied your suggestions

WestLangley:
Thanks, @soadzoor  for your persistence. I would not have come to this solution otherwise.

Do you know now to squash the commits to a single one so this change is easier to follow?

soadzoor:
> Do you know now to squash the commits to a single one so this change is easier to follow?
I think I finally figured it out, please check if it's ok

WestLangley:
@mrdoob @Mugen87 @yomotsu We have a working solution.

mrdoob:
Thanks!

PR #15187, opened 2018-11-01
WebGLBackground: enable code injection

WestLangley:
Shaders are modified by code injection in the `WebGLProgram` constructor.

The code injection only works correctly for built-in-materials where the material property names have been standardized. (e.g., `map`, `envMap`, etc.)

Shader materials,  however, can have arbitrary properties and uniforms.

This PR adds the necessary interface so code injection also works correctly for the background shaders.

### DIFF
--- a/src/renderers/webgl/WebGLBackground.js
+++ b/src/renderers/webgl/WebGLBackground.js
@@ -67,6 +67,17 @@ function WebGLBackground( renderer, state, objects, premultipliedAlpha ) {
 
 				};
 
+				// enable code injection for non-built-in material
+				Object.defineProperty( boxMesh.material, 'envMap', {
+
+					get: function () {
+
+						return this.uniforms.tCube.value;
+
+					}
+
+				} );
+
 				objects.update( boxMesh );
 
 			}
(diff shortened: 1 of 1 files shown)


mrdoob:
Not sure I am following. Is this needed in order for `WebGLProgram` to handle the includes?

WestLangley:
It is required for [this](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLProgram.js#L494-L500) to work. That code injects the necessary decoding and encoding chunks into the shader. It is based on certain, assumed material property names. `ShaderMaterial` does not have these material property names.

`PMREMGenerator` implements [this](https://github.com/mrdoob/three.js/blob/dev/examples/js/pmrem/PMREMGenerator.js#L62-L63) hack, which creates a fake `envMap` property. I have seen this hack elsewhere, too.

The approach here is more elegant, and is a must for the existing merged PRs to work correctly.

mrdoob:
Thanks!

mrdoob:
Just tried hiding the room and adding `scene.background = hdrCubeMap;` in (examples/webgl_tonemapping)[https://threejs.org/examples/webgl_tonemapping.html] and things are starting to work 😊

screen shot 2018-11-05 at 5 16 08 pm

WestLangley:
I think there's something wrong with the HDR cube map files, though. They're pixelated.

mrdoob:
Yeah, I'll investigate tomorrow.

mrdoob:
Hoping that we'll be able to get this kind of quality eventually... 💪

https://grhp.github.io/filament/public/suzanne.html

PR #15188, opened 2018-11-01
Editor: i18n

linbingquan:


sweerwen:
你好，我觉得有些单词还可以专业些，我列举我所知道的：
roughness 粗糙度
metalness 金属度
emissive  自发光
emissivemap 自发光贴图
displacemap  置换贴图
aomap  环境光遮蔽(罩)贴图
back  背面
shininess 高光大小
fragment 片元

linbingquan:
@sweerwen Thanks! 

https://github.com/linbingquan/three.js/commit/dbe72d6bbed73546634f00272b09d616848504f7

editor/js/Sidebar.Project.js
```diff
@@ -71,7 +71,19 @@ Sidebar.Project = function ( editor ) {
 
 		if ( key.indexOf( 'WebGL' ) >= 0 && System.support.webgl === false ) continue;
 
-		options[ key ] = key;
+		var optionTxt = '???';
+
+		if ( key === 'WebGLRenderer' ) {
+			optionTxt = strings.getKey( 'sidebar/project/renderer/webglrenderer' )
```
mrdoob: I wonder if we should be translating API names. Would people know that it's using `THREE.WebGLRenderer` when reading `WebGL渲染器`?

mrdoob: Same thing for the `添加` section... Wouldn't it be better to keep the english names `Group`, `Plane`, `Box`, ... so people understand they're adding a `THREE.Group`, `THREE.Plane`, `THREE.Box`?

linbingquan: > I wonder if we should be translating API names. Would people know that it's using `THREE.WebGLRenderer` when reading `WebGL渲染器`?

I think most of people don't know `THREE.WebGLRenderer` when reading `WebGL渲染器`, maybe people think `Render(渲染器)` has relationship with `WebGL`. I don't commit this part of code until we found the better way.

linbingquan: > Same thing for the `添加` section... Wouldn't it be better to keep the english names `Group`, `Plane`, `Box`, ... so people understand they're adding a `THREE.Group`, `THREE.Plane`, `THREE.Box`?


This is other way like `Group` -> `组 Group`, `Plane` -> `平面 Plane`, `Box` -> `正方体 Box`, but I don't think it is the best way.

How about we add some tips in options, like **Blender software**.

image

PR #15191, opened 2018-11-01
Update Geometry.html 

WestLangley:
As proposed in https://github.com/mrdoob/three.js/pull/14881#issuecomment-433334466.


Mugen87:
In order to encourage contribution, I think it's better when the actual user (in this case @ieskudero) would did this change. Especially since there is no rush for it...

Mugen87 (approved):

Issue #15182, opened 2018-10-31
BufferGeometry drawRange influence both face index & other vertices properties


Makio64:
##### BufferGeometry drawRange influence both face index & other vertices properties

The drawRange influence both face indexing & vertex attributes such as position.

For exemple if my index use the vertex 0 to 10, I was expecting setDrawRange( 0, 30 ) will do the job, but actually if I have more than 10 face indexed, then the extra face after the 10 first will not be draw,

I wonder if its a Webgl limitation or if we can separated setDrawRange in two function setVertexDrawRange & setIndexDrawRange. 

##### Three.js version

- [x] r98

##### Browser

- [x] All of them

##### OS

- [x] All of them

WestLangley:
For non-indexed `BufferGeometry`, `count` in `setDrawRange( start, count )`, is the number of vertices.

For indexed `BufferGeometry`, `count` in `setDrawRange( start, count )`, is the number of indices.

Mugen87:
> I wonder if its a Webgl limitation or if we can separated setDrawRange in two function setVertexDrawRange & setIndexDrawRange.

I do not vote for this. Having a single function `setDrawRange()` is better approach from my point of view.

Makio64:
Thanks for the clarification about it. Also I agree a general setDrawRange is good but cant we have the tree : setDrawRange with the general behavior describe by WestLangley & the two others for deeper control ?

WestLangley:
> cant we have [three] : setDrawRange with the general behavior describe by WestLangley & the two others for deeper control ?

I expect that would be too confusing.

Perhaps you can use `geometry.addGroup()` for additional control in your app:
```javascript
geometry.clearGroups();
geometry.addGroup( start, count, 0 );
material = [ material ]; // if there is just one material
```

Makio64:
@WestLangley So the param start/count in addGroup are about index for indexedGeometry & vertex for non indexed too? 

I'm not sure if this will resolve my needs about discarding by face & vertex at the same time. but thanks to your explanation, I already solved my original problem, this extra control isnt really need anymore. 

I suggest to clarify the documentation with your explanation :)

WestLangley:
>I suggest to clarify the documentation with your explanation :)

Done!

PR #15192, opened 2018-11-01
BufferGeometry docs: clarification

WestLangley:
As discussed in https://github.com/mrdoob/three.js/issues/15182#issuecomment-434838531.


Mugen87 (approved):

mrdoob:
Thanks!

PR #15203, opened 2018-11-04
support Geometry in GLTFExporter

makc:
ok, I know you guys hate old Geometry but, while it's still around, and it takes a single line to support it - why not?

### DIFF
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -1001,7 +1001,7 @@ THREE.GLTFExporter.prototype = {
 		 */
 		function processMesh( mesh ) {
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
Geometry should have been covered by this block:

https://github.com/makc/three.js.fork/blob/fc31f5e634714c9811a8a13046419b369b8ee522/examples/js/exporters/GLTFExporter.js#L1030-L1033

Have you found an example that isn’t working?

makc:
well yes I was getting geometry.getAttribute not defined. unfortunately the scene is dynamic, so I don't have easy to reproduce test case

makc:
looking at the code above it could have been lines

makc:
in which case this is not a proper fix - closing

PR #15220, opened 2018-11-07
update CubeTextureLoader onProgress

lolofy:
update CubeTextureLoader onProgress

### DIFF
--- a/src/loaders/CubeTextureLoader.js
+++ b/src/loaders/CubeTextureLoader.js
@@ -34,6 +34,12 @@ Object.assign( CubeTextureLoader.prototype, {
 				texture.images[ i ] = image;
 
 				loaded ++;
+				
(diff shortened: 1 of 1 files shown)


Mugen87:
Not sure this change makes sense. Normally `onProgress()` is used in context of the `progress` event of `XMLHttpRequest`. That means the respective [event data](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onprogress#Example) are passed to the callback. Your change passes different data (jus the amount of loaded files) to `onProgress()` which is somewhat inconsistent.

I think it's better if `onProgress()` always has the same semantics or is not supported at all.

Mugen87:
Closing. It's no good approach to enhance `CubeTextureLoader` like this.

Issue #15047, opened 2018-10-13
WebGLRenderer: Add support for multiple shader programs with a single material.


gero3:
It would be necessary to have multiple multiple webglprogram on a single material to solve #11400.

This would require several steps to be taken to get this done:
 - [x] Figuring out what parameters are necessary to generate a webglprogram and use a single parameter source to create it. (#18279)
 - [ ] Build single parameter source from 4 parameter sources: material, object, renderer, scene.
     - [ ] Material parameter: gets regenerated when needsUpdate is true
     - [ ] Renderer parameter: gets regenerated every render    
     - [ ] Scene parameter: gets regenerated every render
     - [ ] object parameter: gets regenerated every ??
 - [ ] Create keys based of the 4 parameter sources to figure out if something changed.
 - [ ] Use single parameter source as key to figure out if program needs to be created.
 - [ ] Use single parameter source as key to figure out which program to use.

We would also need to track which program gets used by which material to make sure that the programs get deleted when the material gets disposed of.

Mugen87:
> Create keys based of the 4 parameter sources to figure out if something changed.

I'm actually afraid that generating a key per renderable item per frame will worsen the performance of the renderer because it would have to check a lot of scenarios in order to assign the correct program. Essentially, you need the logic from `WebGLPrograms.getParameters()` and `WebGLPrograms.getProgramCacheKey()` which is currently only executed in `WebGLRenderer.initMaterial()`. And this method is only called if a material needs an update (so it's currently not a hot spot when rendering). 

Even if you optimize the code, the renderer will produce more overhead than before. An interesting side effect: `Material.needsUpdate` should become obsolete. If we compute the program key per renderable item and per frame, the renderer can always figure out if a new program or a recompile is necessary.

Mugen87:
I've experimented in this context for a while and came up with the following minimal solution: https://github.com/Mugen87/three.js/commit/e0099e56abd3cb573c46144096deb70e249a5108

It actually solves the ping-pong effect in certain scenarios e.g. when rendering to a render target with different output encodings (#19056). However, the general performance gets worse since the renderer has to compute a program key per frame for each object right now. This becomes to a problem when rendering scenes with a lot of objects:

https://raw.githack.com/Mugen87/three.js/dev51/examples/index.html#webgl_lights_pointlights2
https://threejs.org/examples/#webgl_lights_pointlights2

I'm trying to figure out a solution to avoid unnecessary calls of `getProgram()`.

PR #15223, opened 2018-11-08
selective lighting

ghost:
selective lighting support based on 
https://github.com/ErikBehar/three.js/commit/ac0499b70b82bc7bb780100a8372fcdf318d1424#diff-5e43a0b5002eb2c419def3baf67d4e67
by @ErikBehar

still need to work on example, and review it thoroughly


looeee:
Here's a [live version](https://raw.githack.com/byondrnd/three.js/dev/examples/webgl_lights_selective_lights.html), although it's not currently working for me:

`Cannot read property 'set' of undefined` 

At this line: 

[`light1.affectedLayers.set(1);`](https://github.com/byondrnd/three.js/blob/8c991219808b714c1e1ceaa55db93a6254c96a56/examples/webgl_lights_selective_lights.html#L115)

byondrnd:
@looeee 
sorry, didn't commit the new build
you can now check again

mrdoob:
Why are we adding more properties to `Material`?
Shouldn't we just compute the relevant lights by testing `mesh.layers` against `light.layers`?

byondrnd:
@mrdoob 
in some cases you don't want the light to affect the material but still present the object

byondrnd:
@mrdoob 
using mesh.layers now

byondrnd:
@looeee @mrdoob 
new example, better implementation
:)
[live example](https://raw.githack.com/byondrnd/three.js/dev/examples/webgl_lights_selective_lights.html)

looeee:
@byondrnd I keep forgetting to check this until I'm on my phone, and your example is not mobile friendly. Buttons would be preferable to keyboard commands for changing layers.

Aside from that, the performance is very low on my Moto G6 - 15 to 20 fps - which seems odd for such a simple scene.

byondrnd:
@looeee 
- updated the example
- some performance fix

looeee:
[Updated live example](https://raw.githack.com/byondrnd/three.js/6c948a068a88fba553fe61272760a41ea8cde120/examples/webgl_lights_selective_lights.html)

I'm now getting about 1fps on my phone :(

mike-nelson:
Where can I get the latest source for this? 

Usage is like this?

    myLight.layers.set(1);
    myMesh.layers.set(1);

byondrnd:
> Where can I get the latest source for this?
> 
> Usage is like this?
> 
> ```
> myLight.layers.set(1);
> myMesh.layers.set(1);
> ```

yes, you can use the [example ](https://raw.githack.com/byondrnd/three.js/6c948a068a88fba553fe61272760a41ea8cde120/examples/webgl_lights_selective_lights.html) as ref

mike-nelson:
@byondrnd Thanks. Got it. Works great!

BTW, i notice this implementation currently only works with light types and not individual lights. I had two directional lights and they were 'sharing' (lights from layer1 were lighting objects on layer2). I needed to change one to a spotlight and then it got its own affectedLayers. This took me several hours to figure out so is a bit of a gotcha.

pboyer:
@byondrnd @mrdoob This is a super cool feature and we'd love to see it in r105. Need any help? Our team (@Oletus and I) would be happy to help bring this to the finish line.

pboyer:
Ping

navenedrob:
@mrdoob Will this be going to dev branch any time soon?

mike-nelson:
I used selective lighting successfully here:
https://www.altherm.co.nz/virtual-home

Click on Entrance Doors. The doors are painted with coloured light, which was the only way we could get the colours looking realistic, but then the handles and glass would get coloured, so we had to use selective lighting to make them non-colourised. 

Its a great pretty much essential feature for creating this.

navenedrob:
@mrdoob Would you like to close #5180 and this PR to consolidate? I am keeping this up to date in my fork/branch since r109 is close.

Mugen87:
From my experience it will be more unlikely to merge such a PR if it's not unclear how the changes affect the performance of existing apps.

navenedrob:
@Mugen87 @looeee @mike-nelson Thoughts on how we can better test performance so we can get this merged? PlayCanvas has this lighting feature currently, other frameworks might, so I know it would be a popular feature to have.

Mugen87:
You can compare frame times for different lighting setups with and without this PR. If you want to be on the safe side, it's recommended to make a performance analysis with Chrome dev tools.

donmccurdy:
Here's an attempted rebase of the PR, updating to r112.

Code: https://github.com/mrdoob/three.js/compare/dev...donmccurdy:feat-selective-lighting

Demo: https://raw.githack.com/donmccurdy/three.js/feat-selective-lighting/examples/webgl_lights_selective_lights.html

An error is thrown when disabling any layer, though... I think something has changed about uniform updates since r100, affecting this PR, but I'm not sure what.

Mugen87:
I'm afraid the entire PR is conceptually wrong. The presented logic only works if objects do not share materials. Check out the following fiddle that demonstrate the limitation:

https://jsfiddle.net/0dhaomk5/1/

There are two cubes in the scene. One is placed in a lit layer, the other one in an unlit layer. Since both cubes share the same material, they are rendered with the same shader program although this shouldn't be the case. This happens because a material can only have a single shader program (see #15047). Depending on which cube is rendered first, the both are both lit or unlit. So the layer configuration on object level is not processed correctly..

To solve this problem one can consider to introduce `Layers` to `Material` and perform the configuration there. Or we can consider to solve #15047 although this will require a deep refactoring in `WebGLRenderer`. 

In any event, I vote to close the PR since moth changes of were made under a wrong assumption. We can continue the discussion in the respective issue #5180.

donmccurdy:
Hmm yeah that's a good point. I'm fine with closing this and discussing elsewhere. I have some concerns with #15047 but it's a longer conversation. It's also possible that features like light probe volumes (https://github.com/mrdoob/three.js/issues/16228) will take care of many of the "lots of lights" use cases, even though it's a fundamentally different feature and wouldn't necessarily preclude selective lighting either.

Mugen87:
Closing. Selective lighting needs to be implemented differently.

Mugen87:
> and discussing elsewhere.

#5180

Issue #15212, opened 2018-11-06
WebGLRenderer convenience methods


WestLangley:
Because we typically use a single renderer instance, we see patterns like this:
```javascript
this.oldClearColor.copy( renderer.getClearColor() );
this.oldClearAlpha = renderer.getClearAlpha();
var oldAutoClear = renderer.autoClear;

// ... change renderer properties ...

renderer.setClearColor( this.oldClearColor );
renderer.setClearAlpha( this.oldClearAlpha );
renderer.autoClear = this.oldAutoClear;
```
In these cases, the app does not care about the current value, but desires only to temporarily reset the value and later restore it.

For these use cases, and new ones, such as rendering inset viewports, I'm thinking it would be nice to have certain `WebGLRenderer` convenience methods such as these:

```javascript
saveViewport()
restoreViewport()

saveScissor()
restoreScissor()

saveScissorTest()
restoreScissorTest()

saveAutoClear() // saves color, depth, stencil, too
restoreAutoClear()

saveClearColor() // saves both color and alpha
restoreClearColor()
```

I'll implement it if there is interest.

pailhead:
ship it

greggman:
How about put the clear color on the camera?

WestLangley:
@mrdoob If you don't like this idea, I will propose adding new methods, instead:

```javascript
// replace methods with new APIs
renderer.getSize( size ); // currently returns a new object
renderer.getDrawingBufferSize( drawingBufferSize ); // currently returns a new object
renderer.getCurrentViewport( currentViewport ); // currently returns reference to private variable

// new methods
renderer.getViewport( viewport ); // alternate API: just return a reference to _viewport
renderer.getScissor( scissor ); // alternate API: just return a reference to _scissor
scissorTest = renderer.getScissorTest();
```
Do you mind returning references to private variables (e.g.,` _viewport`)?

How you feel about this?

WestLangley:
Closing in favor of #15786, #15806.

PR #15243, opened 2018-11-13
SSAO: New implementation

### LINKED ISSUE #8576: SSAO example is broken, just darkening based on normal direction

Mugen87:
This PR introduces a new screen space ambient occlusion (SSAO) shader. Most of the implementation is based on John  Chapman's [article](http://john-chapman-graphics.blogspot.com/2013/01/ssao-tutorial.html) about SSAO.

The code hopefully fixes #8576 since it's provides a more or less real ambient occlusion effect.

As mentioned by a [user](http://john-chapman-graphics.blogspot.com/2013/01/ssao-tutorial.html?showComment=1367624542487#c1527234297842232685) in the linked article, the implemented blur shader produces a thin halo of non-occlusion when there are high frequencies in the depth buffer (steep changes of depth values). I guess the usage of `DepthLimitedBlurShader` might solve this issue, but it's a more complex blur and thus computationally more expensive (and in general not necessary in all use cases).

### DIFF
--- a/examples/js/postprocessing/SSAOPass.js
+++ b/examples/js/postprocessing/SSAOPass.js
@@ -1,190 +1,375 @@
-'use strict';
-
 /**
- * Screen-space ambient occlusion pass.
- *
- * Has the following parameters
- *  - radius
- *  	- Ambient occlusion shadow radius (numeric value).
- *  - onlyAO
- *  	- Display only ambient occlusion result (boolean value).
- *  - aoClamp
- *  	- Ambient occlusion clamp (numeric value).
- *  - lumInfluence
- *  	- Pixel luminosity influence in AO calculation (numeric value).
- *
- * To output to screen set renderToScreens true
- *
- * @author alteredq / http://alteredqualia.com/
- * @author tentone
- * @class SSAOPass
+ * @author Mugen87 / https://github.com/Mugen87
  */
+
 THREE.SSAOPass = function ( scene, camera, width, height ) {
+
+	THREE.Pass.call( this );
+
+	this.width = ( width !== undefined ) ? width : 512;
+	this.height = ( height !== undefined ) ? height : 512;
+
+	this.clear = true;
+
+	this.camera = camera;
+	this.scene = scene;
+
+	this.kernelRadius = 8;
+	this.kernelSize = 64;
+	this.kernel = [];
+	this.noiseTexture = null;
+	this.output = 0;
+
+	this.minDistance = 0.005;
+	this.maxDistance = 0.1;
+
+	//
+
+	this.generateSampleKernel();
+	this.generateRandomKernelRotations();
+
+	// beauty render target with depth buffer
+
+	var depthTexture = new THREE.DepthTexture();
+	depthTexture.type = THREE.UnsignedShortType;
+	depthTexture.minFilter = THREE.NearestFilter;
+	depthTexture.maxFilter = THREE.NearestFilter;
+
+	this.beautyRenderTarget = new THREE.WebGLRenderTarget( width, height, {
+		minFilter: THREE.LinearFilter,
+		magFilter: THREE.LinearFilter,
+		format: THREE.RGBAFormat,
+		depthTexture: depthTexture,
+		depthBuffer: true
+	} );
+
+	// normal render target
+
+	this.normalRenderTarget = new THREE.WebGLRenderTarget( width, height, {
+		minFilter: THREE.NearestFilter,
+		magFilter: THREE.NearestFilter,
+		format: THREE.RGBAFormat
+	} );
+
+	// ssao render target
+
+	this.ssaoRenderTarget = new THREE.WebGLRenderTarget( width, height, {
+		minFilter: THREE.LinearFilter,
+		magFilter: THREE.LinearFilter,
+		format: THREE.RGBAFormat
+	} );
+
+	this.blurRenderTarget = this.ssaoRenderTarget.clone();
+
+	// ssao material
 
 	if ( THREE.SSAOShader === undefined ) {
 
-		console.warn( 'THREE.SSAOPass depends on THREE.SSAOShader' );
-		return new THREE.ShaderPass();
+		console.error( 'THREE.SSAOPass: The pass relies on THREE.SSAOShader.' );
 
 	}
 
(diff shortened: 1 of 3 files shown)


mrdoob:
Sweet!

mrdoob:
Many thanks!

looeee:
On my laptop it looks great:

ssao-laptop

But on my mobile it's very jagged:

ssao-mobile

Is it dependant on device resolution? 

Also, "Beauty" output is just the diffuse color, right? Is there a reason why it's called "Beauty" and not "Diffuse"?

Mugen87:
> Is it dependant on device resolution?

The render result gets better when the device has a higher resolution, sure. The demo does currently not regard the pixel device ratio since `SSAO` is already an expensive post-processing pass and because of #14340.

> Is there a reason why it's called "Beauty" and not "Diffuse"?

Yes. It's actually a common term in context of post processing to name the default render pass "beauty pass". I think this is actually better than "diffuse" since "diffuse" is already used in context of lighting models. 

BTW: You find the term "Beauty" in other demos, too (e.g. [here](https://threejs.org/examples/#webgl_postprocessing_sao)).

looeee:
>  I think this is actually better than "diffuse" since "diffuse" is already used in context of lighting models

OK, makes sense, although the choice of the word "Beauty" still seems a little random to me in this context. 

> The demo does not currently set the pixel device ratio since SSAO is already an expensive post-processing pass 

My mobile (Moto G6) has a pixel ratio of 3, so that's probably the issue here.

examples/js/postprocessing/SSAOPass.js
```diff
@@ -1,190 +1,375 @@
...
-				return this.uniforms[ 'radius' ].value;
+		renderer.setClearColor( 0x000000 );
+		renderer.render( this.scene, this.camera, this.beautyRenderTarget, true );
 
-			},
-			set: function ( value ) {
+		// render normals
 
-				this.uniforms[ 'radius' ].value = value;
+		this.renderOverride( renderer, this.normalMaterial, this.normalRenderTarget, 0x7777ff, 1.0 );
```
WestLangley: This is changing the background color and not being reset to the original value, apparently.

Mugen87: `.renderOverride()` should actually reset the original state at the end of the method.

WestLangley: To demonstrate, in `webgl_postprocessing_ssao.html`, remove the scene background:
```javascript
//scene.background = new THREE.Color( 0xa0a0a0 );
```
and add a call to render
```javascript
effectComposer.render();
```

Mugen87: I think I've found the problem right [here](https://github.com/mrdoob/three.js/blob/e94fd574f52bd8d2e6f03e47913d577806c0f5ad/examples/js/postprocessing/SSAOPass.js#L262). `WebGLRenderer.getClearColor()` returns no copy of the color. If the color is [later](https://github.com/mrdoob/three.js/blob/e94fd574f52bd8d2e6f03e47913d577806c0f5ad/examples/js/postprocessing/SSAOPass.js#L275) modified, it also affects the original value. A simple fix would be:
```js
this.originalClearColor.copy( renderer.getClearColor() );
```
and later
```js
renderer.setClearColor( this.originalClearColor );
```

WestLangley: Would #15212 have been helpful in this case, or are there problems with that, too?

Mugen87: Not sure. This seems app-level logic to me. Maybe it's better if the app saves and restores these values. Especially if you need to save/restore more than just a single pair of clear color and alpha values.

WestLangley: >This seems app-level logic to me.

Granted. I just see the same patterns used over and over...

Mugen87: If these methods are added, users could ask for the possibility to save multiple states (think of vertex array objects that represent binding states) and restore/load them with a ID or a name. I think stuff like that would bloat the API of `WebGLRenderer`.

WestLangley: >If these methods are added, users could ask for the possibility...

Anything could happen. Sorry, I'm not buying that as a very compelling argument.

>This seems app-level logic to me.

That's a reasonable argument.

examples/js/postprocessing/SSAOPass.js
```diff
@@ -1,190 +1,375 @@
...
- * @class SSAOPass
+ * @author Mugen87 / https://github.com/Mugen87
  */
+
 THREE.SSAOPass = function ( scene, camera, width, height ) {
 
+	THREE.Pass.call( this );
+
+	this.width = ( width !== undefined ) ? width : 512;
+	this.height = ( height !== undefined ) ? height : 512;
```
moraxy: `this.width` and `this.height` are set here, but the code below still uses normal `width` and `height` to setup the `WebGLRenderTarget`s and the `resolution` uniforms. This causes `WebGLRenderTarget` to be created with `undefined` dimensions on start before getting resized by `onWindowResize`.  
Could that cause any strange problems on devices with bad WebGL implementations later on (the jaggedness on mobile devices for example)?

Mugen87: Good catch! Let me fix this.

Issue #13664, opened 2018-03-22
Generic WorkerLoader


kaisalmen:
`OBJLoader2` introduced `LoaderSupport` in R88. This separated the more generic Worker related functions from the `OBJ` format loader. My idea was that other loaders could use it to allow worker based loading where it makes sense.

I realized during discussion we had (https://github.com/mrdoob/three.js/pull/13263  and #13648) that this is not the right approach. `LoaderSupport` was over-freighted with functions that have been partially removed in R91, but any loader needs to depend on it.
It seems to be the better idea for to change the direction of the dependency and create a generic `WorkerLoader` that uses existing loaders and impose as little as possible on how Parsing must be organized.

`LoaderSupport` will disappear. I already have a WIP version of `OBJLoader(2)` that no longer depends on `LoaderSupport` (#13663). It should be replaced by a generic `WorkerLoader` integrating the useful Worker tools from `LoaderSupport`.

My overall idea for `WorkerLoader` looks as follows:
Any Loaders feature these methods where `serializeParser` is optional:
```Javascript
THREE.AnyLoader.prototype = function () {

	parse: function ( content ) {
	},
	
	load: function ( url, onLoad, onProgress, onError) {
	},

	serializeParser: function( serializeClassFunction, serializeObjectFunction ) {
	}
}
```
A Parser should ideally be: 
 - independent of any other code (three or other libraries) which makes the Worker code a lot smaller and faster to load and use
 - Should isolate Mesh creation in a function allowing to pass from the Worker back to the Main in Transferables whereever possible

The loader is given to `WorkerLoader` which either uses the `serializeParser` from the provided loader, it allows to use an optional Parser which `WorkerLoader` tries to serialize itself or it loads the given files to the worker, where `optionalParser` could be used as code wrapper (handling Transferables for example). 
```javascript
THREE.WorkerLoader = function ( loader, optionalParser, extraLibs2Load ) {
	
	this.workerSupport = new THREE.WorkerLoader.WorkerSupport();
	this.loader = loader;
	
	var code;
	if ( optionalParser ) {
		
		code = this.workerSupport.serializeClass( optionalParser );
		
	} else {
		
		code = this.loader.serializeParser( this.workerSupport.serializeClass, this.workerSupport.serializeObject ),
		
	}
	this.workerSupport.validate( code, 'Parser', extraLibs2Load );
};

THREE.WorkerLoader.prototype = function () {
	
	constructor: THREE.WorkerLoader,

[...]

Issue #13648, opened 2018-03-21
DRACOLoader: Decoding should happen in a Web Worker.


donmccurdy:
Decoding Draco geometries is a long, blocking task, and shouldn't be happening on the main thread. I'm hoping that with some of the LoaderSupport APIs we should be able to configure it to run just as easily in a web worker as not. Proposed API:

```js
THREE.DRACOLoader.setDecoderPath('path/to/decoders/');

// Main thread
var loader1 = new THREE.DRACOLoader()
  // .setParser(new THREE.DRACOLoader.Parser()) // OPTIONAL
  .load('foo.drc',  () => {...});

// Web worker
var loader2 = new THREE.DRACOLoader()
  .setParser(new THREE.DRACOLoader.WWParser())
  .load('foo.drc', () => {...});

// GLTFLoader + WW
var dracoLoader = new THREE.DRACOLoader()
  .setParser(new THREE.DRACOLoader.WWParser());
var gltfLoader = new THREE.GLTFLoader()
  .setDRACOLoader(dracoLoader)
  .load('foo.gltf', () => {...});
```

The distinction between Loader and Parser exists because the Parser cannot know about any `THREE.*` types, to minimize dependencies in the WW.

/cc @takahirox @kaisalmen

***

Aside — In a perfect world the WASM Draco decoder would use threading, but WASM doesn't support that yet, and the decoder would need changes if it did, so using a web worker seems like the place to start.

takahirox:
> Decoding Draco geometries is a long, blocking task, and shouldn't be happening on the main thread.

Agreed.

About the proposed API, what's the purpose of exporting `.setParser()`? `new THREE.DRACOLoader( { 
worker: true } )`  isn't good? (I'm not familiar with neither `DRACOLoader` nor `LoaderSupport` yet.)

donmccurdy:
That would be fine, too. There may be need for a Parser subclass to isolate code for three.js-independent tasks, but it doesn't need to be exposed to the user necessarily.

Related: https://github.com/kaisalmen/WWOBJLoader/issues/36#issuecomment-374842424

kaisalmen:
`LoaderSupport` will disappear. I am currently working on `OBJLoader(2)` version that no longer depends on `LoaderSupport`. It should be replaced by a generic `WorkerLoader` integrating the useful Worker tools from `LoaderSupport`.

My rough idea for it looks as follows:
Any Loaders features these methods where `serializeParser` is optional:
```Javascript
[...]

PR #15249, opened 2018-11-15
DRACOLoader: Web Worker support

donmccurdy:
Possible solution to https://github.com/mrdoob/three.js/issues/13648 — a drop-in replacement for DRACOLoader, intended for use with GLTFLoader. A glTF file may contain multiple compressed meshes, and this loader allows each mesh to be decompressed by a configurable number of workers, without using the main thread. ~~For simplicity, I omitted some functionality of DRACOLoader that glTF does not use, such as point clouds. The result is 150 lines shorter than DRACOLoader.~~

Usage:

```js 
var loader = new THREE.GLTFLoader();
var decoder = new THREE.DRACOLoader();
decoder.setDecoderPath( '/examples/js/libs/draco' );
loader.setDRACOLoader( decoder );
loadel.load( 'model.glb', (gltf) => ... );
```

~~The alternative would be https://github.com/mrdoob/three.js/issues/13664#issuecomment-422168608, which involves LoaderSupport. That's more code, but also has some additional features. I'm not sure how to weigh those tradeoffs at the moment.~~


Mugen87:
I personally prefer this approach since it's a very clear and compact implementation.

donmccurdy:
@sunag I notice you're using the Draco decoder directly in SEA3DLoader rather than using DRACOLoader — if you'd be interested in using something like this, instead, I'm glad to rename the file as WWDRACOLoader to not be glTF-specific.

donmccurdy:
I think there are a couple open things here:

- [ ] Create a new loader vs modifying DRACOLoader?
  - DRACOLoader's `.load()` method wouldn't benefit from this as directly, since each `.drc` file has only a single mesh.
- [ ] Use LoaderSupport, or not?
  - The worker-related complexity seems manageable as-is, but I'm open to discussing this.
- [ ] This implementation currently downloads the draco decoder library to each worker.
  - Presumably the browser's caching prevents that from being a major issue, but it might be better to download once on the main thread anyway.

kaisalmen:
`WWDRACOLoader` is a wrapper for existing `DRACOLoader` utilizing `WorkerLoader` (`LoaderSupport` should disappear in the future...) that just can be configured to be used by `GLTFLoader`:
https://rawgit.com/kaisalmen/three.js/WorkerLoader_OBJLoader/examples/webgl_loader_worker_gltf.html

PR for the preview version will come soon. Then, people can compare 😉 
WorkerLoader and its toolbox are overkill for a single loader, but if multiple loaders are able to re-use these functions and help them flourish, then it'll pay off

sunag:
> @sunag I notice you're using the Draco decoder directly in SEA3DLoader rather than using DRACOLoader — if you'd be interested in using something like this, instead, I'm glad to rename the file as WWDRACOLoader to not be glTF-specific.

Thank you @donmccurdy. Certainly this could happen in the next updates.

kaisalmen:
@donmccurdy nice that the new decoder code has no longer dependencies to three.js! 👍

examples/js/loaders/GLTFDRACOLoader.js
```diff
@@ -0,0 +1,364 @@
...
+    var numFaces = dracoGeometry.num_faces();
+    var numPoints = dracoGeometry.num_points();
+    var numAttributes = dracoGeometry.num_attributes();
+
+    // Add attributes of user specified unique id.
+    for (var attributeName in attributeIDs) {
+      var attributeType = self[ attributeTypes[ attributeName ] ];
+      var attributeId = attributeIDs[attributeName];
+      var attribute = decoder.GetAttributeByUniqueId( dracoGeometry, attributeId );
+      geometry.attributes.push( this.decodeAttribute( draco, decoder, dracoGeometry, attributeName, attributeType, attribute ) );
```
arpu: this should be 
geometry.attributes.push( decodeAttribute( draco, decoder, dracoGeometry, attributeName, attributeType, attribute ) );

examples/js/loaders/GLTFDRACOLoader.js
```diff
@@ -0,0 +1,364 @@
...
+
+      var draco = module.draco;
+      var decoder = new draco.Decoder();
+
+      var buffer = new draco.DecoderBuffer();
+      buffer.Init( new Int8Array( rawBuffer ), rawBuffer.byteLength );
+
+      try {
+
+        return this.decodeGeometry( draco, decoder, buffer, attributeIDs, attributeTypes );
```
arpu: return decodeGeometry( draco, decoder, buffer, attributeIDs, attributeTypes );

examples/js/loaders/GLTFDRACOLoader.js
donmccurdy: Fixed, thanks!

examples/js/loaders/GLTFDRACOLoader.js
donmccurdy: Fixed!

donmccurdy:
Ok, this PR is in better shape now. I think we can now answer a few of the questions mentioned earlier:

(1) Let's replace DRACOLoader, rather than adding a glTF-specific variant of it.
(2) Let's not use LoaderSupport yet. I'm willing to reconsider this once ES modules are source files perhaps, and there's certainly some potential code overlap between this and BasisTextureLoader.
(3) Issues loading the Draco library in web workers efficiently have been fixed.

There are several breaking changes from the original DRACOLoader:

- Removed TriangleStripDrawMode support.
- Removed quantization support (What was `attribute.isQuantized`? How do I test this?).
- Removed setVerbosity, setDrawMode, and setSkipDequantization methods.
- Changed static initialization methods to instance methods.

None of these affect usage in GLTFLoader, so the new and old versions are both compatible with the existing GLTFLoader. The initialization changes slightly:

```js
// Before
var dracoLoader = new THREE.DRACOLoader();
THREE.DRACOLoader.setDecoderPath( 'js/libs/draco/gltf/' )
loader.setDRACOLoader( dracoLoader );

// After
var dracoLoader = new THREE.DRACOLoader();
dracoLoader.setDecoderPath( 'js/libs/draco/gltf/' )
loader.setDRACOLoader( dracoLoader );
```

donmccurdy:
There are two advantages to this approach:

1. Decoding does not block the main thread. This is true regardless of the model, and significant especially for WebXR applications.
2. Decoding is parallelized across workers, and may or may not complete more quickly depending on the file and the device concurrency. For a model with a single mesh, it will be slower, but I consider that an acceptable tradeoff because of (1). For models with many meshes, and especially many _large_ meshes, the load time improvement is significant.

Running tests on two glTF models with multiple meshes (attached below), here are the results:

* **matilda.glb** (103 meshes, 6MB uncompressed geometry)
  - 3% load time decrease, averaged over 5 tests. (527ms -> 512ms)
* **lovecraftian.glb** (18 meshes, 45MB uncompressed geometry)
  - 60% load time decrease, averaged over 5 tests. (3815ms -> 1509ms)

[draco_perf_tests.zip](https://github.com/mrdoob/three.js/files/3308973/draco_perf_tests.zip)

| matilda | lovecraftian |
|---|---|
| Screen Shot 2019-06-19 at 11 04 48 PM | Screen Shot 2019-06-19 at 11 04 35 PM |

donmccurdy:
The file in this PR is named `DRACOLoader2.js`, but I'd like to update DRACOLoader directly if there are no objections to the breaking changes mentioned in https://github.com/mrdoob/three.js/pull/15249#issuecomment-503886776.

donmccurdy:
I've added the changes directly to DRACOLoader (not a new loader), and I think this is ready  – here's a demo of my viewer, using the web workers:

https://gltf-viewer-experimental.donmccurdy.com/

Mugen87 (approved):

Issue #15250, opened 2018-11-15
Use standingMatrix in WebVR as an option


unilimes:
Now, if we use WebVR, we can't use matrix of camera, to control position of left/right cameras. Our idea is to add some flag in WebVRManager such as `.useStandingMatrix` to define which matrix will be used.

MrMaksymov:
It would be cool, because sometimes you need to animate the camera, and doing it through the camera wrapper is not always convenient.

tnormandao:
The reason why I did not use the solution out of the box, but began to reinvent my wheel.

Mugen87:
WebVR support has been removed, see #18129.

PR #15253, opened 2018-11-15
Added option to use standingMatrix

unilimes:
https://github.com/mrdoob/three.js/issues/15250


Mugen87:
WebVR support has been removed, see #18129.

PR #15268, opened 2018-11-16
Simplifying Object.lookAt() and adding high-level Object.lookAtObject3D() method #15267

lguminski:
Solves #15267

### DIFF
--- a/docs/api/en/core/Object3D.html
+++ b/docs/api/en/core/Object3D.html
@@ -289,11 +289,11 @@
 
 		<h3>[method:null lookAtObject3D]( [param:Object3D object] )</h3>
 		<p>
-			object - An object in the same scene.<br /><br />
+		object - An object in the same scene.<br /><br />
 
-			Rotates the object to face the object given as a parameter.<br /><br />
+		Rotates the object to face the object given as a parameter.<br /><br />
 
-			This method supports all objects in the scene, no matter how rotated and/or translated their parent(s) are.
+		This method supports all objects in the scene, no matter how rotated and/or translated their parent(s) are.
 		</p>
 
 		<h3>[method:Array raycast]( [param:Raycaster raycaster], [param:Array intersects] )</h3>
--- a/docs/api/en/core/Object3D.html
+++ b/docs/api/en/core/Object3D.html
@@ -281,12 +281,19 @@
 		<h3>[method:null lookAt]( [param:Vector3 vector] )</br>
 		[method:null lookAt]( [param:Float x], [param:Float y], [param:Float z] )</h3>
 		<p>
-		vector - A vector representing a position in world space.<br /><br />
-		Optionally, the [page:.x x], [page:.y y] and [page:.z z] components of the world space position.<br /><br />
+		vector - A vector representing a position in parent's space.<br /><br />
+		Optionally, the [page:.x x], [page:.y y] and [page:.z z] components of the parent's space position.<br /><br />
 
-		Rotates the object to face a point in world space.<br /><br />
+		Rotates the object to face a point in parent's space.
+		</p>
 
-		This method does not support objects with rotated and/or translated parent(s).
+		<h3>[method:null lookAtObject3D]( [param:Object3D object] )</h3>
+		<p>
+			object - An object in the same scene.<br /><br />
+
+			Rotates the object to face the object given as a parameter.<br /><br />
+
+			This method supports all objects in the scene, no matter how rotated and/or translated their parent(s) are.
 		</p>
 
 		<h3>[method:Array raycast]( [param:Raycaster raycaster], [param:Array intersects] )</h3>
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -351,19 +351,19 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
     var m1 = new Matrix4();
     var position = new Vector3();
 
-    return function lookAtObject3D(object) {
+    return function lookAtObject3D( object ) {
 
-      this.parent.updateWorldMatrix(true, false);
+      this.parent.updateWorldMatrix( true, false );
 
-      object.updateWorldMatrix(true, false);
+      object.updateWorldMatrix( true, false );
 
       m1
-        .getInverse(this.parent.matrixWorld)
-        .multiply(object.matrixWorld);
+        .getInverse( this.parent.matrixWorld )
+        .multiply( object.matrixWorld );
 
-      position.setFromMatrixPosition(m1);
+      position.setFromMatrixPosition( m1 );
 
-      this.lookAt(position);
+      this.lookAt( position );
 
       return;
     };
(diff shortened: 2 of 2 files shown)


WestLangley:
You are redefining what `Object3D.lookAt( target )` does when you declare that `target` is not in world space but in the object's parent's space. This will break existing three.js apps. Such a redefinition is almost surely not going to be acceptable.

Also, your methods do not appear to keep the object "right-side-up" in world space. Are you aware that `lookAt()` does more than just orient the object to face a particular direction?

Mugen87:
Yes, this parametrization is unusual. Unity's `Transform.lookAt()` method also defines the target position in world space.

lguminski:
@WestLangley @Mugen87 thank you for the comments. Let me first address your concerns about backward compatibility of this reimplementation:

- if object is connected to the scene or is parentless (like a detached camera), the change is fully backwards compatible, because `this.matrix` equals `this.matrixWorld`.
- if the object had a parent, then the change _is not_ backwards compatible. My point, however, is that such objects were never properly handled anyway, and that was explicitly stated in [documentation](https://threejs.org/docs/#api/en/core/Object3D.lookAt): "_This method does not support objects with rotated and/or translated parent(s)_."

The latest version (r98) introduced handling of rotated parents, but the disclaimer in the [documentation](https://threejs.org/docs/#api/en/core/Object3D.lookAt) remained, so I don't think anyone started using the method in that way.

The advantage of this reimplementation is that in case a camera is bound to a moving parent, like:

> movingObject.add(camera);
> controls = new THREE.OrbitControls(camera);

, then the behaviour of `OrbitControls` becomes consistent with what you would expect i.e. it respects the rotation of the parent, allowing you for a smooth navigation around the child ([demo](http://jsfiddle.net/lguminski/24ahxn5w/show)).

In addition, this PR introduces a new method (`lookAtObject3D()`) that correctly rotates object toward any other object in the scene. (The `r98` version of `lookAt()` still requires users to recalculate the target object's position to `x`, `y`, `z` coordinates in world space themselves.  The `lookAtObject3D()` method does the work for them.)

If the backwards compatibility is still a concern, you could introduce this implementation under another name (like `localLookAt()`), but then the implementations of `OrbitControls` and `TrackballControls` would need to be touched, to make use of this method instead of `lookAt()`.

I cannot tell anything about the Unity convention as I don't know that system. I also don't know if the [original problem](https://github.com/mrdoob/three.js/issues/15267) that triggered the PR, exists in that system or not.

WestLangley:
@lguminski So, with this PR, if I want an object to look at the world origin, I can't call `object.lookAt( 0, 0, 0 )`. Instead, I must call `object.lookAt( x, y, z )`, where `( x, y, z )` are the coordinates of the world origin in the frame of reference of the object's parent. Correct? And, how will the object be oriented relative to `object.up`?

Also, can you please respond to my question above?

>Are you aware that lookAt() does more than just orient the object to face a particular direction?

lguminski:
@WestLangley with this PR, 

    object.lookAt( 0, 0, 0 );

rotates object toward the world origin if the object's frame reference is world (i.e. `object.parent` is `null` or `object.parent` is `scene`). So for such objects nothing changes, and the PR is backward compatible.

> Also, your methods do not appear to keep the object "right-side-up" in world space. Are you aware that lookAt() does more than just orient the object to face a particular direction?

I understand that. For objects that I just mentioned (i.e. which are in the world space), "right-side-up" in world space is preserved. So again full backward compatibility.

Please correct me if I am wrong, but what I am saying is that this change is not backward compatible only for users who did not respect your disclaimer ("This method does not support objects with rotated and/or translated parent(s)."). Such users will need to modify their code from `object.lookAt( 0, 0, 0 );` to:
    
    const WORLD_CENTER = new THREE.Object3D();

    o.lookAtObject3D( WORLD_CENTER );

if they want to keep old behaviour.

Obviously it would be nice to support all old behaviours, and I can understand that. In this case, however, it is like trying to support use cases, which you explicitly said that you don't support ("_This method does not support..._"). And if I would have on one hand the decision to support unsupported use cases, and on the other the simplification of the code and fixing broken use cases (this broken `OrbitControls` issue), I would decide for the latter. But obviously I am not maintainer of this code, so it is not my decision.

lguminski:
One more comment. The code

    const WORLD_CENTER = new THREE.Object3D();

    o.lookAtObject3D( WORLD_CENTER );

keeps "right-side-up" orientation relative to parent. Not to world. So in the sense of orientation it is not exactly the equivalent of `object.lookAt( 0, 0, 0 );` behaviour added in r98 version.

But the r98 support for objects with parents is inactive, because the documentation still says that such objects are not supported (the disclaimer "This method does not support objects with rotated and/or translated parent(s)." is still in documentation). Therefore you don't have any obligation to preserve backward compatibility with r98 for this use case.

That's why I still believe, that this PR will make three.js more robust without any side-effects. And this is probably the last moment the code can be fixed in a clean way. If your disclaimer wouldn't be in documentation, then the only way to fix the `OrbitControls` issue without affecting users, would be through the workaround with the second camera (explained in the issue).

WestLangley:
As I said,

>So, with this PR, if I want an object to look at the world origin, I can't call object.lookAt( 0, 0, 0 ). Instead, I must call object.lookAt( x, y, z ), where ( x, y, z ) are the coordinates of the world origin in the frame of reference of the object's parent.

I believe that concept is a bit too complex for most users.

Also, in your `OrbitControls` demo, where I have added world axes for clarity, you can see that the camera is upside down relative to world space.

screen shot 2018-11-17 at 7 25 42 pm

You are treating `Object.up` -- not as a world direction --  but as a relative direction.

lguminski:
@WestLangley thank you for giving it attention. And your patience to my strong opinions.

You are right. I am using `up` as a relative direction. And I believe it is correct. Let me explain why.

The reason is that I tend to think more like an architect than a developer. And when I model software, I always try to make it consistent with the domain it models. Because I know that inconsistencies become design flaws, and they will bite sooner or later.

In this case `three.js` aims to model physical reality. In the reality, when I take a telescope and look at, say, Jupiter, this action doesn't turn me upside down to match my orientation with the orientation of Jupiter. Instead, I keep standing firmly on the ground, and my relative orientation to my parent (the Earth) doesn't change. Why would `three.js` behave differently? That was my reasoning.

WestLangley:
>thank you for giving it attention. And your patience to my strong opinions.

:-)

Since you are redefining `lookAt( point )` to accept a point defined in the parent's coordinate space, and since you are treating `object.up` as a relative direction, this PR is not going to be acceptable.

However, I do understand your approach, and it can be a valid methodology -- just not here.

Adding `Object3D.lookAtObject( object )` as a convenience method is possible. However, the convenience method must work in all cases, and that means it must traverse the scene graph to compute the correct world position of the target `object`. That is inefficient inside the render loop, because the renderer has already traversed the scene graph for you. `Object3D.lookAt( point )` doesn't have that problem.

lguminski:
I can understand. Well, at least now it is a well thought through decision.

PR #15283, opened 2018-11-20
Fix for EXTENSIONS.KHR_LIGHTS

KonstantinEletskiy:
In last Blender GLTF exporter naming was changed

This fix supports previous gltf files and newly created

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -171,7 +171,8 @@ THREE.GLTFLoader = ( function () {
 					var extensionsRequired = json.extensionsRequired || [];
 
 					switch ( extensionName ) {
-
+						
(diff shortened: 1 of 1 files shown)


donmccurdy:
Hi, welcome and thanks for the PR! Unfortunately I don't think we can merge this — the older working name of the extension, `KHR_lights`, was never ratified and has been removed from all tools I know of. In Blender it was (or should have been) marked as experimental. The final name is `KHR_lights_punctual`, and supporting the incorrect name will lead to confusion down the road. In addition, there were more differences from the earlier version of the extension than simply renaming it.

If you need help converting models that contain `KHR_lights` to use `KHR_lights_punctual`, I can probably help with that, but I don't think GLTFLoader should support `KHR_lights` directly.

PR #15289, opened 2018-11-21
GLTFExporter: serialize scene userData

makc:
This is already read by the loader. The code copied from processNode.


donmccurdy (approved):

mrdoob:
Thanks!

PR #15294, opened 2018-11-21
reorder properties for better debugging experience

greggman:
object properties are ordered in JavaScript. Devtools use that to decide how to display them. So, moving name and type to the front seems like it provides a better debugging experience?

Old:

before

New:

after

Maybe few others should be moved too? Seems like uuid isn't that useful for most debugging.

Heck, maybe `type` is not important either as the debugger is already displaying the type. Name being the first property thought seems immensely useful

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -21,10 +21,10 @@ function Object3D() {
 
 	Object.defineProperty( this, 'id', { value: object3DId ++ } );
 
-	this.uuid = _Math.generateUUID();
-
 	this.name = '';
 	this.type = 'Object3D';
+
+	this.uuid = _Math.generateUUID();
 
 	this.parent = null;
 	this.children = [];


Mugen87 (approved):

WestLangley:
I use the UUID most often. It is useful when identifying when a material or geometry is reused. It is important.

The type is the class name, usually, and is used for `shader.name`. It is also useful with the shader editor.

The name defaults to a blank string. Occasionally it is set.

IMO, this PR would not be "better" as far as I am concerned.

donmccurdy:
Putting `type` before `uuid` would be convenient, IMO. 

> maybe type is not important either as the debugger is already displaying the type.

This isn't true in Firefox, and maybe not in minified code generally. Example:

screen shot 2018-11-21 at 9 29 27 am

Issue #15261, opened 2018-11-16
add dispose() to BufferAttribute


yaoyao-cn:
is it possible to add a dispose() method to BufferAttribute?

i need to dispose buffer attribute directly, not through the buffer geometry.
so i will change BufferAttribute and WebGLAttributes a bit, but i am not sure whether i am right.

i will change like this:
1.Make BufferAttribute as a sub classs of EventDispatcher to dispatch the dispose event
2.Handle dispose event in WebGLAttributes

am i right?  thank you !

##### Three.js version

- [x] Dev
- [ ] r98
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
> i need to dispose buffer attribute directly, not through the buffer geometry.

Can you please explain your use case a bit? How does your workflow look like?

yaoyao-cn:
@Mugen87 
ok, let me try to explain my workflow.

i am developing a GIS system like goole earth web and i need to render the whole earth with DOM and DEM. the earth is renderd through dynamic tiles which are subdivided autometically base on the camera distance.
every tile is a buffer geometry initialized with no buffer attributes and index buffer. the buffer attributes and index buffer are set by the tile manager every frame. so the shape of the tile can change every frame depending on the incoming DOM and DEM data from internet.

tiles are cached, every frame the tile manager pick some free tiles, set the buffer attribute, construct meshes(meshes are also cached) and then add the meshes to the threejs scene. after the render process, the tile manager remove the meshes from threejs scene and mark the tiles used as free tiles again.

in a word. meshes , buffer geometries and buffer attributes are separated objects for the system. the system know how to dispose all of them. buffer attributes and index buffers are shared between buffer geometries.

so, i need a dispose method of buffer attribute. just like the Texture.dispose() of material.

Mugen87:
I think that sounds reasonable.

Mugen87:
Closing, see https://github.com/mrdoob/three.js/pull/17063#issuecomment-737993363.

PR #15308, opened 2018-11-23
add dispose method to BufferAttribute

zhaoy-dev:
add dispose method to BufferAttribute
refer to this issues:https://github.com/mrdoob/three.js/issues/15261


src/core/BufferAttribute.js
```diff
@@ -319,6 +312,22 @@ Object.assign( BufferAttribute.prototype, {
 
 		return new this.constructor( this.array, this.itemSize ).copy( this );
 
+	},
+
+	dispose: function () {
```
Mugen87: Can you please also update the documentation page of `BufferAttribute`? You have to add the `dispose()` method similar to the other pages (e.g. https://threejs.org/docs/#api/en/core/BufferGeometry.dispose).

Mugen87 (changes requested):

yaoyao-cn:
geometry.dispose() will dispose another attribute successfully and take no effect on attribute1 when called after attribute1.dispose().

src/core/BufferAttribute.js
yaoyao-cn: i have updated the doc and added InterleavedBufferAttribute.dispose method like BufferAttriubte

Mugen87 (approved):
lgtm. Let's see what the others think about the PR :blush:

Mugen87:
@yaoyao-cn Can you please resolve the merge conflicts? 😇

docs/api/en/core/InterleavedBufferAttribute.html
```diff
@@ -95,7 +95,8 @@ <h3>[method:null setXYZ]( [param:Integer index], [param:Float x], [param:Float y
 		<h3>[method:null setXYZW]( [param:Integer index], [param:Float x], [param:Float y], [param:Float z], [param:Float w] ) </h3>
 		<p>Sets the x, y, z and w components of the item at the given index.</p>
 
-
+		<h3>[method:null dispose]()</h3>
+		<p>Disposes the buffer attribute from memory directly.Also you can dispose the buffer attribute through [page:BufferGeometry.dispose BufferGeometry.dispose] implicitly.</p>
```
donmccurdy: I don't see how disposing an InterleavedBufferAttribute would work, since it would share an underlying InterleavedBuffer with other attributes. If we add this method, I think that invoking it on an interleaved attribute should throw an error.

Mugen87: `InterleavedBufferAttribute` is not derived from `BufferAttribute`. So I think a note in the docs is sufficient.

Mugen87:
Closing, see #17063.

yaoyao-cn:
@Mugen87 i just see it now. is it necessary to resolve the conflicts caused by #15308 now ?

Mugen87:
No, I have prepared everything in the new PR. Thanks!

PR #15316, opened 2018-11-26
Update README.md

Mugen87:
- Changed `BoxGeometry` to `BoxBufferGeometry`.
- Accordingly changed the fiddle. Also switched from rawgit URL to https://threejs.org/build/three.min.js

https://jsfiddle.net/cfLzo0qr/

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ function init() {
 
 	scene = new THREE.Scene();
 
-	geometry = new THREE.BoxGeometry( 0.2, 0.2, 0.2 );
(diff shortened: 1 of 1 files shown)


looeee:
Why did you close this? 

We should update both the readme and the Creating a Scene page to use BufferGeometry.

Mugen87:
I thought there is no interest in this change.

> Creating a Scene page 

I've totally missed this one. You're right, we should also update the guide. If @mrdoob agrees, I'll make a new PR and change both files.

PR #15327, opened 2018-11-27
Use language when linking to documentation source for API.

jsantell:
The docs for API are sorted by language -- I don't know the history of this change that added this (a17a1228da7a168dbee30182e1328d90891fc8b3) because the docs were sorted by language there as well, but this fixes the "Edit" button found in API examples e.g. https://threejs.org/docs/#api/en/scenes/Scene


jsantell:
Oops, looks like there are two other PRs for this #15136 #15157  :smile:

PR #15356, opened 2018-12-03
Implemented raycast method for fat lines.

### LINKED ISSUE #15006: Raycasting support for Line2Segments

I'm not sure if this is a bug or a new feature to be added. When I try to click on a line2segment, I never get an intersection. I have been successful with getting intersections with line segments.

 I have attached a fiddle that demonstrates the problem:

https://jsfiddle.net/aitheorem/gr7e6why/66/

##### Three.js version
 r97

##### Browser

- [x] All of them

##### OS
 I don't know if this a problem only windows. I have not tested the other operating systems.
- [x] Windows

weeklyTea:
Fixes #15006

### DIFF
--- a/examples/js/lines/LineSegments2.js
+++ b/examples/js/lines/LineSegments2.js
@@ -58,6 +58,10 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
 
 		var ray = new THREE.Ray();
 		var inverseMatrix = new THREE.Matrix4();
+		var pointOnSegment = new THREE.Vector3();
+		var pointOnRay = new THREE.Vector3();
+		var startPoint3 = new THREE.Vector3();
+		var endPoint3 = new THREE.Vector3();
 
 		return function raycast( raycaster, intersects ) {
 
@@ -67,10 +71,6 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
 			ray.copy( raycaster.ray ).applyMatrix4( inverseMatrix );
 
 			const geometry = this.geometry;
-			var pointOnSegment = new THREE.Vector3();
-			var pointOnRay = new THREE.Vector3();
-			var startPoint3 = new THREE.Vector3();
-			var endPoint3 = new THREE.Vector3();
 			var count = geometry.attributes.instanceStart.count;
 
 			for ( var i = 0; i < count; ++ i ) {
--- a/examples/js/lines/LineSegments2.js
+++ b/examples/js/lines/LineSegments2.js
@@ -59,13 +59,17 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
 		var ray = new THREE.Ray();
 		var inverseMatrix = new THREE.Matrix4();
 		var pointOnSegment = new THREE.Vector3();
+		var pointOnSegmentNDC = new THREE.Vector3();
 		var pointOnRay = new THREE.Vector3();
+		var pointOnRayNDC = new THREE.Vector3();
 		var startPoint3 = new THREE.Vector3();
 		var endPoint3 = new THREE.Vector3();
 
 		return function raycast( raycaster, intersects ) {
 
-			var precision = raycaster.linePrecision; // todo: calculate proper precision
+			if ( raycaster.camera === undefined ) return;
+
+			var precision = this.material.linewidth / this.material.resolution.y;
 
 			inverseMatrix.getInverse( this.matrixWorld );
 			ray.copy( raycaster.ray ).applyMatrix4( inverseMatrix );
@@ -78,7 +82,13 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
 				startPoint3.fromBufferAttribute( geometry.attributes.instanceStart, i );
 				endPoint3.fromBufferAttribute( geometry.attributes.instanceEnd, i );
 
-				var distSq = ray.distanceSqToSegment( startPoint3, endPoint3, pointOnRay, pointOnSegment );
+				ray.distanceSqToSegment( startPoint3, endPoint3, pointOnRay, pointOnSegment );
+
+				// Set z to 0 because for raycasting we shouldn't consider depth.
+				pointOnSegmentNDC.copy( pointOnSegment ).project( raycaster.camera ).setZ( 0 );
+				pointOnRayNDC.copy( pointOnRay ).project( raycaster.camera ).setZ( 0 );
+
+				var distSq = pointOnSegmentNDC.distanceTo( pointOnRayNDC )
 
 				if ( distSq > precision ) continue;
 
--- a/examples/js/lines/LineSegments2.js
+++ b/examples/js/lines/LineSegments2.js
@@ -54,6 +54,56 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
 
 	}() ),
 
+	raycast: ( function () {
+
+		var ray = new THREE.Ray();
+		var inverseMatrix = new THREE.Matrix4();
+
+		return function raycast( raycaster, intersects ) {
+
+			var precision = raycaster.linePrecision; // todo: calculate proper precision
+
+			inverseMatrix.getInverse( this.matrixWorld );
+			ray.copy( raycaster.ray ).applyMatrix4( inverseMatrix );
+
+			const geometry = this.geometry;
+			var pointOnSegment = new THREE.Vector3();
+			var pointOnRay = new THREE.Vector3();
+			var startPoint3 = new THREE.Vector3();
+			var endPoint3 = new THREE.Vector3();
+			var count = geometry.attributes.instanceStart.count;
+
+			for ( var i = 0; i < count; ++ i ) {
+
+				startPoint3.fromBufferAttribute( geometry.attributes.instanceStart, i );
+				endPoint3.fromBufferAttribute( geometry.attributes.instanceEnd, i );
+
+				var distSq = ray.distanceSqToSegment( startPoint3, endPoint3, pointOnRay, pointOnSegment );
+
+				if ( distSq > precision ) continue;
+
+				pointOnRay.applyMatrix4( this.matrixWorld );
+
+				var distance = raycaster.ray.origin.distanceTo( pointOnRay );
+
+				if ( distance > raycaster.near && distance < raycaster.far ) {
+
+					intersects.push( {
+
+						distance: distance,
+						point: pointOnSegment.clone().applyMatrix4( this.matrixWorld ),
+						object: this
+
+					} );
+
+				}
+
+			}
+
+		};
+
+	}() ),
+
 	copy: function ( source ) {
 
 		// todo
--- a/examples/webgl_lines_fat.html
+++ b/examples/webgl_lines_fat.html
@@ -63,7 +63,8 @@
 
 			var line, renderer, scene, camera, camera2, controls;
 			var line1;
-			var matLine, matLineBasic, matLineDashed;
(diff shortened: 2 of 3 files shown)


Mugen87:
For testing: https://raw.githack.com/weeklyTea/three.js/fatLines_raycast/examples/webgl_lines_fat.html

examples/js/lines/LineSegments2.js
```diff
@@ -54,6 +54,56 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
...
+
+		return function raycast( raycaster, intersects ) {
+
+			var precision = raycaster.linePrecision; // todo: calculate proper precision
+
+			inverseMatrix.getInverse( this.matrixWorld );
+			ray.copy( raycaster.ray ).applyMatrix4( inverseMatrix );
+
+			const geometry = this.geometry;
+			var pointOnSegment = new THREE.Vector3();
```
Mugen87: It's better to create these vectors once in the closure scope and not every time `raycast()` is called.

examples/webgl_lines_fat.html
```diff
@@ -278,6 +335,10 @@
 
 				} );
 
+				gui.add( param, 'interactive' ).onChange( function ( val ) {
+					interactive = val
+				})
```
Mugen87: Missing semicolon.

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,59 @@
...
+
+					var res = intersects.filter( function ( res ) {
+
+						return res && res.object;
+
+					} )[ 0 ];
+
+					if ( res && res.object ) {
+
+						selectedObject = res.object;
```
Mugen87: `selectedObject` is never declared.

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,59 @@
...
+
+				if ( intersects.length > 0 ) {
+
+					if ( prevLinewidth === undefined ) {
+
+						prevLinewidth = matLine.linewidth
+
+					}
+
+					var res = intersects.filter( function ( res ) {
```
Mugen87: Why not just using `intersects[ 0 ]`? Right now, you are creating in each call of `onDocumentMouseMove` an inline function.

Mugen87 (changes requested):

weeklyTea:
@Mugen87 Thanks for review! Should I create new branch with fixes and then create new pull request? Or there is easier way?

Mugen87:
You can just add additional commits to this branch.

examples/js/lines/LineSegments2.js
weeklyTea: Fixed in 96dc01126 .

examples/webgl_lines_fat.html
weeklyTea: Fixed in 96dc01126 .

examples/webgl_lines_fat.html
weeklyTea: Fixed in 96dc01126 .

examples/webgl_lines_fat.html
weeklyTea: Sorry, redundant copypaste from webgl_raycast_sprite.html. Fixed in 96dc01126 .

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,54 @@
...
+
+						prevLinewidth = matLine.linewidth
+
+					}
+
+					var res = intersects[ 0 ];
+
+					if ( res && res.object ) {
+
+						res.object.selectedObject.material.linewidth = prevLinewidth + prevLinewidth / 3;
```
Mugen87: It should be:
```js
res.object.material.linewidth
```
Otherwise you produce a runtime error: `Uncaught TypeError: Cannot read property 'material' of undefined`

weeklyTea: \>.< Sorry.. Fixed in 0a584ae45f

Mugen87:
Okay, let's see what @WestLangley thinks about the PR.

Mugen87 (dismissed):

examples/js/lines/LineSegments2.js
```diff
@@ -54,6 +54,56 @@ THREE.LineSegments2.prototype = Object.assign( Object.create( THREE.Mesh.prototy
...
+		var ray = new THREE.Ray();
+		var inverseMatrix = new THREE.Matrix4();
+		var pointOnSegment = new THREE.Vector3();
+		var pointOnRay = new THREE.Vector3();
+		var startPoint3 = new THREE.Vector3();
+		var endPoint3 = new THREE.Vector3();
+
+		return function raycast( raycaster, intersects ) {
+
+			var precision = raycaster.linePrecision; // todo: calculate proper precision
```
WestLangley: Line width is in pixels. What you are doing here is not correct. You can't mix world units and pixels. Maybe, if you see how sprite raycasting is handled, there is a workaround.

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,54 @@
...
+
+					matLine.linewidth = prevLinewidth
+					prevLinewidth = undefined
+
+				}
+
+			}
+
+			var raycaster = new THREE.Raycaster();
+			var mouseVector = new THREE.Vector3();
```
WestLangley:
```javascript
new THREE.Vector2();
```

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,54 @@
...
+
+			function getIntersects( x, y ) {
+
+				x = ( x / window.innerWidth ) * 2 - 1;
+				y = - ( y / window.innerHeight ) * 2 + 1;
+
+				mouseVector.set( x, y, 0.5 );
```
WestLangley:
```javascript
mouseVector.set( x, y );
```

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,54 @@
...
+				raycaster.setFromCamera( mouseVector, camera );
+
+				return raycaster.intersectObject( scene, true );
```
WestLangley: Perhaps you should pass in the `optionalTarget` in this case. The need for this arg was discussed in a previous PR. There may be different opinions on this.

examples/js/lines/LineSegments2.js
WestLangley: Well, that won't work because lines can be fat and are specified in pixels -- unless you are willing to require to user to set the precision based on the camera settings and screen resolution, and accept that approach as "good-enough".

weeklyTea: @WestLangley , @Mugen87 , yes you rigth. In our proj we have another implementation: we project 3d points into 2d pixel coords using `Vector3.project(camera)` so then it is ok to compare cursor coordinates and linewidth, but also we use our custom raycaster contains camera.. I checked raycast method for sprites, am i right it is possible to project point3 onto a view plane using `this.modelViewMatrix` (seems function `transformVertex` in src/objects/Spite.js does what i described)? If yes our approach can be easily reimplemented without using `camera`.

examples/webgl_lines_fat.html
weeklyTea: Could you link previous PR ?

WestLangley: https://github.com/mrdoob/three.js/pull/12872

examples/js/lines/LineSegments2.js
WestLangley: @weeklyTea I am glad fat lines are working for you! Perhaps you will be able to rework this PR. Also see #14936. Some of the issues may be similar...

examples/webgl_lines_fat.html
weeklyTea: Done in 3196781fb476

examples/webgl_lines_fat.html
weeklyTea: Done in 3196781fb476

examples/js/lines/LineSegments2.js
WestLangley: Yes, I mentioned option (2), but I don't like it, either. If you want to file a PR for (1), we can have a look. I do not know if (1) will be acceptable to others or not.

examples/webgl_lines_fat.html
```diff
@@ -223,6 +226,56 @@
 
 			//
 
+			function onDocumentMouseMove( event ) {
+
+				if ( ! interactive ) return;
+
+				event.preventDefault();
+
+				var intersects = [];
```
WestLangley: Now, you are instantiating an array every mouse move. That defeats the purpose. You will need to create one array in a closure and reuse it. See https://github.com/mrdoob/three.js/blob/dev/docs/api/en/core/Raycaster.html#L151.

Like I said elsewhere, others may have varying opinions on the usefulness of doing this.

/ping @Mugen87 for opinion

WestLangley:
@weeklyTea The raycasting appears to me to have huge positional errors... ???

weeklyTea:
> @weeklyTea The raycasting appears to me to have huge positional errors... ???

As I understand it happens because resolution in shader is not equals to resolution in raycast method. Seems guilty line is https://github.com/mrdoob/three.js/pull/15356/commits/2b056a3148c7b2556291b828c6f82d56d6b0134e#diff-3807f6cd3995839aef359ea88e0eb2f5R220 if you comment it everything should work fine.

weeklyTea:
@Mugen87 @WestLangley could yout give a feedback about this solution? Is it ok or should I change something?

src/core/Raycaster.js
```diff
@@ -6,14 +6,16 @@ import { Ray } from '../math/Ray.js';
  * @author stephomi / http://stephaneginier.com/
  */
 
-function Raycaster( origin, direction, near, far ) {
+function Raycaster( origin, direction, near, far, camera = undefined ) {
```
Mugen87: Default parameters are not allowed in the code base so far.

src/core/Raycaster.js
```diff
@@ -65,12 +67,14 @@ Object.assign( Raycaster.prototype, {
 
 	linePrecision: 1,
 
-	set: function ( origin, direction ) {
+	set: function ( origin, direction, camera = undefined ) {
```
Mugen87: Same here...

src/core/Raycaster.js
```diff
@@ -65,12 +67,14 @@ Object.assign( Raycaster.prototype, {
...
 
 		// direction is assumed to be normalized (for accurate distance calculations)
 
 		this.ray.set( origin, direction );
 
+		this.camera = camera;
```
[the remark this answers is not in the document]
Mugen87: I personally think this tight coupling between `Raycaster` and `Camera` is a bad design. I vote for option 2 (define a precision value).

Mugen87 (changes requested):

WestLangley:
>@weeklyTea The raycasting appears to me to have huge positional errors... ???
>>As I understand it happens because resolution in shader is not equals to resolution in raycast method. Seems guilty line is 2b056a3#diff-3807f6cd3995839aef359ea88e0eb2f5R220 if you comment it everything should work fine.

The raycasting is still not accurate when I test it. I tested with 100 px line width.

src/core/Raycaster.js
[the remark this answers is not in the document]
weeklyTea: But 2 will work only if start of line will be near to end by Z coordinate (I mean line will has small depth).. 
Why saving camera in raycaster is bad? Will it be better if the only camera matrices will be saved?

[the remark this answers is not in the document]
weeklyTea: @WestLangley I've taken a look how to make GPU picking and it seems in this case we still need additional data like renderer and mouse position (the last one is needed only for perspective camera) :(

Mugen87:
Closing since the implementation is not yet satisfying. Moving on with this issue in #17160

PR #15357, opened 2018-12-03
[wip] optimize(?) buffer binding between calls

pailhead:
It's possible to some times end up binding the same buffer over and over between calls. It doesn't seem very likely, and might not happen in such a number to really matter.

## r99:

https://raw.githack.com/pailhead/three.js/test-buffer-binding/examples/test.html

```
bindBuffer: 12
vertexAttribPointer: 8
```

## this branch:
https://raw.githack.com/pailhead/three.js/test-buffer-binding-build/examples/test.html

```
bindBuffer: 9
vertexAttribPointer: 5
```

The example basically renders the same sphere, but with different materials and (not sure how to categorize) instancing. The index and the positions are the same, but other draw calls may include normals and an instancing attribute.

It seems that attributes are treated on the geometry level, but it may be possible that two consecutive geometries share some of the attibutes and locations. I imagine it may not happen often if at all most of the time, but i'm wondering if there is some kind of an approach that would see a benefit from doing this diffing more granular. Any thoughts?

### DIFF
--- a/examples/test.html
+++ b/examples/test.html
@@ -1,0 +1,238 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - geometry - extrude shapes from geodata</title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+		<style>
+		body {
+			color: #ffffff;
+			font-family: Monospace;
+			font-size: 13px;
+			text-align: center;
+			font-weight: bold;
+			background-color: #000000;
+			margin: 0px;
+			overflow: hidden;
+		}
+
+		#info {
+			position: absolute;
+			top: 0px;
+			width: 100%;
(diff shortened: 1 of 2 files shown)


pailhead:
image

It didn't seem to catch the instanced attribute (4 here should be bound only once):

image

And this illustrates the point, where in order to reduce these the most, it would make sense to render the instanced with normals, after the non instanced one with normals.

mrdoob:
Have you tried a scene with, say, 1000 objects to see if this improves performance?

pailhead:
Hmmm, i wasn't sure what a good real world test would be, but i tried to make a synthetic one, where i made a bunch of iterations and render them in the worst possible order. At first it felt like a small increase, now the same scene is actually rendering slower, i don't quite understand why :)

Less commands though:
image
image

Mugen87:
A more promising approach is #14600 which completely eliminates redundant WebGL state changes.

mrdoob:
@pailhead seems like this should be closed?

Issue #14634, opened 2018-08-03
webgl_geometry_colors_lookuptable.html needs to be revisited


WestLangley:
The example is not disposing of geometries properly. Geometry should likely be reused anyway.

Look-up-table should probably be rendered with an orthographic camera.

`examples/math/Lut.js` also needs to be revisited.

Mugen87:
> The example is not disposing of geometries properly.

The linked PR solves at least this problem.

Mugen87:
> examples/math/Lut.js also needs to be revisited.

What do you have in mind for this file?

WestLangley:
>What do you have in mind for this file?

It's just.. different...
```javascript
this.legend.position = parameters.hasOwnProperty( 'position' ) ? parameters[ 'position' ] : { 'x': 21.5, 'y': 8, 'z': 5 };

this.legend.position = new THREE.Vector3( position.x, position.y, position.z );
```
In the main demo,
- the model probably should be loaded only once, updating the vertex colors as needed
- `axis` is not needed
- `normalizeNormals()` is not needed
- `parameters` args use a different convention than we have adopted

Regarding the legend, I suggested using an orthographic camera for that, but maybe for VR support, a floating sprite is the right way to go. I am not sure what the feeling is on that...

I know this is not high-priority, but this code, after all, is presented as an example to be copied.

PR #15360, opened 2018-12-03
add lut example

greggman:
I assumed there was a LUT (look up table) example already but apparently not?

So ported this one (and found a bug in the original 😅)


Mugen87:
For testing: https://raw.githack.com/greggman/three.js/lut-example/examples/webgl_postprocessing_lut.html

examples/webgl_postprocessing_lut.html
```diff
@@ -0,0 +1,289 @@
...
+		<script src="js/postprocessing/EffectComposer.js"></script>
+		<script src="js/postprocessing/RenderPass.js"></script>
+		<script src="js/postprocessing/ShaderPass.js"></script>
+
+		<script src="js/loaders/GLTFLoader.js"></script>
+		<script src="js/libs/dat.gui.min.js"></script>
+
+		<script>
+
+const renderer = new THREE.WebGLRenderer( { canvas: document.querySelector( 'canvas' ) } );
```
Mugen87: Can you please ensure a correct indentation according to the other examples? 

Besides, please use `var` instead of `let` and `const`.

examples/js/shaders/LUTShader.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * @author miibond
+ *
+ * Full-screen tone-mapping shader based on http://www.cis.rit.edu/people/faculty/ferwerda/publications/sig02_paper.pdf
+ */
```
Mugen87: I don't think this comment is correct. Same for the authorship.

examples/js/shaders/LUTShader.js
```diff
@@ -0,0 +1,70 @@
...
+
+THREE.LUTShader = {
+
+	uniforms: {
+
+		"tDiffuse": { value: null },
+		"lutMap":  { value: null },
+        "lutMapSize": { value: 1, },
```
Mugen87: This file has also indentation issues.

examples/webgl_postprocessing_lut.html
```diff
@@ -0,0 +1,289 @@
...
+	magFilter: THREE.LinearFilter,
+	format: THREE.RGBFormat,
+};
+composer1 = new THREE.EffectComposer( renderer, new THREE.WebGLRenderTarget( 1, 1, rtParameters ) );
+
+composer1.addPass( renderBackground );
+composer1.addPass( renderModel );
+composer1.addPass( effectLUT )
+
+function resizeRendererToDisplaySize( renderer ) {
```
Mugen87: I'm not sure the way `resizeRendererToDisplaySize()` works is okay. Why not just using a `resize` event listener like in the other examples?

Mugen87 (changes requested):

Mugen87:
What about [webgl_geometry_colors_lookuptable](https://threejs.org/examples/webgl_geometry_colors_lookuptable.html)? Do we really need two examples? I remember there were some issue with the existing one (see #14634) so maybe we can just have `webgl_postprocessing_lut`?

examples/webgl_postprocessing_lut.html
greggman: Sorry about the indentation. I meant to fix that before I uploaded 😅

I'd prefer to leave them as let and const if possible. This is not part of the library, it's an example. Furthermore the entire rest for JavaScript world has moved on. Using `var` is not only out of date it is nearly universally acknowledged as bad practice. I don't really want to be spreading bad practices.

examples/js/shaders/LUTShader.js
greggman: Doh 😅

examples/js/shaders/LUTShader.js
greggman: fixed

greggman:
[webgl_geometry_colors_lookuptable](https://threejs.org/examples/webgl_geometry_colors_lookuptable.html) is a 2D LUT. This is a 3D LUT. 3D LUTs are used in nearly all AAA games and most modern movies. 3DLUTs really stuck out in the movie 300 but you can see them applied to all most every movie since then. 

3DLUTs can replace many of the other samples. Not suggesting they be replaced just pointing out that 3DLUTs can replace all of these shaders

```
ColorCorrectionShader.js
BrightnessContrastShader.js
ColorifyShader.js
GammaCorrectionShader.js
LuminosityHighPassShader.js
LuminosityShader.js
SepiaShader.js
TechnicolorShader.js
ToneMapShader.js
```

In WebGL1 it would be slower than 1 of those shaders but probably faster than any 2 since you can combine the effects off all those shaders into a single LUT. In WebGL2 (not provided) it would probably be faster than any of those shaders.

Googling 3DLuts brings up lots of sites providing LUTs for movies and photo editing. I suppose another thing to do would be to provide code to read common 3dlut formats but that seemed like too much work. Seems better to provide a converter maybe

Mugen87:
Thanks for the clarification! Against this backdrop, it's of course okay to have two examples.

examples/webgl_postprocessing_lut.html
Mugen87: > This is not part of the library, it's an example. Furthermore the entire rest for JavaScript world has moved on. 

That's true but it's important for us to keep the example code consistent. Even if not all aspects of the code base are up to date.

examples/webgl_postprocessing_lut.html
Mugen87: Like I said in the other comment, it's a matter of consistency. I would prefer a solution where you fix the issue in all examples. Not just in one.

looeee:
Nice example! 

It's seems like there's some strong aliasing around the ears. It's especially apparent with the 'bgy' LUT. 

Perhaps including an FXAA pass would help with that?

examples/webgl_postprocessing_lut.html
looeee: Agreed with @Mugen87 on this one. We're overdue another conversation on whether to start using let and const, but that should be discussed elsewhere. 
In the meantime please keep the code style consistent with other examples.

Mugen87:
Why have you closed the PR?

looeee:
@greggman did you close this by mistake?
It's a nice example, it would be great to add this technique.

PR #15369, opened 2018-12-04
Examples:correct the resize code 

lafflan:
old code while cause blur on mobile when resize


Mugen87:
#14340 already addresses this issue and harmonizes the resolution for all post-processing examples.

sweerwen:
#14340

PR #15390, opened 2018-12-07
PMREMGenerator: #14469 Adding different weighting scheme for weighting of samples

### LINKED ISSUE #14469: Incorrectly weighted samples in PMREMGenerator

I noticed some glb files using PBR materials looked much more metallic in three.js under a particular environment map compared to other viewers (Substance Painter and Unity). After some investigation I believe this is due to PMREMGenerator not weighting lighting samples correctly -- each sample is weighted the same which causes HDR environment maps with large variations in intensity to become unsightly when convolved.

Here's a tonemapped image that shows the problem (notice what happens to the sun in the bottom right of the bottom left face): https://raw.githubusercontent.com/akhosravian/threejs_pmrem_bug/master/pmrem.JPG

Here's a screenshot of a line of sphere of increasing roughness using the map from that image:
https://raw.githubusercontent.com/akhosravian/threejs_pmrem_bug/master/issue.jpg

Live examples (the assets here are rather large so it takes a bit to load): 
https://jsfiddle.net/r07gpq2a/
https://jsfiddle.net/tkqcfraz/

Standalone project: https://github.com/akhosravian/threejs_pmrem_bug

Is there a way to work around this?
 
##### Three.js version

- [ ] Dev
- [x] r94
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

akhosravian:
This is behind a toggle due to some issues with seams, but it works in many cases. Maybe the seam issue is obvious to someone with more familiarity @bhouston 

Here's some comparison screenshots:

Curret Method:
Random Sampling

With this change (fewer artifacts, the first roughness step is much blurrier, yellow tint disappears):
Importance Sampling

Fixes #14469


bhouston:
Where was the yellow tint coming from?

akhosravian:
The sun in the original HDR image is extremely bright (it is a day time scene after all!) and yellow tinted. 

My belief is since all samples in the current implementation are weighted equally this burning day star would skew things just a little bit yellow. I can generate the map with an external tool for comparison to verify if it is a concern.

examples/js/pmrem/PMREMGenerator.js
```diff
@@ -21,11 +19,13 @@ THREE.PMREMGenerator = ( function () {
...
 	scene.add( camera );
 
-	var PMREMGenerator = function ( sourceTexture, samplesPerLevel, resolution ) {
+	var PMREMGenerator = function ( sourceTexture, params = {} ) {
 
 		this.sourceTexture = sourceTexture;
-		this.resolution = ( resolution !== undefined ) ? resolution : 256; // NODE: 256 is currently hard coded in the glsl code for performance reasons
-		this.samplesPerLevel = ( samplesPerLevel !== undefined ) ? samplesPerLevel : 32;
+		this.sourceResolution = ( params.sourceResolution !== undefined ) ? params.sourceResolution : 512;
+		this.targetResolution = ( params.targetResolution !== undefined ) ? params.targetResolution : 256; // NODE: 256 is currently hard coded in the glsl code for performance reasons
```
jsantell: I think this comment can be removed, and `targetResolution` doesn't need to be a property

examples/js/pmrem/PMREMGenerator.js
```diff
@@ -21,11 +19,13 @@ THREE.PMREMGenerator = ( function () {
 	scene.add( planeMesh );
 	scene.add( camera );
 
-	var PMREMGenerator = function ( sourceTexture, samplesPerLevel, resolution ) {
+	var PMREMGenerator = function ( sourceTexture, params = {} ) {
 
 		this.sourceTexture = sourceTexture;
-		this.resolution = ( resolution !== undefined ) ? resolution : 256; // NODE: 256 is currently hard coded in the glsl code for performance reasons
-		this.samplesPerLevel = ( samplesPerLevel !== undefined ) ? samplesPerLevel : 32;
+		this.sourceResolution = ( params.sourceResolution !== undefined ) ? params.sourceResolution : 512;
```
jsantell: Might be helpful to check for `resolution` for folks upgrading to this version if `sourceResolution` not found, e.g. `( params.sourceResolution !== undefined ) ? params.sourceResolution : params.resolution !== undefined ? params.resolution : 512;`

examples/js/pmrem/PMREMGenerator.js
```diff
@@ -21,11 +19,13 @@ THREE.PMREMGenerator = ( function () {
...
+		this.targetResolution = ( params.targetResolution !== undefined ) ? params.targetResolution : 256; // NODE: 256 is currently hard coded in the glsl code for performance reasons
+		this.samplesPerLevel = ( params.samplesPerLevel !== undefined ) ? params.samplesPerLevel : 32;
+		this.useImportanceSampling = ( params.useImportanceSampling !== undefined ) ? params.useImportanceSampling : false;
```
jsantell: `new PMREMGenerator({ useImportanceSampling: false })` would not use importance sampling in this case which would be unexpected

examples/js/pmrem/PMREMGenerator.js
```diff
@@ -59,25 +59,18 @@ THREE.PMREMGenerator = ( function () {
 
 		}
 
+		var extraUniforms = this.useImportanceSampling? getImportanceSamplingUniforms() : getRandomSamplingUniforms();
+		shader.uniforms = Object.assign(getBaseUniforms(), extraUniforms);
+
+		var shaderMain = this.useImportanceSampling? getImportanceSamplingFragmentShader() : getRandomSamplingShader();
+		shader.fragmentShader = getBaseFragmentShader() + shaderMain;
```
jsantell: `shader` is used by all instances of PMREMGenerator -- rational was due to a memory leak (https://github.com/mrdoob/three.js/pull/15330) that was mostly from WebGLState, but some build up from creating ShaderMaterial. Shader materials with the same shaders and uniforms should roll out to the same program under the hood AFAIK, so this shouldn't be the case, and I didn't dig into it anymore.

All that being said, maybe we don't need to use the same material instance across all PMREMGenerators, or at the very least store up to 2 (random and importance) materials, since right now it'd be impossible to test both the importance PMREM alongside the random sampling version

jsantell (commented):
Great stuff! I'm thrilled to see the snowy/sparky artifacts in the first roughness level smoothed out. Looking forward to trying this out

examples/js/pmrem/PMREMGenerator.js
bhouston: Why not?   I think some people may want a low resolution if they want to have multiple maps in a large environment so that one can blend between them.

examples/js/pmrem/PMREMGenerator.js
bhouston: good idea.

examples/js/pmrem/PMREMGenerator.js
bhouston: Regard this style, I believe Three.JS does not use objects passed in as parameters, rather it uses ordered arguments.  The reason for this is that in generally objects used as parameter sets on a regular basis , because they are memory allocations on the heap, introduces GC churn.  Where as order arguments I believe are just stack based instead and thus do not impact GC.

examples/js/pmrem/PMREMGenerator.js
bhouston: People will generally not switch between random and importance at run time except for testing purposes (which we do not have to optimize for.)  In fact, I think that one the bugs are worked out with importance, random could potentially be removed.

bhouston:
This is an amazing contribution.  Sorry for the earlier brief message, was traveling.

bhouston:
> The sun in the original HDR image is extremely bright (it is a day time scene after all!) and yellow tinted.

The sampling pattern is not wide enough to cause even the first level to have a yellow shift in the original.  I am sure of that.  There must a logical reason for the source of the yellow shift and why it went away.  I would think the issue is in regards to encode/decode somewhere along the lines.

examples/js/pmrem/PMREMGenerator.js
jsantell: @bhouston agree that switching is an unlikely use case and ultimately only need one implementation! Just the current implementation would result in args not matching what shader would be used

examples/js/pmrem/PMREMGenerator.js
jsantell: Makes sense! Maybe a breaking API change would be useful when/if moving to importance sampling method for consistency with core (1 object allocation won't add to overhead in the PMREM case) -- that being said, `useImportanceSampling: false` should use random sampling (use case, me flipping a boolean testing cubemaps back to back with importance vs random sampling)

examples/js/pmrem/PMREMGenerator.js
jsantell: To clarify, it doesn't need to be stored on the PMREMGenerator instance and could just be a local variable -- agreed it is valuable to specify the output resolution!

WestLangley:
Thanks everyone for working on this. :-)

Here are some notes I made to myself some time ago:

1. `PMREMGenerator` constructor changes the properties of the input `sourceTexture`. IMO, we should never allow that.

2. `PMREMGenerator` `matrixFromVector( vec3 n )` appears to be a nonorthogonal transform -- _presumably_ that is not correct... I noticed this because the blurring appeared to shear the image in my example.

3. `PMREMGenerator` - `PlaneBufferGeometry( 2, 2, 0 )`

4. `PMREMCubeUVPacker` - orthographic camera top and bottom are swapped

5. `PMREMCubeUVPacker` - the computed mipmaps at each blur level do not appear to be used - at least as far as I was able to tell. In other words, the test case needs to be improved so it is clear that the algorithm is working correctly.

Assuming I am correct in my observations, (2) and (5) are the most serious.

TheophileMot (commented):
Nice work! This will be a great feature. I had some trouble getting it to work, though. The code currently has importance sampling turned off by default, but when I turn it on, the result looks strange. Here's a comparison of importance off vs. on:

screenshot from 2018-12-18 15-13-46
screenshot from 2018-12-18 15-18-58

Also, when metalness and roughness are both set to 1, the object texture looks dimpled like a golf ball. Any ideas what's causing this?

akhosravian:
Apologies for the dead air its been a hectic bit. Looking at fixing up this change now.

jingwenhz:
Hello, not sure if you already have a fix in mind, but I noticed the following when trying out the code.
It appears that there are not enough samples at the higher mip levels so rough materials look very spotty. This is consistent with what @spidersharma03 noted in his previous implementation. Sampling from the previous mip level instead of the source texture appears to improve results a lot, although this may not be strictly 'correct'.

Changing
`if ( !this.useImportanceSampling && i < 5 ) shader.uniforms[ 'envMap' ].value = this.cubeLods[ i ].texture;`
to
`if ( i < 5 ) shader.uniforms[ 'envMap' ].value = this.cubeLods[ i ].texture;`

Random sampling (existing implementation)
random_sampling

Importance sampling
importance_sampling_nonlod

Importance sampling using previous mip level
importance_sampling_lod

The materials above have roughness 0.6 and look more accurate with the importance sampling, but I haven't yet tested with e.g. the MetalRoughSphere scene.

akhosravian:
Did some digging on this there seems to be issues with seams between cube map faces. I'm at a loss for what the cause of this is though. Here's a gif I grabbed of the Khronos MetalRoughSpheres scene using 2048 samples per level (the artifacts become more pronounced with higher numbers of samples).

There's a mention in the comment at the top of the file about creating an extra pixel around each face to avoid seams. Can anyone elaborate on how that works?

jingwenhz:
Looking at this a bit more, it seems that the suggested change is performing mipmap-filtered importance sampling described here, is this true? https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch20.html

But since the format of HDR files is RGBE which is not monotonic, PMREMGenerator is setting both min and mag filtering to THREE.NearestFilter. In that case, reading from the higher mips would not reduce noise. Could this be a source of error that we're seeing?

bhouston:
Automatic generation of mipmaps is only possibly if the encoding is monotonic otherwise it will produce garbage.

jingwenhz:
I see, that makes sense. But if the sampling procedure is then reading from the higher mipLevels, what kind of behavior does that result in?

bhouston:
Thus with our encoded RGBA values we support monotonic and non-monotonic encodings via the use of the explicit mipmapping approach used in CubeUV map.

jingwenhz:
Thanks for the great explanation! The example you linked is really helpful for visualizing what is going on. My only question is that the importance sampling code seems to be reading from higher mip levels of the sourceTexture when constructing the cubeLods, before they get passed to the CubeUVPacker. 

Line 385:
`rgbColor.rgb += envMapTexelToLinear(textureCube(envMap, L, mipLevel)).rgb * NdotL;`

Unless there is something else going on that I don't realize? (Sorry for the many questions! I really hope to see this feature added :) )

mrdoob:
#18004

PR #15394, opened 2018-12-07
Adding a function to OrbitControls.js to control camera position

johnson2heng:
Since OrbitControls only have functions getPolarAngle and getAzimuthalAngle that can get radians, but they don't set the location of OrbitControls, I add one.

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -98,6 +98,20 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	};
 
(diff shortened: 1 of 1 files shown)


Mugen87:
What about #15376?

johnson2heng:
I forgot the test case:  https://johnson2heng.github.io/GitChat-Three.js/threejs_pull_demo/orbitControl/

Mugen87:
Thanks for the effort but in the meanwhile we try to limit the changes to `OrbitControls`. Please read https://github.com/mrdoob/three.js/pull/16042#issuecomment-475757052 for more information.

PR #15397, opened 2018-12-08
Adding LightningStrike, LightningStorm with examples

yomboprime:
Support to create lightning strikes and other voltaic arcs, animated in time.

These are 3 examples in one html, and two new classes:
`THREE.LightningStrike` is a BufferGeometry which varies over time (a call to `.update( time )` is needed) Rays can have a limited lifespan or can be 'ethernal'.
`THREE.LightningStorm` is a higher class object, easy to use and drop-in as a storm in a game. It creates rays on its own over time. (Also needs `.update()`)

The two classes are created with parameter objects (details in the file headers). You can define some 'callbacks' to control the form of the rays, do something when a ray hits the ground, and other things.

Edit: Also the rays have a read-only state, so you know each phase of the life cycle including propagation and vanishing.

Time can go forwards and backwards, be slowmotion, etc.

I had the need to do this for some time. 😊

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -58,6 +58,7 @@ var files = {
 		"webgl_kinect",
 		"webgl_layers",
 		"webgl_lensflares",
+		"webgl_lightningstrike",
 		"webgl_lights_hemisphere",
 		"webgl_lights_physical",
 		"webgl_lights_pointlights",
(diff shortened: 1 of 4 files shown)


jsantell:
Heh, pretty cool :smile: 

screenshot from 2018-12-10 10-50-25

mrdoob:
Thanks!

PR #15430, opened 2018-12-14
Editor: Fixed shadow casting bug.

linbingquan:


Mugen87:
Can you please shortly explain what this bug fix is about?

linbingquan:
@Mugen87 I wanna to make shadow effect, I toggle `Checkbox of  ReceiveShadow `, but `Editor` doesn't render immediately.

image

linbingquan:
@Mugen87 I wanna to make shadow effect, when I toggle `Checkbox of  ReceiveShadow`, but `Editor` doesn't render immediately.

image

Mugen87:
Good catch!

Mugen87 (approved):

mrdoob:
Hmm, I'm not able to reproduce the issue in https://threejs.org/editor/.
What are the steps to reproduce?

linbingquan:
@mrdoob I make a `DirectionalLight` and a `Box` with `Shadow cast`,  and `Plane` without `Shadow receive`. When I toggle `Shadow receive` of `Plane`, the shadow receiveing doesn't work immediately.

Result screen shot:


and then I switch `CheckBox` tow times, the shadow receiveing works correctly.

Result screen shot:

mrdoob:
Ah, I see. I can reproduce that and I understand why it's happening.
I think the fix is good.

mrdoob:
Thanks!

Issue #15321, opened 2018-11-27
GLTFLoader: Disable cache


rpadovani:
We are working on an huge project with thousands of objects, and every object is unique. We load data from a gltf file, and the loader checks if any object is equal to something already created.

Disabling the check on the cache [here](https://github.com/mrdoob/three.js/blob/3665fb0ea0237ebab9798c5847861086782c778d/examples/js/loaders/GLTFLoader.js#L1566) (`return null` instead of calling isPrimitiveEqual over thousands of objects) cuts the loading time by a factor of 4, in our case.

I know this is a corner case, but would you be interested in an option to disable the cache? My company will sponsor my work on this, because we have to implement it anyway, and makes more sense to do it upstream than keeping our own version of the loader

mrdoob:
How about creating new loader instance per object?

donmccurdy:
First of all, thanks for opening the issue and sharing your performance numbers — that’s invaluable. :)

The geometry cache is kept at the parser level, and isn’t shared across .load() requests, so one-per-model loaders won’t help here unless I’ve missed something. Splitting your single glTF file into many files would help, but understandably you may not want to do that for other reasons.

I think there are two choices:

1. An option to disable the cache as you suggest.
2. Reimplement the cache to create a unique string key for each primitive, so cache lookups are O(1) rather than O(n).

My preference would be (2). Are you able to test that and see if the results are acceptable? If it’s still prohibitively slow I’m also open to (1), but I suspect the overhead of a faster cache would be negligible.

rpadovani:
Hi, thanks for your answers :-)

Splitting the file isn't really an option - our customer target's browser doesn't support HTTP/2, so loading hundreds (or thousands) of different files have some overhead. TBH I haven't tested it, but it something I'd like to avoid: we should also keep a track of all the files and sending them to the client in some way.

I like the idea to make the cache lookups O(1), but I don't know how to implement it in a generic way: I can always add something like an `id` in our GLTF, but then I have to use our specific `id`.

At the moment a new object in the cache is added as:

```javascript
cache.push( { primitive: primitive, promise: geometryPromise } );
```

There is any way to create an unique signature from the `primitive`?

The only generic implementation I can think of is using something like [object-hash](https://www.npmjs.com/package/object-hash) to create an unique hash of objects and use that as key. 
Any other suggestion?

donmccurdy:
> Splitting the file isn't really an option...

Understood 👍

[...]

PR #15431, opened 2018-12-14
GLTFLoader: use object as cache instead of array

rpadovani:
Implements #15321: use object as cache for GTLF loader instead of array.

For our specific case the loading time goes from ~220 seconds to ~7 seconds (we have ~60000 geometries with ~30000 mesh). 

Completely disabling the cache (as we have done so far) has more or less the same impact.

Signed-off-by: Riccardo Padovani <rpadovani@nextbit.it>

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1555,9 +1555,9 @@ THREE.GLTFLoader = ( function () {
 
 		var keys = Object.keys( attributes ).sort();
 
-		for ( var key of keys ) {
+		for ( var i = 0, il = keys.length; i < il; i ++ ) {
 
-			attributesKey += key + attributes[ key ];
+			attributesKey += keys[ i ] + attributes[ keys[ i ] ];
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1541,7 +1541,7 @@ THREE.GLTFLoader = ( function () {
 
 		} else {
 
-			geometryKey = primitiveDef.indices + ':' + createAttributesKey( primitiveDef.attributes ) + primitiveDef.mode;
+			geometryKey = primitiveDef.indices + ':' + createAttributesKey( primitiveDef.attributes ) + ':' + primitiveDef.mode;
 
 		}
 
@@ -1557,7 +1557,7 @@ THREE.GLTFLoader = ( function () {
 
 		for ( var i = 0, il = keys.length; i < il; i ++ ) {
 
-			attributesKey += keys[ i ] + attributes[ keys[ i ] ];
+			attributesKey += keys[ i ] + ':' + attributes[ keys[ i ] ] + ';';
 
 		}
 
@@ -1565,13 +1565,27 @@ THREE.GLTFLoader = ( function () {
 
 	}
 
-	function createArrayKey( a ) {
+	function createArrayKeyBufferGeometry( a ) {
 
 		var arrayKey = '';
 
 		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			arrayKey += i + createAttributesKey( a[ i ] );
+			arrayKey += i + a[ i ].uuid;
+
+		}
+
+		return arrayKey;
+
+	}
+
+	function createArrayKeyGLTFPrimitive( a ) {
+
+		var arrayKey = '';
+
+		for ( var i = 0, il = a.length; i < il; i ++ ) {
+
+			arrayKey += i + createGeometryKey( a[ i ] );
 
 		}
 
@@ -1581,7 +1595,7 @@ THREE.GLTFLoader = ( function () {
 
 	function createMultiPassGeometryKey( geometry, primitives ) {
 
-		return createGeometryKey( geometry ) + createArrayKey( primitives );
+		return createGeometryKey( geometry ) + createArrayKeyGLTFPrimitive( primitives );
 
 	}
 
@@ -2610,7 +2624,7 @@ THREE.GLTFLoader = ( function () {
 
 				// See if we've already created this combined geometry
 				var cache = parser.multiplePrimitivesCache;
-				var cacheKey = createArrayKey( geometries );
+				var cacheKey = createArrayKeyBufferGeometry( geometries );
 				var cached = cache[ cacheKey ];
 
 				if ( cached ) {
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1553,7 +1553,9 @@ THREE.GLTFLoader = ( function () {
 
 		var attributesKey = '';
 
-		for ( var key in attributes ) {
+		var keys = Object.keys( attributes ).sort();
+
+		for ( var key of keys ) {
 
 			attributesKey += key + attributes[ key ];
 
@@ -1569,13 +1571,7 @@ THREE.GLTFLoader = ( function () {
 
 		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			arrayKey += i;
-
-			for ( var j = 0, jl = a[ i ].lenght; j < jl; j ++ ) {
-
-				arrayKey += j + a[ i ][ j ];
-
-			}
+			arrayKey += i + createAttributesKey( a[ i ] );
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1595,11 +1595,11 @@ THREE.GLTFLoader = ( function () {
 
 	function createMultiPassGeometryKey( geometry, primitives ) {
 
-		var key = createPrimitiveKey( geometry );
+		var key = geometry.uuid;
 
 		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
 
-			key += i + primitives[ i ].uuid;
+			key += i + createPrimitiveKey( primitives[ i ].uuid );
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1528,7 +1528,7 @@ THREE.GLTFLoader = ( function () {
 
 	}
 
-	function createGeometryKey( primitiveDef ) {
+	function createPrimitiveKey( primitiveDef ) {
 
 		var dracoExtension = primitiveDef.extensions && primitiveDef.extensions[ EXTENSIONS.KHR_DRACO_MESH_COMPRESSION ];
 		var geometryKey;
@@ -1585,7 +1585,7 @@ THREE.GLTFLoader = ( function () {
 
 		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			arrayKey += i + createGeometryKey( a[ i ] );
+			arrayKey += i + createPrimitiveKey( a[ i ] );
 
 		}
 
@@ -1595,7 +1595,15 @@ THREE.GLTFLoader = ( function () {
 
 	function createMultiPassGeometryKey( geometry, primitives ) {
 
-		return createGeometryKey( geometry ) + createArrayKeyGLTFPrimitive( primitives );
+		var key = createPrimitiveKey( geometry );
+
+		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
+
+			key += i + primitives[ i ].uuid;
+
+		}
+
+		return key;
 
 	}
 
@@ -2520,7 +2528,7 @@ THREE.GLTFLoader = ( function () {
 		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
 
 			var primitive = primitives[ i ];
-			var cacheKey = createGeometryKey( primitive );
+			var cacheKey = createPrimitiveKey( primitive );
 
 			// See if we've already created this geometry
 			var cached = cache[ cacheKey ];
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1571,21 +1571,7 @@ THREE.GLTFLoader = ( function () {
 
 		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			arrayKey += i + a[ i ].uuid;
-
-		}
-
-		return arrayKey;
-
-	}
-
-	function createArrayKeyGLTFPrimitive( a ) {
-
-		var arrayKey = '';
-
-		for ( var i = 0, il = a.length; i < il; i ++ ) {
-
-			arrayKey += i + createPrimitiveKey( a[ i ] );
+			arrayKey += ':' + a[ i ].uuid;
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1541,7 +1541,7 @@ THREE.GLTFLoader = ( function () {
 
 		} else {
 
-			geometryKey = primitiveDef.indices + ':' + createAttributesKey( primitiveDef.attributes );
+			geometryKey = primitiveDef.indices + ':' + createAttributesKey( primitiveDef.attributes ) + primitiveDef.mode;
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1585,7 +1585,7 @@ THREE.GLTFLoader = ( function () {
 
 		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
 
-			key += i + createPrimitiveKey( primitives[ i ].uuid );
+			key += i + createPrimitiveKey( primitives[ i ] );
 
 		}
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -338,7 +338,7 @@ THREE.GLTFLoader = ( function () {
 
 			case 'directional':
 				lightNode = new THREE.DirectionalLight( color );
-				lightNode.target.position.set( 0, 0, -1 );
+				lightNode.target.position.set( 0, 0, - 1 );
 				lightNode.add( lightNode.target );
 				break;
 
@@ -356,7 +356,7 @@ THREE.GLTFLoader = ( function () {
 				lightDef.spot.outerConeAngle = lightDef.spot.outerConeAngle !== undefined ? lightDef.spot.outerConeAngle : Math.PI / 4.0;
 				lightNode.angle = lightDef.spot.outerConeAngle;
 				lightNode.penumbra = 1.0 - lightDef.spot.innerConeAngle / lightDef.spot.outerConeAngle;
-				lightNode.target.position.set( 0, 0, -1 );
+				lightNode.target.position.set( 0, 0, - 1 );
 				lightNode.add( lightNode.target );
 				break;
 
@@ -964,6 +964,7 @@ THREE.GLTFLoader = ( function () {
 					uniforms.refractionRatio.value = material.refractionRatio;
 
 					uniforms.maxMipLevel.value = renderer.properties.get( material.envMap ).__maxMipLevel;
+
 				}
 
 				uniforms.specular.value.copy( material.specular );
@@ -1356,8 +1357,10 @@ THREE.GLTFLoader = ( function () {
 				var accessor = target.POSITION !== undefined
 					? parser.getDependency( 'accessor', target.POSITION )
 						.then( function ( accessor ) {
+
 							// Cloning not to pollute original accessor below
 							return cloneBufferAttribute( accessor );
+
 						} )
 					: geometry.attributes.position;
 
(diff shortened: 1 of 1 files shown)


examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1523,57 @@ THREE.GLTFLoader = ( function () {
 
 	}
 
-	function isArrayEqual( a, b ) {
+	function createGeometryKey( geometry ) {
```
donmccurdy: minor: could you name this parameter `primitiveDef` instead? I find it easier to distinguish whether we're working with the glTF primitive object or `THREE.Geometry` that way.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1523,57 @@ THREE.GLTFLoader = ( function () {
...
+		if ( dracoExtension ) {
 
-			if ( a[ i ] !== b[ i ] ) return false;
+			geometryKey = 'draco:' + dracoExtension.bufferView
+				+ ':' + dracoExtension.indices
+				+ ':' + createAttributesKey( dracoExtension.attributes );
+
+		} else {
+
+			geometryKey = geometry.indices + ':' + createAttributesKey( geometry.attributes );
```
donmccurdy: I realize it wasn't covered in the original method, but for good measure could you include `primitiveDef.mode` here? It applies only to the non-Draco case.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1523,57 @@ THREE.GLTFLoader = ( function () {
...
-	function getCachedCombinedGeometry( cache, geometries ) {
-
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var arrayKey = '';
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			if ( isArrayEqual( geometries, cached.baseGeometries ) ) return cached.geometry;
+			arrayKey += i + a[ i ];
```
donmccurdy: I don't think this `createArrayKey` function will work as intended — the way this method is used now, the array will contain either `THREE.Geometry` or glTF primitive definitions. In either case, the string for an array element will be `[object Object]`.

examples/js/loaders/GLTFLoader.js
rpadovani: As far as I understand, that is a Enum right? So I can simply concatenate it?

examples/js/loaders/GLTFLoader.js
rpadovani: I will save the values of the keys, but at the momentthe `isArrayEqual` just does a direct comparison between elements of the arrays, so I thought there were inside strings or something like that

rpadovani:
@donmccurdy thank you very much for your review :-)

I implemented the changes you requested, and I've also run eslint which edited others parts of the file, I hope that is okay - otherwise let me know and I will revert them!

takahirox:
Thanks for the PR and performance test. Good to know that it's much scalable as we expected.

Just in case, can you also check the performance with standard glTF files if this change affects the performance? Standard glTF files mean the glTF file including not so many geometries. Probably `webgl_loader_gltf.html` and `webgl_loader_gltf_extensions.html` examples are good to check.

I'm concerned about string concatanation performance, it can be relatively slow. We might see performance degradation on standard glTF files. It shouldn't be big deal but just in case I wanna check.

examples/js/loaders/GLTFLoader.js
donmccurdy: Yep!

examples/js/loaders/GLTFLoader.js
donmccurdy: Currently `a` passed into this method could be an `Array<THREE.BufferGeometry>` or `Array<GLTF.Primitive>` here in which case we're doing something equivalent to:

```js
arrayKey += 25 + new THREE.BufferGeometry();
```

The direct comparison was possible because if the mesh had identical primitives in the same order, then earlier caching would ensure it has the same exact geometry objects.

donmccurdy:
> I've also run eslint which edited others parts of the file, I hope that is okay

The other changes look OK to me, thanks!

> I'm concerned about string concatanation performance, it can be relatively slow. We might see performance degradation on standard glTF files. It shouldn't be big deal but just in case I wanna check.

This PR creates a string key for each of `n` primitives, making cache lookups `O(1)` instead of `O(n)`, for a net cache time complexity of `O(n)` instead of `O(n^2)`. For any n>2 this seems like a strictly good change, and for 1-2 primitives the difference should be small. Assuming we did measure some very small degradation on the 1-2 primitive case, how else would we design the cache? Feel free to profile this, but I would not spend too much effort on it.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,64 @@ THREE.GLTFLoader = ( function () {
...
 	}
 
-	function getCachedGeometry( cache, newPrimitive ) {
+	function createAttributesKey( attributes ) {
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var attributesKey = '';
 
-			var cached = cache[ i ];
+		for ( var key in attributes ) {
```
takahirox: I think here is assumed that keys of objects gotten by "for ( var key in object )" is consistent, correct?

My question is, is that guaranteed? It seems practically yes on the latest Chrome and FireFox, but I'm not really sure if it's guaranteed in the language spec. If no, probably we need to explicitly sort keys. Anyone knows?

donmccurdy: Good point, I think keys should probably be sorted yes.

takahirox: Just in case, let me confirm my understanding. glTF spec doesn't mention the order of attributes then the order can be different across primitives, correct?

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes

donmccurdy: The spec doesn’t (and probably couldn’t) require an order in the JSON, no.

takahirox: OK, thanks

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,64 @@ THREE.GLTFLoader = ( function () {
...
 
 		}
 
-		return null;
+		return attributesKey;
 
 	}
 
-	function getCachedCombinedGeometry( cache, geometries ) {
+	function createArrayKey( a ) {
```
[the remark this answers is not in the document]
rpadovani: So I assume I can call `createAttributesKey` over the elements of the array?

examples/js/loaders/GLTFLoader.js
rpadovani: I updated the code to extract the keys and sort them before iterating over them

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,64 @@ THREE.GLTFLoader = ( function () {
...
+		var arrayKey = '';
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			if ( isArrayEqual( geometries, cached.baseGeometries ) ) return cached.geometry;
+			arrayKey += i;
 
-		}
+			for ( var j = 0, jl = a[ i ].lenght; j < jl; j ++ ) {
```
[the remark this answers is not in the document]
rpadovani: Fixed, thanks!

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,60 @@ THREE.GLTFLoader = ( function () {
...
-	function getCachedGeometry( cache, newPrimitive ) {
+	function createAttributesKey( attributes ) {
+
+		var attributesKey = '';
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var keys = Object.keys( attributes ).sort();
 
-			var cached = cache[ i ];
+		for ( var key of keys ) {
```
takahirox: We don't use `for ... of` yet. Please use `for ( var i = 0, il = keys.length; i < il; i  ++ )` instead.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,60 @@ THREE.GLTFLoader = ( function () {
...
 
 	}
 
-	function getCachedGeometry( cache, newPrimitive ) {
+	function createAttributesKey( attributes ) {
+
+		var attributesKey = '';
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var keys = Object.keys( attributes ).sort();
```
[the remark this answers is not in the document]
rpadovani: It depends on browsers, anyhow I don't think a custom implementation would be better:

https://stackoverflow.com/questions/234683/javascript-array-sort-implementation
https://v8.dev/blog/array-sort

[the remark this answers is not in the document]
takahirox: Thanks. Yeah perhaps native code is faster.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,60 @@ THREE.GLTFLoader = ( function () {
...
+		if ( dracoExtension ) {
 
-			if ( a[ i ] !== b[ i ] ) return false;
+			geometryKey = 'draco:' + dracoExtension.bufferView
+				+ ':' + dracoExtension.indices
+				+ ':' + createAttributesKey( dracoExtension.attributes );
+
+		} else {
+
+			geometryKey = primitiveDef.indices + ':' + createAttributesKey( primitiveDef.attributes ) + primitiveDef.mode;
```
donmccurdy: super minor but i think there should be a `:` between the attributes and the mode, since both are numeric types.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,60 @@ THREE.GLTFLoader = ( function () {
...
+		var attributesKey = '';
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var keys = Object.keys( attributes ).sort();
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = keys.length; i < il; i ++ ) {
 
-			if ( isPrimitiveEqual( cached.primitive, newPrimitive ) ) return cached.promise;
+			attributesKey += keys[ i ] + attributes[ keys[ i ] ];
```
donmccurdy: Let's put an infix here, so the attributes key will be something like: `POSITION:0,NORMAL:1`. Some attributes have indices, like `TEXCOORD_1...TEXCOORD_N`. I can't think of any situation where the code above would actually hash incorrectly, but just to be safe. :)

examples/js/loaders/GLTFLoader.js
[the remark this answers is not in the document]
donmccurdy: As written here, the parameter to `createArrayKey` will be one of:

- Array<THREE.BufferGeometry>
- Array<GLTF.Primitive>

I would just split this into two methods, or do this in a `for` loop  inline (since each method would only be called once). For each geometry you can hash on `geometry.uuid`. For each primitive you'll need to use the `createGeometryKey` method above. The mesh/primitive and mesh/geometry concepts in glTF and three.js are a little different, which makes naming these things hard... 😅

takahirox:
Performance analysis and evaluation.

**Time complexity:**

Time complexity for current and new cache system can be represented as BigO with three parameters, n, m, g. n is the number of primitives in an asset, m is the number of attributes in a primitive, and g is the number of distinct geometries. g should be equal to or smaller than n.

The current array based cache lookup is O(g\*m), while the new object based cache lookup is O(m\*logm). In time complexity analysis we can regard m and logm as 1 because it's small enough, up to eight.

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes

Then they'll be O(g) and O(1) respectively. Therefore total time complexity for an asset will be O(n\*g) and O(n). The new one should be more scalable.

**Benchmark test:**

I made benchmark test and confirmed the performance.

https://raw.githack.com/takahirox/three.js/CacheTest/test/benchmark/cache.html?n=256&m=8&g=256

The numbers represent the total elapsed times [ms] of n primitives cache lookup where g = n. The smaller number is faster. The table shows that the current array based cache follows O(g\*n) (=O(n^2) in this case) while the new object based cache does O(n) as we expected. And also it shows that users would see the benefit from the new cache system in practical where n is greater than 100-200.

|  | Chrome |   |   |   | FireFox |   |   |  
|-- | -- | -- | -- | -- | -- | -- | -- | -- |
|  | Array | Object | Array | Object | Array | Object | Array | Object |
|  | m=4 | 4 | 8 | 8 | 4 | 4 | 8 | 8 |
|n=32 | 0.2 [ms]| 0.15 | 0.16 | 0.24 |   |   |   |  |
|64 | 0.32 | 0.26 | 0.35 | 0.35 |   |   |   |  |
|128 | 0.5 | 0.26 | 0.8 | 0.35 | 2.36 |   | 2.76 |  |
|256 | 1.86 | 0.43 | 2.34 | 0.68 | 8.94 |   | 10.76 |  |
|512 | 6.64 | 0.73 | 8.02 | 1.2 | 35.06 |   | 39.64 | 2.39 |
|1024 | 23.56 | 1.38 | 30.72 | 2.25 | 142.91 | 2.61 | 159.95 | 4.72 |
|2048 | 92.6 | 2.76 | 125 | 4.44 | 552.65 | 5.31 | 654.42 | 8.71 |
|4096 | 380 | 5 | 498 | 9 | 2264.08 | 12.08 | 2624.15 | 20.32 |
|8192 | 1526 | 12.7 | 1965 | 20 | 9090.62 | 23.29 | 10621.78 | 44.61 |

**Degradation:**

The only performance degradation I'm aware of is the case where g is small because sorting (and string concatenation) may be relatively much slower than m times integer comparisons. For example, the new cache system is 10x slower where n=1024, m=8, and g=1 on my Windows Chrome. 0.3ms for the current one and 3.0ms for the new one.

https://raw.githack.com/takahirox/three.js/CacheTest/test/benchmark/cache.html?n=1024&m=8&g=1

Perhaps it might be ok. Because it may be rare case and sorting up to eight elements per primitive isn't so slow(?).

donmccurdy:
Thanks for benchmarking @takahirox! 

Do you know why `n=1024,g=1` should give different results from `n=1,g=1024`? The primitive-level caching should be the same either way — presumably that's a result of the multi-material cache?<sup>1</sup>

Because the overhead of sorting is only visible for small `g`, where overall impact is likely to be low, I think I'm still in favor of this change.

***

<small><sup>1</sup> It's a question for another time, but I'm having second thoughts about keeping the multi-material logic in the loader at all... it adds complexity, and arguably could be done in a SceneUtils-ish optimization function instead.</small>

rpadovani:
Hi @takahirox, thanks for your amazing benchmark :)

I have updated the code with the last changes you asked for, let me know if there is something else that needs to be addressed!

Have a nice day!

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,74 @@ THREE.GLTFLoader = ( function () {
...
 		}
 
-		return null;
+		return arrayKey;
+
+	}
+
+	function createMultiPassGeometryKey( geometry, primitives ) {
+
+		return createGeometryKey( geometry ) + createArrayKeyGLTFPrimitive( primitives );
```
donmccurdy: Because the `createArrayKeyGLTFPrimitive` function is only used in one place, I think we can just roll its for loop into this function. Also, `createGeometryKey` expects a GLTF primitive as an argument, not a BufferGeometry. Just `geometry.uuid` should suffice here.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,74 @@ THREE.GLTFLoader = ( function () {
 
 	}
 
-	function isArrayEqual( a, b ) {
+	function createGeometryKey( primitiveDef ) {
```
donmccurdy: Note that this method is being called below with a `BufferGeometry` as an argument, which it doesn't support. For clarity the method could be renamed as `createPrimitiveKey`.

donmccurdy (commented):
@takahirox do you have a model available to test that multi-geometry caching is working as before, or should we create one?

If we can test both single- and multi-geometry caching, then after the comments are addressed I think we can merge this and leave any additional changes for another time.

mrdoob:
@donmccurdy 

> 1 It's a question for another time, but I'm having second thoughts about keeping the multi-material logic in the loader at all... it adds complexity, and arguably could be done in a SceneUtils-ish optimization function instead.

👍

takahirox:
> @takahirox do you have a model available to test that multi-geometry caching is working as before, or should we create one?

I don't have. I had before but seems like I lost. I think we should create.

rpadovani:
@donmccurdy sorry for the delay, I implemented your requests!

Let me know if there is something else I can improve

rpadovani:
@mrdoob @takahirox @donmccurdy any update on this? :)

examples/js/loaders/GLTFLoader.js
```diff
@@ -1509,97 +1514,96 @@ THREE.GLTFLoader = ( function () {
...
+	function createArrayKeyBufferGeometry( a ) {
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var arrayKey = '';
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = a.length; i < il; i ++ ) {
 
-			if ( isPrimitiveEqual( cached.primitive, newPrimitive ) ) return cached.promise;
+			arrayKey += i + a[ i ].uuid;
```
donmccurdy: minor: why use `i` in the key?

examples/js/loaders/GLTFLoader.js
```diff
@@ -1509,97 +1514,96 @@ THREE.GLTFLoader = ( function () {
...
-		return null;
+		return arrayKey;
 
 	}
 
-	function getCachedMultiPassGeometry( cache, geometry, primitives ) {
+	function createMultiPassGeometryKey( geometry, primitives ) {
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var key = createPrimitiveKey( geometry );
```
donmccurdy: Note that a geometry is not a glTF primitive – i think you want `geometry.uuid` here?

examples/js/loaders/GLTFLoader.js
```diff
@@ -1509,97 +1514,96 @@ THREE.GLTFLoader = ( function () {
...
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
 
-			if ( geometry === cached.baseGeometry && isArrayEqual( primitives, cached.primitives ) ) return cached.geometry;
+			key += i + primitives[ i ].uuid;
```
donmccurdy: note that glTF primitives do not have `.uuid` like Geometry, I think you want `createPrimitiveKey` here.

examples/js/loaders/GLTFLoader.js
rpadovani: Just as a separator, do you think I can drop it completely?

examples/js/loaders/GLTFLoader.js
rpadovani: Uhm, I have confused the two of them, thanks!

rpadovani:
@donmccurdy thanks for the review and the patience :) Fixed them!

examples/js/loaders/GLTFLoader.js
donmccurdy: As a separator I'd use something that can't occur in the UUID itself, like a ':'.

rpadovani:
@donmccurdy I changed the separator, I think we are there :)

mrdoob:
@donmccurdy looks good?

examples/js/loaders/GLTFLoader.js
```diff
@@ -1547,59 +1528,68 @@ THREE.GLTFLoader = ( function () {
...
+	function createMultiPassGeometryKey( geometry, primitives ) {
 
-		for ( var i = 0, il = cache.length; i < il; i ++ ) {
+		var key = geometry.uuid;
 
-			var cached = cache[ i ];
+		for ( var i = 0, il = primitives.length; i < il; i ++ ) {
 
-			if ( geometry === cached.baseGeometry && isArrayEqual( primitives, cached.primitives ) ) return cached.geometry;
+			key += i + createPrimitiveKey( primitives[ i ].uuid );
```
donmccurdy: Note here that `createPrimitiveKey` accepts a primitiveDef, which won't have a `.uuid` until it's been parsed into a BufferGeometry. So just `createPrimitiveKey( primitives[ i ] )` is what we want. I realize there's no easy way to test this bit of code sorry. 🙃

rpadovani:
@donmccurdy thanks for your patience and your review :)

I fixed that last bit!

donmccurdy (approved):
Looks good to me now, thanks! :)

rpadovani:
@mrdoob do you think it will land for r102? :)

mrdoob:
Yes 😇

mrdoob:
Thanks!

PR #15471, opened 2018-12-24
Math  quat and vec3

Thinkia:


examples/css2d_label.html
```diff
@@ -134,10 +135,8 @@
 
 				requestAnimationFrame( animate );
 
-				var elapsed = clock.getElapsedTime();
-
-				moon.position.set( Math.sin( elapsed ) * 5, 0, Math.cos( elapsed ) * 5 );
-
+				moon.position.applyAxisAngle(new THREE.Vector3(0,2,0),Math.PI/120);
```
WestLangley: >new THREE.Vector3(0,2,0)

Avoid doing this. You are instantiating a new vector every frame unnecessarily. Create one and reuse it.

src/core/Object3D.js
```diff
@@ -164,9 +164,10 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 
 		var q1 = new Quaternion();
 
-		return function rotateOnAxis( axis, angle ) {
+		return function rotateOnAxis( vec3, angle ) {
```
WestLangley: `axis` vectors in three.js are always assumed to be unit length to prevent unnecessary normalization. We do not want to change this signature.

examples/css2d_label.html
Thinkia: Thank you for your comment

src/core/Object3D.js
Thinkia: use vec3 maybe simpler  ?

examples/css2d_label.html
```diff
@@ -133,11 +135,8 @@
 			function animate() {
 
 				requestAnimationFrame( animate );
-
-				var elapsed = clock.getElapsedTime();
-
-				moon.position.set( Math.sin( elapsed ) * 5, 0, Math.cos( elapsed ) * 5 );
-
+				moon.position.applyAxisAngle( moonDir, Math.PI / 120 );
+				earth.rotateOnAxis( earthDir, Math.PI / 120 );
```
yomotsu: css2d_label.html is just a behavior example and I think this change makes it slightly more complex. Could we keep this as it is?

mrdoob:
Please, do not include builds in PRs.

src/core/Object3D.js
WestLangley: >axis vectors in three.js are always assumed to be unit length

I do not support changing this convention unless it is changed uniformly throughout the library.

examples/css2d_label.html
Thinkia: sure

PR #15494, opened 2018-12-31
Allows specifying arc length of minor arc; angle of tube section

egrotzke:
Lets the user create partial arcs for the minor arc (the arc of the tube).


Mugen87:
We have been a bit more restrictive with changes like this in the past in order to focus the geometry generators on the most important things. Hence, I'm convinced that  `TorusGeometry` is okay as it is...

PR #15498, opened 2019-01-01
BufferGeometry: Support morph targets in .toNonIndexed()

Mugen87:
Ensure morph attributes are honored in `BufferGeometry.toNonIndexed()`.


hakimmazigh:
في الثلاثاء، ١ يناير، ٢٠١٩ ١٣:١٣، كتب Michael Herzog <
notifications@github.com>:

Mugen87:
I highly recommend to merge this PR since the current implementation of `Buffer.toNonIndexed()` is just incomplete and thus buggy. It took me quite a lot of time to realize that missing morph attributes caused the problem  when using this method with an animated mesh.

PR #15500, opened 2019-01-01
Always use strings for uniform names in examples

Oletus:
Uniforms of materials should be set and accessed by indexing with a string rather than using the dot operator.

This makes the code more compatible with JavaScript tools such as Closure compiler. If the uniform names are not strings, Closure compiler might try to optimize the uniform names, which would create a mismatch between the shader code and the JS code.

This commit should fix the majority of instances where uniforms were being accessed using the dot operator.

### DIFF
--- a/examples/js/objects/Fire.js
+++ b/examples/js/objects/Fire.js
@@ -58,7 +58,7 @@ THREE.Fire = function ( geometry, options ) {
 
 		}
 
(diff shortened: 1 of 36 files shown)


mrdoob:
Interesting...

mrdoob:
Thanks!

Oletus:
Thanks for the review!

PR #15510, opened 2019-01-03
Use current reference to draco loader

webprofusion-chrisc:
Use the current draco loader reference instead of a hard coded THREE.DRACOLoader reference.

This continues to work as normal but also enables use of an alternative draco loader (such as a modular one that's not in the THREE namespace).

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -508,7 +508,7 @@ THREE.GLTFLoader = ( function () {
 		this.name = EXTENSIONS.KHR_DRACO_MESH_COMPRESSION;
 		this.json = json;
 		this.dracoLoader = dracoLoader;
-		THREE.DRACOLoader.getDecoderModule();
+		this.dracoLoader.getDecoderModule();
 
 	}
 


donmccurdy:
Are you sure this change works correctly? I don't think static methods on the DRACOLoader class can be invoked on an instance. Example:

```js
class Gizmo {}
Gizmo.foo = 25;
g = new Gizmo();
console.log(g.foo); // undefined
```

I think you'll have to create a new instance method on the official DRACOLoader class.

webprofusion-chrisc:
@donmccurdy hmm, I see what you mean, my mistake - as I'm fully converted to modules I'm using https://github.com/evanfrawley/three-draco-loader/blob/master/index.js which in turn has getDecoderModule on the prototype.

As this is effectively initializing the DRACOLoader, could that init be moved out of the GLTFLoader? Or can there be an official es6 modular version of DRACOLoader? Currently there doesn't appear to be an official way to use GLTFLoader+ DRACOLoader in a modular way,

donmccurdy:
There's no official way to use any of `examples/js` as ES modules, but [we're working on it](https://github.com/mrdoob/three.js/pull/15518).

In the meantime, the init is there because it substantially reduces loading time... exposing a similar method on the instance would be OK I think though.

webprofusion-chrisc:
Thanks, will close this for now as more thought is required.

PR #15514, opened 2019-01-03
WebGLRenderer: Stop rendering Geometry

Mugen87:
With this PR, `WebGLRenderer` only supports `BufferGeometry`. If users still use `Geometry` in their scene, they have to manually convert their geometries to `BufferGeometry` before rendering like so:
```js
const geometry = new THREE.BufferGeometry().fromGeometry( legacyGeometry );
```
I've tested all WebGL examples, the docs and the editor with this change.

Related #15387

### DIFF
--- a/docs/api/en/core/BufferGeometry.html
+++ b/docs/api/en/core/BufferGeometry.html
@@ -18,9 +18,6 @@
 		<p>
 		To read and edit data in BufferGeometry attributes, see [page:BufferAttribute] documentation.
 		</p>
-		<p>
-		For a less efficient but easier-to-use representation of geometry, see [page:Geometry].
-		</p>
 
 		<h2>Example</h2>
 		<code>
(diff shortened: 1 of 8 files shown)


aardgoose:
Can/will all the ray casting support for Geometry() in Mesh() and Line() be removed as well?
Also computeLineDistances() in Line() and LineSegments()?

Mugen87:
I think so. But please discuss such topics here: #15387

Mugen87:
Should we give it a go in this `dev` (`r102`) cycle?

Mugen87:
Um, https://github.com/mrdoob/three.js/issues/16928 and similar reports on the forum/ stackoverflow showed that still a lot of people create `Points` with `Geometry`. 

Before merging this PR, I suggest to announce the stop of `Geometry` support in `WebGLRenderer` with a console warning. It can be logged when `projectObject()` encounters `Geometry` while processing the scene. One release later, this or a similar PR can be merged.

Mugen87:
Closing due to merge conflicts. It might be easier to setup a new PR when everybody feels okay with stopping support of `Geometry` in  `WebGLRenderer`. As mentioned before, the first change in this process should be an announcement of this change with a console warning.

Issue #15420, opened 2018-12-12
WebGLRenderer.render() should respect previously set render target


maccesch:
A little while ago `clearRenderTarget()` was (rightly) deprecated because it internally changed the render target. Now you have to do it explicitely:

```javascript
renderer.setRenderTarget(target);
renderer.clear();
```

This makes the internally used concept of a state machine more clear.
But by that logic a call to `render()` should use the current render target as well:

```javascript
renderer.setRenderTarget(target);
renderer.render(scene, camera);      // should render to target
```

But that doesn't work because of this line:
https://github.com/mrdoob/three.js/blob/3013895bf68ad0c27fb0f9d3ca0beea2c0e7a87a/src/renderers/WebGLRenderer.js#L1106

Right now it only works like this:
```javascript
renderer.setRenderTarget(target);
renderer.render(scene, camera, target);
```

Please let me know what you think. I'd be happy to submit a PR for this.

pailhead:
I just ran into this. I can't actually figure out how to manually clear the target, since your first snippet doesnt work for me.

maccesch:
@Mugen87 Yes just removing the line would probably break a lot of existing code. But couldn't we deprecate the use of the `renderTarget` parameter and change the code so if you don't provide one then the current active one is used instead of setting it to `null`?

Here's a quick code draft:

```javascript
this.render = function ( scene, camera, renderTarget, forceClear ) {
  if ( renderTarget === true ) {
    forceClear = true;
  }
  ...
  if ( renderTarget instanceof RenderTarget ) {
    console.warn('Deprecated ...');
    this.setRenderTarget( renderTarget );
  }
}
```

Mugen87:
Let's see what @mrdoob and @WestLangley thinks about this. I remember we have recently done similar ctor/method signature changes e.g. #14790

WestLangley:
The WebGL API is a state machine, but the three.js API is not.

The suggestion proposed in this issue would require the render target to be set prior to every render call:
```javascript
renderer.setRenderTarget( null );
renderer.render( scene, camera );
```
because there is no guarantee the render target had not been changed by some other method.

maccesch:
@WestLangley I'm a little confused now. In deprecating `clearRenderTarget()` (#14842) you made basically the opposite argument regarding statefulness. And the `WebGLRenderer.clear()` method definitely follows a state machine concept now as the initial examples in this issue demonstrate.

[...]

PR #15525, opened 2019-01-04
CubeCamera:  restore previously set render target

maccesch:
See discussion in #15420.

### DIFF
--- a/src/cameras/CubeCamera.js
+++ b/src/cameras/CubeCamera.js
@@ -58,6 +58,8 @@ function CubeCamera( near, far, cubeResolution, options ) {
 
 		if ( this.parent === null ) this.updateMatrixWorld();
 
+		var prevRenderTarget = renderer.getRenderTarget();
+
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

src/cameras/CubeCamera.js
```diff
@@ -58,6 +58,8 @@ function CubeCamera( near, far, cubeResolution, options ) {
 
 		if ( this.parent === null ) this.updateMatrixWorld();
 
+		var prevRenderTarget = renderer.getRenderTarget();
```
WestLangley: This pattern is used elsewhere in the library, and the variable name is `currentRenderTarget`. For consistency, I suggest using the same name.

maccesch: Makes sense. Done.

Mugen87 (approved):

mrdoob:
Thanks!

PR #15538, opened 2019-01-07
Docs: Clean up

linbingquan:


looeee:
The examples use this style:

``` js
if ( WEBGL.isWebGLAvailable() === false ) {

	document.body.appendChild( WEBGL.getWebGLErrorMessage() );

}
```

Perhaps we should do the same in the docs?

linbingquan:
> The examples use this style:
> 
> ```js
> if ( WEBGL.isWebGLAvailable() === false ) {
> 
> 	document.body.appendChild( WEBGL.getWebGLErrorMessage() );
> 
> }
> ```
> 
> Perhaps we should do the same in the docs?

@looeee I think it's OK now.

```js
		<code>
		if ( WEBGL.isWebGLAvailable() ) {

			// Initiate function or other initializations here
			animate();

		} else {

			var warning = WEBGL.getWebGLErrorMessage();
			document.getElementById( 'container' ).appendChild( warning );

		}
		</code>
```

Mugen87:
I think @looeee feedback was not about the code style but to change the code section analogous to the examples.

Mugen87:
Let's merge this. We can still align the code according to the examples with a different PR. This one improves the code style after all.

PR #15542, opened 2019-01-07
Docs: Added conversion warning in geometry browser

Mugen87:
see https://github.com/mrdoob/three.js/pull/15504#issuecomment-452093247


mrdoob:
Thanks!

PR #15552, opened 2019-01-09
legacythree2gltf: Add legacy JSON to glTF converter

### LINKED ISSUE #15547: Consider providing a converter for old JSON models
With JSONLoader moving out of the core library and requiring BufferGeometry conversion to render, we should consider providing a converter (based on https://github.com/mrdoob/three.js/tree/dev/utils/converters) to upgrade old JSON models to the new JSON format or glTF. 

For example:

```
node legacythree2gltf.js model.json
```

donmccurdy:
Will probably need some tweaks and bug fixes but this seems like a good start. Fixes https://github.com/mrdoob/three.js/issues/15547.

Related PRs:

- [x] Requires https://github.com/mrdoob/three.js/pull/15551 for skinning
- [x] Requires https://github.com/mrdoob/three.js/pull/15011 for morph targets
- [x] Requires https://github.com/mrdoob/three.js/pull/14116 to avoid increased file sizes

### DIFF
--- a/utils/converters/legacythree2gltf.js
+++ b/utils/converters/legacythree2gltf.js
@@ -1,0 +1,146 @@
+#!/usr/bin/env node
+
+const fs = require( 'fs' );
+const path = require( 'path' );
+const Canvas = require( 'canvas' );
+const { Blob, FileReader } = require( 'vblob' );
+THREE = require( '../../build/three.js' );
+require( '../../examples/js/loaders/deprecated/LegacyJSONLoader.js' );
+require( '../../examples/js/exporters/GLTFExporter.js' );
+
+if ( process.argv.length <= 2 ) {
+
+	console.log( `Usage: ${path.basename( __filename )} model.json` );
+	process.exit( - 1 );
+
+}
+
+//
+
+// Patch global scope to imitate browser environment.
+global.window = global;
+global.Blob = Blob;
+global.FileReader = FileReader;
+global.THREE = THREE;
+global.document = {
+	createElement: ( nodeName ) => {
+
+		if ( nodeName !== 'canvas' ) throw new Error( `Cannot create node ${nodeName}` );
+
+		const canvas = new Canvas( 256, 256 );
+		// This isn't working — currently need to avoid toBlob(), so export to embedded .gltf not .glb.
+		// canvas.toBlob = function () {
+		//   return new Blob([this.toBuffer()]);
+		// };
+		return canvas;
+
+	}
+};
+
+//
+
+// Load legacy JSON file and construct a mesh.
+var file = process.argv[ 2 ];
+var resourceDirectory = THREE.LoaderUtils.extractUrlBase( file );
+var loader = new THREE.LegacyJSONLoader();
+
+var jsonContent = fs.readFileSync( file, 'utf8' );
+var { geometry, materials } = loader.parse( JSON.parse( jsonContent ), resourceDirectory );
+
+var mesh;
+var boneDefs = geometry.bones || [];
+var animations = geometry.animations || [];
+var hasVertexColors = geometry.colors.length > 0;
+
+geometry = new THREE.BufferGeometry().fromGeometry( geometry );
+
+// Remove unnecessary vertex colors and groups added during BufferGeometry conversion.
+if ( ! hasVertexColors ) geometry.removeAttribute( 'color' );
+if ( geometry.groups.length === 1 ) geometry.clearGroups();
+
+if ( ! materials ) {
+
+	materials = new THREE.MeshStandardMaterial( { color: 0x888888, roughness: 1, metalness: 0 } );
+
+}
+
+if ( boneDefs.length ) {
+
+	var { roots, bones } = initBones( boneDefs );
+	mesh = new THREE.SkinnedMesh( geometry, materials );
+	roots.forEach( ( bone ) => mesh.add( bone ) );
+	mesh.updateMatrixWorld( true );
+	mesh.bind( new THREE.Skeleton( bones ), mesh.matrixWorld );
+	mesh.normalizeSkinWeights();
+
+} else {
+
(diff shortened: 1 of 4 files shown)


donmccurdy:
If we want to have the converter, I think it's OK to review and merge this without waiting on #15011 (which shouldn't require any converter changes).

utils/converters/legacythree2gltf.js
```diff
@@ -0,0 +1,155 @@
...
+
+if ( ! materials ) {
+
+	materials = new THREE.MeshStandardMaterial( { color: 0x888888, roughness: 1, metalness: 0 } );
+
+}
+
+if ( optimize ) {
+
+	geometry = THREE.BufferGeometryUtils.mergeVertices( geometry );
```
Mugen87: I'm not sure about this line. If the geometry has uv-coordinates, you might merge vertices with different uv-coordinates. In this case, texture coordinates will be wrong.

(BTW: This does not happen with `Geometry.mergeVertices()` because of the decoupled structure of faces and vertices.

donmccurdy: I think https://github.com/mrdoob/three.js/pull/14116 compares all attributes before merging vertices. But in any case, this is behind the `--optimize` flag here so it can be optionally skipped. Without that flag, the Geometry->BufferGeometry conversion was increasing the number of vertices significantly on these examples.

Mugen87 (approved):

titansoftime:
Wanted to let you know. I used this converter on my models and had mostly a success. SkinnedMesh's work as well. Very nice.

Only issue I see is that for some reason, some files get quite a bit larger. For example I have a Castle model and the exact same model decimated (in Blender) by like 50% or so for mobile version. Oddly enough the converted decimated file is significantly larger than the converted non-decimated file. Very strange. Perhaps due to unnecessary exported attributes as some have reported? I can provide my files if wanted.

I used the GLTF-Pipeline tool to convert the gltf to glb afterwards. Works great, but weird file bloat remains.

So yea this mostly does the job, which is good because my animated json models do not work at all with the LegacyJSONLoader due to it handling bones completely differently (no initBones etc.) and I do not like being more than a couple versions behind the latest three build.

donmccurdy:
Glad it's (almost) working! 

There is an `--optimize` flag, are you still seeing the size increase when using that? That handles the only source of bloat I'm aware of. If you can share a JSON example that doesn't convert well I'll take a look. 

Assuming the export is 1:1 (which isn't guaranteed, with the Geometry->BufferGeometry step involved) I'd expect GLB to be strictly smaller than JSON, with some exceptions for very tiny files and maybe morph targets.

titansoftime:
Indeed.

Yes I am using the --optimize flag. I will upload the json files when I get back home =]

titansoftime:
So I maybe have an idea why the gltf/glb file size is not as reduced (or sometimes larger) than a json file.

The blender json exporter allows for adjustable float precision (I usually use 4-6 myself). However since buffergeometry uses float32arrays for attribute precision it's not really alterable and is locked into max precision (found this out while attempting to modify three2gltf.js, thought I was losing my mind).

So anyway, of course 4x the text length from extended floats are going to cause a size increase... right?

Attached is an example json/glb/gltf zip.
[58.zip](https://github.com/mrdoob/three.js/files/2984788/58.zip)

donmccurdy:
It's not the vertex data that's larger - the Float32Array binary data is still going to be smaller than the JSON representation, even when the JSON only goes to 4 digits precision. If you split the file it's easier to see what's going on:

```
gltf-pipeline -i 58.gltf -o split/58.gltf -s
```

```
ls -lh
-rw-r--r--  1 donmccurdy  eng   678K Mar 19 12:52 58.bin
-rw-r--r--  1 donmccurdy  eng   1.1M Mar 19 12:52 58.gltf
```

In the file above, all of the vertex data is in the `.bin` file, and it's nice and compact at ~50% of the size of the original `.json` file. That remaining 1.1MB of JSON in the `.gltf` file is metadata identifying where each bone's animation comes from in the binary payload – while it's larger than you'd expect with 50 bones and animations on each, the good thing is that this metadata is very repetitive and GZIP-friendly, and the whole 1.1M blob zips down to just 60KB.

Comparing final results:

| file | original size | gzipped |
|---|---|---|
| 58.json | 1.3MB | 0.4MB |
| 58.gltf | 1.4MB | 0.45MB |
| 58.glb + Draco | 0.73MB | 0.13MB |

To create the compressed `.glb` I used `gltf-pipeline`. For some reason the `.glb` in the folder you shared seems to be invalid, and that may be an exporter bug we should look into.

titansoftime:
Ah ok I understand.

I use the gltf-pipeline as well.

Hmm, will have to look into why that that glb is invalid, I've tested several using this method and they are ok. Will look into.

Draco is fantastic at compression but I have found that the decompress time outweighs the advantage, especially for files that are already cached in the browser. Super impressive tech nonetheless.

donmccurdy:
> Hmm, will have to look into why that that glb is invalid, I've tested several using this method and they are ok. Will look into.

I have a custom script I'm using to inspect file size, so it could be that I have a bug in my script too, rather than the file being bad. It does open OK for me.

> Draco is fantastic at compression but I have found that the decompress time outweighs the advantage, especially for files that are already cached in the browser

Can't remember if I mentioned this before but just in case, if you're loading multiple models in parallel then the version of DRACOLoader in https://github.com/mrdoob/three.js/pull/15249 should get you a 4x decoding speed improvement using workers. But yeah, I agree it depends on the use case whether that compression is worthwhile.

Mugen87:
Closing due to #17976. Still thanks for this code!

donmccurdy:
Sounds good! I've copied this over to a Gist that uses the `three` npm package rather than the local file structure: https://gist.github.com/donmccurdy/c090dc53c7bfb704ef9de654ecc07632

PR #15555, opened 2019-01-10
allow optionalTargets in Curve methods to save memory. [MIGRATE] getTangent/getTangentAt required targets

ngokevin:


ngokevin:
Updated.

docs/api/en/extras/core/Curve.html
```diff
@@ -31,32 +31,39 @@ <h3>[property:Integer arcLengthDivisions]</h3>
...
+			otherwise a new Vector will be created.
+    </p>
+		<p>
 			Returns a vector for a given position on the curve according to the arc length.
 		</p>
 
-		<h3>[method:Array getPoints]( [param:Integer divisions] )</h3>
+		<h3>[method:Array getPoints]( [param:Integer divisions], [param:Array optionalTarget] )</h3>
+    <p>
+		  [page:Array optionalTarget] — (optional) target to set the result. Otherwise a new [page:Array] is instantiated. If set, you must clear this array prior to each call (i.e., array.length = 0;).
```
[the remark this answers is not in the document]
ngokevin: Updated.

src/extras/core/Curve.js
```diff
@@ -237,29 +259,48 @@ Object.assign( Curve.prototype, {
...
 	// which seems to give a reasonable approximation
 
-	getTangent: function ( t ) {
+	getTangent: ( function () {
 
-		var delta = 0.0001;
-		var t1 = t - delta;
-		var t2 = t + delta;
+		var pt1 = new Vector3();
+		var pt2 = new Vector3();
```
Mugen87: `.getTangent()` will also be used with `Vector2`. This was the main reason why I have not added an optional target so far to this method. I think it's wrong when the method returns `Vector3` in context of a 2 dimensional data space.

ngokevin: Will try to get it to work with both then?

ngokevin: How about do what the rest of codebase does:

1. Console message saying that a target is required. 
2. Return a Vector3 for now.

We strongly recommend devs explicitly pass in a target, which lets us not have to infer Vector2 or Vector3.

If devs do not pass a target, which is a deprecated route, we just return a Vector3, which is a superset of a Vector2, and can just safely ignore the Z.

Mugen87: I suggest the following option:

Let's create two new class `Curve2` and `Curve3` which derive from `Curve` and overwrite `.getTangent()`. All concrete 3D curve classes (`LineCurve3`, `CatmullRomCurve3`, `CubicBezierCurve3` and `QuadraticBezierCurve3`) would derive from `Curve3`. All other classes from `Curve2`. In this way, it's easier to handle the differences between the 2D and 3D data space. And we should have a clear implementation which is also backwards compatible (since the current implementation of `Curve.getTanget()` would stay as it is).

@WestLangley What do you think about this? I guess this approach should provide us more flexibility with defining the correct type.

WestLangley: I would't worry about backwards compatibility. Just fix it. You'll have to see what seems to make sense...

    2D: Curve -> Curve2 -> CurvePath2 -> Path -> Shape
    3D: Curve -> Curve3 -> CurvePath3

ngokevin: Simplest approach that'd work might be just an internal property like `this.context = '2d'` or `this.context = '3d'`.

WestLangley: You could make the target non-optional when the dimension is unclear.

ngokevin: I'm okay with that.

Mugen87: At some point all optional targets should be mandatory so I think this suggestion is okay.

BTW: If `Curve.getTangent()` gets a new mandatory parameter, then it's necessary to adjust example files like `RollerCoasterGeometry` or `webgl_geometry_extrude_splines`.

ngokevin:
Updated:

- getTangent/getTangentAt is required. Detects target is Vector2 or Vector3 to see which aux vector to use for `pt1`/`pt2`.
- Update docs and examples and tests with the new API change.
- [ ] Migration guide, not sure if I have access to this, but I made it loud in the PR title.

Mugen87:
Closing in favor of #18985.

PR #15580, opened 2019-01-14
Helpers: Fix usage of THREE namespace.

### LINKED ISSUE #15579: RectAreaLightHelper not working on npm distribution
RectAreaLightHelper from 'three' package installed via npm is not working, it seems to try to use at some point THREE.BackSide, but THREE is not existent in the file (it has not been imported) so the construction of a RectAreaLightHelper fails.

To recreate the bug simply install the npm version of three as indicated in the documentation
`npm install three`
and run some code like:
```
import {RectAreaLight, RectAreaLightHelper} from "three";

var light = new RectAreaLight();
var helper = new RectAreaLightHelper(light);
```

When reaching the construction of the new helper, the script fails.

##### Screenshot
Here we can see the error in build/three.module.js inside the node_modules folder, seen from Chrome developer tools:
Imgur

##### Three.js version
npm distribution, three 0.100.0

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Will happen on any

Mugen87:
Fixed #15579

### DIFF
--- a/src/helpers/ArrowHelper.js
+++ b/src/helpers/ArrowHelper.js
@@ -32,8 +32,8 @@ function ArrowHelper( dir, origin, length, color, headLength, headWidth ) {
 
 	Object3D.call( this );
 
-	if ( dir === undefined ) dir = new THREE.Vector3( 0, 0, 1 );
-	if ( origin === undefined ) origin = new THREE.Vector3( 0, 0, 0 );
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

Issue #14921, opened 2018-09-16
Object3D: Introduce onBeforeShadow().


Samsy:
Object3D.onBeforeRender not triggered when depthmaterial for shadows are rendered.

renderBufferDirect method does not trigger the onBeforeRender function

##### Three.js version

- [ ] Dev
- [ ] r96
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
`WebGLShadowMap` has a function [renderObject()](https://github.com/mrdoob/three.js/blob/54071fa06c24461ce11db94ad2e67f5885c09117/src/renderers/webgl/WebGLShadowMap.js#L373) which could call `onBeforeRender()` for the current object. But I don't think should happening since it's an internal rendering for creating shadow maps. What is your use case for this?

Samsy:
Let say you got 20 meshes sharing the same material but needs different uniforms.

Using onBeforeRender function is useful to refresh thoses uniforms associated to each mesh just before they are rendered ( avoiding the creation of 20 materials for 20 meshs, as you just need to refresh the uniforms )

That works, but not when shadow are rendered. If you got some uniforms that influences mesh position for example, it won't get refresh because onBeforeRender is never called when rendering shadow

Mugen87:
Yeah, that makes sense. Let's see what the others think about this issue.

WestLangley:
>Let say you have 20 meshes sharing the same material but need different values of uniforms.

That is what cloning the material is for. They still share a single shader.

> using a same customdepthmaterial shared across all thoses meshes, uniforms won't get refresh per-mesh because onBeforeRender is never called when rendering shadow

Right, but each object can clone the custom depth material, correct?

--

The other solution, presumably, would be to add support for `object.onBeforeShadow()` and `object.onAfterShadow()`, but I have not given it enough thought to know if there is a compelling reason to add that feature.

Samsy:
Main problem with the cloning material thing is, it also clones textures and you end up having as much textures as clone.

Textures in material * number of clones.

For few reasons It would be nice to have more controls over those things. I mean the current system is working, the only thing missing in to get that onBeforeRender function working when shadows are rendered.

WestLangley:
[...]

PR #15581, opened 2019-01-14
WebGLRenderer: Avoid duplicate uniform update with ShaderMaterial

ghost:
When a Material does force a refreshMaterial but also has uniformsNeedUpdate set, it will upload Uniforms twice. This happens with Textures too, which then can result in too many Textures uploaded. (maxTextures are reached)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "three",
-  "version": "0.106.1",
+  "version": "0.106.2",
   "description": "JavaScript 3D library",
   "main": "build/three.js",
   "repository": "mrdoob/three.js",


Mugen87:
The issue does only happen with `ShaderMaterial`, right? The following code section is the only place where `ShaderMaterial.uniformsNeedUpdate` is evaluated:

https://github.com/mrdoob/three.js/blob/c07c597153e7eaaa7eae49f4afb27505c7d702ce/src/renderers/WebGLRenderer.js#L1883-L1886

`ShaderMaterial.uniformsNeedUpdate` should be set by the user. Can you please describe your use case where you set the flag and uniforms are upload twice.

redplant3d:
Here is one scenario:

You got Objects which has an onBeforeRender callback set. You got two Objects which uses the same Shader and applies uniform changes through onBeforeRender. onBeforeRender sets uniformsNeedUpdate to true.
Then when Rendering a Scene:

Render Object A -> onBeforeRender sets uniformsNeedUpdate to true  -> setProgram sets material as active (here it will get called twice, refreshMaterial and uniformsNeedUpdate is true)

Render Object B -> onBeforeRender sets uniformsNeedUpdate to true  -> setProgram sets material as active (here it works, only uniformsNeedUpdate is true)

The whole point is that before setting uniformsNeedUpdate to true their most be knowledge if the Material has been set before. (Make sure refreshMaterial is false)


I think their can be other situations like program update etc. as this always happens when refreshMaterial is true and uniformsNeedUpdate is true.

Mugen87:
Okay, I see. I think the fix is okay if you change the line to the following since only `ShaderMaterial` has `uniformsNeedUpdate` as a property:

```js
if ( material.isShaderMaterial ) material.uniformsNeedUpdate = false;
```

redplant3d:
Done. Dismissed the comment though. Is this okay?

Mugen87:
I think a comment is useful in this case. If someone reads the code, it's easier to understand why this line of code was added.

redplant3d:
Okay, re-added it.

Mugen87 (approved):

WestLangley:
Quite frankly, I was not even aware of #13279.

>You got two Objects which uses the same Shader and applies uniform changes through onBeforeRender

`onBeforeRender()` can be dangerous. Users can mess up the renderer's state. I do not think we should be modifying the library to accommodate edge cases created by a user's use of that method.

@redplant3d Are you cloning your uniforms? Two object instances should not be sharing the same uniforms if the intention is that the values should be different. That is how the system was originally designed, anyway.

Please provide a simple live example to demonstrate your issue.

@Mugen87 Let's avoid approving PRs without an example you can test, unless of course, it is trivial.

Mugen87:
I think this change is simple and avoids an obvious fault.

The problem with a flexibel API is that it's hard to predict how it will be used by the community. It's easy to overlook an obvious usage that might cause issue in other parts of the library. If we don't to restrict the API of `three.js` (e.g. due to refactoring), I prefer to merge such changes in order to make things more consistent. That's the reason why I also vote for changes like mentioned in #14921.

WestLangley:
What is the purpose of `.uniformsNeedUpdate` ?

When should a user set it?

When should a user not set it?

How can you approve a PR without testing it?

Mugen87:
> How can you approve a PR without testing it?

I have tested this with the lensflare example. The change is correct.

WestLangley:
>You can answer all other question by yourself

If you do know the answers to those questions, I would be grateful if you would explain.

Mugen87:
> When should a user set it?

This was actually already explained by @redplant3d. You have multiple objects that use the same `ShaderMaterial` and change uniform values in `onBeforeRender()`.

The flag was introduced when the lensflare plugin was removed from the renderer. A lensflare object uses a single material in order to render an arbitrary amount of lensflare elements. Since it's necessary to change the respective uniforms for rendering a lensflare element, you have to set `uniformsNeedUpdate` to true in order to force a uniform update. Without this flag, no update would happen since shader program and material are identical. Using a single material and `uniformsNeedUpdate` is just a more efficient option than using a material per lensflare element.

Mugen87:
@redplant3d Fixed with `R110` via #17591 😇

makc:
> multiple objects that use the same ShaderMaterial and change uniform values in onBeforeRender()

why not just use two ShaderMaterial-s with the same shaders, @Mugen87 ? I thought 3js would understand multiple materials that produce the same code

PR #15596, opened 2019-01-16
Fixed multi-group geometry bug in GLTFExporter

bovesan:
Uses group object rather than geometry index as cache key.

### DIFF
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -1320,14 +1320,14 @@ THREE.GLTFExporter.prototype = {
 
 				if ( geometry.index !== null ) {
 
-					if ( cachedData.attributes.has( groups[ i ] ) ) {
+					if ( cachedData.attributes.has( geometry.index ) ) {
 
 						primitive.indices = cachedData.attributes.get( geometry.index );
 
 					} else {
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
Hm, unfortunately this has the effect of breaking the cache for single-material geometries. We want to ensure that if a geometry is shared by multiple meshes, its index attribute is written only once.

/cc @takahirox

bovesan:
Aha. We could add a few extra lines to maintain caching for single-material geometries, but I guess proper caching of groups would require bigger changes.

```
let cacheKey = geometry.index;

if ( groups[ i ].start || groups[ i ].count ) {

	cacheKey = groups[ i ];

}

if ( cachedData.attributes.has( cacheKey ) ) {

	primitive.indices = cachedData.attributes.get( geometry.index );

} else {

	primitive.indices = processAccessor( geometry.index, geometry, groups[ i ].start, groups[ i ].count );
	cachedData.attributes.set( cacheKey, primitive.indices );

}
```

I don't have a short and elegant solution for how to cache identical groups in different geometries. If geometry.index had an id field, we could create a cache key like this:
```
cacheKey = geometry.index.id+'-'+groups[ i ].start+'-'+groups[ i ].count ;
```

Issue #15589, opened 2019-01-15
Missing DirectionalLight example


ieskudero:
I have realized that there is no DirectionalLight example, and I can't find an issue explaining why that is, if there is a reason for it or maybe because no one took care of it yet.

Mugen87:
There is also no example for `THREE.AmbientLight` since it's a very basic type of light. `THREE.DirectionalLight` is used all over the place. `webgl_loader_fbx` demonstrates a nice setup with shadows.

https://github.com/mrdoob/three.js/blob/70438028775dd4d539ebdfdaf1aafd6fbcac43c7/examples/webgl_loader_fbx.html#L85-L92

ieskudero:
I see both lights differently in terms of complexity, especially when you need to set the shadows in the DirectionalLight. Anyway, it is true that many example uses directionalLight.

Mugen87:
> I see both lights differently in terms of complexity, especially when you need to set the shadows in the DirectionalLight.

That's true. If you like to add a new example similar to `webgl_lights_hemisphere`, go for it^^.

PR #15600, opened 2019-01-17
added directional light example

ieskudero:
based on the issue #15589. I have tried to create a simple example to show the use of a `THREE.DirectionalLight` and it's shadow, using the `THREE.HemisphereLight` as base.

### DIFF
--- a/examples/webgl_lights_directional.html
+++ b/examples/webgl_lights_directional.html
@@ -1,0 +1,405 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - lights - directional light</title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+		<style>
+			body {
+				background-color: #fff;
+				color: #111;
+				margin: 0px;
+				overflow: hidden;
+				font-family: Monospace;
+				font-size: 13px;
+			}
+
+			#info {
+				position: absolute;
+				top: 0px; width: 100%;
+				padding: 5px;
+				text-align: center;
+			}
+
+			a {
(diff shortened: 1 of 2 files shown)


Mugen87:
Can you please rebase your branch to the latest version of `dev`? If you do this, it will be possible to check the new example with this link:

https://raw.githack.com/ieskudero/three.js/ieskudero-patch-1/examples/webgl_lights_directional.html

BTW: Please add the new example to [files.js](https://github.com/mrdoob/three.js/blob/dev/examples/files.js). Otherwise it's not selectable in the side navigation.

examples/webgl_lights_directional.html
```diff
@@ -0,0 +1,405 @@
...
+
+				//SCENE
+
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color().setHSL( 0.6, 0, 1 );
+				scene.fog = new THREE.Fog( scene.background, 1, 5000 );
+
+				// LIGHTS
+
+				dirLight = new THREE.DirectionalLight( 0xffffff, 1 );
```
Mugen87: I think you can omit the ctor parameters since you change the color in the next line.

Mugen87:
Normally you solve the conflicts offline and then push the merge-commit. Maybe the easiest way is to start with a new branch from scratch. This would also mean you make a fresh PR.

ieskudero:
> Normally you solve the conflicts offline and then push the merge-commit

I should have done so, yes, sorry about it. The conflict is in the `file.js`. The solution to this conflict is to accept both lines (`"webgl_lightningstrike"` and `"webgl_lights_directional"`) but I don't have write access.

I think it would be easier to solve it that way, but I can make another PR if you want, @Mugen87 .

And again, sorry for all the trouble.

PR #15601, opened 2019-01-17
New GLTFExporter exporter options "maxTextureWidth" and "maxTextureHeight"

RuiBEAT:
New export options added to the GLTFExporter allowing to resize the embedded textures restricting them to a maximum of 'maxTextureWidth' x 'maxTextureHeight'. Example and documentation updated according to it.

### DIFF
--- a/docs/examples/exporters/GLTFExporter.html
+++ b/docs/examples/exporters/GLTFExporter.html
@@ -96,6 +96,8 @@
 			<li>animations - Array<[page:AnimationClip AnimationClip]>. List of animations to be included in the export.</li>
 			<li>forceIndices - bool. Generate indices for non-index geometry and export with them. Default is false.</li>
 			<li>forcePowerOfTwoTextures - bool. Export with images resized to POT size. This option works only if embedImages is true. Default is false.</li>
+			<li>maxTextureWidth - int. Restricts the image maximum width to the given value. This option works only if embedImages is true. Default is 4096.</li>
+			<li>maxTextureHeight - int. Restricts the image maximum height to the given value. This option works only if embedImages is true. Default is 4096.</li>
 		</ul>
 		</p>
 		<p>
--- a/docs/examples/exporters/GLTFExporter.html
+++ b/docs/examples/exporters/GLTFExporter.html
@@ -96,8 +96,7 @@
 			<li>animations - Array<[page:AnimationClip AnimationClip]>. List of animations to be included in the export.</li>
 			<li>forceIndices - bool. Generate indices for non-index geometry and export with them. Default is false.</li>
 			<li>forcePowerOfTwoTextures - bool. Export with images resized to POT size. This option works only if embedImages is true. Default is false.</li>
-			<li>maxTextureWidth - int. Restricts the image maximum width to the given value. This option works only if embedImages is true. Default is 4096.</li>
-			<li>maxTextureHeight - int. Restricts the image maximum height to the given value. This option works only if embedImages is true. Default is 4096.</li>
+			<li>maxTextureSize - int. Restricts the image maximum size (both width and height) to the given value. This option works only if embedImages is true. Default is Infinity.</li>
 		</ul>
 		</p>
 		<p>
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -711,14 +711,13 @@ THREE.GLTFExporter.prototype = {
 				canvas.width = Math.min( image.width, options.maxTextureSize );
 				canvas.height = Math.min( image.height, options.maxTextureSize );
 
-				if ( options.forcePowerOfTwoTextures ) {
+				if ( options.forcePowerOfTwoTextures && ! isPowerOfTwo( canvas ) ) {
 
-					if(! isPowerOfTwo( image )) {
-						console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
-					}
+					console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
 
 					canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
 					canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
+
 				}
 
 				var ctx = canvas.getContext( '2d' );
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -81,8 +81,7 @@ THREE.GLTFExporter.prototype = {
 			animations: [],
 			forceIndices: false,
 			forcePowerOfTwoTextures: false,
-			maxTextureHeight: 4096,
-			maxTextureWidth: 4096
+			maxTextureSize: Infinity
 		};
 
 		options = Object.assign( {}, DEFAULT_OPTIONS, options );
@@ -709,21 +708,27 @@ THREE.GLTFExporter.prototype = {
 
 				var canvas = cachedCanvas = cachedCanvas || document.createElement( 'canvas' );
 
-				if ( image.width > options.maxTextureWidth || image.height > options.maxTextureHeight ) {
+				canvas.width = Math.min( image.width, options.maxTextureSize );
+				canvas.height = Math.min( image.height, options.maxTextureSize );
 
-					console.warn( 'GLTFExporter: Resized too big image.', image );
-				}
+				if ( options.forcePowerOfTwoTextures) {
 
-				canvas.width = Math.min( image.width, options.maxTextureWidth );
-				canvas.height = Math.min( image.height, options.maxTextureHeight );
+					if(! isPowerOfTwo( image )) {
+						console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
+					}
 
-				if ( options.forcePowerOfTwoTextures && ! isPowerOfTwo( image ) ) {
+					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
+						console.warn( 'GLTFExporter: power-of-two textures requested, but the given maxTextureSize is not power-of-two.')
+					}
 
-					console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
-
-					canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
-					canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
-
+					// REMARK: These 'if' checks are performed just to avoid the unnecessary POW+LOG calls behind floorPowerOfTwo
+					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
+						canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
+					}
+					
+					if(! THREE.Math.isPowerOfTwo( canvas.height )) {
+						canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
+					}
 				}
 
 				var ctx = canvas.getContext( '2d' );
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -711,24 +711,14 @@ THREE.GLTFExporter.prototype = {
 				canvas.width = Math.min( image.width, options.maxTextureSize );
 				canvas.height = Math.min( image.height, options.maxTextureSize );
 
-				if ( options.forcePowerOfTwoTextures) {
+				if ( options.forcePowerOfTwoTextures ) {
 
 					if(! isPowerOfTwo( image )) {
 						console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
 					}
 
-					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
-						console.warn( 'GLTFExporter: power-of-two textures requested, but the given maxTextureSize is not power-of-two.')
-					}
-
-					// REMARK: These 'if' checks are performed just to avoid the unnecessary POW+LOG calls behind floorPowerOfTwo
-					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
-						canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
-					}
-					
-					if(! THREE.Math.isPowerOfTwo( canvas.height )) {
-						canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
-					}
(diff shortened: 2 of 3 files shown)


donmccurdy:
Nice! Although, I'm not sure if this functionality belongs in the exporter, or in a *Utils class... thoughts @takahirox / @fernandojsg?

takahirox:
We concluded that we keep the exporter simple as much as possible. Optimization or additional work , e.g. draco compression, should be done in post-process pipe-line tool. So I prefer resizing in pipe-line tool unless there's any special reason we should do in the exporter.

RuiBEAT:
Hi guys! Thank you for your quick response :) I've decided to add it to the exporter options because It keeps a close relationship with the POT option. Both are size restrictions to the final texture size when embedding it as binary, and both options just set the size of the canvas where the texture will be blitted.

In addition, we must remember that many mobile devices have a max texture size restriction, not just for performance but for compatibility.

In my use case, I need to re-export models that my web platform users upload in different formats to be later downloaded for use from mobile devices. Conceptually, I simply need to re-export them to ensure compatibility, so I though about it as an exporter option. Since it has no performance impact on the process, and having seen the similarity with the POT option, it seemed to me that the exporter was the right place... so I discarded doing it in additional utility tools that would involve additional drawing ops and computational cost :)

looeee:
@RuiBEAT due to the way that browsers write textures, in general the textures created by the exporter are very large. This means that an image optimization step is usually needed anyway, otherwise your file sizes will be very large. 

To see this, import a simple .glb file like a cube containing well optimised textures using the glTF loader, then export it with the exporter. The file size will increase quite a bit.

This is an unfortunate limitation of the loader and it would be great if we could find a workaround, but until then I'm not sure how useful other image optimizations are.

More information here:
https://discourse.threejs.org/t/three-gltfexporter-causing-huge-size-difference/3842/10

looeee:
> We concluded that we keep the exporter simple as much as possible. 

@takahirox where was this concluded? 

Obviously there are limits to the complexity we want to deal with here, but currently the `GLTFExporter` is the best way to share scenes created with three.js, so we should be willing to at least consider any optimisations that makes this process easier.

RuiBEAT:
@looeee Yes, file size could be maybe increased, but not the size of the texture in memory (something specially critical for mobile devices). If the texture is not hw-accelerated (.jpg, .png, etc...) the memory consumption when the image is loaded will be width*height*pixel_width, depending directly on the image final resolution. A 32 bit texture 2048x2048 (it doesn't matter if is a .jpg, .png or whatever) it will cost 32bit*2048px*2048px = 16 MB on the device memory. I have found some of my users using models with embedded textures up to 8192x8192 pixels... so 256MB on memory (even when the file size is reasonable)!! 

So we are dealing with two problems here: **file size** and **memory consumption**

Ideally, textures could be exported depending on the final platform (.DDS .PVR, .ETC) to support hw acceleration and optimize memory consumption. But when the same model file needs to be loaded in different devices, this is not a choice.

donmccurdy:
> >  We concluded that we keep the exporter simple as much as possible.
> 
> @takahirox where was this concluded?

I don't know that we've set any hard rule here, but see https://github.com/mrdoob/three.js/pull/14485 for previous discussion. There's a long list of optimizations that could be done to glTF models (resize textures, pack textures, compressed textures, draco compression, interleaved attributes, merging vertices, merging meshes, flattening the node hierarchy, ...) and we'd much rather leave most of that out of GLTFExporter if it can be implemented some other way. The file is already large, and (as an `examples/js/*` class) can't really be refactored into smaller files.

In particular, if optimizations can be applied before passing an object into the exporter, then the optimizations are available at runtime and when using the other six three.js exporters. If optimizations can be applied after export with tools like [glTF-Pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/) or [glTF-Transform](https://github.com/donmccurdy/glTF-Transform), they'll be available to users bringing glTF into three.js from other sources. Both of those are preferable to a new option in GLTFExporter.

One possibility would be to move the [existing texture resize methods from the renderer](https://github.com/mrdoob/three.js/blob/b53cec23cd91f8f84a6f149b5debe0650575d6b3/src/renderers/webgl/WebGLTextures.js#L16-L75) into a utility class.

The POT option exists because Facebook 3D Posts rejected many files without it, and we wanted exports from the Editor to be compatible with that.

examples/js/exporters/GLTFExporter.js
```diff
@@ -707,8 +709,13 @@ THREE.GLTFExporter.prototype = {
 
 				var canvas = cachedCanvas = cachedCanvas || document.createElement( 'canvas' );
 
-				canvas.width = image.width;
-				canvas.height = image.height;
+				if ( image.width > options.maxTextureWidth || image.height > options.maxTextureHeight ) {
+
+					console.warn( 'GLTFExporter: Resized too big image.', image );
+				}
```
donmccurdy: Let's omit the warning here, but change the default max size to `Infinity`. If the user has to opt in there's no need to warn.

examples/js/exporters/GLTFExporter.js
```diff
@@ -80,7 +80,9 @@ THREE.GLTFExporter.prototype = {
 			embedImages: true,
 			animations: [],
 			forceIndices: false,
-			forcePowerOfTwoTextures: false
+			forcePowerOfTwoTextures: false,
+			maxTextureHeight: 4096,
+			maxTextureWidth: 4096
```
donmccurdy: Let's merge these into a single option, `maxTextureSize`.

examples/misc_exporter_gltf.html
```diff
@@ -112,6 +117,22 @@
...
+    			if (e.target.value == '') {
+      				e.target.value = parseInt(document.getElementById('option_maxwidth').getAttribute('max'), 10)
+    			}
+			  } );
+			  
+			document.getElementById('option_maxheight').addEventListener('change', function (e) {
+				if (e.target.value == '') {
+					e.target.value = parseInt(document.getElementById('option_maxheight').getAttribute('max'), 10)
+				}
+			} );
```
donmccurdy: I don't think we need the event handlers for validation here — use `inputEl.value || Infinity`, if they're empty they can be ignored.

examples/misc_exporter_gltf.html
```diff
@@ -36,6 +36,8 @@
 			<label><input id="option_binary" name="visible" type="checkbox">Binary (<code>.glb</code>)</label>
 			<label><input id="option_forceindices" name="visible" type="checkbox">Force indices</label>
 			<label><input id="option_forcepot" name="visible" type="checkbox">Force POT textures</label>
+			<label><input id="option_maxwidth" name="maxWidth" type="number" value="4096" min="2" max="4096"> Max texture width</label>
+			<label><input id="option_maxheight" name="maxHeight" type="number" value="4096" min="2" max="4096"> Max texture height</label>
```
donmccurdy: can add `step="1"` to make this accept only integers.

examples/misc_exporter_gltf.html
```diff
@@ -56,7 +58,10 @@
 					truncateDrawRange: document.getElementById( 'option_drawrange' ).checked,
 					binary: document.getElementById( 'option_binary' ).checked,
 					forceIndices: document.getElementById( 'option_forceindices' ).checked,
-					forcePowerOfTwoTextures: document.getElementById( 'option_forcepot' ).checked
+					forcePowerOfTwoTextures: document.getElementById( 'option_forcepot' ).checked,
+					maxTextureHeight: parseInt(document.getElementById('option_maxheight').nodeValue, 10) || 4096, // To prevent NaN value
+					maxTextureWidth: parseInt(document.getElementById('option_maxheight').nodeValue, 10) || 4096 // To prevent NaN value
```
donmccurdy: See above for requiring an integer, I think this can be simplified to:

```
Number(document.getElementById('option_maxheight').value) || 4096
```

donmccurdy:
I think I'm OK with this. After the changes suggested inline (a single `maxTextureSize` option) it's a 1-line addition, since we're already assigning a width to the canvas. @takahirox sound OK?

examples/js/exporters/GLTFExporter.js
```diff
@@ -707,8 +709,13 @@ THREE.GLTFExporter.prototype = {
...
-				canvas.height = image.height;
+				if ( image.width > options.maxTextureWidth || image.height > options.maxTextureHeight ) {
+
+					console.warn( 'GLTFExporter: Resized too big image.', image );
+				}
+
+				canvas.width = Math.min( image.width, options.maxTextureWidth );
+				canvas.height = Math.min( image.height, options.maxTextureHeight );
 
 				if ( options.forcePowerOfTwoTextures && ! isPowerOfTwo( image ) ) {
```
fernandojsg: I believe we should check `isPowerOfTwo( canvas )` in case the user has set a `maxTextureWidth/Height` that are not POT and we have resized the canvas to that new size.

fernandojsg (commented):
Overall it looks good to me. I agree that the size of this PR is small enough for the benefit it could bring.

RuiBEAT:
@donmccurdy I have updated the code following most of your advices. However, I keep a warning message in case the user has set a maxTextureSize that is not POT, and the forcePOT is checked (as @fernandojsg suggested in his last comment)

takahirox:
Hm I still would like to focus on just exporting valid glTF in the exporter and lean toward removing options that can be done out of the exporter. Actually I've been thinking if we can remove trs, forceIndices, and forcePowerOfTwoTextures options (regarding the last two options, once FB supports non-index and non-POT).

The exporter is still growing. To keep it simple, I'd like to move out the stuff which can be done out of the exporter. And if we continue to add options even which are small, the test might become complex. We may need to take care some combinations.

But yeah the change in this PR isn't big and the exporter isn't so complex yet. So merging for now and revisiting the options when necessary?

docs/examples/exporters/GLTFExporter.html
```diff
@@ -96,6 +96,8 @@ <h3>[method:null parse]( [param:Object3D input], [param:Function onCompleted], [
 			<li>animations - Array<[page:AnimationClip AnimationClip]>. List of animations to be included in the export.</li>
 			<li>forceIndices - bool. Generate indices for non-index geometry and export with them. Default is false.</li>
 			<li>forcePowerOfTwoTextures - bool. Export with images resized to POT size. This option works only if embedImages is true. Default is false.</li>
+			<li>maxTextureWidth - int. Restricts the image maximum width to the given value. This option works only if embedImages is true. Default is 4096.</li>
+			<li>maxTextureHeight - int. Restricts the image maximum height to the given value. This option works only if embedImages is true. Default is 4096.</li>
```
donmccurdy: Can be just `maxTextureSize` here now.

examples/js/exporters/GLTFExporter.js
```diff
@@ -707,16 +708,27 @@ THREE.GLTFExporter.prototype = {
...
-					console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
+					if(! isPowerOfTwo( image )) {
+						console.warn( 'GLTFExporter: Resized non-power-of-two image.', image );
+					}
 
-					canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
-					canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
+					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
+						console.warn( 'GLTFExporter: power-of-two textures requested, but the given maxTextureSize is not power-of-two.')
+					}
```
donmccurdy: I don't think this requires a warning – we should just use the largest size that is both less than maxTextureSize and still a power of two. We also don't know here that the canvas width came from maxTextureSize. The `THREE.Math.ceilPowerOfTwo( value )` and `THREE.Math.floorPowerOfTwo( value )` functions may be helpful here.

examples/js/exporters/GLTFExporter.js
```diff
@@ -707,16 +708,27 @@ THREE.GLTFExporter.prototype = {
 
 				var canvas = cachedCanvas = cachedCanvas || document.createElement( 'canvas' );
 
-				canvas.width = image.width;
-				canvas.height = image.height;
+				canvas.width = Math.min( image.width, options.maxTextureSize );
+				canvas.height = Math.min( image.height, options.maxTextureSize );
 
-				if ( options.forcePowerOfTwoTextures && ! isPowerOfTwo( image ) ) {
+				if ( options.forcePowerOfTwoTextures) {
```
donmccurdy: nit: please keep spaces around parens, e.g. `( options.forcePowerOfTwoTextures )`.

examples/js/exporters/GLTFExporter.js
```diff
@@ -707,16 +708,27 @@ THREE.GLTFExporter.prototype = {
...
+					}
 
+					// REMARK: These 'if' checks are performed just to avoid the unnecessary POW+LOG calls behind floorPowerOfTwo
+					if(! THREE.Math.isPowerOfTwo( canvas.width )) {
+						canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
+					}
+					
+					if(! THREE.Math.isPowerOfTwo( canvas.height )) {
+						canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
+					}
```
donmccurdy: After the suggestion above, these two if blocks can be removed.

examples/js/exporters/GLTFExporter.js
donmccurdy: You're warning that the user gave a non-POT `maxTextureSize`, but you're checking the value of `canvas.width` and it came from `Math.min( image.width, options.maxTextureSize );`. So, it's possible that the user requested a POT max size, but the actual image was already non-POT and less than that. Still, let's just omit the warning.

And ok, you're using floorPowerOfTwo anyway and ceilPowerOfTwo isn't needed, so that's fine.

examples/js/exporters/GLTFExporter.js
donmccurdy: Just this would be cleaner, without the if blocks:

```js
canvas.width = THREE.Math.floorPowerOfTwo( canvas.width );
canvas.height = THREE.Math.floorPowerOfTwo( canvas.height );
```

The cost of computing a POT is trivial compared to writing out the data.

donmccurdy (approved):
Thanks!

Mugen87:
@RuiBEAT Any chances to resolve the merge conflicts? Unfortunately, this PR was overlooked.

Mugen87:
Already solved here #17553

PR #15602, opened 2019-01-17
WebGLTextures: Refactoring

Mugen87:
Since there are some issues related to textures, I would like to refactor `WebGLTextures` a bit so it's easier to implement new features (e.g. avoid duplicate texture uploads).

The first thing I'd like to clean up is the usage of `__image__webglTextureCube` which stores the `WebGLTexture` object for a cube texture. I can't think of a reason why `__webglTexture` is not used to store the reference. Having a single property that stores the reference to a `WebGLTexture` object makes it easier to init and delete it.


mrdoob:
Lets merge this at the start of 102 just in case.

Mugen87:
👍

mrdoob:
Thanks!

PR #15605, opened 2019-01-18
Sprite text

anhr:
A sprite based text component. Because a sprite is being used, the text will always face the camera, and have its orientation fixed relative to the camera.
[Example](https://raw.githack.com/anhr/three.js/SpriteText3/examples/webgl_sprites_text.html)


looeee:
[Live example](https://raw.githack.com/anhr/three.js/SpriteText3/examples/webgl_sprites_text.html)

looeee:
@anhr thanks for making this contribution. I agree that an example showing how to draw text onto a `CanvasTexture` is useful. 

However, given that this can be done in just a couple of lines I'm inclined to think that adding `SpriteText.js` is unnecessary. Also, there's no particular reason to assume that the user would only want to draw text onto a `Sprite`, they are just as likely to want to use a normal planar mesh instead. 

Here's a quick mock up of how a simple example that demonstrates this would look:

https://codepen.io/looeee/pen/PXrPKW

anhr:
@looeee thanks for comment. I have played with your [code](https://codepen.io/looeee/pen/PXrPKW) . Of course your code is simple. I had wrote similar code in my project before.
But I do not like size of rectangle around text. I see same size of rectangle if I cahange text from '^_^ :D :--) :0' to short '^' or to '^_^ :D :--) :0 very long string'.
I have spend long time for resolving this issue. Fortunately, I found a solution [three-spritetext](https://github.com/vasturiano/three-spritetext) . Now I can change text dinamically and I see size of rectangle around text is changing also.
I think some developers have same problem and can use my SpriteText.js in his projects.

looeee:
Ok, that's a valid argument. I'm still inclined to think that's something we can leave up to people to deal with in their own apps, but let's see what other people have to say.

Mugen87:
>  I'm still inclined to think that's something we can leave up to people to deal with in their own apps

I agree with that.

PR #15609, opened 2019-01-18
Add directionalLight example

ieskudero:
based on the issue #15589. I have tried to create a simple example to show the use of a `THREE.DirectionalLight `and it's shadow, using the `THREE.HemisphereLight ` example as base.

### DIFF
--- a/examples/webgl_lights_directional.html
+++ b/examples/webgl_lights_directional.html
@@ -94,7 +94,7 @@
 
 			}
 
-			var controls, camera, scene, renderer, dirLight, dirLightHeper, shadowCameraHelper;
+			var controls, camera, scene, renderer, dirLight, dirLightHelper, shadowCameraHelper;
 			var mixers = [];
 			var stats;
 			var gui = new dat.GUI();
@@ -124,12 +124,12 @@
 
 				dirLight = new THREE.DirectionalLight( );
 				dirLight.color.setHSL( 0.1, 1, 0.95 );
-				dirLight.position.set( - 1, 1.75, 1 );
-				dirLight.position.multiplyScalar( 30 );
+				dirLight.position.set( - 30, 50, 30 );
+
 				scene.add( dirLight );
 
-				dirLightHeper = new THREE.DirectionalLightHelper( dirLight, 10 );
-				scene.add( dirLightHeper );
+				dirLightHelper = new THREE.DirectionalLightHelper( dirLight, 10 );
+				scene.add( dirLightHelper );
 
 				//SHADOW
 
@@ -368,7 +368,7 @@
 					case 68: // d
 
 						dirLight.visible = ! dirLight.visible;
-						dirLightHeper.visible = ! dirLightHeper.visible;
+						dirLightHelper.visible = ! dirLightHelper.visible;
 						break;
 
 				}
--- a/examples/webgl_lights_directional.html
+++ b/examples/webgl_lights_directional.html
@@ -195,7 +195,6 @@
 					var s = 0.35;
 					mesh.scale.set( s, s, s );
 					mesh.position.y = 15;
-					//mesh.rotation.y = - 1;
 
 					mesh.castShadow = true;
 
@@ -204,6 +203,9 @@
 					var mixer = new THREE.AnimationMixer( mesh );
 					mixer.clipAction( gltf.animations[ 0 ] ).setDuration( 1 ).play();
 					mixers.push( mixer );
+
+					dirLight.target = mesh;
+					dirLightHelper.update();
 
 				} );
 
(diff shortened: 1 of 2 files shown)


Mugen87:
Live link: https://raw.githack.com/ieskudero/three.js/ieskudero-patch-1/examples/webgl_lights_directional.html

Mugen87:
It would be great if the following UI controls for the shadow camera's frustum setting would be range sliders. A range of maybe 500 should be sufficient. Besides, it would be great if the initial camera angle is a bit lower so you can see the horizon.

Mugen87 (approved):

examples/webgl_lights_directional.html
```diff
@@ -0,0 +1,405 @@
...
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color().setHSL( 0.6, 0, 1 );
+				scene.fog = new THREE.Fog( scene.background, 1, 5000 );
+
+				// LIGHTS
+
+				dirLight = new THREE.DirectionalLight( );
+				dirLight.color.setHSL( 0.1, 1, 0.95 );
+				dirLight.position.set( - 1, 1.75, 1 );
+				dirLight.position.multiplyScalar( 30 );
```
WestLangley: This is confusing IMO. Just set the position where you want it.

WestLangley:
I'd also demonstrate directional light `target` functionality.
```javascript
scene.add( new THREE.AxesHelper( 100 ) );
...
dirLight.target = mesh
dirLightHeper.update();

```
Then be able to vary `mesh.position` with a slider.

examples/webgl_lights_directional.html
ieskudero: Yes, it is. This code is taking from THREE.HemisphereLight example, should I clean it too?

examples/webgl_lights_directional.html
```diff
@@ -0,0 +1,405 @@
...
+					'512': 512,
+					'1024': 1024,
+					'2048': 2048
+				} ).onChange( function ( value ) {
+
+					dirLight.shadow.mapSize.width = dirLight.shadow.mapSize.height = value;
+
+				} );
+
+				shadowf.add( datas, 'left', -250, 250 ).step( 1 ).onChange( function ( value ) {
```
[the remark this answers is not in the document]
ieskudero: OK, done.

examples/webgl_lights_directional.html
WestLangley: In a different PR.

WestLangley:
>I haven't add a THREE.AxesHelper()

OK.

>The Y position of the target is 15, but it always starts from 0.

You will have to debug that.

Also, the shadow frustum can't have zero area, you are generating console errors.

ieskudero:
> Also, the shadow frustum can't have zero area, you are generating console errors

I didn´t understand this. I can see the error but I don´t know why is happening. @WestLangley  Can you explain this a little bit more, please?

WestLangley:
>I think it would be better to constrain left from -250 to zero -- or to -10.

Try - 10.

ieskudero:
frustum errors removed. There is only one error that this, and almost all of light examples, are giving:

`WebGL warning: clear: This operation requires zeroing texture data. This is slow`

For what I have read, this error happens only in Firefox, and the scene renders anyway.

@WestLangley , @Mugen87  any other changes you might want me to add?

Mugen87:
LGTM. If @WestLangley think it's okay, we can merge it.

examples/webgl_lights_directional.html
```diff
@@ -0,0 +1,444 @@
...
+			}
+
+			function setShadowCamera() {
+
+				if( shadowCameraHelper ) {
+
+					scene.remove( shadowCameraHelper );
+
+				}
+				shadowCameraHelper = new THREE.CameraHelper( dirLight.shadow.camera );
```
WestLangley: Calling this every time the GUI is modified is causing hundreds of geometries to be allocated on the GPU.

Type `renderer.info` into the console.

I expect you only need one shadow camera helper.

ieskudero: calling `shadowCameraHelper.update()` didn´t change the camera lines, so in order to show the helper changes I choose to create the camera helper everytime. I don't like it either, so If there is a way to avoid this, tell me and I will gladly change it.

WestLangley: Please try to track it down. Users copy these examples.

WestLangley:
To update the shadow camera helper when the shadow camera frustum is modified:
```javascript
light.shadow.camera.left = - 200;
light.shadow.camera.updateProjectionMatrix();
light.shadowCameraHelper.update()
```
Please use this pattern instead of instantiating a new helper each time.

examples/webgl_lights_directional.html
WestLangley: See note below.

WestLangley:
To update the shadow camera helper when the shadow camera frustum is modified:
```javascript
light.shadow.camera.left = - 200;
light.shadow.camera.updateProjectionMatrix();
shadowCameraHelper.update();
```
Please use this pattern instead of instantiating a new helper each time.

---
Maybe we should have the helper update the projection matrix automatically.

Mugen87:
@mrdoob I think this PR is ready to go 🚀

Mugen87:
Closing. If there is still interest in this example, it should be updated first so it uses modules only.

Issue #15603, opened 2019-01-17
Add Stencil Function Material Properties


gkjohnson:
At the moment writing to the stencil buffer is a bit inconvenient (and undocumented). Would it make sense to add material property that defines the stencil operations? Here's an idea of what the API might look like:

```js
// Render where a mesh intersects the rest of the world
const frontFaceMat = new MeshBasicMaterial();
frontFaceMat.side = FrontSide;
frontFaceMat.depthWrite = false;
frontFaceMat.colorWrite = false;
frontFaceMat.stencil = {
    func: ALWAYS,
    ref: 0,
    mask: 0xff,

    fail: KEEP,
    zfail: KEEP,
    zpass: INCR_WRAP
};

const backFaceMat = new MeshBasicMaterial();
backFaceMat.side = BackSide;
backFaceMat.depthWrite = false;
backFaceMat.colorWrite = false;
backFaceMat.stencil = {
    func: ALWAYS,
    ref: 0,
    mask: 0xff,

    fail: KEEP,
    zfail: KEEP,
    zpass: DECR_WRAP
};

const intersectionMat = new MeshBasicMaterial();
intersectionMat.depthWrite = false;
intersectionMat.depthTest = false;
intersectionMat.opacity = 0.25;
intersectionMat.color.set(0xff0000);
intersectionMat.stencil = {
    func: NOTEQUAL,
    ref: 0,
    mask: 0xff,

    fail: REPLACE,
    zfail: REPLACE,
    zpass: REPLACE
};

const materials = [ frontFaceMat, backFaceMat, intersectionMat ];
const intersectionGroup = SceneUtils.createMultiMaterialObject( geometry, materials );
```

Of course this depends on the meshes in the `intersectionGroup` being run _in order_ and _after everything we want to intersect_, which the default webgl renderer object sort would probably mess up but maybe there are ways around that I'm unaware of?

WestLangley:
>Would it make sense to add material property that defines the stencil operations?

Yes. I am at a loss to recall why we did not do that in the first place.

In fact, adding an example along the lines of your suggestion would be very beneficial.

---

Adding a link to [this SO answer](https://stackoverflow.com/a/54238011/1461008) and, in particular, my subsequent comments. I think making gl calls behind the back of the renderer is unadvisable.

gkjohnson:
> I think making gl calls behind the back of the renderer is unadvisable.

[...]

PR #15611, opened 2019-01-19
Add Stencil Parameter to Materials

gkjohnson:
Adds a stencil field to materials to enable simpler use of stencil operations as suggested in #15603.

The stencil state initializes to `null`, which means the stencil operation is disabled. You can enable it by setting the field to an object like so:

```js
var frontMat = new THREE.MeshBasicMaterial();
frontMat.stencilWrite = true;

frontMat.stencilFunc = THREE.AlwaysStencilFunc;
frontMat.stencilRef = 0;
frontMat.stencilMask = 0xff;

frontMat.stencilFail = THREE.KeepStencilOp;
frontMat.stencilZFail = THREE.KeepStencilOp;
frontMat.stencilZPass = THREE.IncrementWrapStencilOp;
```

I made a basic example showing that it works [here](https://raw.githack.com/gkjohnson/three.js/stencil-example/examples/webgl_stencil.html). ([source](https://github.com/gkjohnson/three.js/blob/stencil-example/examples/webgl_stencil.html)).

image

#### Documentation Updates

I've updated the docs with information about the stencil operations:

[materials/Material.html](https://raw.githack.com/gkjohnson/three.js/add-material-stencil/docs/api/en/materials/Material.html)

[constants/Materials.html](https://raw.githack.com/gkjohnson/three.js/add-material-stencil/docs/api/en/constants/Materials.html)


gkjohnson:
@WestLangley Absolutely! If it starts looking like this is going to get merged I'll look into updating the docs and creating a better example.

WestLangley:
In the mean time, I suggest the following change to your temporary example so it is clearer to others what your program is doing.
```javascript
// intersection highlight
var geometry = new THREE.BoxBufferGeometry( 2, 2, 0.1 );
geometry.translate( 1, 0, 0 );
...
var displayMat = new THREE.MeshPhongMaterial();
displayMat.opacity = 0.5;
```
Just a suggestion...

pailhead:
Why put it in `stencil ={}` ? I think this is different than what is there with blending now, and you have to pass the entire object with everything in it otherwise you don't know what defaults are used internally. How about just default values for all of them? (like blending does)

gkjohnson:
@pailhead

You're right -- I think I prefer the nested style here but it's not necessarily in family with the other fields. Same with the constants. So should they be named more like so?

```js
const frontFaceMat = new MeshBasicMaterial();
frontFaceMat.stencilWrite = true;

frontFaceMat.stencilFunc = AlwaysStencilFunc;
frontFaceMat.stencilRef = 0;
frontFaceMat.stencilMask = 0xff;

frontFaceMat.stencilFail = KeepStencilOp;
frontFaceMat.stencilZFail = KeepStencilOp;
frontFaceMat.stencilZPass = IncrementWrapStencilOp;
```

pailhead:
yeah I think the other tricky thing here is that it’s not just the material saying what the state should be but the renderer too. I’m talking about the opaque and then transparent rendering. If using the same api as transparent maybe there should be a stencil pass that doesn’t make a whole lot of sense. 

I’m also not sure what happens when you have the test just set to pass vs no test. Ie if you have just one material with some stencil write should all the others also perform an always passing stencil test?

gkjohnson:
> yeah I think the other tricky thing here is that it’s not just the material saying what the state should be but the renderer too. I’m talking about the opaque and then transparent rendering. If using the same api as transparent maybe there should be a stencil pass which may not make a whole lot of sense.

I don't think a separate stencil pass quite makes sense because the order in which you render your stencil passes is completely dependent on the desired effect. `renderOrder` is really the only way to get that control. The transparent vs opaque binning in the renderer complicates this a little bit and means the user will have to be aware that all stencil renders will have to be binned in the same `transparent` or `opaque` grouping to properly guarantee render order.

> I’m also not sure what happens when you have the test just set to pass vs no test. Ie if you have just one material with some stencil write should all the others also perform an always passing stencil test? Or should you be enabling and disabling as you go through the draw list?

Can you elaborate on this a bit more? If the the stencil test is disabled then it shouldn't matter what values have been set via `setFunc` and `setOp` functions because the stencil test isn't being done at all. I think it's best to only turn on the stencil test "as-needed" by the material. I'm not sure if there are any performance differences, though.

pailhead:
>Can you elaborate on this a bit more?

Let me give it a shot, i was thinking:

```
mesh.onBeforeRender = ()=>{
  gl.enable( gl.STENCIL_TEST )
  ....
  gl_disable( gl.STENCIL_TEST)
}
```
vs just having `gl.ALWAYS` and always enabled.  In 100 meshes, we could have 30 writing and doing some kind of stencil test and 70 not doing any writes or tests.

 The difference between enabling the stencil once before all 100, and then having some just always pass (but doing the test), vs enabling / disabling 30 times. but not doing any kind of stencil test for the other 70.

gkjohnson:
I see. So basically is it better to just keep the stencil test enabled throughout the entire draw instead of turning it on and off between draws. Even if you decide to keep the stencil test enabled you'll still have to be at least changing the `stencilFunc` state when going from drawing a stenciled material to a non stenciled one. In the worst case you'll have to set the function _and_ the operation with two calls so it doesn't seem like you'd be saving a whole lot to me.

gkjohnson:
I've also updated the material and constants documentation to reflect the new stencil capabilities:

[materials/Material.html](https://raw.githack.com/gkjohnson/three.js/add-material-stencil/docs/api/en/materials/Material.html)

[constants/Materials.html](https://raw.githack.com/gkjohnson/three.js/add-material-stencil/docs/api/en/constants/Materials.html)

WestLangley:
@gkjohnson Thanks for this!

Mugen87:
I think it's also important to update `Material.toJSON()` and `Material.copy()` otherwise the change seems incomplete to me. Adding an example would also be good. As far as I can see, you already have the code for this.

gkjohnson:
@Mugen87 I agree! Added.

Regarding the example I plan to add one that includes clip planes when I get a chance.

Thanks!

Mugen87:
Okay! Many thanks for the other changes.

gkjohnson:
I haven't been able to to get around to creating an example yet but I may have time tomorrow to start. Would it be preferable to modify one of the existing examples ([webgl_clipping](https://threejs.org/examples/webgl_clipping.html) or [webgl_clipping_advanced](https://threejs.org/examples/webgl_clipping_advanced.html)) add add a toggle to enable clipped surfaces or create a new one that focuses on stenciled clip planes?

Mugen87:
I would create a new one.

gkjohnson:
I added an example with solid clip planes here (it requires a new build so the link is to a different branch):

https://raw.githack.com/gkjohnson/three.js/add-material-stencil-example/examples/webgl_clipping_stencil.html

image

It doesn't handle the case where the camera enters the geometry but I'll have to look into that another time.

WestLangley:
The example is awesome. Thanks!

I'd suggest referring to the plane constant as "constant" in the API (instead of "distance"), and as in https://threejs.org/examples/webgl_clipping_intersection.html, adding a GUI toggle to show the clipping planes. This gives the viewer a frame-of-reference.

```js
// helpers

var helpers = new THREE.Group();
helpers.add( new THREE.AxesHelper( 5 ) );
helpers.add( new THREE.PlaneHelper( planes[ 0 ], 2, 0xff0000 ) );
helpers.add( new THREE.PlaneHelper( planes[ 1 ], 2, 0x00ff00 ) );
helpers.add( new THREE.PlaneHelper( planes[ 2 ], 2, 0x0000ff ) );
helpers.visible = true;
scene.add( helpers );
```

Screen Shot 2019-03-09 at 8 00 35 PM

examples/webgl_clipping_stencil.html
```diff
@@ -0,0 +1,311 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - clipping planes</title>
```
WestLangley: `webgl - clipping stencil` -- or whatever

examples/webgl_clipping_stencil.html
```diff
@@ -0,0 +1,311 @@
...
+					po.renderOrder = i + 1.1;
+
+					object.add( stencilGroup );
+					poGroup.add( po );
+					planeObjects.push( po );
+					scene.add( poGroup );
+
+				}
+
+				// Handle camera entering model
```
WestLangley: Can you explain what is going on here? Is it important, or can you just set `controls.min/maxDistance`?

gkjohnson: Heh, an extraneous comment. I'll clamp the camera position so it can't enter the geometry, though.

gkjohnson:
@WestLangley Thanks! I've added the helpers, camera clamp, and other suggestions.

https://raw.githack.com/gkjohnson/three.js/b8146f2b6/examples/webgl_clipping_stencil.html

examples/webgl_clipping_stencil.html
WestLangley: also `maxDistance`, please.

examples/webgl_clipping_stencil.html
```diff
@@ -0,0 +1,325 @@
...
+
+			}
+
+			function init() {
+
+				clock = new THREE.Clock();
+
+				scene = new THREE.Scene();
+
+				camera = new THREE.PerspectiveCamera( 36, window.innerWidth / window.innerHeight, 1 );
```
WestLangley: Please add constructor `far` argument.

examples/webgl_clipping_stencil.html
gkjohnson: good idea!

examples/webgl_clipping_stencil.html
gkjohnson: Done

tonifds:
> I added an example with solid clip planes here (it requires a new build so the link is to a different branch):
> 
> https://raw.githack.com/gkjohnson/three.js/add-material-stencil-exam
Caps are rendered with sharp pixels. Anti-aliasing does not seem to work. Specifying antialias:true on WebGLRenderer is not enough.

Is there any way of doing this?


> I added an example with solid clip planes here (it requires a new build so the link is to a different branch):
> 
> https://raw.githack.com/gkjohnson/three.js/add-material-stencil-example/examples/webgl_clipping_stencil.html
> 
> image
> 
> It doesn't handle the case where the camera enters the geometry but I'll have to look into that another time.

Caps are rendered with sharp pixels. Anti-aliasing does not seem to work. Specifying antialias:true on WebGLRenderer is not enough.

Is there any way of doing this?

gkjohnson:
> Caps are rendered with sharp pixels. Anti-aliasing does not seem to work. Specifying antialias:true on WebGLRenderer is not enough.

I noticed this, as well. My best guess is that it's an artifact of the way that the stencil buffer and anti aliasing work and that more or less fragments affected by stencil operations can't be anti aliased in the same way as other pixels.

I'd suggest setting "antialias" to false and using any other type of AA approach to smooth edges out, including [FXAA](https://threejs.org/examples/?q=fxaa#webgl_postprocessing_fxaa) or [TAA](https://threejs.org/examples/?q=postp#webgl_postprocessing_taa). You can easily improve the edges by bumping the pixel ratio of the renderer, but that will definitely be more performance intensive.

```js
renderer.setPixelRatio( 2 );
```

image

gkjohnson:
Fixed typescript lint errors

gkjohnson:
/ping @mrdoob 

Just a gentle reminder about this PR! Let me know if there's anything I should fix or how I can make it easier to review -- I'd like to start using this in an upcoming project.

WestLangley:
@mrdoob It has been 6 months. I would like to merge this.

@Mugen87 Would you be willing to review one more time?

Mugen87:
One obvious thing: The example is outdated and needs to be written as module code.

WestLangley:
>I still wonder if we should really introduce an API for this use case.

Really? Why should `Material` support `depthFunc`, `depthTest`, `depthWrite`, and `colorWrite` and not support `stencilWrite`?

gkjohnson:
@Mugen87 

I've modified the the example to use modules.

> I still wonder if we should really introduce an API for this use case

Of course I respect the right of project maintainers to change their minds but as it is I've put quite a bit of effort into writing docs, making an example, and updating the code at the encouragement and suggestions of project collaborators (and being added to a project milestone) so I hope you understand if I feel a little miffed at the idea that this suddenly might not be something that the project is interested in merging. I would definitely appreciate a little more insight into what the right way is to use stencils if this is the case or what the vision for supporting them is because it feels like hacking around the library to get them to work at the moment. Or at least it's not clear why the use case of drawing to the stencil buffer should not be worthy of proper support.

I use three.js professionally and always enjoy giving back to the project but it's been discouraging over the last several months when it occasionally feels like making a PR is like throwing them into a void. I'm left wondering what the right way is to follow the process here? I want to give back to the project but I feel like I'm not sure how to effectively do so. So I suppose I'm asking for help to understand how I can best give back to the project through PRs and contribute changes that I need for my professional work. I know there are a lot of PRs and issues that come through here but 6 months seems like an unreasonable amount of time with no insight into what the plan is.

I hope this comes across more as trying to help than it does frustrated. I really like the community here and I feel I've learned a lot through this project so I just want to understand how to collaborate more effectively. What can I do to help make this easier / faster for everyone?

Thanks!

Mugen87:
> Really? Why should Material support depthFunc, depthTest, depthWrite, and colorWrite and not support stencilWrite?

I personally think that performing stencil operations in this way are quite complicated. Hence, such a low level API won't be beneficial for most users. 

Enabling/disabling e.g. the depth test is something different than this.

gkjohnson:
> I personally think that performing stencil operations in this way are quite complicated. Hence, such a low level API won't be beneficial for most users. Enabling/disabling e.g. the depth test is something different than this.

I think supporting both high level and low level APIs is important considering the low level APIs are what's needed to develop complex, specialized rendering techniques and build up abstractions and extensions. I think the blend constants are a more comparable example: while individual control over all blend function constants may be too much for most users it's provided as an option for advanced cases.

> If you need to wait that long for feedback, the feature probably does not convince everyone. In any event, I leave the decision up to @mrdoob.

Understood. It would just be nice to have a little more insight into what the concerns are but I understand it's not an easy problem and time is limited.

WestLangley:
Many thanks!

src/renderers/webgl/WebGLState.js
```diff
@@ -677,6 +677,15 @@ function WebGLState( gl, extensions, utils, capabilities ) {
 		depthBuffer.setMask( material.depthWrite );
 		colorBuffer.setMask( material.colorWrite );
 
+		var stencilWrite = material.stencilWrite;
+		stencilBuffer.setTest( stencilWrite );
```
WestLangley: It seems that this has broken the Masking examples in r.107. It is overwriting the stencil test independently set by the app.

A quick fix is to comment the line out.

/ping @gkjohnson

Mugen87: E.g.: https://threejs.org/examples/webgl_postprocessing_masking

Mugen87: Instead of just commenting out code and I vote to revert the PR for now. It can be merged again if existing code does not break.

gkjohnson: Good catch -- I didn't realize other code was using the stencil state considering it's not documented.

I've made two PRs: one that comments out the code as @WestLangley suggested (#17137) and one that fixes MaskPass (#17136) so it works with the new stencil materials.

My concern with just commenting out the code is that the documentation will still reflect the new stencil fields and depending on how long we expect this to be removed (until 108?) it could be confusing.

WestLangley: >A quick fix is to comment the line out.

The suggestion to comment the line out was only intended to verify the offending line of code. I would not consider that as a hot fix.

makc:
I thought I would ask in this thread, since all the relevant people are here and I do not remember my stack overflow password... so I was trying to do fake CSG with this feature, following three.js clipping planes example and https://github.com/AlexGreulich/HRTFVR/blob/master/include/glut/progs/advanced/csg.c ...so far Ive got this http://makc.github.io/three.js/csg/ but I kind of feel it is too hard to set up. Am I missing something? Could this be simplified? Please advise, oh great stencil masters 🙏

makc:
p.s. my main issue with this is how rigid the setup is. I can not use the scene graph for anything other than transformations, i.e. to switch between different csg operations like in C demo I would have to create and swap multiple scenes. same when I need to replace one of the object materials. less than ideal, if I may say.

PR #15615, opened 2019-01-19
Slice AnimationClip

Fyrestar:
To create a new clip from a existing from-to frame with optional new name. Some tools only export animations as a single clip.

### DIFF
--- a/src/animation/AnimationClip.js
+++ b/src/animation/AnimationClip.js
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
 		return this;
 
 	},
+	
+	slice: function(from, to, name) {
+
+            var frames = to - from, duration = 0;
+
+            var tracks = [];
+
+            for ( var i = 0, l = this.tracks.length; i < l; i ++ ) {
+
+                var track = this.tracks[i];
+
+                var stride = track.values.length / track.times.length;
+
+
+                var times = new track.TimeBufferType(frames);
+                var values = new track.ValueBufferType(frames * stride);
+
+                var start = track.times[from];
+
+                for ( var t = 0; t < frames; t ++ ) {
+
+                    times[t] = track.times[ from + t ] - start;
+
+                    for ( var k = 0; k < stride; k ++ )
+                        values[t * stride + k] = track.values[ ( from + t ) * stride + k ];
+
+                }
+
+                duration = Math.max(times[ to ], duration);
+
+                tracks.push(
+                    new track.constructor( track.name, times, values, track.createInterpolant )
+                );
+
+            }
+
+            return new AnimationClip( name || this.name, duration, tracks );
+
+        },
 
 	validate: function () {
 


src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
...
+
+            var frames = to - from, duration = 0;
+
+            var tracks = [];
+
+            for ( var i = 0, l = this.tracks.length; i < l; i ++ ) {
+
+                var track = this.tracks[i];
+
+                var stride = track.values.length / track.times.length;
```
donmccurdy: Would use `track.getValueSize()` here.

src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
...
+                    for ( var k = 0; k < stride; k ++ )
+                        values[t * stride + k] = track.values[ ( from + t ) * stride + k ];
+
+                }
+
+                duration = Math.max(times[ to ], duration);
+
+                tracks.push(
+                    new track.constructor( track.name, times, values, track.createInterpolant )
+                );
```
donmccurdy: It's possible that some of the tracks won't have any keyframes in the sliced range, so it might be best to keep empty tracks out of the new clip.

src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
...
+                for ( var t = 0; t < frames; t ++ ) {
+
+                    times[t] = track.times[ from + t ] - start;
+
+                    for ( var k = 0; k < stride; k ++ )
+                        values[t * stride + k] = track.values[ ( from + t ) * stride + k ];
+
+                }
+
+                duration = Math.max(times[ to ], duration);
```
donmccurdy: Minor: Duration is computed automatically, so it might be simpler just to pass it `undefined` in the constructor.

donmccurdy (commented):
Related: #13430

I'm OK with merging either of the PRs; mine would need a rebase anyway. Do you have an example model with multiple subclips, to test that this works as expected and gives smoothly-looping clips? There are some edge cases to be aware of — like:

```
-------------clip1-------------↓-------------clip2-------------↓-------------clip3-------------|
track1 ----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|---|
track2----|------------------------------------------------------------------------------------|
track3----|----|----|----|----|-------------------------------------|----|----|----|----|----|-|
```

... where track2 and track3 actually do have values in clip2, but no keyframes. For a correctly authored animation that wouldn't be a problem, but automatic optimizations at export or import can mess things up...

src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
 		return this;
 
 	},
+	
+	slice: function(from, to, name) {
+
+            var frames = to - from, duration = 0;
```
takahirox: Code style: We use hard tabs, not soft tabs.

src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
 		return this;
 
 	},
+	
+	slice: function(from, to, name) {
```
takahirox: Code style: Extra spaces are necessary after opening bracket and before closing bracket `slice: function( from, to, name ) {`

https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style%E2%84%A2#blocks

src/animation/AnimationClip.js
```diff
@@ -420,6 +420,45 @@ Object.assign( AnimationClip.prototype, {
...
+
+            var frames = to - from, duration = 0;
+
+            var tracks = [];
+
+            for ( var i = 0, l = this.tracks.length; i < l; i ++ ) {
+
+                var track = this.tracks[i];
```
takahirox: Code style: Extra spaces are necessary after opening bracket and before closing bracket.

https://github.com/mrdoob/three.js/wiki/Mr.doob's-Code-Style%E2%84%A2#arrays

Fyrestar:
When i export animations with optimizations it will basically remove the start/end keyframe of the desired sequence of a track when smoothing it out. I'll change it to compute the values at these times,

Mugen87:
Thanks for the work here but I think it's better to focus on #13430 (which is already tested and approved).

PR #15622, opened 2019-01-21
 Set if the bottom and top should have a lid 

thuijzer:
Sometimes you don't want a bottom or top lid/cap.
Added two optional boolean options: bottomLid and topLid so you can turn off the lid.

### DIFF
--- a/src/geometries/ExtrudeGeometry.js
+++ b/src/geometries/ExtrudeGeometry.js
@@ -558,13 +558,15 @@ function ExtrudeBufferGeometry( shapes, options ) {
 
 				// Bottom faces
 
-				if(bottomLid) {
+				if ( bottomLid ) {
+
 					for ( i = 0; i < flen; i ++ ) {
 
 						face = faces[ i ];
 						f3( face[ 2 ] + offset, face[ 1 ] + offset, face[ 0 ] + offset );
 
 					}
+
 				}
 
 				layer = steps + bevelSegments * 2;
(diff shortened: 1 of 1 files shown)


src/geometries/ExtrudeGeometry.js
```diff
@@ -565,36 +574,51 @@ function ExtrudeBufferGeometry( shapes, options ) {
 
 				// Top faces
 
-				for ( i = 0; i < flen; i ++ ) {
+				if ( topLid ) {
```
WestLangley: formatting: add blank line

If this merged, please also update the docs.

examples/webgl_geometry_spline_editor.html
```diff
@@ -151,7 +151,7 @@
 
 				controls.addEventListener( 'start', function () {
 
-					cancelHideTransorm();
+					cancelHideTransform();
```
WestLangley: Shouldn't the changes in this file be in a separate PR?

Mugen87:
Um, `CylinderBufferGeometry` has `openEnded` which defines whether the top/bottom segments of the cylinder are opened or not. I would prefer a consistent approach and not something special for each geometry generator.

WestLangley:
>I would prefer a consistent approach and not something special for each geometry generator.

This is about `ExtrudeGeometry`, not `CylinderGeometry`. Boxes also have tops and bottoms. I would not worry about it.

You can add the same options to `CylinderGeometry` if you want. I prefer `options` to a long list of args, anyway: it's much easier to change the API.

Plus, this new feature is hidden in the `options` and is backwards compatible.

Mugen87:
Closing for now. Let's keep `ExtrudeGeometry` as it is since the class is already the most complex geometry generator. If @mrdoob thinks it's still worth, we can reopen and merge the PR. In this case, I'll update the docs and TS file (unless the OP does this in the meanwhile).

Issue #7489, opened 2015-10-29
Specular Environment BRDF is opaque


bhouston:
Right now the indirect specular BRDF, aka the environmental BRDF, is opaque and can not be modified.  This is limiting our ability to improve our lighting model because we can not properly modify our responses to indirect specular lighting.

Given a normal microfacet BRDF model of D \* G \* F, I believe that the lod mip maps encode the D model for the most part, thus we should just be left with G and F in the indirect specular BRDF.  I believe that the environmental BRDF also includes other terms from the render equation such as dotNL (which is equivalent to dotNV in the case of an envMap reflection.)

Thus I believe that the Environment BRDF is something like `dotNL * G * F`.

WestLangley:
[This post](http://brabl.com/tag/ibl/) and the embedded links may be helpful in understanding what is going on.

WestLangley:
Also [this](http://blog.selfshadow.com/publications/s2013-shading-course/lazarov/s2013_pbs_black_ops_2_slides_v2.pdf).

tschw:
@bhouston 
> lod mip maps encode the D model for the most part [ ... ]. 

Plus stuff we have in the rendering equation (or the *visibility term* V, as used in some papers). Most notably normalization factors (`1 / PI` weighting) that stem from the hemispherical integral are already in the data.

> Thus I believe that the Environment BRDF is something like `dotNL * G * F`

That's basically how I understand it too. I don't believe we need the `dotNL` part (at least not at this level, it might be *inside* of G), but I might be wrong and a careful re-read the SIGGRAPH stuff will probably clarify it. 

BTW I'd always think of `dotNL` it as `dotNV` in this context;  cosine of reflection angle / angle of emission rather than incidence.

For transparent surfaces, we have to consider the critical angle where we get [total internal reflection](https://en.wikipedia.org/wiki/Total_internal_reflection) in the Fresnel term (see [these figures](https://upload.wikimedia.org/wikipedia/commons/thumb/a/a4/Fresnel_reflection.svg/1200px-Fresnel_reflection.svg.png)).

A simple way to implement it is to scale the Schlick-style approximation to the appropriate range (since we're looking at cosines the angle axis is flipped):

```glsl
float eta = material.refractionIndex / trace.refractionIndex;
float cosCritical = sqrt(max(0., 1. - sqr(eta)));
cosSchlick = (cosReflect - cosCritical) / (1. - cosCritical);
```
[...]

PR #15644, opened 2019-01-26
Restore energy loss in indirect (IBL) specular lighting via multiscattering approximation

jsantell:
Currently, for metal surfaces (MeshPhysicsMaterial, MeshStandardMaterial), as the roughness of the material increases, the surface darkens if lit via IBL+PMREM (possibly #7489?). For fully metal surfaces (conductors), all incoming irradiance should equal outgoing radiance (F0=1) -- like a mirror (metalness=1, roughness=0). Even if the metal is rough (metalness=1, roughness=1), the same amount of energy should be reflected off of the material.

There is the [white furnace test](https://google.github.io/filament/Filament.md.html#toc4.7.2), which has a uniformly lit enironment with fully metal, white spheres with varying degrees of roughness. As these surfaces are fully metal, all incoming light must be reflected, and should be indistinguishable from the environment. If the spheres are visible, then some energy is either lost or gained.

Currently, energy is lost as metal surfaces increase in roughness:

furnace-before

BRDFs only account for a single scattering of light -- for rough, metal surfaces, light will bounce around microfacets after the initial scattering before reaching the viewer, which isn't accounted for in BRDF models.

diagram_single_vs_multi_scatter
via [Filament's beautiful docs](https://google.github.io/filament/Filament.md.html)

There are several sampling and preprocessing methods that aid in real-time approximation of multiscattering light, but ultimately, I was able to get Fdez-Agüera's method ([A Multiple-Scattering Microfacet
Model for Real-Time Image-based Lighting](http://jcgt.org/published/0008/01/03/)) working well, and used a [DFG approximation from Filament](https://google.github.io/filament/Filament.md.html#lighting/imagebasedlights/distantlightprobes) instead of a DFG LUT .

comparison-beforeaftersmall

Above, on the left is r100, on the right is this branch -- from top to bottom, metal surface with increasing roughness, rough surface with increasing metalness, metal surface with increasing roughness, nonmetal increasing rough surface). Much more energy is retained on rough, metal objects with this change.

specular-multiscatter-pieces-small

Above is the single-scattering specular indirect lighting only (left), only the multiscattering (center), and both (combined). Currently in r100, only the single-scattering is visible.

furnace-after

Perhaps the least exciting image, above is the furnace test with these changes. All incoming light equals the outgoing light, making the spheres invisible.

Now a handful of questions:

* Should this be MeshPhysicalMaterial only?
* What should the conditions for this be? Currently, it's `defined( USE_ENVMAP ) && defined( ENVMAP_TYPE_CUBE_UV )`, but mostly I'm checking to see if a radiance map exists.
* Needs a good correctness check
* Performance check (seems good, but needs a rigorous test)
* Is the DFG approximation valid?
* Do we need to separate the indirect specular and diffuse lighting functions? They're combined here to save computations.
* Maybe this should only be on desktop, depending on perf

### DIFF
--- a/src/renderers/shaders/ShaderChunk/bsdfs.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/bsdfs.glsl.js
@@ -1,4 +1,22 @@
-export default /* glsl */`
+export default /* glsl */ `
+
+// Analytical approximation of the DFG LUT, one half of the
+// split-sum approximation used in indirect specular lighting.
+// via 'environmentBRDF' from "Physically Based Shading on Mobile"
+// https://www.unrealengine.com/blog/physically-based-shading-on-mobile - environmentBRDF for GGX on mobile
+vec2 integrateSpecularBRDF( const in float dotNV, const in float roughness ) {
+	const vec4 c0 = vec4( - 1, - 0.0275, - 0.572, 0.022 );
+
+	const vec4 c1 = vec4( 1, 0.0425, 1.04, - 0.04 );
+
+	vec4 r = roughness * c0 + c1;
+
+	float a004 = min( r.x * r.x, exp2( - 9.28 * dotNV ) ) * r.x + r.y;
+
+	return vec2( -1.04, 1.04 ) * a004 + r.zw;
+
+}
+
 float punctualLightIntensityToIrradianceFactor( const in float lightDistance, const in float cutoffDistance, const in float decayExponent ) {
 
 #if defined ( PHYSICALLY_CORRECT_LIGHTS )
@@ -239,20 +257,35 @@ vec3 BRDF_Specular_GGX_Environment( const in GeometricContext geometry, const in
 
 	float dotNV = saturate( dot( geometry.normal, geometry.viewDir ) );
 
-	const vec4 c0 = vec4( - 1, - 0.0275, - 0.572, 0.022 );
+	vec2 brdf = integrateSpecularBRDF( dotNV, roughness );
 
-	const vec4 c1 = vec4( 1, 0.0425, 1.04, - 0.04 );
-
-	vec4 r = roughness * c0 + c1;
-
-	float a004 = min( r.x * r.x, exp2( - 9.28 * dotNV ) ) * r.x + r.y;
-
-	vec2 AB = vec2( -1.04, 1.04 ) * a004 + r.zw;
-
-	return specularColor * AB.x + AB.y;
+	return specularColor * brdf.x + brdf.y;
 
 } // validated
 
+// Fdez-Agüera's "Multiple-Scattering Microfacet Model for Real-Time Image Based Lighting"
+// Approximates multiscattering in order to preserve energy.
+// http://www.jcgt.org/published/0008/01/03/
+void BRDF_Specular_Multiscattering_Environment( const in GeometricContext geometry, const in vec3 specularColor, const in float roughness, inout vec3 singleScatter, inout vec3 multiScatter ) {
+
+	float dotNV = saturate( dot( geometry.normal, geometry.viewDir ) );
+
+	vec3 F = F_Schlick( specularColor, dotNV );
+	vec2 brdf = integrateSpecularBRDF( dotNV, roughness );
+	vec3 FssEss = F * brdf.x + brdf.y;
+
+	float Ess = brdf.x + brdf.y;
+	float Ems = 1.0 - Ess;
+
+	// Paper incorrect indicates coefficient is PI/21, and will
+	// be corrected to 1/21 in future updates.
+	vec3 Favg = specularColor + ( 1.0 - specularColor ) * 0.047619; // 1/21
+	vec3 Fms = FssEss * Favg / ( 1.0 - Ems * Favg );
+
+	singleScatter += FssEss;
+	multiScatter += Fms * Ems;
+
+}
 
 float G_BlinnPhong_Implicit( /* const in float dotNL, const in float dotNV */ ) {
 
--- a/src/renderers/shaders/ShaderChunk/lights_physical_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/lights_physical_pars_fragment.glsl.js
@@ -96,11 +96,17 @@ void RE_Direct_Physical( const in IncidentLight directLight, const in GeometricC
 
 void RE_IndirectDiffuse_Physical( const in vec3 irradiance, const in GeometricContext geometry, const in PhysicalMaterial material, inout ReflectedLight reflectedLight ) {
 
-	reflectedLight.indirectDiffuse += irradiance * BRDF_Diffuse_Lambert( material.diffuseColor );
+	// Defer to the IndirectSpecular function to compute
+	// the indirectDiffuse if energy preservation is enabled.
+	#ifndef ENVMAP_TYPE_CUBE_UV
+
+		reflectedLight.indirectDiffuse += irradiance * BRDF_Diffuse_Lambert( material.diffuseColor );
+
+	#endif
 
 }
 
-void RE_IndirectSpecular_Physical( const in vec3 radiance, const in vec3 clearCoatRadiance, const in GeometricContext geometry, const in PhysicalMaterial material, inout ReflectedLight reflectedLight ) {
+void RE_IndirectSpecular_Physical( const in vec3 radiance, const in vec3 irradiance, const in vec3 clearCoatRadiance, const in GeometricContext geometry, const in PhysicalMaterial material, inout ReflectedLight reflectedLight) {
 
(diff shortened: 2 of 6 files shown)


mrdoob:
@bhouston @spidersharma03 do these changes look good to you?

jsantell:
Some perf testing, it looks like this might be slighty slightly less performant than the current implementation, but practically the same. Haven't given too much of an eye to optimization yet. It'd be helpful if someone could check it out on iOS.

Test case, 64 objects mostly filling the screen

* https://indirect-perf.glitch.me/ (for testing PR)
* https://indirect-perf.glitch.me/?r100 (for testing current r100 build)

Mobile Results

device | browser | current FPS | proposed FPS |
----- | --- | --- | --- |
Pixel 2 | Chrome r71 | ~59 | ~59
Pixel 2 | Firefox v64 | ~57 | ~57
Pixel 1 | Chrome r71 | ~39 | ~38
Pixel 1 | Firefox v64 | ~47 | ~46

Linux Laptop, 800x800

browser | DPR | current FPS | proposed FPS |
----- | --- | --- | --- |
Firefox v64 | 1 | ~58 | ~58
Firefox v64 | 2 |  ~32 | ~30
Firefox v64 | 3 |  ~19 | ~19
Chromium v71 | 2 | ~44 | ~43

> If you want to refactor the shader code, that would be fine with me. I find the `#define RE_Direct` chunks, etc., somewhat confusing.

Sounds great, I can take care in a follow up

donmccurdy:
iPhone SE, Safari: 33fps before and after.

bhouston:
Is there any way this could be precalculated into the ibl map, even in an
approximate fashion just to maintain energy?

jsantell:
Updated PR with `preserveEnergy` boolean on MeshPhysicalMaterial that `#define PRESERVE_ENERGY`'s when set with a PMREM so it could possibly be more opt in. Also consolidated the integrated BRDF generation code which was mostly what the current indirect specular was generating.

Also as a quick double check, this is what the DFG approximation looks like generated in `integrateBRDF`:

localhost_8080_examples_texture html 5

mrdoob:
Shouldn't `preserveEnergy` be `true` by default? The improved output seems worth the slight loss of fps.

WestLangley:
>Shouldn't preserveEnergy be true by default? The improved output seems worth the slight loss of fps.

I'd hardwire it, then, and remove the `preserveEnergy` flag.

jsantell:
> Shouldn't `preserveEnergy` be `true` by default? The improved output seems worth the slight loss of fps.

Wasn't sure about the (small) performance delta, and figured this would be the best way for opt in -- happy to remove the flag if should be enabled for all MeshPhysicalMaterials with PMREMs -- which maybe MeshPhysicalMaterial over MeshStandardMaterial is sufficient opt-in for potentially higher cost for more accurate renderings

mrdoob:
I think this should apply to both `MeshStandardMaterial` and `MeshPhysicalMaterial` and be enabled by default. Good looking by default is best. If people need extra ms they could opt-out.

jsantell:
> I think this should apply to both `MeshStandardMaterial` and `MeshPhysicalMaterial` and be enabled by default. Good looking by default is best. If people need extra ms they could opt-out.

SGTM -- should it continue being a boolean for folks to disable?

mrdoob:
Is there any other reason for that boolean apart from a few extra ms? If not, I would remove the boolean altogether (and avoid complexity).

src/renderers/shaders/ShaderChunk/bsdfs.glsl.js
```diff
@@ -239,20 +257,41 @@ vec3 BRDF_Specular_GGX_Environment( const in GeometricContext geometry, const in
...
+} // validated
 
-	float a004 = min( r.x * r.x, exp2( - 9.28 * dotNV ) ) * r.x + r.y;
+// Fdez-Agüera's "Multiple-Scattering Microfacet Model for Real-Time Image Based Lighting"
+// Approximates multiscattering in order to preserve energy.
+// http://www.jcgt.org/published/0008/01/03/
+void BRDF_Specular_Multiscattering_Environment( const in GeometricContext geometry, const in vec3 specularColor, const in float roughness, inout vec3 singleScatter, inout vec3 multiScatter ) {
 
-	vec2 AB = vec2( -1.04, 1.04 ) * a004 + r.zw;
+  float dotNV = saturate( dot( geometry.normal, geometry.viewDir ) );
```
WestLangley: Formatting is messed up... It appears you have introduced spaces for tabs in this PR. Please see if you can fix that.

src/renderers/shaders/ShaderChunk/lights_physical_pars_fragment.glsl.js
```diff
@@ -110,14 +114,35 @@ void RE_IndirectSpecular_Physical( const in vec3 radiance, const in vec3 clearCo
...
 	#endif
 
-	reflectedLight.indirectSpecular += ( 1.0 - clearCoatDHR ) * radiance * BRDF_Specular_GGX_Environment( geometry, material.specularColor, material.specularRoughness );
+  float clearCoatInv = 1.0 - clearCoatDHR;
+
+  // Both indirect specular and diffuse light accumulate here
+  // if energy preservation enabled, and PMREM provided.
+  #if defined( ENVMAP_TYPE_CUBE_UV )
+
+    vec3 singleScattering = vec3(0.0);
```
WestLangley: minor nit... white space code style. multiple instances in this PR. I know it takes awhile to get used to. :)

jsantell: Oops, looks like `./node_modules/eslint/bin/eslint.js --fix [file]` didn't correct the string glsl which makes sense now -- will fix :+1:

jsantell:
Updated the styling of the glsl strings :+1:

mrdoob:
Thanks!

mrdoob:
The `webgl_materials_standard` looks definitely much better.

Before:
screen shot 2019-01-29 at 9 17 43 pm
https://raw.githack.com/mrdoob/three.js/r100/examples/webgl_materials_standard.html

After:
screen shot 2019-01-29 at 9 16 48 pm
https://raw.githack.com/mrdoob/three.js/dev/examples/webgl_materials_standard.html

However, `webgl_materials_envmaps_exr` and `webgl_materials_envmaps_hdr` now have some sort of dark fresnel?

Before:
screen shot 2019-01-29 at 9 30 21 pm
https://raw.githack.com/mrdoob/three.js/r100/examples/webgl_materials_envmaps_exr

After:
screen shot 2019-01-29 at 9 30 04 pm
https://raw.githack.com/mrdoob/three.js/dev/examples/webgl_materials_envmaps_exr

Before:
screen shot 2019-01-29 at 9 30 46 pm
https://raw.githack.com/mrdoob/three.js/r100/examples/webgl_materials_envmaps_hdr

After:
screen shot 2019-01-29 at 9 31 05 pm
https://raw.githack.com/mrdoob/three.js/dev/examples/webgl_materials_envmaps_hdr

jsantell:
@mrdoob noted, will take a look!

WestLangley:
>`webgl_materials_standard` looks definitely much better.

What exactly are you seeing? I see no visible difference. Maybe I am missing something...

I have only seen significant visible difference in real-world examples (that is, non-furnace tests) for rough-metallic surfaces. Maybe @jsantell has seen otherwise...

jsantell:
> > `webgl_materials_standard` looks definitely much better.
> 
> What exactly are you seeing? I see no visible difference. Maybe I am missing something...
> 
> I have only seen significant visible difference in real-world examples (that is, non-furnace tests) for rough-metallic surfaces. Maybe @jsantell has seen otherwise...

The only lighting that's changed is an extra specular lobe for metal surfaces so the change is most noticeable on very rough metallic surfaces, where previously the specular light was lost

jsantell:
Potential fix for the dark fresnel ring: https://github.com/mrdoob/three.js/pull/15673

Mugen87:
@jsantell I think we should note in the [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide) that the rendering of `MeshPhysicsMaterial`/ `MeshStandardMaterial` slightly changed. How would you phrase the most important aspects of the PR?

jsantell:
@mugen87 to take a crack at it, "MeshStandardMaterial and MeshPhysicalMaterial now preserve energy for IBL lighting, resulting in brighter, more accurate colors for metallic materials with high roughness values when lit via a map generated by PMREMGenerator."

Mugen87:
Perfect! I've added your suggestion to the guide.

mrdoob:
@looeee 

> The difference is extremely subtle.

I guess my eyes can see extremely subtle things at this point 😅

PR #15647, opened 2019-01-26
ObjectLoader: fix duplicate materials

### LINKED ISSUE #15635: ObjectLoader still creates multiple materials with same uuid
looks like [this](https://github.com/mrdoob/three.js/issues/15152) is still a problem in one case.

So to remind. In old format with MultiMaterials, ObjectLoader may create multiple materials with the same uuid. After [this PR](https://github.com/mrdoob/three.js/pull/15128) Something like

```
MultiMaterialA = [X, Y]
MultiMaterialB = [X, Z]
```

no longer triggers that problem

but this does

```
Material = X
MultiMaterial = [X, Y]
```

mixtur:
Fixes #15635

### DIFF
--- a/src/loaders/ObjectLoader.js
+++ b/src/loaders/ObjectLoader.js
@@ -497,8 +497,13 @@ Object.assign( ObjectLoader.prototype, {
 
 				} else {
 
-					materials[ data.uuid ] = loader.parse( data );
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #15650, opened 2019-01-27
FirstPersonControls: Added .lookAt()

### LINKED ISSUE #15629: FirstPersonControls should repect intial object rotation
Consider this to be a feature request, I guess? Both #3010 and #6551 were closed as help, and [all](https://stackoverflow.com/questions/14638135/firstpersoncontrols-start-by-looking-at-an-object) [of](https://stackoverflow.com/questions/13474434/three-js-first-person-controle-makes-objects-disappears-from-scene) [relevant](https://stackoverflow.com/questions/24819945/three-js-loaded-model-is-disappearing-while-switch-to-firstperson-control-from-o/25045425#25045425) [SO](https://stackoverflow.com/questions/20905101/three-js-firstperson-control-start-orientation/20906487#20906487) [questions](https://stackoverflow.com/questions/14638135/firstpersoncontrols-start-by-looking-at-an-object/14638753#14638753) are either unanswered, or the answer is "set lat and lon".

So why dont you set lat and lon to the values corresponding to object.rotation rather than to 0 in FirstPersonControls constructor?

Mugen87:
Fixed #15629

### DIFF
--- a/examples/js/controls/FirstPersonControls.js
+++ b/examples/js/controls/FirstPersonControls.js
@@ -34,9 +34,6 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 	this.mouseX = 0;
 	this.mouseY = 0;
 
-	this.lat = 0;
-	this.lon = 0;
-
 	this.moveForward = false;
 	this.moveBackward = false;
 	this.moveLeft = false;
@@ -46,6 +43,17 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 
 	this.viewHalfX = 0;
 	this.viewHalfY = 0;
+
+	// private variables
+
+	var lat = 0;
+	var lon = 0;
+
+	var lookDirection = new THREE.Vector3();
+	var spherical = new THREE.Spherical();
+	var target = new THREE.Vector3();
+
+	//
 
 	if ( this.domElement !== document ) {
 
@@ -181,6 +189,26 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 
 	};
 
+	this.lookAt = function ( x, y, z ) {
+
+		if ( x.isVector3 ) {
+
+			target.copy( x );
+
+		} else {
+
+			target.set( x, y, z );
+
+		}
+
+		this.object.lookAt( target );
+
+		setOrientation( this );
+
+		return this;
+
+	};
+
 	this.update = function () {
 
 		var targetPosition = new THREE.Vector3();
@@ -229,13 +257,13 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 
 			}
 
-			this.lon -= this.mouseX * actualLookSpeed;
-			if ( this.lookVertical ) this.lat -= this.mouseY * actualLookSpeed * verticalLookRatio;
+			lon -= this.mouseX * actualLookSpeed;
+			if ( this.lookVertical ) lat -= this.mouseY * actualLookSpeed * verticalLookRatio;
 
-			this.lat = Math.max( - 85, Math.min( 85, this.lat ) );
+			lat = Math.max( - 85, Math.min( 85, lat ) );
 
-			var phi = THREE.Math.degToRad( 90 - this.lat );
-			var theta = THREE.Math.degToRad( this.lon );
+			var phi = THREE.Math.degToRad( 90 - lat );
+			var theta = THREE.Math.degToRad( lon );
 
 			if ( this.constrainVertical ) {
 
@@ -295,6 +323,20 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 
 	}
 
+	function setOrientation( controls ) {
+
+		var quaternion = controls.object.quaternion;
+
+		lookDirection.set( 0, 0, - 1 ).applyQuaternion( quaternion );
+		spherical.setFromVector3( lookDirection );
+
+		lat = 90 - THREE.Math.radToDeg( spherical.phi );
+		lon = THREE.Math.radToDeg( spherical.theta );
+
+	}
+
 	this.handleResize();
+
+	setOrientation( this );
 
 };
--- a/examples/webgl_nearestneighbour.html
+++ b/examples/webgl_nearestneighbour.html
@@ -90,6 +90,9 @@
 				controls.movementSpeed = 100;
 				controls.lookSpeed = 0.1;
 
+				controls.lookAt( 500, 500, 500 );
+
+
 				// add a skybox background
 				var cubeTextureLoader = new THREE.CubeTextureLoader();
 
--- a/examples/webgl_shadowmap.html
+++ b/examples/webgl_shadowmap.html
@@ -85,10 +85,16 @@
 				container = document.createElement( 'div' );
 				document.body.appendChild( container );
 
-				// SCENE CAMERA
+				// CAMERA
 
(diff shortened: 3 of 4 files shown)


looeee:
Wow, this is literally the exact thing that's been bugging me for the last half hour. Perfect timing 😀

Mugen87:
Wonderful! 👍

examples/js/controls/FirstPersonControls.js
```diff
@@ -30,6 +30,7 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 	this.verticalMax = Math.PI;
 
 	this.autoSpeedFactor = 0.0;
+	this.forward = new THREE.Vector3( 0, 0, - 1 );
```
WestLangley: What happens if the user changes this?

WestLangley:
Sorry, I do not see how this can work. It is also not consistent with the three.js API.

Are you saying you can't call `setOrientation()` without calling `camera.lookAt()` first?

Is `camera.lookAt()` and `setOrientation()` supposed to be called only once prior to initializing the controls?

I hope that is the intention -- users should not be setting the camera behind the back of the controls.

The proper solution is to make `setOrientation()` private, and to set the camera orientation in the controls constructor.

examples/js/controls/FirstPersonControls.js
Mugen87: Then the forward vector is differently. This is necessary if the user passes a ordinary `Object3D` to the controls.

Mugen87:
> Are you saying you can't call setOrientation() without calling camera.lookAt() first?

No, it's just an example.

> Is camera.lookAt() and setOrientation() supposed to be called only once prior to initializing the controls?

You can also call if after initializing the controls.

> Sorry, I do not see how this can work. It is also not consistent with the three.js API.

It worked perfectly on my computer. I think the stackoverflow post show that there is a demand for action. So your attitude seems a bit counterproductive to me.

examples/js/controls/FirstPersonControls.js
makc: if used to control other object, it becomes ThirdPersonControls :)

makc:
@WestLangley they suggested the method rather than inline code because someone might want to change the camera direction after the controls were activated.

examples/js/controls/FirstPersonControls.js
looeee: > I believe FirstPersonControls is designed to control a camera

Why put a limit on what it can control, if it's as simple as exposing `this.forward`?

looeee:
I've had a chance to test this. It's an improvement IMO, but what I would prefer (and what I believe the other controls do) is to respect the camera's orientation and position at the time when the controls are instantiated. 

If I do: 

```js
camera.lookAt( someObject.position );

const controls = new FirstPersonControls( camera );
```

The current behaviour is that the camera jumps to a new and unexpected orientation. 
With this change, I can now do `controls.setOrientation( camera.quaternion )` to fix this, which is an improvement. But I'd rather that just happens automatically, which again, I think is the case for other controls.

examples/js/controls/FirstPersonControls.js
looeee: If we do decide to impose this limit then we should change this line:

``` js 
THREE.FirstPersonControls = function ( object, domElement ) 
```
to
``` js
THREE.FirstPersonControls = function ( camera, domElement ) 
```

And probably also for other controls too, because `object` suggests that any `Object3D` can be used.

Mugen87: @looeee Absolutely!

Mugen87:
`FirstPersonControls.lookAt()` is a good idea! I've changed to PR accordingly and also ensured that @looeee suggestion is implemented. This works because `setOrientation()` is now a private function and called when the controls are instantiated.

Mugen87:
`FirstPersonControls.lookAt()` sounds good. I have changed the PR and also implemented @looeee's suggestion. `setOrientation()` is now a private function and called in the ctor. This means it's now possible to respect the original position of the camera.

Mugen87:
I'll make a note in the migration guide about the changes related to `FirstPersonControls`.

Mugen87:
BTW: I have noticed that `FirstPersonControls.constrainVertical` has somewhat a confusing effect when using `.lookAt()`. It alters the value of `phi` and thus the result of `.lookAt()`. I mean it produces the expected result but users should aware about this fact when turning `constrainVertical` to `true`.

examples/webgl_nearestneighbour.html
```diff
@@ -90,6 +90,9 @@
 				controls.movementSpeed = 100;
 				controls.lookSpeed = 0.1;
 
+				controls.lookAt( new THREE.Vector3( 500, 500, 500 ) );
```
WestLangley: For consistency, I think this API should be supported, too:
```javascript
controls.lookAt( 500, 500, 500 );
```

examples/js/controls/FirstPersonControls.js
```diff
@@ -295,6 +313,20 @@ THREE.FirstPersonControls = function ( object, domElement ) {
 
 	}
 
+	function setOrientation( controls ) {
+
+		var quaternion = controls.object.quaternion;
+
+		lookDirection.copy( controls.forward ).applyQuaternion( quaternion );
```
WestLangley: As I said previously, 
```javascript
lookDirection.set( 0, 0, - 1 ).applyQuaternion( quaternion );
```
This control only supports cameras.

examples/js/controls/FirstPersonControls.js
makc: @looeee @Mugen87 I think the proper question is why the hell forward needs to be different for camera vs other objects? Why can't it just be 0,0,-1 for everything?

examples/js/controls/FirstPersonControls.js
Mugen87: I have tested it and there are some limitations, yes. 

Demo: https://jsfiddle.net/q50o1r97/1/

It would be possible with a few modification, though. But for now, I hardwire the forward vector.

examples/js/controls/FirstPersonControls.js
Mugen87: That's more or less a WebGL standard. (0,0,1) for ordinary objects and (0,0,-1) for cameras, see https://stackoverflow.com/questions/33407209/why-does-the-camera-face-the-negative-end-of-the-z-axis-by-default?rq=1

Mugen87: @looeee I have hardwired the forward vector for now since the controls do not fully support arbitrary objects yet, see https://jsfiddle.net/q50o1r97/1/. If there is any interest in changing this, I suggest to do this with a separate PR. Same for renaming `object` to `camera` in control classes.

mrdoob:
Thanks!

Mugen87:
I think we should update the stackoverflow posts, right?

examples/js/controls/FirstPersonControls.js
looeee: @Mugen87 OK, but I think it would be good to have clarity on this. Either we do support controlling arbitrary objects with controls, or we don't. 

If we do, then it should be noted somewhere - even just as a comment in the files. 

If we don't, then we can just rename `object` to `camera`. 
I'm OK with either.

PR #15651, opened 2019-01-27
Audio: Fix usage of detune

Mugen87:
Right now, `Audio.play()` in `dev` is broken since `Audio.detune()` tries to access `Audio.source` before it is set in `Audio.play()`. Besides, the actual detune value is never set to the `AudioBufferSourceNode` since `Audio.playing` is still `false`.

The PR fixes both bugs and ensures a consistent setting of a `AudioBufferSourceNode` properties which have a `AudioParam` value (`detune` and `playbackRate`).

### DIFF
--- a/src/audio/Audio.js
+++ b/src/audio/Audio.js
@@ -95,16 +95,17 @@ Audio.prototype = Object.assign( Object.create( Object3D.prototype ), {
 		var source = this.context.createBufferSource();
 
 		source.buffer = this.buffer;
-		this.setDetune( this.detune );
 		source.loop = this.loop;
 		source.onended = this.onEnded.bind( this );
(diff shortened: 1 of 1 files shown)


src/audio/Audio.js
```diff
@@ -228,12 +229,7 @@ Audio.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 		this.detune = value;
 
-		if ( this.source.detune === undefined ) {
-
-			console.warn( 'THREE.Audio: AudioBufferSourceNode.detune not supported by the browser.' );
```
Mugen87: I have remove this log since the amount of warnings in Safari was just too high. We should keep in mind that `Audio.setDetune()` is called in `Audio.play()`.

PR #15659, opened 2019-01-28
Update TypeScript definition typings to match @types/three

Methuselah96:
This a port of [all the changes](https://github.com/DefinitelyTyped/DefinitelyTyped/commits/master/types/three) that have been made to @types/three since the [typings were generated from @types/three](https://github.com/mrdoob/three.js/pull/15597). We should make sure to [delete the typing on the DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package) once r101 is released to avoid having to do this in the future.


mrdoob:
Thanks!

PR #15661, opened 2019-01-29
Cache and reuse attribute buffer bindings

bunnybones1:
Improved rendering speed by preventing rebinding of identical attribute buffers, if for instance you reuse some buffers across objects, like duplicated meshes whose only difference are the UV buffer.

Note: this only improves performance if you have geometries that share attributes whose render order is back to back.

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -648,6 +648,12 @@ function WebGLRenderer( parameters ) {
 
 	};
 
+	var boundBufferCache = []
+	for(var i = 0; i < 64; i++) {
+		boundBufferCache[i] = null
+	}
+	var boundIndexBuffer = null;
+
 	this.renderBufferDirect = function ( camera, fog, geometry, material, object, group ) {
 
 		var frontFaceCW = ( object.isMesh && object.normalMatrix.determinant() < 0 );
@@ -706,9 +712,10 @@ function WebGLRenderer( parameters ) {
 
 			setupVertexAttributes( material, program, geometry );
 
-			if ( index !== null ) {
+			if ( index !== null && boundIndexBuffer !== attribute.buffer) {
 
 				_gl.bindBuffer( _gl.ELEMENT_ARRAY_BUFFER, attribute.buffer );
+				boundIndexBuffer = attribute.buffer;
 
 			}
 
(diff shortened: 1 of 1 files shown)


takahirox:
Can you share the performance gain(in case attributes are reused) and loss(in case attribute aren't reused) numbers? When I tried fine grain check and cache like this PR, I realized it has negative performance impact because of adding if statement in loop.

bunnybones1:
I hear you.
This optimization may be not good in general because it is designed for an edge case. In the case of our game, we have many instances of the same mesh except for one difference that they all have unique UVs. We can ensure they render in the same cluster by giving them the same render order. Then when it switches from one mesh to another to render them, we can skip binding the reused attributes. In our case we reduce our webGL commands from 1000 to about 940 and gives us a modest 5-10% fps increase.
That said, I'll put together a test this weekend to get a better sense of the price of redundant gl.bindBuffer calls.
Anyways, this approach might be a poor substitute for instanced rendering, which can solve our problems maybe better.

takahirox:
My impression right now is, as you said, this is for edge case and we have alternative solution, instancing. `setupVertexAttributes()` is one of hot spot function. When I tried fine-grain cache I realized adding if statement in loop has kinda non small performance penalty. (Sorry I lost the number) So better to fork and edit in user side for the case where fine-grain cache works fine, and we go with coarse-grain cache(VAO) rather than fine-grain cache?

But of course decision should very depend on the performance numbers. Especially I wanna know the performance penalty. I'm looking forward to seeing the test result.

bunnybones1:
Sounds good to me. Thanks @takahirox 
I'll post performance results here by end of the weekend

Usnul:
@bunnybones1 has the week ended yet? I'm quite curious about the results too :)

takahirox:
BTW are this PR and the test blocking #14600?

bunnybones1:
Hey @Usnul @takahirox, I'm sorry for forgetting about this. We've gotten really busy with the launch of our game. I won't have time to make the performance tests I wanted until we finish our current sprint.
Anyways, on recent review, I think the benefits of this approach are not worth the modification. We're maintaining this optimization in our game-specific branch of threejs, and taking advantage of it with game code that looks like this:
```
    const newGeo = mesh.geometry.clone() as BufferGeometry
    const originalGeo = mesh.geometry as BufferGeometry
    newGeo.attributes.position = originalGeo.attributes.position
    newGeo.attributes.normal = originalGeo.attributes.normal
    newGeo.index = originalGeo.index
    mesh.geometry = newGeo
    const range = textureAtlas.getOrCreateRange(
      getCardAtlasKey(card),
      586,
      1024
    )
    textureAtlas.updateUVAttribute(mesh.geometry, range)
```
While it does technically reduce gl calls (attribute bindings), it's not a significant performance gain.

Issue #15654, opened 2019-01-28
Pass alphaTest as a uniform, rather than a macro.


donmccurdy:
A material's `.alphaTest` property is currently translated into:

```glsl
#ifdef ALPHATEST

if ( alpha <= ALPHATEST ) discard;

#endif
```

As a result of using a macro, the value cannot be changed or animated without recompiling. I'd like to propose that we allow animating the value, which may require an API addition to skip the conditional when it isn't needed. For example:

```javascript
// Mask
material.alphaMode = THREE.AlphaMask; // or AlphaClip
material.alphaTest = 0.5;

// Blend
material.alphaMode = THREE.AlphaBlend; // or AlphaTransparent

// Both
material.alphaMode = THREE.AlphaBlendMask; // or AlphaTransparentMask
```

In the example above, `alphaMode` replaces `.transparent`. Alternatively, we could use separate boolean flags like `.transparent=true` and `.mask=true`.

Animating `alphaTest` is of limited use with fixed material inputs, but it becomes much more interesting when paired with NodeMaterial, where procedural offset, opacity, alphaTest, and emissive values can be used for effects like these:

| dissolve | smoke |
|-----|-----|
| ![51575747-353af880-1e68-11e9-8852-eebcaa14fa1a](https://user-images.githubusercontent.com/1848368/51819767-682a2580-2288-11e9-9fcb-e65b9ea776af.gif) | ![ezgif-4-e5b3bd8afcf8](https://user-images.githubusercontent.com/1848368/51820052-6dd43b00-2289-11e9-80c2-90ed3ab77436.gif) |

For further reading, see http://glowfishinteractive.com/dissolving-the-world-part-2/. Previous discussion in https://github.com/mrdoob/three.js/issues/5623.

Do we want a PR for this, and if so: (1) any preference on the API, and (2) should it affect core materials or just NodeMaterial?

/cc @sunag

takahirox:
According to the https://github.com/mrdoob/three.js/issues/5623#issuecomment-62963656 the purpose of using macro seems avoiding if-else statement and having better performance. (/cc @WestLangley)

If it's correct, (if necessary) we evaluate the negative performance impact of replacing macro with uniform. (I'm also curious to know how big performance impact one if-else statement has.) And then if we don't like negative impact we introduce `.useAlphaTest` flag property to material?

    // in WebGLProgram.js

    parameters.useAlphaTest ? '#define USE_ALPHATEST' : '',

    // in glsl

[...]

PR #15663, opened 2019-01-29
NodeMaterial - Intro to boolean, mask and dissolve example

sunag:
### Issue
https://github.com/mrdoob/three.js/issues/15654

### New examples

**dissolve** 
https://rawgit.com/sunag/three.js/dev-mask/examples/webgl_materials_nodes.html?e=dissolve

**dissolve - fire**
https://rawgit.com/sunag/three.js/dev-mask/examples/webgl_materials_nodes.html?e=dissolve-fire

### Description
- `material.mask` show `fragment` if `mask` is `true`
- Automatic `float` and `vectors` convertion to `boolean` and vice versa
- `CondNode` returns boolean values if `ifNode` and `elseNode` is `undefined`
- new `BoolNode`

### Usages
```javascript
// CondNode returns a boolean value
// show fragment only if CondNode returns true
material.mask = new THREE.CondNode(
	maskNode, // mask node
	new THREE.FloatNode( .1 ), // threshold
	THREE.CondNode.GREATER // condition
);
```

```javascript
// show fragment if mask is true
material.mask = new THREE.BoolNode( true );
```

image

### DIFF
--- a/examples/js/nodes/Nodes.js
+++ b/examples/js/nodes/Nodes.js
@@ -20,6 +20,7 @@ export { NodeBuilder } from './core/NodeBuilder.js';
 
 // inputs
 
+export { BoolNode } from './inputs/BoolNode.js';
 export { IntNode } from './inputs/IntNode.js';
 export { FloatNode } from './inputs/FloatNode.js';
 export { Vector2Node } from './inputs/Vector2Node.js';
--- a/examples/js/nodes/THREE.Nodes.js
+++ b/examples/js/nodes/THREE.Nodes.js
@@ -20,6 +20,7 @@ import {
 
 	// inputs
 
+	BoolNode,
 	IntNode,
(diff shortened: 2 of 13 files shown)


sunag:
image

donmccurdy:
This looks great, and solves the use case I needed – thank you! 👏

Do you want to keep the `.alphaTest` property on StandardNodeMaterial? It seems like the MeshStandardNodeMaterial wrapper could just create a `.mask` socket when it sees `.alphaTest`, and then `.alphaTest` wouldn't need to exist on the new API. Assuming I understand the difference in the two materials correctly. Up to you, though, I'm happy with this PR as-is.

mrdoob:
Very nice!

mrdoob:
Thanks!

donmccurdy:
This is working well, thanks!

PR #14087, opened 2018-05-18
Support shadow intensity

WestLangley:
As proposed in #8238.

The shadow intensity (darkness) can be changed without affecting the overall illumination of the scene.

This implementation adds a single uniform, `renderer.shadowIntensity`, that takes a value in [ 0, 1 ], defaulting to 1.

The approach works with (Gouraud) `MeshLambertMaterial` -- whose shadows are an approximation -- and it also works with the other built-in materials, where shadows are implemented correctly as the absence of light. (`MeshLambertMaterial ` does not accommodate a per-light shadow intensity.)

I can add docs later if this is merged.

PR #15670, opened 2019-01-30
shadows - introduce light shadow intensity

godknowsiamgood:
So this is a followup to #14087.
Instead of making intensity property of renderer, I made this as a light property.

### DIFF
--- a/examples/webgl_lights_spotlight.html
+++ b/examples/webgl_lights_spotlight.html
@@ -164,6 +164,8 @@
 					penumbra: spotLight.penumbra,
 					decay: spotLight.decay,
 					'shadow intensity': spotLight.shadow.intensity,
+					'shadow near': spotLight.shadow.camera.near,
+					'shadow far': spotLight.shadow.camera.far,
 				};
 
 				gui.addColor( params, 'light color' ).onChange( function ( val ) {
@@ -212,6 +214,22 @@
 				gui.add( params, 'shadow intensity', 0, 1.0 ).onChange( function ( val ) {
 
 					spotLight.shadow.intensity = val;
+					render();
+
+				} );
+
+				gui.add( params, 'shadow near', 10, 50 ).onChange( function ( val ) {
+
+					spotLight.shadow.camera.near = val;
+					spotLight.shadow.camera.updateProjectionMatrix();
+					render();
+
+				} );
+
+				gui.add( params, 'shadow far', 10, 200 ).onChange( function ( val ) {
+
+					spotLight.shadow.camera.far = val;
+					spotLight.shadow.camera.updateProjectionMatrix();
 					render();
 
 				} );
--- a/examples/webgl_lights_spotlight.html
+++ b/examples/webgl_lights_spotlight.html
@@ -183,7 +183,7 @@
 				} );
 
 
-				gui.add( params, 'distance', 50, 200 ).onChange( function ( val ) {
+				gui.add( params, 'distance', 0, 200 ).onChange( function ( val ) {
 
 					spotLight.distance = val;
 					render();
--- a/examples/webgl_lights_spotlight.html
+++ b/examples/webgl_lights_spotlight.html
@@ -96,6 +96,7 @@
 				spotLight.shadow.mapSize.height = 1024;
 				spotLight.shadow.camera.near = 10;
 				spotLight.shadow.camera.far = 200;
+				spotLight.shadow.intensity = 1.0;
 				scene.add( spotLight );
 
 				lightHelper = new THREE.SpotLightHelper( spotLight );
@@ -161,7 +162,8 @@
 					distance: spotLight.distance,
 					angle: spotLight.angle,
 					penumbra: spotLight.penumbra,
-					decay: spotLight.decay
+					decay: spotLight.decay,
+					'shadow intensity': spotLight.shadow.intensity,
 				};
 
 				gui.addColor( params, 'light color' ).onChange( function ( val ) {
@@ -203,6 +205,13 @@
 				gui.add( params, 'decay', 1, 2 ).onChange( function ( val ) {
 
 					spotLight.decay = val;
+					render();
+
+				} );
+
+				gui.add( params, 'shadow intensity', 0, 1.0 ).onChange( function ( val ) {
+
+					spotLight.shadow.intensity = val;
 					render();
 
 				} );
--- a/docs/api/en/lights/shadows/LightShadow.html
+++ b/docs/api/en/lights/shadows/LightShadow.html
@@ -84,7 +84,7 @@ scene.add( helper );
 
 		<h3>[property:Float intensity]</h3>
 		<p>
-			Intensity value controls how shadow influences result surface color. In other words it rules shadow strength. 1.0 (default) makes shadows completely black.
+			Intensity value controls how shadow influences result surface color. In other words it determines how dark shadows are. 1.0 (default) makes shadows completely black.
 		</p>
 
 		<h3>[property:WebGLRenderTarget map]</h3>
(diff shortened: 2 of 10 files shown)


WestLangley:
Related: #8238.

With the exception of `MeshLambertMaterial`, please understand that shadows in three.js are modeled correctly as the absence of light -- shadows do not "darken the color", and should not be described as such.

Also, be aware that _per-light_ shadow intensity is not physically correct. _Per-renderer_ shadow intensity is not physically correct either, but I guess it can be considered a poor-man's way to accommodate for a lack of a global illumination model.

godknowsiamgood:
@WestLangley, yeah I understand first part of your comment, but not quite well second one. Anyway, I just want to make life easier for someone like me :) Does it mean that description in docs is not correct? I didn't use work 'darken'.

WestLangley:
Thank you for your contribution!

>Does it mean that description in docs is not correct?

Yes. Just say it controls the darkness of the shadows.

>Anyway, I just want to make life easier for someone like me

...And you need different lights to have different shadow darkness?

Please test this using a very shiny ground surface and see if you think the specular hot spots look OK when they are in shadow and the shadow intensity is reduced. I don't expect that to look good at all. When I tried previously, the intensity could be reduced, but the hot spots would not change brightness. That was the primary reason I lost interest in this feature.

looeee:
A potential workaround is to use a [https://threejs.org/docs/#api/en/materials/ShadowMaterial](https://threejs.org/docs/#api/en/materials/ShadowMaterial) and adjust the opacity/color. This works well for setting shadow strength on ground surfaces that receive shadow.

WestLangley:
There is also the workaround posted in https://github.com/mrdoob/three.js/pull/14087#issuecomment-431003830.

godknowsiamgood:
@WestLangley 

> see if you think the specular hot spots look OK 
They just dimmed. I guess in what we say 'real world' they shouldn't be visible at all?

image

> There is also the workaround ...
They are work well in most case, but they are still workarounds. They are not obvious for new comers and I guess less performant as this solution.

> And you need different lights to have different shadow darkness?
Actually, for me it will be enough per-renderer settings, but I thought it will be more general solution make it per-light.

So, should I update docs or make per-renderer? Is there a chance to be approved for merge?

Mugen87:
I vote to merge the PR since other engine support similar features. For instance [Unity's shadow strength](https://docs.unity3d.com/Manual/ShadowOverview.html):

> This determines how dark the shadows are. In general, some light is scattered by the atmosphere and reflected off other GameObjects, so you usually don’t want shadows to be set to maximum strength.

Mugen87:
Unity Shadow Intensity of 1:

image

Shadow Intensity of 0.5:

image

Shadow Intensity of 0.2:

image


Even if `three.js` does not render in the same quality, it's still better than workarounds.

examples/webgl_lights_spotlight.html
```diff
@@ -161,7 +162,8 @@
 					distance: spotLight.distance,
 					angle: spotLight.angle,
 					penumbra: spotLight.penumbra,
-					decay: spotLight.decay
+					decay: spotLight.decay,
+					'shadow intensity': spotLight.shadow.intensity,
```
looeee: Seems like if you're adding shadow intensity to the controls, then the other shadow params should be there too, at least `spotLight.shadow.camera.near` and `spotLight.shadow.camera.far`

looeee:
This is a common feature request so I support adding this.

examples/webgl_lights_spotlight.html
godknowsiamgood: I tried it, but I don't see any benefits from changing near plane for example. What do you think?
1. It clutters control interface
2. Far plane is somehow overridden by light's `distance` parameter
3. Don't see any case when it useful in this particular example

looeee: > Far plane is somehow overridden by light's distance parameter

Yes, the `spotLight.shadow.camera.far` has no effect unless  `spotLight.distance = 0`. Demonstrating that is actually the main reason why I want to add `near` and `far` to this example, since it's something that we should make clearer.

godknowsiamgood: I've added two parameters. Please, check 🤖

WestLangley:
I have tried to keep an open mind, but I am NOT in favor of adding this feature because it looks terrible. https://github.com/mrdoob/three.js/pull/15670#issuecomment-459239484.

It also looked terrible when I tried it in #14087, so this is not the fault of the OP.

You can make the shadows faint, but the hotspot remains. The implementation would have to be different -- somehow.

>And you need different lights to have different shadow darkness?
>>Actually, for me it will be enough per-renderer settings, but I thought it will be more general solution make it per-light.

And I certainly would not add a per-light feature for the reasons I explained. I do not think a per-light shadow intensity is possible with `MeshLambertMaterial`, anyway.

>I vote to merge the PR since other engines support similar features. For instance Unity's shadow strength:

...only if you figure out how to make it look decent. It is easy to support a feature because another engine supports it. It is a different matter to figure out how to implement it correctly.

Mugen87:
> It is easy to support a feature because another engine supports it. It is a different matter to figure out how to implement it correctly.

I'm aware that `three.js` does not produce the results of Unity. I still think this PR is a good start. Things can polished and improved at a later point, too.

> I do not think a per-light shadow intensity is possible with MeshLambertMaterial, anyway.

We might consider to exclude `MeshLambertMaterial` from this feature and only support materials that perform per-fragment shading.

Mugen87:
Okay, I have seen that Babylon.js support `shadow darkness`. It might be worth to evaluate this feature and check the respective source code.

WestLangley:
#14087 implements per-renderer shadow intensity and has the same artifacts this PR does. #14087 also supports `MeshLambertMaterial`. I would prefer that approach to this one.

@mrdoob has always handled this issue by adding ambient light to the scene and simultaneously reducing the intensity of the other direct lights. That approach works well, and is likely why shadow intensity did not get much traction. It is likely why #14087 was not merged.

Mugen87:
I see. But let's say we find a way to handle the mentioned artifacts. Do you think this would raise the chances of getting shadow intensity into the library?

examples/webgl_lights_spotlight.html
looeee: Looks good, although could you also change the minimum value for `.distance` to 0?

looeee:
> I have tried to keep an open mind, but I am NOT in favor of adding this feature because it looks terrible. 

The specular highlights are already showing incorrectly over shadows though, right? Meaning that for that case, the current implementation already looks incorrect, and adding shadow intensity hardly makes it look worse.

godknowsiamgood:
> The specular highlights are already showing incorrectly over shadows though, right?

Actually, no.
image

Even though PR would satisfy most of users, I started agree with WestLangley as it starts to generate wrong image, not what 3d-library intended to do. But my lack of skills won't let me rewrite spotlight shader or smth.

Anyway, last word for @mrdoob

WestLangley:
Shadows in three.js are the absence of light. If you want so-called "lighter shadows", then you need to increase the ambient light in the scene, and compensate by reducing the intensity of the existing direct lights.

Making only the shadows lighter via a hack makes your PBR materials non-PBR.

WestLangley:
>But let's say we find a way to handle the mentioned artifacts. Do you think this would raise the chances of getting shadow intensity into the library?

@Mugen87 The only acceptable solution I can think of is the one I posted in my previous comment. So I just don't know the answer to your question...

>The specular highlights are already showing incorrectly over shadows though, right?

@looeee Not to my knowledge.

Mugen87:
Closing. A concept of shadow intensity or darkness needs a different implementation.

mrdoob:
Sorry for the late reply. I agree with everything @WestLangley said.

PR #15673, opened 2019-01-30
Indirect specular dark fresnel fix

jsantell:
Diffuse color has already been calculated relative to the specular, so use that directly to fix dark fresnel for indirect specular lighting. I *believe* this is already handled here, but could use a second set of eyes:

https://github.com/mrdoob/three.js/blob/5283af5e535c0b0f0602edc9f5ef413c6acd1185/src/renderers/shaders/ShaderChunk/lights_physical_fragment.glsl.js#L3

This still passes the furnace test and looks similar to the previous implementation, except slightly brighter fresnel; here's a gif flipping between the two, not terribly exciting for dielectrics, but better than the dark fresnel border! (Look at the upper left edge, and the edges inside the knot)

output_7d0qm6

### DIFF
--- a/src/renderers/shaders/ShaderChunk/lights_physical_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/lights_physical_pars_fragment.glsl.js
@@ -128,7 +128,11 @@ void RE_IndirectSpecular_Physical( const in vec3 radiance, const in vec3 irradia
 
 		BRDF_Specular_Multiscattering_Environment( geometry, material.specularColor, material.specularRoughness, singleScattering, multiScattering );
 
-		vec3 diffuse = material.diffuseColor * ( 1.0 - ( singleScattering + multiScattering ) );
+		// The multiscattering paper uses the below formula for calculating diffuse 
+		// for dielectrics, but this is already handled when initially computing the 
+		// specular and diffuse color, so we can just use the diffuseColor directly.
+		//vec3 diffuse = material.diffuseColor * ( 1.0 - ( singleScattering + multiScattering ) );
+		vec3 diffuse = material.diffuseColor;
 
 		reflectedLight.indirectSpecular += clearCoatInv * radiance * singleScattering;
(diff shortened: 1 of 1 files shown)


WestLangley:
I'm a bit suspicious. The models may be incompatible.

From https://github.com/mrdoob/three.js/pull/15644, the second row on the right ( _rough surface with increasing metalness_) appears too dark when metalness is 1/2 -- at least, to me.

mrdoob:
Thanks!

mrdoob:
@WestLangley Yes. I noticed that too. The second row seems incorrect but the other rows seem better. Definitely more work to do here.

jsantell:
I wonder if it's a matter of using a different roughness scale?

jsantell:
From left to right: r100 (pre-multiscattering), r103, and r103 *without* #15673; metalness from bottom to top, roughness from left to right:

Screenshot from 2019-03-20 11-57-07

While energy preservation is still achieved for conductors, this change introduced a large change for most notably smooth dielectrics, which appear to be reflecting too much energy. This fix did not address dielectrics (and at the time, I wasn't too sure of the behaviour of non-conductors), but seems necessary for preserving energy.

What are the thoughts on backing out this change, plus a different fresnel fix? It's possible that roughness-dependent F90 values need to be used to address that. This will be back to what the screenshots shown in https://github.com/mrdoob/three.js/pull/15673#issuecomment-459182098 look like.

The issue mentioned in https://github.com/mrdoob/three.js/pull/15673#issuecomment-459182098 with the 'dip' in brightness is to be expected (or at least, undefined) without [computing both dielectric and conductor](https://polycount.com/discussion/comment/2452989/#Comment_2452989) outputs; I think achieving energy preservation for fully conductor and dielectric materials is more important than the inbetween materials.

WDYT? @bhouston @WestLangley

bhouston:
I can not comment on this as I haven't researched it.  But I would suggest
following the example of other real time engines and adopt their approach.

PR #15681, opened 2019-02-01
THREE JS Editor: Added Buffer geometry commands

Temdog007:
Added numerous features to THREE JS Editor regarding buffer geometry.

1) Allow an object with an 'empty' Buffer Geometry to be added.
2) Allow geometry from one object able to be copied to another.
3) Allow geometry from one object to be cloned and merged into another.
ezgif-4-a01bf40ada51

4) Allow buffer geometry to be transformed. 
transform demo


Temdog007:
Since I've discovered [clara.io](https://clara.io), I find these features will just clutter up the Editor. So, I'm closing this PR.

PR #15683, opened 2019-02-01
Editor: Load scene via url

Temdog007:
Allow scenes to be loaded if defined in the URL
parameter demo


Temdog007:
This feature currently already exists if `#file` is used instead of `?param`. Closing this PR.

PR #15687, opened 2019-02-01
Clean up rollup.config.js

Mugen87:
Adding the `sourceMap` parameter at the current place leads to a warning:

> You have passed an unrecognized option

The correct approach would look like so (at least with rollup version `1.0.0`):

```js
{
   format: 'umd',
   name: 'THREE',
   file: 'build/three.js',
   indent: '\t',
   sourcemap: true
}
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -196,7 +196,6 @@ export default {
 		glconstants(),
 		glsl()
 	],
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

Issue #12115, opened 2017-09-03
Review THREE.Pass specification


tentone:
Hi

I started working this week on post processing effects for my project.

Noticed that there are a bunch of alternative projects for post processing in threejs (https://github.com/vanruesc/postprocessing, https://github.com/spite/Wagner) and that there is a lot of repeated code in  the "PostprocessingPasses" (creation of quad scene, rendering of depthbuffer, etc).

There is also some lack of flexibility since a EffectComposes is created the render a specific scene with a specific camera and renderer (for me this was the deal breaker).

We should also consider WebGL2 in these reviews (since we are going to have separate renderers probably EffectComposer, ShaderPass, etc should also have 2 versions?).

I think we should review the specification for THREE.Pass and THREE.EffectComposer and create a solid documented specification for new "Passes".

I have started doing some work on this matter to integrate post processing in my project. Im totally available to work on this.

Cumps

WestLangley:
Related: https://github.com/mrdoob/three.js/issues/10144#issuecomment-261118347

tentone:
One thing that could be also implemented is a set of "pre-passes" when adding a new Pass to the Composer the composer checks for pre-requirements (example, DepthBuffer RGBA, DepthBuffer, G-Buffer XPTO, etc)

The Effect composer renders all these pre-requirements and makes the results available to all passes, (if 2 passes need a depth buffer they can use the same buffer). With WebGL2 since we can generate most of these at the "same time" that should improve performance quite a bit.

(This was just an idea that occurred, i will try to implement this in nunuStudio and check if there is some real performance gains with the current renderer).

tentone:
A fast way to fix the flexibility problem with THREE.Pass would be to simply add scene and camera parameters to render.

https://github.com/tentone/nunuStudio/blob/master/source/core/postprocessing/Pass.js

I'm also making some tests with the pre-passes for shared buffers.

Usnul:
personally i see more suitability in a 2 data-structure approach:

- Pass
- Buffer

a pass would define input and output buffers. In "Postprocess" system simply provide some input buffers (or passes that without inputs) and declare what buffer you want to be rendered to screen. Whole thing with swapping is pretty fiddly.

Mugen87:
[...]

PR #15694, opened 2019-02-02
EffectComposer2

Oletus:
EffectComposer2 is a successor to the post-processing stack EffectComposer. Its advantages are:

1. A simpler but more flexible Pass API. Passes specify a list of input and output buffers that they need instead of having "readBuffer" and "writeBuffer" passed in. It's intended that in the future buffers handled by EffectComposer2 can contain depth, normal etc. information in addition to color information. There's no more "renderToScreen" flag for each pass - instead the final color render target for EffectComposer2 gets passed to the render() function.

2. Common functionality is shared between render passes.

The API for the passes has changed so that passes written against EffectComposer are not compatible with EffectComposer2, so this is added alongside the old EffectComposer instead of replacing it. However, the outward-facing API is mostly the same with the exception of removal of the "renderToScreen" flag.

### DIFF
--- a/examples/js/postprocessing2/AfterimagePass2.js
+++ b/examples/js/postprocessing2/AfterimagePass2.js
@@ -1,0 +1,89 @@
+/**
+ * @author HypnosNova / https://www.threejs.org.cn/gallery/
+ * @author Oletus / http://oletus.fi/ - ported to EffectComposer2
+ */
+
+THREE.AfterimagePass2 = function ( damp ) {
+
+	THREE.Pass2.call( this );
+
+	this.colorWriteBufferConfig = new THREE.IntermediateBufferConfig();
+	this.colorWriteBufferConfig.clear = true;
+
+	this.colorReadBufferConfig = new THREE.IntermediateBufferConfig();
+	this.colorReadBufferConfig.isOutput = false;
+	this.colorReadBufferConfig.isInput = true;
+
+	this.bufferConfigs = [ this.colorWriteBufferConfig, this.colorReadBufferConfig ];
+
+	if ( THREE.AfterimageShader === undefined )
+		console.error( "THREE.AfterimagePass relies on THREE.AfterimageShader" );
+
+	this.shader = THREE.AfterimageShader;
+
+	this.uniforms = THREE.UniformsUtils.clone( this.shader.uniforms );
+
+	this.uniforms[ "damp" ].value = damp !== undefined ? damp : 0.96;
+
+	this.textureComp = new THREE.WebGLRenderTarget( window.innerWidth, window.innerHeight, {
+
+		minFilter: THREE.LinearFilter,
+		magFilter: THREE.NearestFilter,
+		format: THREE.RGBAFormat
+
+	} );
+
+	this.textureOld = new THREE.WebGLRenderTarget( window.innerWidth, window.innerHeight, {
+
+		minFilter: THREE.LinearFilter,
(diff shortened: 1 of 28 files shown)


donutcoffee:
I like how you've removed the confusing "readBuffer" and "writeBuffer" from EffectComposer.

I also like that you've sort of added the ability to pass a camera to fillQuad (renderWithClear) e.g. in ShaderPass. There still is a problem with that that it breaks use of Layers just like the original postprocessing.

Would you consider passing a camera to ShaderPass, just like any other pass?
`THREE.ShaderPass2 = function ( shader, textureID, camera ) {`
then at line 63:
`THREE.Pass2.renderWithClear( renderer, this.fillQuad.scene, this.camera || this.fillQuad.camera, writeBuffer, this.colorWriteBufferConfig.clear );`

fillQuad.camera could be renamed to fillQuad.defaultCamera for clarity.

I can't quite comment on the rest at this time - it's been a while too since you submitted this. Are you still working on this?

Oletus:
I'm not actively working on this right now. Some of the smaller enhancements were merged to the existing EffectComposer in PRs #15976 and #16033 . It would still be nice to create an upgraded version of EffectComposer that would get rid of the confusing usage of "readBuffer", but this is a tricky change to make since all the postprocessing passes need to be upgraded, and this change didn't seem to create much interest back when I started on it. Maybe we could make further small changes while keeping things backward compatible like renaming the parameters so that the names would be more intuitive.

Oletus:
I'm not quite sure what you refer to when you talk about problems with layers. Applying regular camera layers doesn't make sense when a postprocessing pass is filling the screen with a single quad, and as far as I know when some passes perform scene rendering with an override material regular camera layers are being applied like they should.

donutcoffee:
> 
> 
> I'm not quite sure what you refer to when you talk about problems with layers. Applying regular camera layers doesn't make sense when a postprocessing pass is filling the screen with a single quad, and as far as I know when some passes perform scene rendering with an override material regular camera layers are being applied like they should.

I have a project where I need to turn groups of objects and some postprocessing elements off at times. Using Layers is a good way to do that, by design. I guess Pass.enable could be used, but in my experience that leads to unexpected results due to buffers not being swapped when they need to etc.

The Layers of the fillQuad is easily set as it's a public property already. Camera used to be hidden.

I agree it's an uncommon case, and it could be seen as a workaround the real problem which is that Pass.enable is tricky to use. :)

DavidPeicho:
With MRT now in the core, I think this should be revisted again. We also need some pooling for render targets

DavidPeicho:
I don't use the EffectComposer a lot (I in general use a custom one). I wrote a course though and I wanted the students to be able to use post-processing easily. I actually discovered that a lot of post-process weren't following the same API.

Some post-process don't handle the resizing correctly, example with the [Bokeh Pass](https://github.com/mrdoob/three.js/blob/master/examples/jsm/postprocessing/BokehPass.js).

Should we give a try to an improvement to the effect composer? Or should we track down the effect that aren't following conventions?

Mugen87:
> Or should we track down the effect that aren't following conventions?

I would start with that since we easier have results. Discussing a new API for post processing is a bigger thing.

DavidPeicho:
I can give a show to fixing some of the post-process. Not the same issue, but I still think we need to add an API to re-use / share the render targets

Mugen87:
Closing, see https://github.com/mrdoob/three.js/issues/12115#issuecomment-2429240760.

Issue #13882, opened 2018-04-17
Texture panel for all textures in Threejs Editor


atul-mourya:
Hello Everyone,

Hope you all would find this feature request interesting.

At this point, threejs editor supports the texture implementation but there is no easier way to set various texture properties like anisotropy, filters, repeat, offset etc. 

I would propose a texture panel to have all the texture properties for each selected object and its texture type.

This will ease up the process of scene and asset creation. Without this feature, the material and texture assignment panels look somewhat incomplete.

I'm sure that this feature will be highly valuable for so many different people and will make the editor more powerful than ever.

Let me know if this feature makes any sense. Will be glad to have a positive response.

sdinesh86:
I am working on a project, for which i am in very dire need of something like this.

@mrdoob I was planning on writting something for this purpose, so i would be glad to contribute on this. Let me know if i can help with it and if you have any ideas on how you would like it to be implemented.

a few ideas :
1. We can extend the current texture row with all the required properties.
2. Every Texture row as a collapsable panel, with all the required controls.
3. Edit button besides the canvas preview which launches the modal dialog where one can edit the parameters.

I personally would prefer the second option.

sdinesh86:
I just reliazed after posting above comment we have an open issue #5876 which i think needs to be addressed before we think of this, 

since currently only one texture offset & repeat settings is honoured, and this #8278 per map per material PR by @bhouston is still open

PR #15695, opened 2019-02-02
THREE JS Editor: Added texture's properties to UI Texture object 

Temdog007:
In reference to issue # #13882.

A map's wrapS, wrapT, anisotropy, offset, repeat, center, and rotation can now be changed in the THREE JS Editor. The panel showing this data can be collapsed/expanded and the button will only show up if a texture is loaded for that map.
texturesidebar
texturesidebar2


Temdog007:
Closing due to lack of interest.

PR #15696, opened 2019-02-02
THREE JS Editor: Allow grid to be modified

Temdog007:
grid

### DIFF
--- a/editor/js/Sidebar.Settings.Viewport.js
+++ b/editor/js/Sidebar.Settings.Viewport.js
@@ -12,8 +12,32 @@ Sidebar.Settings.Viewport = function ( editor ) {
 
 	container.add( new UI.Text( strings.getKey( 'sidebar/settings/viewport/grid' ) ).setWidth( '90px' ) );
 
+	var div = new UI.Div().setPaddingLeft('90px');
+	container.add( div );
+
+	div.add(new UI.Text('Visible').setPaddingRight('8px').setWidth('90px'));
 	var show = new UI.THREE.Boolean( true ).onChange( update );
-	container.add( show );
+	div.add(show, new UI.Break());
+
+	div.add(new UI.Text('Size').setPaddingRight('8px').setWidth('90px'));
+	var size = new UI.Integer(30).setWidth('45px').onChange(update);
+	div.add(size, new UI.Break());
+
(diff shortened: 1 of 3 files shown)


Mugen87:
For testing: https://raw.githack.com/Temdog007/three.js/feature/GridHelperSize/editor/index.html

Mugen87:
Nice feature but it seems that changes to the grid are not saved. Changing other project settings like the language or the theme are still present after a reload.

editor/js/Sidebar.Settings.Viewport.js
```diff
@@ -12,8 +12,32 @@ Sidebar.Settings.Viewport = function ( editor ) {
...
+	var div = new UI.Div().setPaddingLeft( '90px' );
+	container.add( div );
+
+	div.add( new UI.Text( 'Visible' ).setPaddingRight( '8px' ).setWidth( '90px' ) );
 	var show = new UI.THREE.Boolean( true ).onChange( update );
-	container.add( show );
+	div.add( show, new UI.Break() );
+
+	div.add( new UI.Text( 'Size' ).setPaddingRight( '8px' ).setWidth( '90px' ) );
+	var size = new UI.Integer( 30 ).setWidth( '45px' ).onChange( update );
```
Mugen87: It's a bit unfortunate that you can set a value smaller than 0. Same for divisions and dividers:

image


I think it's better to allow just positive integer values.

Mugen87 (approved):

mrdoob:
I'm not sure about this one. The `GridHelper` API doesn't seem very intuitive yet and these settings highlight that fact...

Temdog007:
This feature would be useful for large scenes where the grid doesn't reach all the way or when the colors make it hard to see it.

When you say the `GridHelper` API isn't intuitive, what exactly do you mean? I'd be glad to fix the issues in this PR or another PR.

vujadin:
Here's my version https://jsfiddle.net/vuja22/vfoyn6ek/20/

Temdog007:
Since it appears that @mrdoob wants the `GridHelper` API to be refactored, this feature will need to be updated after those changes occur. So, I'll close this PR and open a new one when those `GridHelper` changes are merged.

Issue #15671, opened 2019-01-30
Improve choosing render target in postprocessing EffectComposer


Oletus:
The current mechanisms for handling render targets in EffectComposer are fairly difficult to use. A lot of the passes write to the "readBuffer" which makes the code a bit hard to follow. The "renderToScreen" flag is also a bit odd - nothing prevents setting it true on a Pass in the middle of the EffectComposer stack, but it ever really makes sense to only set that on the last pass in the stack.

The code would be simpler and more flexible if:

1. The final RGBA output render target is passed to EffectComposer's render() function. There's no need for a "renderToScreen" flag on each Pass.
2. Passes have a flag for telling if they only need one color buffer (only write or both read/write on the same buffer) or two color buffers, one being the read buffer and other one the write buffer. This could replace the needsSwap flag.
3. EffectComposer contains all the logic for managing the intermediate buffers and will pass the right color buffers to each Pass.

This way the EffectComposer render() call would be more similar to regular WebGLRenderer.render(), and it would also be easier to call another EffectComposer from a Pass to set up more complex effects. I've prototyped this kind of changes on a project and they made the EffectComposer much nicer to work with.

This is a breaking change, so it may be necessary to create a separate EffectComposer2 for it, though I'd be happy to change the current EffectComposer as well if others think it's OK.

This is somewhat related to Issue #12115, which outlines some good ideas for a larger overhaul of EffectComposer. These smaller changes would get us a bit closer to that while not being too much work to implement. We could design the new API so that more intermediate buffers like a normal buffer or 
a depth buffer could be passed from EffectComposer to Passes in the future.

Oletus:
@WestLangley @tentone @mrdoob any thoughts? If you think this direction is good I could work on a PR.

tentone:
I totally agree the current post-processing pipeline needs some rethinking. Currently, it is not flexible, each pass ends up calculating everything that it needs (depth, normals etc).

I also don't like the "renderToScreen" flag I would preferer some "RenderTarget" abstraction instead.

WestLangley:
I agree. I would recommend that you mock up `EffectComposer2`. That way you have complete flexibility.

Also, see https://github.com/mrdoob/three.js/issues/10144#issuecomment-261117232.

Mugen87:
> A lot of the passes write to the "readBuffer" which makes the code a bit hard to follow.

[...]

PR #15699, opened 2019-02-03
Small postprocessing cleanups

Oletus:
This improves handling of render target sizes in the current EffectComposer.

### DIFF
--- a/examples/js/postprocessing/AfterimagePass.js
+++ b/examples/js/postprocessing/AfterimagePass.js
@@ -83,6 +83,13 @@ THREE.AfterimagePass.prototype = Object.assign( Object.create( THREE.Pass.protot
 			
 		}
 
+	},
+
+	setSize: function ( width, height ) {
+
+		this.textureComp.setSize( width, height );
+		this.textureOld.setSize( width, height );
+
 	}
 
 } );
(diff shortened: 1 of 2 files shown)


Mugen87:
I think it's a good idea if `EffectComposer` managed its size as a property and does not immediately bypass it to render targets or passes. In this context, it would be great if you find a way so the composer handles the pixel ratio more consistently than before, see #14340

`WebGLRenderer.getDrawingBufferSize()` regards the pixel ratio so initial render targets are sized correctly according to the renderer settings. But since `EffectComposer` has no concept of `pixelRatio`, `EffectComposer.setSize()` behaves a bit unexpected. If an app sets the size like in the following listing, the sizes of the renderer and the effect composer do not match anymore if the pixel ratio is greater than 1.
```js
composer.setSize( width, height );
```
Normally you would expect both methods do the same but that is currently not the case. You can see the problem with one of the basic post processing examples:

https://threejs.org/examples/webgl_postprocessing

Oletus:
Agreed that it would be nice to unify WebGLRenderer and EffectComposer behavior when it comes to setSize, but since this changes the behaviour of an existing API I'd rather do that as a part of moving to a new incompatible version of EffectComposer, see #15671

Mugen87:
> but since this changes the behaviour of an existing API I'd rather do that as a part of moving to a new incompatible version of EffectComposer

I would classify the issue with the pixel ratio as a bug that should be fixed in the existing API.

Oletus:
A lot of this is outdated - will do a new pull request with just the parts that are needed.

PR #15703, opened 2019-02-04
Feature/scene screenshot

Temdog007:
Add button to store screenshot of the scene. The screenshot is saved in the userData of the scene so it is stored when converted to JSON. Screenshot can also be previewed in the editor.
screenshot


Temdog007:
I've had a thought that saving the screenshot as a downloadable file would be more useful than what is implemented in this PR. But, I also discovered that [Firefox](https://support.mozilla.org/en-US/kb/firefox-screenshots) and Chrome both have a `Screenshot Node` option in their respective developer consoles. 

So, I think this PR can be closed do to this feature being unnecessary.

PR #15710, opened 2019-02-04
Add setResolveUrl to GLTFLoader

robertlong:
In Mozilla Hubs we rewrite urls contained within glTF files so that they are requested from our CORS proxy ([see here](https://github.com/mozilla/hubs/blob/71975e49d8f67764add2596585c7da500f920a6c/src/utils/media-utils.js#L80)). This PR adds a function `setResolveUrl` to the GLTFLoader class so that you can provide your own url resolution function.


donmccurdy:
This was the intended goal of [LoadingManager.setURLModifier](https://threejs.org/docs/#api/en/loaders/managers/LoadingManager.setURLModifier). Does that provide what you need?

robertlong:
Yeah, I think it does. Thanks for pointing that out. I'll see if I can move to that in Hubs and reopen this PR if it's not possible.

PR #15711, opened 2019-02-04
Fix BoxHelper clone

robertlong:
The `object` property is not passed into the constructor when cloning a BoxHelper. This PR passes the BoxHelper's `object` property to the constructor when cloning so that `.update()` is called correctly. It also clones the Object3D fields onto the new BoxHelper object.

### DIFF
--- a/src/helpers/BoxHelper.js
+++ b/src/helpers/BoxHelper.js
@@ -8,6 +8,7 @@ import { LineSegments } from '../objects/LineSegments.js';
 import { LineBasicMaterial } from '../materials/LineBasicMaterial.js';
 import { BufferAttribute } from '../core/BufferAttribute.js';
 import { BufferGeometry } from '../core/BufferGeometry.js';
+import { Object3D } from '../core/Object3D.js';
 
 function BoxHelper( object, color ) {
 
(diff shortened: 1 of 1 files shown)


WestLangley:
I am not sure why we need a `clone()` method for this helper... So an option is to remove the `clone()` method altogether.

Can you please provide an example showing a use case supporting this change?

robertlong:
I was using this in [Spoke](https://github.com/MozillaReality/Spoke) (Our 3D editor for Hubs), however it looks like I'm no longer directly cloning it. I'm currently going through and submitting PRs for each of the changes we've made for our fork of ThreeJS. I think I can close this one if you think this object doesn't need a clone method.

That does strike me as a bit odd though. I'd expect all objects to have clone and copy that works as expected.

Mugen87:
> I'd suggest removing `clone()` from `BoxHelper`.

Even if you remove `clone()` from the class, user can still invoke the method. A more robust approach is to overwrite `copy` and `clone()` and then log that copy or cloning is not supported.

We might want to do this in all helpers since this issue confused user multiple times in the past. Otherwise we should consider to properly implement `copy()` and `clone()` but that will cause some refactoring, right?^^

WestLangley:
>A more robust approach is to overwrite `clone()` and then log that cloning is not supported.

Yes.

robertlong:
Ok, I'm going to close this in favor of one of the solutions you have recommended above. Thanks!

PR #15712, opened 2019-02-04
GLTFExporter: Fix exporting uv2 for aoMap

robertlong:
Currently GLTFExporter always uses the second uv set for the ambient occlusion map. If a geometry doesn't have a uv2 attribute it should use the first uv set instead.

### DIFF
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -837,7 +837,7 @@ THREE.GLTFExporter.prototype = {
 		 * @param  {THREE.Material} material Material to process
 		 * @return {Integer}      Index of the processed material in the "materials" array
 		 */
-		function processMaterial( material ) {
+		function processMaterial( material, aoMapUVIndex ) {
 
 			if ( cachedData.materials.has( material ) ) {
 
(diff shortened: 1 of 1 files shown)


examples/js/exporters/GLTFExporter.js
```diff
@@ -837,7 +837,7 @@ THREE.GLTFExporter.prototype = {
 		 * @param  {THREE.Material} material Material to process
 		 * @return {Integer}      Index of the processed material in the "materials" array
 		 */
-		function processMaterial( material ) {
+		function processMaterial( material, aoMapUVIndex ) {
```
donmccurdy: I think calling this parameter `uvCount` or `numUVs` would be (1) more general, and (2) more representative of how it's computed. Its values could be `1 | 2` in that case.

donmccurdy:
Note that if the geometry doesn't have uv2, the AO will not render in three.js, and the exported result will appear different than the original model.

/cc @06wj

06wj:
Should we not export the AO map in this case?

robertlong:
I'll look into this again. I remember this being needed for some models, but we may no longer support that use case.

Mugen87:
Closing for now until it's clear what to do.

PR #15716, opened 2019-02-05
Added instructions for vertical aspect direction.

Muzzarino:
I just thought this would be useful to help others use the formula to changing vertical aspect ratio and not only the horizontal one.

### DIFF
--- a/docs/api/en/cameras/PerspectiveCamera.html
+++ b/docs/api/en/cameras/PerspectiveCamera.html
@@ -53,7 +53,7 @@ scene.add( camera );</code>
 		</p>
 
 		<h3>[property:Float aspect]</h3>
-		<p>Camera frustum aspect ratio, usually the canvas width / canvas height. Default is *1* (square canvas).</p>
+		<p>Camera frustum aspect ratio in horizontal direction of view, usually the canvas width / canvas height. Default is *1* (square canvas).</p>
 
 		<h3>[property:Float far]</h3>
 		<p>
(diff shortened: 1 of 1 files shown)


docs/api/en/cameras/PerspectiveCamera.html
```diff
@@ -75,7 +75,26 @@ <h3>[property:Float focus]</h3>
...
+				var h = 9;
+				var aspectlimit = w / h;
+				var aspectYscale = 1;
+				if (.aspect < aspectlimit)
+				{
+					aspectYscale = .aspect / aspectlimit;
+				}
+				
+				// Recalculate the new degrees with the aspect Y scale multiplier.
+				fov = 2 * Math.atan(1 / (1 / Math.tan(fov * Math.PI / 180 / 2) * aspectYscale) ) * 180 / Math.PI;
```
Mugen87: Can you please link to a source where this formula comes from?

docs/api/en/cameras/PerspectiveCamera.html
```diff
@@ -75,7 +75,26 @@ <h3>[property:Float focus]</h3>
...
+			
+			The [page:.aspect aspect] direction can be vertical using the formula like this:
+			
+			<code>
+				// Calculate the vertical scale if the horizonal aspect is below the specified aspect ratio.
+				var w = 16;
+				var h = 9;
+				var aspectlimit = w / h;
+				var aspectYscale = 1;
+				if (.aspect < aspectlimit)
```
Mugen87: The variable should be `aspect` (without the dot). Besides, please respect the code style of the project. Especially the spaces insides parens. For example it should look like this:

```js
if ( aspect < aspectlimit ) {

    aspectYscale = aspect / aspectlimit;

}
```

WestLangley:
>I just thought this would be useful to help others use the formula to changing vertical aspect ratio and not only the horizontal one.

Sorry, the docs are not the right place for this.

You can ask a question on the three.js forum and answer your own question if you want to help others.

Mugen87:
Um, I think I agree with @WestLangley.

@Muzzarino There [forum](https://discourse.threejs.org/) offers a "Resources" category that would be perfect for your topic 👍

PR #15722, opened 2019-02-06
ShaderChunk: rename in `default_normal_vertex`

linbingquan:


WestLangley:
beginnormal_vertex
    morphnormal_vertex
    skinnormal_vertex
    defaultnormal_vertex

Thank you for the PR, but I think I would leave the code as-is for consistency. Plus, this PR breaks the examples, and user's existing code.

PR #15750, opened 2019-02-10
THREE JS Editor: Added Context Menu to Objects

Temdog007:
1. Added context menu to objects.
 * **Copy**: Clones the currently selected object and stores it
 * **Cut**: Stores the currently selected object and removes it from the scene
 * **Paste**: Adds the stored object to the current selected object
 * **Clone**: Clones the currently selected object and adds it to the scene
 * **Delete**: Removes the currently selected object

contextmenudarkcontextmenulight

2. Changed Outliner to be resizable.
outlinerresize


looeee:
Nice!

You'll need to add the context menu to Scene as well though. If you cut an object you need to be able to paste it into the scene, at the moment you can only paste it as a child of other objects. The scene context menu only needs the "paste" option for now, an this can be done in a separate PR if you want. 

It would be nice if you could open the same context menu from objects in the viewport as well, but that can be saved for another PR too.

looeee (approved):

Temdog007:
Closing for now. Will be re-implemented in a simpler manner at a later time.

PR #15755, opened 2019-02-11
FBXLoader: fix .setPath()

### LINKED ISSUE #15717: FBX loader .setPath() is having no effect
As title says, FBX loader is not using path defined with .setPath function. Instead it looks for resources in root folder.
IMO, the loader is missing line

`loader.setPath( self.path );`

as line 48, which fixes the issue.

##### Three.js version

- [ ] Dev
- [x] r101
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

hujiulong:
fix #15717

Consistent with other loaders, such as:
https://github.com/mrdoob/three.js/blob/76e6d92085e5a8a6217de4e3271d4c173d87b080/examples/js/loaders/ColladaLoader.js#L25
https://github.com/mrdoob/three.js/blob/76e6d92085e5a8a6217de4e3271d4c173d87b080/examples/js/loaders/GLTFLoader.js#L68

### DIFF
--- a/examples/js/loaders/FBXLoader.js
+++ b/examples/js/loaders/FBXLoader.js
@@ -44,6 +44,7 @@ THREE.FBXLoader = ( function () {
 			var path = ( self.path === undefined ) ? THREE.LoaderUtils.extractUrlBase( url ) : self.path;
 
 			var loader = new THREE.FileLoader( this.manager );
+			loader.setPath( self.path );
 			loader.setResponseType( 'arraybuffer' );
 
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #15769, opened 2019-02-13
fixed render target restoration after shadow map rendering

maccesch:
Completes #15571

### DIFF
--- a/src/renderers/webgl/WebGLShadowMap.js
+++ b/src/renderers/webgl/WebGLShadowMap.js
@@ -220,6 +220,8 @@ function WebGLShadowMap( _renderer, _objects, maxTextureSize ) {
 
 			}
 
+			var currentRenderTarget = _renderer.getRenderTarget();
+
 			_renderer.setRenderTarget( shadowMap );
 			_renderer.clear();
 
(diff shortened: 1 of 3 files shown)


Mugen87:
Can you please remove the build files of your PR? ;)

src/renderers/webgl/WebGLShadowMap.js
```diff
@@ -256,6 +258,8 @@ function WebGLShadowMap( _renderer, _objects, maxTextureSize ) {
 
 		scope.needsUpdate = false;
 
+		_renderer.setRenderTarget( currentRenderTarget );
```
Mugen87: I'm a bit confused about your scoping since you refer to variable that is create inside an if statement. What if there are no lights in the scene? You would set `_renderer.setRenderTarget( undefined ); `, right?

maccesch: Oh! Overlooked that, sorry

maccesch: Fixed

maccesch:
Build files removed

Mugen87 (approved):

mrdoob:
Thanks!

PR #15785, opened 2019-02-14
Add plane orientation example

nikolas:
This adds a new example for users to play with and visualize a plane's normal vector.

2019-02-14-140749_1100x802_scrot

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -183,6 +183,7 @@ var files = {
 		"webgl_materials_video",
 		"webgl_materials_video_webcam",
 		"webgl_materials_wireframe",
+		"webgl_math_orientation_plane",
 		"webgl_math_orientation_transform",
 		"webgl_mirror",
 		"webgl_mirror_nodes",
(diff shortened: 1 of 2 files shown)


WestLangley:
Normal vectors must have unit length. Also, the plane constant is important. The three.js forum may be of help to you if you have questions about these concepts.

Also, this example may already be sufficient: https://threejs.org/examples/webgl_clipping_intersection.html.

Mugen87:
For testing: https://raw.githack.com/nikolas/three.js/plane-visualization/examples/webgl_math_orientation_plane.html

Mugen87:
> Normal vectors must have unit length.

@nikolas This example won't be merged as long as the UI allows the selection of normal vectors with no unit length. If you can come up with a solution for this, you should also ensure the example uses only modules. See the other examples on `dev` as a reference.

WestLangley:
Sorry, I do not think this example has sufficient merit. I would suggest closing it.

PR #15793, opened 2019-02-16
moved cube map specific code from WebGLRenderer.setRenderTarget() to …

maccesch:
…WebGLRenderTargetCube.beforeRender()

Relates to #15571

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1144,6 +1144,12 @@ function WebGLRenderer( parameters ) {
 
 		}
 
+		if ( _currentRenderTarget !== null ) {
+
+			_currentRenderTarget.beforeRender( _gl, properties );
+
+		}
+
 		//
 
 		background.render( currentRenderList, scene, camera, forceClear );
@@ -2528,7 +2534,6 @@ function WebGLRenderer( parameters ) {
 		}
 
 		var framebuffer = _framebuffer;
-		var isCube = false;
 
 		if ( renderTarget ) {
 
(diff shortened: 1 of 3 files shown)


src/renderers/WebGLRenderer.js
```diff
@@ -1144,6 +1144,20 @@ function WebGLRenderer( parameters ) {
...
+
+			var textureProperties = properties.get( _currentRenderTarget.texture );
+			_gl.framebufferTexture2D(
+				_gl.FRAMEBUFFER,
+				_gl.COLOR_ATTACHMENT0,
+				_gl.TEXTURE_CUBE_MAP_POSITIVE_X + _currentRenderTarget.activeCubeFace,
+				textureProperties.__webglTexture, _currentRenderTarget.activeMipMapLevel
+			);
+
+			_currentRenderTarget.beforeRender( _gl, properties );
```
Mugen87: There is still a reference to `beforeRender()`.

src/renderers/WebGLRenderer.js
```diff
@@ -1144,6 +1144,20 @@ function WebGLRenderer( parameters ) {
 
 		}
 
+		if ( _currentRenderTarget !== null && _currentRenderTarget.isWebGLRenderTargetCube ) {
+
+			var textureProperties = properties.get( _currentRenderTarget.texture );
+			_gl.framebufferTexture2D(
```
Mugen87: I guess you need to cache the current active cube face, otherwise  `framebufferTexture2D()` is called at each invocation of `WebGLRenderer.render()` even if the activeCubeFace has not changed.

Mugen87:
Solved via #15808

Issue #14730, opened 2018-08-17
BufferAttribute.dynamic: revisited


WestLangley:
This discussion also applies to `InterleavedBuffer.dynamic`.

It seems that the `dynamic` property is being used for multiple purposes:

1. as a _hint_ as to how frequently the data will be updated, [[docs](https://github.com/mrdoob/three.js/blob/master/docs/api/core/BufferAttribute.html#L61-L69)], and

2. as a flag, that when `true`, allows the updating of only a sub-rage, as specified in the `updateRange` property. [[code](https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLAttributes.js#L66-L96)].

The examples are not consistent here. Even in examples in which the entire range is being updated each frame, sometimes the dynamic property is set to `true`, and sometimes it is not.

This has been discussed [before](https://github.com/mrdoob/three.js/pull/9631#issuecomment-254375571), and the purpose of the property was modified at that time.

What should we advise users to do in setting this property?

/ping @benaadams 
/ping @mattdesl 
/ping @mrdoob 

- [ x ] Dev
- [ x ] r95

aardgoose:
The change for case 2. was really to handle changes in size of the buffer via setArray(), rather than choosing between glBufferData and glBufferSubData.  I'm not sure why a simple sizeChanged flag which would be set in the setArray() method couldn't be used, and would restore the expected properties of the dynamic flag.

Indeed setting a BufferAttribute as dynamic and the attempting to increase the size with setArray() leads to an overflow error as you would expect, which I think is really a bug. Initially a sizeChanged flag was rejected as fragile, but I think the current behaviour is already fragile.

WestLangley:
@aardgoose I have no idea what to do with this. Can you propose an API change that makes sense to you -- with an example, perhaps?

@mrdoob Do you understand this? I don't. At all.

aardgoose:
I think the dynamic flag should be solely for the documented purpose, it is to overloaded to make sense as it is,

attribute.array size changes should be flagged in the setArray() and copy() methods which are the only supported paths to change the array length, so I don't see that as being fragile.

For user's who really want to choose between glBufferData and glBufferSubData a flag defaulting to the later (Khronos recommend) could be added.

The would remove the failure path when changing the size of a dynamic buffer noted above and make updateRange work as expected and not be dependent on the dynamic flag.

WestLangley:
[...]

PR #15798, opened 2019-02-18
add attribute resize tracking

aardgoose:
Untangle overloading of BufferAttribute.dynamic property, re #14730 

.dynamic sets buffer usage expectations as documented.

.updateRange works as documented if dynamic = false. (currently updates entire buffer)

.setArray() and .copy() are safe to use if dynamic = true. (currently produces error)

By default gl.bufferSubData is used for updates except in resize operations.

(a flag to select update mechanism could be added if required )

### DIFF
--- a/src/core/BufferAttribute.js
+++ b/src/core/BufferAttribute.js
@@ -24,6 +24,7 @@ function BufferAttribute( array, itemSize, normalized ) {
 
 	this.dynamic = false;
 	this.updateRange = { offset: 0, count: - 1 };
+	this.resized = false;
 
 	this.version = 0;
 
@@ -53,6 +54,7 @@ Object.assign( BufferAttribute.prototype, {
 
 		}
 
+		this.resized = ( array.length !== this.array.length );
 		this.count = array !== undefined ? array.length / this.itemSize : 0;
 		this.array = array;
 
@@ -71,6 +73,7 @@ Object.assign( BufferAttribute.prototype, {
 	copy: function ( source ) {
 
 		this.name = source.name;
+		this.resized = ( source.array.length !== this.array.length );
 		this.array = new source.array.constructor( source.array );
 		this.itemSize = source.itemSize;
 		this.count = source.count;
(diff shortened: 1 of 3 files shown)


Mugen87:
I'm a bit irritated now. I always thought `three.js` does not allow to resize buffer attributes. If an app needs a bigger attribute buffer for some reasons, it has to create a new attribute. Or create a sufficient sized buffer in the forefront.

aardgoose:
Some developers (not myself) were resizing them it seems, which is what started the initial discussion referred to in #14730.  

After the changes I added which allowed you to drop the arrays from the JS side, after the GPU had been updated ( via the onUpload() callback) - this user 'requirement' came out the woodwork resulted in dynamic flag overloading.

src/core/BufferAttribute.js
```diff
@@ -24,6 +24,7 @@ function BufferAttribute( array, itemSize, normalized ) {
 
 	this.dynamic = false;
 	this.updateRange = { offset: 0, count: - 1 };
+	this.resized = false;
```
WestLangley: I think `.resized` should be private... somehow...

Mugen87:
Does the following error message still make sense?

https://github.com/mrdoob/three.js/pull/15798/files#diff-54f00f0a09b102e7d4bffbf907becd3aR89

src/core/BufferAttribute.js
Mugen87: Maybe `._resized`?

aardgoose: is there a convention to indicate private properties? 
_resized looks reasonable.

WestLangley: Yes, prefixing with underscore has been a convention used in this library.

aardgoose: ok.will do.

aardgoose:
Nothing bad appears to happen passing an empty TypedArray in the count = 0 error case.
Tested on Chrome/FF/IE11 on Windows 10, so maybe a pointless error other than pointing to programming error elsewhere.

PR #15799, opened 2019-02-18
MarchingCubes: Per-blob color support.

chaht01:
### Background ###
Since the material is only applied once to each MarchingCube instance, a color example has only shown colored vertex color according to its vertex position. In some cases, I believe that a color example should provide a method to define its vertex color according to each blob's predefined color.

### Changes ###
Thus, in MarchingCubes.js,  `addBall( ballx, bally, ballz, strength, subtract, colors )` method is now able to define the color of each blob. For compatibility of previous version and example, without `colors` argument, `addBall` computes color according to its vertex position as usual.

### Related ###
[Optimize color calculation ](http://www.geisswerks.com/ryan/BLOBS/blobs.html)

### DIFF
--- a/examples/js/MarchingCubes.js
+++ b/examples/js/MarchingCubes.js
@@ -47,7 +47,7 @@ THREE.MarchingCubes = function ( resolution, material, enableUvs, enableColors )
 
 		this.field = new Float32Array( this.size3 );
 		this.normal_cache = new Float32Array( this.size3 * 3 );
-		this.palette = new Float32Array( this.size3 * 4 );
+		this.palette = new Float32Array( this.size3 * 3 );
 
 		// immediate render mode simulator
 
@@ -86,7 +86,7 @@ THREE.MarchingCubes = function ( resolution, material, enableUvs, enableColors )
 
 	}
 
-	function VIntX( q, offset, isol, x, y, z, valp1, valp2, c_valp1, c_valp2 ) {
+	function VIntX( q, offset, isol, x, y, z, valp1, valp2, c_offset1, c_offset2 ) {
 
 		var mu = ( isol - valp1 ) / ( valp2 - valp1 ),
 			nc = scope.normal_cache;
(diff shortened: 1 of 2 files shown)


Mugen87:
For testing: https://raw.githack.com/chaht01/three.js/multi-color-marchingcubes/examples/webgl_marchingcubes.html

mrdoob:
This PR seems to have a big performance penalty.

On my machine, the [current example](https://threejs.org/examples/webgl_marchingcubes.html) runs at 60fps and after this PR it runs at 20fps.

Mugen87:
I experience this: 60 FPS -> 36 FPS

examples/js/MarchingCubes.js
```diff
@@ -170,6 +187,31 @@ THREE.MarchingCubes = function ( resolution, material, enableUvs, enableColors )
 			field6 = scope.field[ qyz ],
 			field7 = scope.field[ q1yz ];
 
+		var c_field0 = scope.palette
+				.slice( q * 4, q * 4 + 3 )
+				.map( v => v / scope.palette[ q * 4 + 3 ] ), // slicing (r, g, b) part
```
mrdoob: I bet these `.slice()` and `.map()` calls have a lot to do with the performance regression.

examples/js/MarchingCubes.js
```diff
@@ -484,10 +565,36 @@ THREE.MarchingCubes = function ( resolution, material, enableUvs, enableColors )
...
+	this.addBall = function ( ballx, bally, ballz, strength, subtract, colors ) {
 
 		var sign = Math.sign( strength );
 		strength = Math.abs( strength );
+		var userDefineColor = ! ( colors === undefined || colors === null );
+		console.log( ballx, bally, ballz );
+		var ballColor = new THREE.Color( ballx, bally, ballz );
+		if ( userDefineColor ) {
+
+			try {
```
mrdoob: This method is called many times per frame. Adding a `try` inside is for sure going to make it slower.

examples/js/MarchingCubes.js
```diff
@@ -484,10 +565,36 @@ THREE.MarchingCubes = function ( resolution, material, enableUvs, enableColors )
 	// Adds a reciprocal ball (nice and blobby) that, to be fast, fades to zero after
 	// a fixed distance, determined by strength and subtract.
 
-	this.addBall = function ( ballx, bally, ballz, strength, subtract ) {
+	this.addBall = function ( ballx, bally, ballz, strength, subtract, colors ) {
 
 		var sign = Math.sign( strength );
 		strength = Math.abs( strength );
+		var userDefineColor = ! ( colors === undefined || colors === null );
+		console.log( ballx, bally, ballz );
```
mrdoob: Remove all `console.log()`s please.

examples/js/MarchingCubes.js
chaht01: This part was critical to enhancing performance(thank you). I deleted this part and pass start index of color array to `VIntX`, `VIntY` and `VIntZ`.

examples/js/MarchingCubes.js
chaht01: I was not able to move this part out of `addBall` function. However, it seems this part is not that critical to performance.

examples/js/MarchingCubes.js
chaht01: I removed it. Thanks.

chaht01:
Now, `VIntX`, `VIntY`, and `VIntZ` receive parameter start index of the color arrays instead of passing pre-sliced arrays. It would be glad to run [example](https://raw.githack.com/chaht01/three.js/multi-color-marchingcubes/examples/webgl_marchingcubes.html) on your machine and see the result.

looeee:
Much better, I'm not able to see any performance difference with the updated example on my phone (Moto G6). ~27fps before and after.

chaht01:
Could you test the example on another device or browser? I tested it on Iphone6 and result was ~60fps.
The way you can see the other difference between before and after is render of `floor`. 
Recent change renders `floor` color more naturally.


Before

Imgur

After

Imgur

mrdoob:
Thanks!

PR #15413, opened 2018-12-11
SkinnedMesh/Skeleton: Added serialization/deserialization

Mugen87:
PR based on #11603 (thanks to @takahirox)

I have tested this PR with multiple examples (e.g. `webgl_animation_skinning_blending`, `webgl_loader_fbx` or `webgl_loader_collada_skinning`) and the serialization/deserialization seems to work. Unfortunately, I have encountered some problems that were really cumbersome to identify and fix^^.

BTW: Before calling `SkinnedMesh.toJSON()` it might be necessary to call `Object3D.updateMatrixWorld()` in order to ensure correct transformation matrices. This might be necessary if `Object3D.position`, `Object3D.scale` and `Object3D.quaternion` are not composed into `Object3D.matrix` yet.

PR #15819, opened 2019-02-21
Allow for animation deserialization

Temdog007:
**What this PR does**

- Add static `fromJSON` method to `KeyframeTrack`
- Add `KeyframeTracks` file
- Allows editor to save animations on refresh (saves to local storage)

**What is PR does NOT do**

- Save animations to files (Clicking `Export Scene` or `Publish` in the editor will not save the animations to those files)

**Concerns**

- Adding the `KeyframeTracks` file introduced a circular dependency with all of the `KeyframeTrack` files. I don't know if this is acceptable. I could remove the circular dependency if `KeyframeTrack.fromJSON` wasn't static. But then, it wouldn't match `KeyframeTrack.toJSON` which is static. I could change that method to not be static but I'm guessing there is a reason it was static in the first place. I do know that there was already a circular dependency with `Vector3`, `Sphere`, `Box3`, and `Quaternion`.
- I didn't add storing the animations to scenes because I don't know if there is a specific way it should be handled. I figured that a discussion could be had in this PR. There are 2 ways I thought of doing this:

1. Add all animations to `Scene`, update `Scene.toJSON`, and implement`Scene.fromJSON` to parse the animations.
2. Add the associated animation to the `Object3D` and update `ObjectLoader` (or update `Object3D.toJSON` and `Object3D.fromJSON`) to parse the animation.


donmccurdy:
Note that some glTF models will use a custom interpolant for cubic spline curves – see example here: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/InterpolationTest, and [this code](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L1008-L1085). This may cause errors on serialization. Some possible solutions:

- [ ] Resample animation to standard cubic curve.
- [ ] Support (de)serializing glTF CUBICSPLINE curves.
- [ ] Omit these animations.

Temdog007:
**Updates**

- Added (de)serialization of custom interpolations

**Concerns**

- Animations cannot be loaded properly on refresh for glTF models that use SkinnedMeshes because those objects are not (de)serialized properly in ObjectLoader. This should be fixed when #15413 is merged.

examples/jsm/loaders/GLTFLoader.js
```diff
@@ -3182,19 +3194,7 @@ var GLTFLoader = ( function () {
...
-							// must be divided by three to get the interpolant's sampleSize argument.
-
-							return new GLTFCubicSplineInterpolant( this.times, this.values, this.getValueSize() / 3, result );
-
-						};
-
-						// Workaround, provide an alternate way to know if the interpolant type is cubis spline to track.
-						// track.getInterpolation() doesn't return valid value for custom interpolant.
-						track.createInterpolant.isInterpolantFactoryMethodGLTFCubicSpline = true;
+						track.createInterpolant = GLTFCubicSplineInterpolant.FactoryMethod;
```
donmccurdy: It's fine if you want to, but you don't need to include changes to `examples/jsm` at this time. The copy in `examples/js` is enough and the change will be carried through later.

src/animation/KeyframeTrack.js
```diff
@@ -64,11 +64,22 @@ Object.assign( KeyframeTrack, {
 
 			};
 
-			var interpolation = track.getInterpolation();
+			// Assumptions for custom interpolations:
+			// 1) The constructor is a part of the THREE module at the time of serialization
+			// 2) The constructor will be a part of the THREE module at the time of deserialization
+			if ( track.hasCustomInterpolation() ) {
```
donmccurdy: Another significant assumption here is that `constructor.name` will be stable with minification. Unfortunately I don't think that is true, and `THREE.MyClass.constructor.name !== 'MyClass'` is possible, although perhaps it doesn't affect the Editor?

donmccurdy (commented):
Nice solution for serialization, thank you! I wasn't sure that would work.

Looking at the change in b8558906b7bcd2a3d62a5d371c5101af53d3032f, maybe we should merge the PR without custom interpolation first, assuming everyone approves this PR, and then consider the changes for custom interpolation serialization in a separate PR? I worry that part will require more discussion. @mrdoob what do you think?

src/animation/KeyframeTrack.js
Temdog007: The Editor doesn't use the minified version of THREE js so it wouldn't affect the editor. However, I think we'd want this to work with all versions anyway.

I could add a property to store the class name of all `Interpolants` similar to how all instances of `Object3D` have a `type` property to store their names.

examples/jsm/loaders/GLTFLoader.js
Temdog007: Ok

Temdog007:
I'll close this PR and separate this PR into two based on @donmccurdy's [suggestion](https://github.com/mrdoob/three.js/pull/15819#pullrequestreview-209739443).

PR #15824, opened 2019-02-22
Add function for querying whether OrbitControls have updated

JakeGilger:
This PR adds a function to OrbitControls that allows for querying whether orbit controls have updated. I have added and used this on a local fork in my render loop as follows:

```typescript
  private startRenderingLoop() {
  // ...
  (function render() {
      requestAnimationFrame(render);
      if (component.needsRender()) {
        component.animateMeshes(); // animateMeshes is a function I want to call when "intensive" assets are enabled
        component.renderer.render(component.scene, component.camera);
      }
    }());
  }

  private needsRender(): boolean {
    const controlUpdate = this.controls.wasUpdated();
    const result = controlUpdate || this.sceneChanged || this.intensiveAssetsEnabled;
    this.sceneChanged = false;
    return result;

    // return true; // this is the "strawman" case, usually renderer.render() is always called.
  }
```

In my particular use case, I have a game that only needs to be redrawn if there has been some user interaction, like fiddling with the camera or manipulating the scene. In making this rendering change I have no side effects, and the application when idling uses 0% of my GPU according to Task Manager (previously, it would hover around 15% usage).

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -82,6 +82,9 @@ THREE.OrbitControls = function ( object, domElement ) {
 	this.position0 = this.object.position.clone();
 	this.zoom0 = this.object.zoom;
 
+    // for querying if update calls have made changes
+	this.updated = false;
+
 	//
 	// public methods
 	//
@@ -120,6 +123,13 @@ THREE.OrbitControls = function ( object, domElement ) {
 		state = STATE.NONE;
 
 	};
+
+    // Allows querying for whether the controls have been updated since last query.
+    this.wasUpdated = function () {
+        var temp = this.updated;
+        this.updated = false;
+        return temp;
+    };
 
 	// this method is exposed, but perhaps it would be better if we can make it private...
 	this.update = function () {
@@ -210,6 +220,8 @@ THREE.OrbitControls = function ( object, domElement ) {
 				lastPosition.copy( scope.object.position );
 				lastQuaternion.copy( scope.object.quaternion );
 				zoomChanged = false;
+
+				this.updated = true;
 
 				return true;
 


looeee:
You want to use this to render the scene when the user moves the camera with OrbitControls, is that right?

In that case you can use the control's change event:

`controls.addEventListener( 'change', ()=> {  controlUpdate = true } );`

JakeGilger:
Although that could "work," there are reasons why an event handler is not the way to go here. Event handlers run asynchronously to animation frame time (I believe they schedule tasks, but I could be wrong on this one).  Thus, they are either not guaranteed to run every animation frame, or could potentially run multiple times per frame. In doing so we can get stutter effects in both cases as the frame either advances 0 control updates or 2+ updates, respectively. In the latter case we also have wasted computation.

Putting the logic within the render loop ensures exactly one check per frame, meaning no chance for stuttering.

A helpful video that helped me understand a little more about the javascript event loop and why this makes sense (from Jake Archibald, a pretty entertaining speaker and a guy whose hobby is javascript internals): https://vimeo.com/254947206

WestLangley:
Orbit controls was designed to be used as follows when there is no animation loop:
```js
controls.addEventListener( 'change', render ); // call this only in static scenes (i.e., if there is no animation loop)
```

If an animation loop is used, then calling `controls.update()` is still not always necessary.

```js
controls.update(); // only required if controls.enableDamping = true, or if controls.autoRotate = true
```

This is confusing enough as it is.

Since `OrbitControls` is part of the examples, you are free to modify your copy as you see fit.

JakeGilger:
> If an animation loop is used, then calling `controls.update()` is still not always necessary.

Definitely. I wholeheartedly agree with you (and the comment atop the update function wondering whether it should be made private). This is why I am not calling `controls.update()` within my loop at all.

Rather, I am using this function to check and see if any internal updates to the OrbitControls were made. Does it make sense that one should be able to query the OrbitControls in this manner?

WestLangley:
In this PR, you are proposing this in the `update()` method:
```js
scope.dispatchEvent( changeEvent );
this.updated = true;
```

I think having both is confusing for users. That is my opinion.

gkjohnson:
> Event handlers run asynchronously to animation frame time (I believe they schedule tasks, but I could be wrong on this one).

This isn't the case with the event dispatcher that OrbitControls is using. The THREE EventDispatcher guarantees that the events are fired immediately (which may be confusing on its own and should be clarified in the docs), as you can see in [the dispatchEvent source](https://github.com/mrdoob/three.js/blob/dev/src/core/EventDispatcher.js#L60). That should mean you can do this manually as @looeee suggested without worrying about the events being delayed for some reason:

```js
let controlsNeedUpdate = false;
controls.addEventListener( 'change', () => controlsNeedUpdate = true )

function render() {

  if (controlsNeedUpdate ) controls.update();
  controlsNeedUpdate = false;
  
  // ...

}
```

JakeGilger:
Oh! Thanks for that link @gkjohnson. I agree with you that it is synchronous after checking that source. Yeah some clarity in the docs would probably be helpful that the eventDispatcher (and thus the handler attached to it) is synchronous.

Thanks, I feel ok closing this PR now that my concern was addressed :)

P.S. I'm not so sure my loop would end up looking like that suggestion... On change events I don't need to make another .update() call now do I ;)

JakeGilger:
For reference, here's the render loop with the GPU-saving optimization (in case anyone wishes to use it):

```typescript
export class MyComponent {
  // These determine whether we need to render every animation frame.
  private intensiveAssetsEnabled: boolean;
  private controlsUpdated: boolean;
  private sceneUpdated: boolean;

  private initControls(): void {
    this.controls = new THREE.OrbitControls(this.camera);
    // init controls settings
    this.controls.addEventListener( 'change', () => { this.controlsUpdated = true } );
  }

  private startRenderingLoop(): void {
    /* Renderer */
    this.renderer = new THREE.WebGLRenderer({ alpha: true, canvas: this.canvas });
    // init renderer settings

    let component = this;
    (function render() {
      requestAnimationFrame(render);
      if (component.needsRender()) {
        // ...
        component.renderer.render(component.scene, component.camera);
      }
    }());
  }

  private needsRender(): boolean {
    const result = this.controlsUpdated || this.sceneUpdated || this.intensiveAssetsEnabled;
    this.controlsUpdated = false;
    this.sceneUpdated = false;
    return result;
  }
}
```

gkjohnson:
Sorry, you're right!

I've more or less been using the `change` event to set a dirty flag or to schedule a render in quite a few of my applications and it works just fine.

Glad to help!

PR #15835, opened 2019-02-25
OrbitControls. - Ignore keyboard event if some other input element has focus.

kkolinko:
It is a fix for bug #15834.

Issue #15834, opened 2019-02-25
OrbitControls eats cursor movement keyboard events for other input fields


kkolinko:
##### Description of the problem

OrbitControls registers a listener for keydown events on a window and processes the events regardless of which input element has focus. When a user uses keyboard to move cursor in an input field, the events for the left and right arrow keys of keyboard are handled by OrbitControls and do not move the cursor. Better behaviour would be to ignore the event if some other input element has the focus.

Steps to reproduce, using current `dev` branch and examples:

1. Open documentation (`docs\index.html`) and navigate to `TorusGeometry` document.
2. In the iframe that displays an example of a torus, click on "THREE.TorusGeometry" button to open a properties editor for the object.
3. Click on one of input boxes in the properties editor, e,g, on the value "10" for radius.
4. Try to use your keyboard to move cursor position to the left and to the right in the input field.
Expected behaviour: the cursor moves, as you press left or right arrows on your keyboard.
Actual behaviour: the image of the torus moves.

The behaviour of `r101` is to suppress all keydown events, regardless of key code (known as #15810, already fixed in `dev`). The behaviour of `r100` was to move both the cursor and the object.

This issue could be fixed by either a) registering the keydown listener on the domElement instead of a window, b) filtering the event according to the actual value of `document.activeElement`. The "b)" way seems to provide better compatibility with older versions of three.js.

##### Three.js version

- [x] Dev
- [x] r101

##### Browser

- [x] Chrome
- [x] Firefox

##### OS

- [x] Windows

Mugen87:
The former behavior of `OrbitControls` was changed because of #15619. The original PR introduced a problem since it prevents the usage of any keys when `OrbitControls` was active. #15693 fixed this issue by preventing the default action of a key press only for used keys.

I think we should keep the current behavior of `OrbitControls`. The perfect solution that makes everybody happy is obviously not possible. Any modifications in context of key input should be done on application level.

kkolinko:
> Any modifications in context of key input should be done on application level.

Nice to have, but it means that an application has to either call the existing methods used in this feature, or have to re-implement them. The current `handleKeyDown()` method, or the `pan()` method that is called by it, are not public API. (The `pan()` method is non trivial, and calls other non-public methods).

Mugen87:
> Nice to have, but it means that an application has to either call the existing methods used in this feature, or have to re-implement them.

TBH, I don't understand the relation of this to the actual title of your issue.

[...]

PR #15836, opened 2019-02-25
Examples. - Allow domElement controlled by OrbitControls to receive focus.

kkolinko:
Allow domElement controlled by OrbitControls to receive focus, by setting tabindex attribute on it.

This is related to PR #15835 (issue #15834). Unless those are applied first OrbitControls does not care what element has the focus.

This PR updates examples and live examples in documentation. I skip examples where either of the following is true:

1.  `OrbitControls` do not process keyboard events (`controls.enablePan` is false), or
2.  `OrbitControls` handle the whole document (one-argument constructor is used, e.g. `new THREE.OrbitControls( camera )`).


Mugen87:
Closing in favor of #16676

PR #15848, opened 2019-02-27
Gltf lights in documentation

pranav-ap:
Hi,

Are you sure GLTF supports lights?  Don McCurdy says otherwise in the following thread.

https://stackoverflow.com/questions/50879461/lighting-to-add-lights-in-gltf-2-0

Regards,
Pranav


donmccurdy:
Punctual lights (spot, point, directional) have been added to glTF since then – I just corrected my comment on Stack Overflow, thanks. 😅

https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual

donmccurdy:
Closing since the documentation here is OK.

Issue #15704, opened 2019-02-04
Parallax-Corrected CubeMap


sneha-belkhale:
So I was creating a room scene and wanted to attach a cube texture environment map to the ground ( taken with the three.js CubeCamera ). The environment map showed up but since environment maps are used under the assumption that all objects are at an infinite distance, there was a major parallax issue that made the env map unusable for my scene.

This article covers the topic pretty clearly: 
https://www.clicktorelease.com/blog/making-of-cruciform/

I think it would be nice to include this little patch for those who desire it. I've already wrote a patch for the current physical shader, which simply applies the correction if the #define PARALLAX_CORRECT exists. The user would have to specify two uniforms, maybe like so : 

```
var parallaxCorrectMat = new THREE.MeshPhysicalMaterial({
envMap: someEnvMap,
cubeMapSize: new Vector3(100,200,100),
cubeMapPos: new Vector3(0,0,0)
})
```

Let me know your thoughts ---

Sneha

sneha-belkhale:
for clarity - I made a demo where you can see the difference between the parallax corrected / default cube maps. 

https://codercat.tk/tmp/parallaxCubeMap/

Kif11:
I would like to have similar functionality for creating realistic ground reflections.

sneha-belkhale:
cool, can one of the threejs guys let me know if this sounds good to add? I already have the basic code ready for a PR.

Mugen87:
I've a question about the article. It says:

> We want to correct the normal that we use to lookup the cubemap. A cubemap assumes that the images rendered in the different faces are far far away. What we do with the parallaxCorectNormal function is take that vector, and look where it intersects a volume that defines the scene. **I'm using a cuboid because it accurately models the hall model.**

What if a cuboid is not suitable for other scenarios?

sneha-belkhale:
@Mugen87 - 

Yeah so this specific implementation only works for cuboids. 

It's not optimal but with cuboids alone you could cover a range of scenarios by segmenting your envmap into a set of cuboids.

looeee:
> What if a cuboid is not suitable for other scenarios?

Cuboids would work for the majority of indoor scenes, since most rooms are cuboids or combinations of cuboids (e.g. hallways with a 90 degree bend).

Mugen87:
@looeee Yeah, that's true.

[...]

PR #15897, opened 2019-03-03
Examples: Box Projected CubeMap Environment Mapping

sneha-belkhale:
I made an example demo of Box Projected CubeMap Environment Mapping, as a follow up to the discussion in #15704. 

The scene uses a `CubeCamera` to generate an environment map - and applies the environment map to a ground mesh that has a `MeshPhysicalMaterial` with custom shader injections that implement Box Projection, following this article: 
http://devlog-martinsh.blogspot.com/2011/09/box-projected-cube-environment-mapping.html

There is also a toggle control to contrast what the environment map would look like without Box Projection. 

screenshot of the scene: 

screen shot 2019-03-04 at 12 48 45 am

Let me know if the shader injection snippets should be moved to another file -- although I am not sure where it should go .. maybe in the `examples/js/shaders` directory?


Mugen87:
https://raw.githack.com/sneha-belkhale/three.js/bpcem/examples/webgl_box_projected_cubemap_environment_mapping.html

examples/webgl_box_projected_cubemap_environment_mapping.html
```diff
@@ -0,0 +1,484 @@
...
+
+				var defaultMat = new THREE.MeshPhysicalMaterial({
+					roughness: 1,
+					envMap: cubeCamera.renderTarget.texture,
+					roughnessMap: rMap,
+				});
+
+				var boxProjectedMat = new THREE.MeshPhysicalMaterial({
+					color: new THREE.Color('#ffffff'),
+			    roughness: 1,
```
Mugen87: There are some minor formatting issues in your file. Please ensure to use only hard tabs.

examples/webgl_box_projected_cubemap_environment_mapping.html
```diff
@@ -0,0 +1,484 @@
...
+					shader.fragmentShader = shader.fragmentShader.replace(
+						'#include <envmap_physical_pars_fragment>',
+						envmapPhysicalParsReplace
+					);
+
+					materialShader = shader;
+
+				};
+
+			  groundPlane = new THREE.Mesh( new THREE.PlaneGeometry( 200, 100, 100 ), boxProjectedMat );
```
Mugen87: Please use `PlaneBufferGeometry` here.

examples/webgl_box_projected_cubemap_environment_mapping.html
```diff
@@ -0,0 +1,484 @@
...
+				groundPlane.rotateX( - Math.PI / 2 );
+				groundPlane.position.set( 0, - 49, 0 );
+				groundPlane.geometry.computeVertexNormals();
```
Mugen87: Why are you computing the vertex normals here?

examples/webgl_box_projected_cubemap_environment_mapping.html
```diff
@@ -0,0 +1,484 @@
...
+				cameraControls.target.set( 0, -10, 0 );
+				cameraControls.maxDistance = 400;
+				cameraControls.minDistance = 10;
+				cameraControls.update();
+
+				// cube camera for environment map
+				cubeCamera = new THREE.CubeCamera( 1, 1000, 512 );
+				cubeCamera.renderTarget.texture.generateMipmaps = true;
+				cubeCamera.renderTarget.texture.minFilter = THREE.LinearMipMapLinearFilter;
+				cubeCamera.renderTarget.texture.magFilter = THREE.LinearFilter;
```
Mugen87: `THREE.LinearFilter` should be already the default value. No need to set it then.

examples/webgl_box_projected_cubemap_environment_mapping.html
sneha-belkhale: My bad, forgot to remove this line, I think it's remnants from when I was using another geo.

Mugen87 (approved):

Mugen87:
Oh, I've noticed on more thing: I would rename the example to `webgl_box_projected_cubemap` in order to avoid the line breakage here:

image

sneha-belkhale:
@Mugen87 yes agreed, thanks for reviewing:)

examples/webgl_materials_envmaps_bpcem.html
```diff
@@ -0,0 +1,482 @@
...
+			var FAR = 800;
+
+			var camera, scene, renderer;
+
+			var cameraControls;
+
+			var groundPlane, smallSphere, wallMat;
+
+			init();
+			animate();
```
WestLangley: You shouldn't need an animation loop for a static scene. Instead, do this:
```js
controls.addEventListener( 'change', render );
```

WestLangley:
How about
`webgl_materials_envmaps_bpcem` -> `webgl_materials_envmaps_box_projected`

Mugen87:
If it fits in one line I'm okay with that^^

WestLangley:
>If it fits in one line I'm okay with that^^

`webgl_materials_envmaps_parallax` is shorter.

examples/webgl_materials_envmaps_bpcem.html
sneha-belkhale: oh nice, never thought of it!

sneha-belkhale:
`webgl_materials_envmaps_box_projected` didn't fit on one line, so I went with `webgl_materials_envmaps_parallax` ~

mrdoob:
Thanks!

PR #15909, opened 2019-03-05
Subsurface lighting example

wizgrav:
playing with light is fun

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -308,6 +308,7 @@ var files = {
 		"webgl_raymarching_reflect",
 		"webgl_shadowmap_pcss",
 		"webgl_simple_gi",
+		"webgl_subsurface_lighting",
 		"webgl_tiled_forward",
 		"webgl_worker_offscreencanvas"
 	],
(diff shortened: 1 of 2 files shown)


donmccurdy:
Demo: https://rawcdn.githack.com/wizgrav/three.js/dev/examples/webgl_subsurface_lighting.html

WestLangley:
We already have this: https://threejs.org/examples/webgl_materials_translucency.html

I am reluctant to support examples that inject non PBR-based code into the PBR materials.

wizgrav:
@WestLangley the surface scattering itself was just to highlight the concept. You can think about it as a king of procedural light cookie. There's no similar example and it basically highlights cases where traditional shadow mapping completely breaks. It's not something that should go in core obviously, just food for thought

Mugen87:
> I am reluctant to support examples that inject non PBR-based code into the PBR materials.

Agreed. Even if the PR is not merged, I'd like to promote the [forum](https://discourse.threejs.org/) as an excellent place for sharing such interesting work. Topics in the `Resource` category often demonstrate similar features/enhancements for `three.js`.

PR #12456, opened 2017-10-21
ImageBitmapLoader: Support Firefox's partial createImageBitmap implementation.

donmccurdy:
Update for ImageBitmapLoader:

1. Add Firefox support. The current implementation only works in Chrome, as Firefox does not accept an `options` argument to `createImageBitmap()`. Also depends on https://github.com/mrdoob/three.js/pull/12455.
2. ~~Fetch and create the ImageBitmap in a worker by default. I believe (and please correct me if I'm wrong) there isn't much value in creating an ImageBitmap on the main thread.~~

***

> **UPDATE:** The performance measurements below were done in Chrome using a WebWorker. Because it turns out `createImageBitmap` is already async, that Web Worker just adds overhead, and so actual results should be at least as good what I've reported here.

This can be used with any loader through a `THREE.Loader.Handlers` hook:

```js
class CustomTextureLoader {

	constructor () {

		this.loader = new THREE.ImageBitmapLoader();

	}

	load ( url, onLoad, onProgress, onError ) {

		this.loader.load( url, ( imageBitmap ) => {

			onLoad( new THREE.CanvasTexture( imageBitmap ) );

		}, onProgress, onError );

	}

}

THREE.Loader.Handlers.add( /\.(png|jpg|jpeg)$/, new CustomTextureLoader() );
```

## Performance: Load and upload time (ms)

| Preview | [Boombox](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoomBox) | Boombox + ImageBitmapLoader |
|---|---|---|
| screenshot | screen shot 2017-10-20 at 10 47 29 pm | screen shot 2017-10-20 at 10 47 34 pm |

| Preview | [DamagedHelmet](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/DamagedHelmet) | DamagedHelmet + ImageBitmapLoader |
|---|---|---|
| screenshot | screen shot 2017-10-20 at 10 47 38 pm | screen shot 2017-10-20 at 10 47 43 pm |
> In all charts, green is loading duration (non-blocking), and red is duration of first render (blocking), in milliseconds. Results measured in Chrome.

[...]

PR #15910, opened 2019-03-05
Add workaround for FireFox in ImageBitmapLoader

takahirox:
FireFox has an unrecognized `ImageBitmap` options issue in `createImageBitmap()`.

If you pass an second argument for options, it causes an error.

```javascript
createImageBitmap(new ImageData(1,1), {});
TypeError: 2 is not a valid argument count for any overload of Window.createImageBitmap.
```

Even if you pass `undefined`, it still causes.

```javascript
createImageBitmap(new ImageData(1,1), undefined);
TypeError: 2 is not a valid argument count for any overload of Window.createImageBitmap.
```

So the following line in `ImageBitmapLoader` always causes an error on FireFox.

```javascript
return createImageBitmap( blob, scope.options );
```

I heard that it may still take time to fix this issue.

So what do you think of adding the following workaround to `ImageBitmapLoader` for FireFox? In case users don't need options for `ImageBitmap` they can use `ImageBitmapLoader` even on FireFox.

```javascript
if ( scope.options === undefined ) {

	// Workaround for FireFox. It causes an error if you pass options.
	return createImageBitmap( blob );

} else {

	return createImageBitmap( blob, scope.options );

}
```

### DIFF
--- a/src/loaders/ImageBitmapLoader.js
+++ b/src/loaders/ImageBitmapLoader.js
@@ -71,7 +71,16 @@ ImageBitmapLoader.prototype = {
 
 		} ).then( function ( blob ) {
 
-			return createImageBitmap( blob, scope.options );
+			if ( scope.options === undefined ) {
+
+				// Workaround for FireFox. It causes an error if you pass options.
+				return createImageBitmap( blob );
+
+			} else {
+
+				return createImageBitmap( blob, scope.options );
+
+			}
 
 		} ).then( function ( imageBitmap ) {
 


mrdoob:
Thanks!

takahirox:
Thanks for merging. I'll keep an eye on this FireFox issue and let's remove the workaround when it's fixed.

takahirox:
FYI, I just noticed that this workaround was added before once #12456, but after that somehow reverted cb20c474f8592f21a396604d123519e595e79987

mrdoob:
Not sure...

bastienrobert:
It started to refuse again here too. But only when Devtools are open under Firefox 80.0.1.

bastienrobert:
When you try to access [this example](https://threejsfundamentals.org/threejs/threejs-load-gltf-rotate-cars.html) using Firefox, everything is ok. But if you open the Responsive Design Mode with iPhone device, it shows a `TypeError: Window.createImageBitmap: 2 is not a valid argument count for any overload`.

Firefox version is 80+.

image

mrdoob:
@bastienrobert I see. Please, report the issue to Mozilla.

PR #15911, opened 2019-03-05
Export types as namespace

Methuselah96:
The types should be exported as a namespace since three.js is a UMD library. This is in line with how they are currently [exported from @types/three](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/three/index.d.ts) and fixes its usages in other @types libraries. This is required for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33477 to get merged.


bhouston:
makes perfect sense.  sorry about missing this.

mrdoob:
Thanks!

Issue #11746, opened 2017-07-11
Non-blocking assets loaders


fernandojsg:
As discussed in https://github.com/mrdoob/three.js/issues/11301 one of the main problems that we have in WebVR, although is annoying in non-VR experiences too, is blocking the main thread while loading assets. 

With the recent implementation on [link traversal](https://blog.mozvr.com/link-traversal/) in the browser non-blocking loading is a must to ensure a satisfying user experience. If you jump from one page to another and the target page start to load assets blocking the mainthread, it will block the render function so no frames will be submitted to the headset and after a small period of grace the browser will kick us out from VR and it will require the user to take out the headset, click enter VR again (user gesture required to do so) and go back to the experience.

Currently we can see two implementations of non-blocking loading of OBJ files:
- (1) Using webworkers to parse the obj file and then return the data back to the main thread [WWOBJLoader](https://threejs.org/docs/index.html#examples/loaders/WWOBJLoader2):
Here the parsing is done concurrently and you could have several workers at the same time. The main drawback is that once you've loaded the data you need to send the payload back to the mainthread to reconstruct the THREE objects instances and that part could block the main thread:
https://github.com/kaisalmen/WWOBJLoader/blob/master/src/loaders/WWOBJLoader2.js#L312-L423

- (2) Mainthread promise with deferred parsing using setTimeOut:[Oculus ReactVR](https://github.com/facebook/react-vr/tree/master/ReactVR/js/Loaders/WavefrontOBJ): This loader keeps reading lines using small time slots to prevent blocking the main thread by calling setTimeout: https://github.com/facebook/react-vr/blob/master/ReactVR/js/Loaders/WavefrontOBJ/OBJParser.js#L281-L298
With this approach the loading will be slower as we're just parsing some lines on each time slot, but the advantage is that once the parsing is completed, we'll have the THREE objects ready to use without any additional overhead.

Both has their pros and cons and I'm honestly not an expert of webworkers to evalute that implementation but It's an interesting discussion that ideally would lead to a generic module that could be used to port the loaders to a non-blocking version.

Any suggestions?

/cc @mikearmstrong001 @kaisalmen @delapuente @spite

spite:
You can have a Promise+worker+incremental based loader (a bit like a mix of both points)

[...]

PR #15919, opened 2019-03-06
[WIP] Suggestion: Add a way to enable ImageBitmapLoader to GLTFLoader

takahirox:
**Background**

`ImageBitmap` has a very good advantage to regular images, the main thread blocking time caused by uploading to texture is shorter. For example, VR app will have better presence with `ImageBitmap`.

But `ImageBitmap` has two problems in Three.js. First, compatibility. IE, Edge, and Safari don't seem to support yet.

https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap

And FireFox doesn't handle `ImageBitmap` options now #15910. We tried to fix in FireFox side https://github.com/mrdoob/three.js/issues/11746#issuecomment-414845303 but I heard it may still take time to fix.

Second, `ImageBitmap` doesn't fit to some current Three.js `Texture` API. `flipY` and `premultiplyAlpha` should be specified on `ImageBitmap` creation. Changing `texture.flipY/premultiplyAlpha` later has no effect.

```javascript
texture.flipY = ! texture.flipY;
texture.premultiplyAlpha = ! texture.premultiplyAlpha;
texture.needsUpdate = true; // no effect if texture.image is ImageBitmap
```

So it's too early to switch from `ImageLoader` to `ImageBitmapLoader` as default now. But I think there are users who want to use `ImageBitmap` where it works fine. Especially they may want to use in case they don't need the `ImageBitmap` options because it works even on FireFox. Loading glTF assets may be a good use case. glTF core spec doesn't require `flipY` and `premultiplyAlpha`.

But `TextureLoader` uses `ImageLoader` inside. And there is no way to switch to `ImageBitmapLoader`.

**Suggestion**

I'd like to suggest providing users a way to switch from `ImageLoader` to `ImageBitmapLoader` in `TextureLoader` and `GLTFLoader` under their responsibility.

As I mentioned above, using `ImageBitmap` in Three.js may be problematic for some use cases. So I think it's better that users, who understand the limitation/restriction, explicitly enable `ImageBitmapLoader` rather than we automatically detect `ImageBitmap` availability and switch.

Suggested API.

```javascript
var loader = new THREE.TextureLoader();
if ( 'createImageBitmap' in window ) loader.enableImageBitmapLoader();
loader.load( ... ); // using ImageBitmapLoader instead of ImageLoader inside

var gltfLoader = new THREE.GLTFLoader().enableImageBitmapLoader();
if ( 'createImageBitmap' in window ) gltfLoader.enableImageBitmapLoader();
gltfLoader.load( ... );  // using ImageBitmapLoader instead of ImageLoader inside
```

What do you think? Sounds hacky? Any better solutions?

I haven't updated the document yet in this PR. I'll update and remove "[WIP]" from the title if this suggestion looks ok.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -12,6 +12,7 @@ THREE.GLTFLoader = ( function () {
 
 		this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
 		this.dracoLoader = null;
+		this._useImageBitmapLoader = false;
 
 	}
 
@@ -118,6 +119,13 @@ THREE.GLTFLoader = ( function () {
 
 		},
 
+		enableImageBitmapLoader: function () {
+
+			this._useImageBitmapLoader = true;
+			return this;
+
+		},
+
 		parse: function ( data, path, onLoad, onError ) {
 
 			var content;
@@ -214,7 +222,8 @@ THREE.GLTFLoader = ( function () {
 
 				path: path || this.resourcePath || '',
 				crossOrigin: this.crossOrigin,
-				manager: this.manager
+				manager: this.manager,
+				enableImageBitmapLoader: this._useImageBitmapLoader
 
 			} );
 
@@ -1604,6 +1613,8 @@ THREE.GLTFLoader = ( function () {
 
 		this.textureLoader = new THREE.TextureLoader( this.options.manager );
 		this.textureLoader.setCrossOrigin( this.options.crossOrigin );
+
+		if ( options.enableImageBitmapLoader ) this.textureLoader.enableImageBitmapLoader();
 
 		this.fileLoader = new THREE.FileLoader( this.options.manager );
 		this.fileLoader.setResponseType( 'arraybuffer' );
--- a/src/loaders/TextureLoader.js
+++ b/src/loaders/TextureLoader.js
@@ -4,6 +4,7 @@
 
 import { RGBAFormat, RGBFormat } from '../constants.js';
 import { ImageLoader } from './ImageLoader.js';
+import { ImageBitmapLoader } from './ImageBitmapLoader.js';
 import { Texture } from '../textures/Texture.js';
 import { DefaultLoadingManager } from './LoadingManager.js';
 
@@ -11,6 +12,7 @@ import { DefaultLoadingManager } from './LoadingManager.js';
 function TextureLoader( manager ) {
 
 	this.manager = ( manager !== undefined ) ? manager : DefaultLoadingManager;
+	this._useImageBitmapLoader = false;
 
 }
 
@@ -22,7 +24,8 @@ Object.assign( TextureLoader.prototype, {
 
 		var texture = new Texture();
 
-		var loader = new ImageLoader( this.manager );
+		var loader = this._useImageBitmapLoader
+			? new ImageBitmapLoader( this.manager ) : new ImageLoader( this.manager );
 		loader.setCrossOrigin( this.crossOrigin );
 		loader.setPath( this.path );
 
@@ -45,6 +48,13 @@ Object.assign( TextureLoader.prototype, {
 		}, onProgress, onError );
 
 		return texture;
+
+	},
+
+	enableImageBitmapLoader: function () {
+
+		this._useImageBitmapLoader = true;
+		return this;
 
 	},
 


donmccurdy:
Note that workarounds like this are possible today:

```js
class CustomTextureLoader {

	constructor () {

		this.loader = new THREE.ImageBitmapLoader();

	}

	load ( url, onLoad, onProgress, onError ) {

		this.loader.load( url, ( imageBitmap ) => {

			onLoad( new THREE.CanvasTexture( imageBitmap ) );

		}, onProgress, onError );

	}

}

```

Would that be sufficient, without modifying the different loaders?

takahirox:
Thanks, I was missing `Loader.Handlers`. I didn't really like adding the same new method across multiple loaders. `Loader.Handlers` approach should be nicer.

I don't see `Loader.Handlers` in the doc (unless I'm missing). I'll add with `GLTFLoader` + `ImageBitmapLoader` example code.

Mugen87:
> I don't see Loader.Handlers in the doc (unless I'm missing).

No, it's actually missing. Let me add some information about `Loader.Handlers` to the `Loader` page.

Mugen87:
> I'll add with GLTFLoader + ImageBitmapLoader example code.

Sry, I've missed that you want to make the PR^^. Anyway, the descriptions from the closed one might be a bit of a help ;)

netpro2k:
One problem with using "handlers" for this is that all images in a glb end up being blob urls, so there would be no way to use ImageBitmapLoader for those. In general we have found a bunch of cases where we want to take further control over how things are loaded in GLTFLoader, so something that exposes that would be super useful.

donmccurdy:
That's a good point. This is also a problem for using GLTFLoader in Node.js, where Blob URLs aren't available. If we can come up with an API that's more general than `enableImageBitmapLoader()`, ideally flexible enough for Node.js, ImageBitmap, compressed textures, etc., I'm fine with adding a method here. I'd prefer an approach that doesn't rely on the global `THREE` namespace, but instead passes something into the loader.

Naive example:

```js
const loader = new THREE.GLTFLoader();
loader.setTextureProvider( new CustomTextureProvider() );

class CustomTextureProvider {
  /**
    * @param {THREE.GLTFParser} parser
    * @param {GLTF.Texture} textureDef
    * @returns {Promise<THREE.Texture>}
    */
  load( parser, textureDef )  {

    const imageDef = parser.json.images[ textureDef.source ];
    const textureLoader = new THREE.TextureLoader();
    return new Promise( ( resolve ) => textureLoader.load( imageDef.uri, resolve ) );

  }
}
```

takahirox:
> One problem with using "handlers" for this is that all images in a glb end up being blob urls, so there would be no way to use ImageBitmapLoader for those.

Doesn't this work?

```javascript
THREE.Loader.Handlers.add( /^blob:.*$/, new CustomTextureLoader() );
```

netpro2k:
> Doesn't this work?
> 
> ```js
> THREE.Loader.Handlers.add( /\.(png|jpg|jpeg)$/, new CustomTextureLoader() );
> THREE.Loader.Handlers.add( /^blob:.*$/, new CustomTextureLoader() );
> ```

Haven't tried, but It's possible that would work in a well constrained example, but in general I don't think that's a handler you would want to register globally since there are many non-texture things that can be blob urls. In Hubs for example all parts of sketchfab models (including the gltf and bin) end up being blob urls since we read them from a zip file on the client. 

Something like Don suggested above where we actually get the texture definition would be really nice. If we expand it a bit I bet this would be useful for other types of entries like meshes, nodes, etc as well.

Mugen87:
Quick update: `THREE.Loader.Handlers` has been removed with the last release `R109`. Instead, you can now use a new API of `LoadingManager` for loader configuration. 

- `LoadingManager.addHandler()`
- `LoadingManager.removeHandler()`
- `LoadingManager.getHandler()`

In this way, you do not register handler globally anymore but on loading manager level.

Mugen87:
Closing. Please open a new issue if `LoadingManager.addHandler()` is not sufficient to configure the usage of `ImageBitmapLoader` in `GLTFLoader`.

Please discuss the idea of `GLTFLoader.setTextureProvider()` in a separate issue if it's still relevant (see https://github.com/mrdoob/three.js/pull/15919#issuecomment-471082093).

PR #15922, opened 2019-03-07
SVGLoader: Added fill-opacity style support.

flo-Ty:
Working example here: https://btevc.cf/repos/three.js/examples/webgl_loader_svg_update_1.html

### DIFF
--- a/examples/js/loaders/SVGLoader.js
+++ b/examples/js/loaders/SVGLoader.js
@@ -701,7 +701,7 @@ THREE.SVGLoader.prototype = {
 			if ( node.style.fill !== '' ) style.fill = node.style.fill;
 
 			if ( node.hasAttribute( 'fill-opacity' ) ) style.fillOpacity = node.getAttribute( 'fill-opacity' );
-      if ( node.style.fillOpacity !== '' ) style.fillOpacity = node.style.fillOpacity;
+			if ( node.style.fillOpacity !== '' ) style.fillOpacity = node.style.fillOpacity;
 
 			return style;
 
--- a/examples/js/loaders/SVGLoader.js
+++ b/examples/js/loaders/SVGLoader.js
@@ -8,8 +8,6 @@ THREE.SVGLoader = function ( manager ) {
 
 	this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
 
-	this.rootNode = null;
-
 };
 
 THREE.SVGLoader.prototype = {
@@ -1018,14 +1016,16 @@ THREE.SVGLoader.prototype = {
 		console.time( 'THREE.SVGLoader: Parse' );
 
 		parseNode( xml.documentElement, { fill: '#000' } );
-		this.rootNode = xml.documentElement;
+		
+		var data = { paths: paths, xml: xml.documentElement };
 
 		// console.log( paths );
 
 
 		console.timeEnd( 'THREE.SVGLoader: Parse' );
 
-		return paths;
+
+		return data;
 
 	}
 
--- a/examples/js/loaders/SVGLoader.js
+++ b/examples/js/loaders/SVGLoader.js
@@ -7,6 +7,9 @@
 THREE.SVGLoader = function ( manager ) {
 
 	this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
+	
+	this.nodeMap = new Map();
+	this.rootNode = null;
 
 };
 
@@ -1016,6 +1019,7 @@ THREE.SVGLoader.prototype = {
 		console.time( 'THREE.SVGLoader: Parse' );
 
 		parseNode( xml.documentElement, { fill: '#000' } );
+		this.rootNode = xml.documentElement;
 
 		// console.log( paths );
 
--- a/examples/js/loaders/SVGLoader.js
+++ b/examples/js/loaders/SVGLoader.js
@@ -122,6 +122,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 
 			var point = new THREE.Vector2();
 			var control = new THREE.Vector2();
@@ -545,6 +546,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 			path.moveTo( x + 2 * rx, y );
 			path.lineTo( x + w - 2 * rx, y );
 			if ( rx !== 0 || ry !== 0 ) path.bezierCurveTo( x + w, y, x + w, y, x + w, y + 2 * ry );
@@ -591,6 +593,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 
 			var index = 0;
 
@@ -623,6 +626,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 
 			var index = 0;
 
@@ -645,6 +649,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 			path.subPaths.push( subpath );
 
 			return path;
@@ -663,6 +668,7 @@ THREE.SVGLoader.prototype = {
 
 			var path = new THREE.ShapePath();
 			path.color.setStyle( style.fill );
+			path.fillOpacity = style.fillOpacity;
 			path.subPaths.push( subpath );
 
 			return path;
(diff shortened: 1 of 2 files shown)


examples/js/loaders/SVGLoader.js
```diff
@@ -694,6 +700,9 @@ THREE.SVGLoader.prototype = {
 			if ( node.hasAttribute( 'fill' ) ) style.fill = node.getAttribute( 'fill' );
 			if ( node.style.fill !== '' ) style.fill = node.style.fill;
 
+			if ( node.hasAttribute( 'fill-opacity' ) ) style.fillOpacity = node.getAttribute( 'fill-opacity' );
+      if ( node.style.fillOpacity !== '' ) style.fillOpacity = node.style.fillOpacity;
```
WestLangley: formatting: needs spaces -> tabs

yomboprime:
Seems all right to me 👍

examples/js/loaders/SVGLoader.js
BTEVC: Ah, missed that. Updated now.

yomboprime:
Now that I think on it, the SVGLoader code is adding the `fillOpacity` member to the `ShapePath` instances.
Is that okay?

WestLangley:
I am not sure what is best... Someone who is more-familiar with this will have to speak up.

yomboprime:
(I think @Mugen87 can throw light on this.)

According to mrdoob: https://github.com/mrdoob/three.js/pull/15170#issuecomment-436099412

It seems the `SVGLoader` should not export the SVG tags, but better the SVG node itself, in the object `ShapePath.userData`.

So the SVGLoader should put `path.userData = { node: node }`. And `fillOpacity` should be obtained by the user  with `path.userData.node.getAttribute( 'fill-opacity' )` instead of `path.fillOpacity`.

Additionally I think new functions `svgLoader.getRootNode()` and `svgLoader.getNodePathMap()` (map from nodes to paths) could be called (after the SVG is loaded) if the SVG node tree is needed (see linked PR above)

Exporting the SVG tags allows the user to render things like strokes, which are a nice feature to have. For example, the `tiger.svg` has some shape outlines and not-closed paths that have got tags like `stroke-color` and `stroke-width`. They are not rendered currently.

To not break old user code, a new flag `svgLoader.exportOnlyVisible` or `exportOnlyFillShapes`, which defaults to `true`, could be added to let export (when `false`) the stroke paths which have the `fill-color` set to `none` and thus are currently non-visible and are not exported.

If this is done I can also implement a function that generates, given the path points and the stroke SVG properties, the stroke shape (with miter line joining and no linecaps). I can implement later the tags `strokeLinejoin`, `strokeLinecap` and `strokeMiterLimit` among others.

The stroke creation could also be useful for shapes generated from `Font` text.

BTEVC:
Okay, this is a little above my experience level, but I would be willing to put in the work to figure out how to do it if it would be useful.

yomboprime:
Please, let's wait for some feedback. I am no authority here 😊

If my idea is approved, these are the changes:
 - Keep your changes to the `parseStyle()` function, they are all right.
 - Put `path.userData = { node: node };` just after the line `paths.push( path );`.
 - Revert your other changes to the `SVGLoader`.

With only this, you can access in your application any SVG tag as I said: `path.userData.node.getAttribute( 'fill-opacity' )`

Later other tags can be added to the style and I can do a PR for the stroke creation. 

About the node map, you can also add it if you want:
 - Add a new `this.nodeMap = new Map();` instance variable to the constructor.
 - Add a new `this.rootNode = null;` instance variable to the constructor.
 - Put `nodeMap.set( node, path );` just after the line `paths.push( path );`.
 - Put `this.rootNode = xml.documentElement;` just after the line `parseNode( xml.documentElement, { fill: '#000' } );`

That's all. A getter function for the map and another for the rootNode is optional.

WestLangley:
>To not break old user code

I wouldn't worry about that in this case. You can always add a migration note.

>Please, let's wait for some feedback. I am no authority here

...and I'll defer to @mrdoob and @Mugen87. :-)

Mugen87:
The suggestion of @mrdoob was to introduce `userData` to `ShapePath` which is also done in other classes of `three.js`. Exposing the entire node is probably the most flexible approach rather than exposing individual node properties over `ShapePath`. 

The problems was that the OP of #15170 never implemented his feedback. This PR could be a good opportunity to finally do so 😉

BTEVC:
Okay great, I'll get started on implementing the changes @yomboprime suggested in the next couple of days then.

yomboprime:
Another thought: As the node properties are not inherited between the nodes, I would also add the `style` to `userData`, i.e.:
`path.userData = { node: node, style: style };`

(The styles do inherit properties from upper nodes)

yomboprime:
Looks good. I got already working the stroke creation. In some days I can make a PR.

Screenshot_20190313_012207

I need to implement `stroke-miterlimit` to limit the miter that goes away in some places as you can see. Will implement also round and bevel joins and also linecaps.

BTEVC:
Awesome. 👍

BTEVC:
Woops, had left in the line `this.rootNode = xml.documentElement;` from the nodeMap suggestion. The final decision was to exclude the nodeMap correct?

yomboprime:
Even if the map is out, the root node can be useful. I would leave it. (And declare it in the constructor also, please)

BTEVC:
K, how does that look?

BTEVC:
K, how does that look?

yomboprime:
All right. You keep toggling code 😊

BTEVC:
Haha, yeah, sorry, had a long day today. :)

Mugen87 (approved):

examples/js/loaders/SVGLoader.js
```diff
@@ -8,6 +8,8 @@ THREE.SVGLoader = function ( manager ) {
 
 	this.manager = ( manager !== undefined ) ? manager : THREE.DefaultLoadingManager;
 
+	this.rootNode = null;
```
mrdoob: Instead of adding this here, I would change `parse()` so it returns an object instead of the current `paths` array.

```js
var data = { paths: [], xml: xml.documentElement };
```

mrdoob:
@WestLangley @Mugen87 @yomboprime thanks for the help here guys!

mrdoob:
@yomboprime these strokes are looking great!

yomboprime:
> @yomboprime these strokes are looking great!

I'm making progress... 😃
strokes_20190315_022443

yomboprime:
I've got the strokes working right.

It is a long function (720 js lines) in `THREE.SVGLoader.pointsToStroke( points, style, arcDivisions )` which returns a BufferGeometry.

I could move it to another place like its own class `StrokeGeometry` in `examples/js/geometries/`. What do you think?

mrdoob:
@yomboprime I was expecting that 😁 I would add it to SVGLoader so it's self contained

yomboprime:
> @yomboprime I was expecting that grin I would add it to SVGLoader so it's self contained

Do you mean that I have a verbose coding habit? 😃

examples/js/loaders/SVGLoader.js
BTEVC: Okay, updated, does that look right?

yomboprime:
@BTEVC You should also modify the svgloader example for the new change in returned data.

BTEVC:
@yomboprime Oh right, so far only updated my own [example](https://btevc.tk/repos/three.js/examples/webgl_loader_svg_update_1.html).

mrdoob:
> Do you mean that I have a verbose coding habit? 😃

Not at all! I just suspected converting strokes to geometry was going to be arduous 🙈

yomboprime:
@BTEVC You should do:
```
loader.load( url, function ( data ) {
var paths = data.paths;
```
And leave the `for` loop as it was before

BTEVC:
@yamboprime Ah, mkay. Will fix this shortly.

examples/webgl_loader_svg.html
```diff
@@ -161,9 +161,9 @@
 					group.position.y = 70;
 					group.scale.y *= - 1;
 
-					for ( var i = 0; i < paths.length; i ++ ) {
+					for ( key in paths['paths'] ) {
```
mrdoob: This seems over complicated. You should be able to keep the previous code and just add a these lines 
```js
loader.load( url, function ( data ) {
```
```js
var paths = data.paths;

for ( var i = 0; i < paths.length; i ++ ) {
```

BTEVC:
So checks still not passing, anyone know what might be going on here?

yomboprime:
> So checks still not passing, anyone know what might be going on here?

It seems something related to webvr renderer files. I have no Idea, could be a Github bug.

yomboprime:
Bonus: Stroke `Font` text. I copied the `webgl_geometry_text_shapes` example and converted the line text to stroke text with minor changes.
strokesText

Mugen87:
> So checks still not passing, anyone know what might be going on here?

Don't worry about that. The travis build was  temporarily broken. Should be fixed now.

yomboprime:
@Mugen87 Can I make my PR if it depends on this one? I'm not a git expert.

Mugen87:
I guess it's better to wait until @mrdoob approves this PR. It has to merged before yours in any event.

BTW: This PR here needs to be mentioned in the migration guide since the parameter of the `onLoad()` function changes. User will have to change their code similar to `webgl_loader_svg`.

mrdoob:
Thanks!

BTEVC:
@yomboprime @WestLangley @Mugen87 @mrdoob Thank you all for helping me through this PR, exciting to participate in this library, even if in a very small way.

PR #15933, opened 2019-03-08
Remove Loader inheritance from CTMLoader

takahirox:
`CTMLoader` is the only loader inheriting `Loader` but it doesn't look it really needs to inherit. `CTMLoader` calls `scope(this).createMaterial()` but it can be replaced with `THREE.Loader.prototype.createMaterial()` without inheriting `Loader` as similarly `LegacyJSONLoader` does.

With this change, no loader will inherit `Loader`. We may remove `Loader` if we move `initMaterials()`, `createMaterial()` and `Handlers` to somewhere(`LoaderUtils`?).

### DIFF
--- a/examples/js/loaders/ctm/CTMLoader.js
+++ b/examples/js/loaders/ctm/CTMLoader.js
@@ -10,11 +10,8 @@
 
 THREE.CTMLoader = function () {
 
-	THREE.Loader.call( this );
-
 };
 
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):
It's important to keep the loaders consistent so I think this change is a good one.

mrdoob:
Thanks!

PR #15946, opened 2019-03-11
AnimationLoader: Fix .parse().

Mugen87:
`AnimationLoader.load()` is currently broken since `AnimationLoader.parse()` does not return an array of animation clips. The change makes the loader more consistent to `BufferGeometryLoader`.

BTW: `AnimationLoader` is not used in the code base yet so it was not necessary to refactor code.


mrdoob:
Thanks!

PR #15947, opened 2019-03-11
Add TypeScript definitions for example modules

Methuselah96:
This addresses https://github.com/mrdoob/three.js/issues/9562#issuecomment-471359980. This PR only adds TypeScript definitions for the examples that have been converted to modules. I will add more of them as they get converted. The source is [@types/three](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/three) and I did not check for correctness.


mrdoob:
Thanks!

PR #15958, opened 2019-03-13
ObjectLoader: Fix TS definition of .load()

jeffreylanters:
The Object3D typings we're missing the material.

### DIFF
--- a/src/core/Object3D.d.ts
+++ b/src/core/Object3D.d.ts
@@ -39,6 +39,11 @@ export class Object3D extends EventDispatcher {
   name: string;
 
   type: string;
+  
+  /**
+   * The material of the object.
+   */
+  material: Material;
 
   /**
    * Object's parent in the scene graph.
--- a/src/core/Object3D.d.ts
+++ b/src/core/Object3D.d.ts
@@ -39,11 +39,6 @@ export class Object3D extends EventDispatcher {
   name: string;
 
   type: string;
-  
-  /**
-   * The material of the object.
-   */
-  material: Material;
 
   /**
    * Object's parent in the scene graph.
--- a/src/loaders/ObjectLoader.d.ts
+++ b/src/loaders/ObjectLoader.d.ts
@@ -13,7 +13,7 @@ export class ObjectLoader {
 
   load(
     url: string,
-    onLoad?: (object: Object3D) => void,
+    onLoad?: <ObjectType extends Object3D>(object: ObjectType) => void,
     onProgress?: (event: ProgressEvent) => void,
     onError?: (event: Error | ErrorEvent) => void
   ): void;


Mugen87:
This seems to be a mistake. `Object3D` does not have a `material` property.

jeffreylanters:
> This seems to be a misunderstanding. `Object3D` does not have a `material` property.

But it does work, when I log an `Object3D` object, I'll see and can set the material.

jeffreylanters:
Screenshot 2019-03-13 at 11 18 31

Mugen87:
It looks like you are working with an instance of `Mesh`.

jeffreylanters:
The variable comes straight out of the object loader, it's untouched. There might my another mistyping?

image

jeffreylanters:
When I log the Object3D
Screenshot 2019-03-13 at 11 30 52

The material does exist
Screenshot 2019-03-13 at 11 31 09

Mugen87:
`Mesh` is derived from `Object3D`. I think if you study the source code of `Object3D` and `Mesh`, things should become more clear:

https://github.com/mrdoob/three.js/blob/09e1fde724b89e2de4d5e774de27f137ab13d97c/src/core/Object3D.js#L21-L101
https://github.com/mrdoob/three.js/blob/09e1fde724b89e2de4d5e774de27f137ab13d97c/src/objects/Mesh.js#L20-L33

jeffreylanters:
Okay, I see what you mean. But still, the ObjectLoader packed in the module, returns an object with the Typescript Type `Object3D`, and it does contain `material`, but it's not typed.

Mugen87:
> returns an object with the Typescript Type Object3D, and it does contain material, but it's not typed.

Then there is a problem in `ObjectLoader.d.ts`. The `onLoad()` callback of `ObjectLoader.load()` can return many different types of objects e.g. `Scene`, `Mesh`, `LineSegments` etc. It would not make sense to put all the properties of these classes to `Object3D.d.ts`, right? Instead, the following line needs to be changed:

https://github.com/mrdoob/three.js/blob/09e1fde724b89e2de4d5e774de27f137ab13d97c/src/loaders/ObjectLoader.d.ts#L16

jeffreylanters:
I see what is going on. Thanks for clearing it up! It could be a `generic key of type`, function to keep it more strict.

Mugen87:
Like so?

https://github.com/mrdoob/three.js/blob/09e1fde724b89e2de4d5e774de27f137ab13d97c/src/loaders/ObjectLoader.d.ts#L22

mrdoob:
> It would not make sense to put all the properties of these classes to `Object3D.d.ts`, right?

Wouldn't this work?

```
onLoad?: (object: Scene | Mesh | Line | LineSegments | Points) => void,
```

Mugen87:
As an alternative, I think it would also work to define a generic type derived from `Object3D`.
```
onLoad?: <T extends Object3D>(object: T) => void,
```
Not sure which approach is better  🤔

mrdoob:
> Not sure which approach is better 🤔

@jeffreylanters you choose 🙂

jeffreylanters:
> As an alternative, I think it would also work to define a generic type derived from `Object3D`.
> 
> ```
> onLoad?: <T extends Object3D>(object: T) => void,
> ```
> Not sure which approach is better 🤔

Like I said before, I think a generic approach would be the best way to go. I think this looks is perfect. Thanks for thinking with me!

jeffreylanters:
I've updated the PR.

Mugen87 (approved):

mrdoob:
Thanks!

PR #15960, opened 2019-03-13
remove unrequired condition in checkIntersection for Mesh

thinkimlazy:
I think it's obvious here that 
`material.side !== DoubleSide` 
already includes 
`material.side === BackSide`

### DIFF
--- a/src/objects/Mesh.js
+++ b/src/objects/Mesh.js
@@ -6,7 +6,7 @@ import { Matrix4 } from '../math/Matrix4.js';
 import { Object3D } from '../core/Object3D.js';
 import { Triangle } from '../math/Triangle.js';
 import { Face3 } from '../core/Face3.js';
-import { DoubleSide, BackSide, TrianglesDrawMode } from '../constants.js';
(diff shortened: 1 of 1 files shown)


src/objects/Mesh.js
```diff
@@ -135,17 +135,7 @@ Mesh.prototype = Object.assign( Object.create( Object3D.prototype ), {
...
-			if ( material.side === BackSide ) {
-
-				intersect = ray.intersectTriangle( pC, pB, pA, true, point );
-
-			} else {
-
-				intersect = ray.intersectTriangle( pA, pB, pC, material.side !== DoubleSide, point );
-
-			}
+			var intersect = ray.intersectTriangle( pA, pB, pC, material.side !== DoubleSide, point );
```
Mugen87: I'm afraid this change is not correct. Notice the difference in the order of parameters in the first and second call of `intersectTriangle()`:
```js
pC, pB, pA
```
and
```
pA, pB, pC
```

Mugen87 (changes requested):

Issue #15973, opened 2019-03-14
Editor: Looking through cameras


EtagiBI:
Hello,

It would be useful to have an option to look at the scene through PerspectiveCamera in Three.js Editor.

Here's Blender way to do it:

![Camera](https://user-images.githubusercontent.com/20855725/54351738-f0fef400-4671-11e9-8289-664fed5de695.gif)

##### Three.js version

- [ ] Dev
- [x] r102
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

PR #15983, opened 2019-03-15
Editor: Select Viewport Camera

Temdog007:
Adds feature from #15973

cameraView


Temdog007:
For testing: https://raw.githack.com/Temdog007/three.js/feature/CameraSwitch/editor/index.html

Mugen87:
Right now, you leave the custom camera view mode when you move the mouse away from the `View` button. TBH, this handling feels a bit strange^^. It might be better to have a checkbox in order to enable/disable the view.

I'm also not sure it's necessary to animate the transition. The code might be less complex if you remove the animation related code.

looeee:
> Right now, you leave the custom camera view mode when you move the mouse away from the View button. TBH, this handling feels a bit strange^^

Yup, this is a bit odd. I also don't think the animation is needed. 

Most 3D editors add a menu to the viewport itself, allowing you to switch between view types/cameras and other settings. A small menu in top left of the viewport seems to be standard. Check out how Clara.io does it, for example.

Temdog007:
Updates

camera view

camera view2

mrdoob:
I was thinking of implementing this way instead:

Screenshot 2019-03-15 at 12

A drop-down in the top-right that allows the user selecting other cameras. `EditorControls` should only be able to modify the default camera for now.

looeee:
> A drop-down in the top-right that allows the user selecting other cameras. I think this is also what @looeee described

Yup, that's exactly what I was suggesting, except top left. But whatevs 🤷‍♀️

At some point it would be good to add the standard orthographic views to this list as well, top, left, bottom, right. And maybe in the far future allow for the viewport to be split into two or four views. 
 
@Temdog007 thanks for working on this, it's great to see the editor getting some love.

Temdog007:
I didn't mean to close this PR. I couldn't push changes to the branch so I tried to delete it and push it.

PR #15991, opened 2019-03-15
Start WebVRManager animation when setting the animation callback

dmarcos:
VR render loop never kicks off If the event `vrdisplaypresentchange` fires before `setAnimationLoop` is invoked.

Add `animation.start` call in VRManager `setAnimationLoop`. Consistent with [WebGLRenderer `setAnimationLoop` method](https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L1077)


mrdoob:
Thanks!

PR #15994, opened 2019-03-15
GLTFExporter: improve error message when exporting unsupported light

garyo:


examples/js/exporters/GLTFExporter.js
```diff
@@ -1754,7 +1754,7 @@ THREE.GLTFExporter.prototype = {
 
 			} else if ( object.isLight ) {
 
-				console.warn( 'THREE.GLTFExporter: Only directional, point, and spot lights are supported.' );
+				console.warn( `THREE.GLTFExporter: light ${object.name} of type ${object.constructor.name}: Only directional, point, and spot lights are supported.` );
```
mrdoob: We usually do this instead:
```js
console.warn( 'THREE.GLTFExporter: Only directional, point, and spot lights are supported.', object );
```

garyo: Would that show the type though? That's what a user would need to search for.

Mugen87: The warning looks in Chrome like so:

image

I think it's okay like that. Users can still debug the object for more details.

garyo:
done.

donmccurdy (approved):

mrdoob:
Thanks!

PR #7513, opened 2015-11-02
Ability to limit pan based on distance from origin

agnivade:
- Added a flag which controls the limiting of panning

PR #16004, opened 2019-03-17
add pan limit controls to orbitController & fix controller when in iFrame

mshamaseen:


WestLangley:
As in #7513, I think this pan limit too use-case-specific.

`OrbitControls` is part of the examples so users can modify it to accommodate app requirements.

Mugen87:
@mshamaseen Thanks for the PR though!

PR #16015, opened 2019-03-19
Update vr drawing buffer size when the WebVRManager device is set.

dmarcos:
It fixes scenarios where buffer size is set to the screen resolution instead of headset. See below

Buffer will end up with the wrong size if `vrdisplaypresentchange`  event fires before user code has called `renderer.vr.setDevice`.

This PR factors out the buffer resizing logic to be also invoked when the device is set.

**Before**

Before

**After**

After

### DIFF
--- a/src/renderers/webvr/WebVRManager.js
+++ b/src/renderers/webvr/WebVRManager.js
@@ -65,26 +65,13 @@ function WebVRManager( renderer ) {
 
 	function onVRDisplayPresentChange() {
 
+		updateDrawingBufferSize();
+
 		if ( isPresenting() ) {
-
-			var eyeParameters = device.getEyeParameters( 'left' );
-			var renderWidth = eyeParameters.renderWidth * framebufferScaleFactor;
-			var renderHeight = eyeParameters.renderHeight * framebufferScaleFactor;
-
-			currentPixelRatio = renderer.getPixelRatio();
-			renderer.getSize( currentSize );
-
-			renderer.setDrawingBufferSize( renderWidth * 2, renderHeight, 1 );
 
 			animation.start();
 
 		} else {
-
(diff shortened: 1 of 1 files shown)


src/renderers/webvr/WebVRManager.js
```diff
@@ -119,6 +106,33 @@ function WebVRManager( renderer ) {
 
 	}
 
+	function updateDrawingBufferSize() {
```
dmarcos: This is the same logic previously contained in the `onVRDisplayPresentChange` handler

src/renderers/webvr/WebVRManager.js
```diff
@@ -210,6 +222,10 @@ function WebVRManager( renderer ) {
 
 		animation.setContext( value );
 
+		updateDrawingBufferSize();
+
+		if ( isPresenting() ) animation.start();
```
dmarcos: Start animation. If `vrdisplaypresentchange` fires after `requestPresent` and no device has been set the render look will never kick off.

src/renderers/webvr/WebVRManager.js
```diff
@@ -65,27 +65,14 @@ function WebVRManager( renderer ) {
 
 	function onVRDisplayPresentChange() {
 
-		if ( isPresenting() ) {
-
-			var eyeParameters = device.getEyeParameters( 'left' );
-			var renderWidth = eyeParameters.renderWidth * framebufferScaleFactor;
-			var renderHeight = eyeParameters.renderHeight * framebufferScaleFactor;
+		updateDrawingBufferSize();
```
dmarcos: Factor out drawing before resizing logic and call it.

Mugen87:
WebVR support has been removed, see #18129

PR #16016, opened 2019-03-19
Editor: Time render operation

### LINKED ISSUE #16007: Editor: add FPS counter

Temdog007:
Fixes #16007.

Based on @mrdoob's [comment](https://github.com/mrdoob/three.js/issues/16007#issuecomment-473988100).


Temdog007:
For testing: https://raw.githack.com/Temdog007/three.js/feature/renderTime/editor/index.html

looeee:
`render time 1.7,249,999,982,886,948ms`

I think it would be OK to round this down a bit 😁 
I doubt anyone needs more than 4 or 5 decimal places on milliseconds, so `1.7249ms`. 

Also, "last frame time" might be more a descriptive label.

Temdog007:
Turns out that Firefox always rounds to the nearest ms with `performance.now`. I had to use Chrome to see any decimals.

Mugen87 (approved):

Temdog007:
I'm not sure how useful this really is. Unless there is a continuous render mode (where the Editor will render on every `requestAnimationFrame` rather than on an event), this may not provide useful information. In practice, one would want to know how rendering the scene affects the FPS. Closing for now.

PR #16039, opened 2019-03-22
Quaternion: Remove unneeded closure

WestLangley:
As suggested in https://github.com/mrdoob/three.js/pull/16022#pullrequestreview-216380290.


Mugen87 (approved):

mrdoob:
Thanks!

PR #16042, opened 2019-03-22
Add getState method to OrbitControls

mess110:
Depending on the state, I would like certain actions to be enabled/disabled in my game. I think knowing the 
state of OrbitControls can be helpful for others as well, that's why I am submitting this PR

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -97,6 +97,12 @@ THREE.OrbitControls = function ( object, domElement ) {
 		return spherical.theta;
 
 	};
+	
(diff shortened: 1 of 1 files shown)


WestLangley:
Thanks for your suggestion, but the existing method `saveState()` uses a different meaning of the word 'state' than the proposed `getState()`, which returns the value of a private variable.

There have been many requests to modify `OrbitControls` to accommodate specific application use cases. I think at this point, we need to limit these changes.`OrbitControls` is part of the examples, so users are free to modify it.

mess110:
Fair enough, will close this PR. Thanks for the review.

Issue #16026, opened 2019-03-20
ObjectLoader doesn't parse all geometries


Temdog007:
##### Description of the problem

`ObjectLoader` doesn't support parsing the following geometries:

- `TextGeometry`
- `ParametricGeometry`
- `InstancedGeometry`
- ~~`WireframeGeometry`~~

This causes an issue where the Editor cannot load any scenes or objects that use these geometries.

##### Three.js version

- [x] Dev
- [x] r102

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
There was already a PR (see #14357) in order to introduce support for `EdgesGeometry` and `WireframeGeometry` but it was not accepted. The problem is that adding support for all those geometries blows up `ObjectLoader` and I'm not sure this is desired.

Temdog007:
I understanding not wanting to serialize `EdgesGeometry` and `WireframeGeometry` since those take geometry as a parameter. I would still suggest that `TextGeometry` and `InstancedGeometry` be serialized. This is so that the Editor can import and export these geometries.

Temdog007:
Duplicate of #16340

PR #16045, opened 2019-03-23
Serialize TextGeometry

Temdog007:
This is specifically so that the editor can create and load `TextGeometry`.

See #16026


Mugen87:
I'm afraid we need a different approach for serializing/deserializing `TextGeometry`. The idea is to decouple `Font` from `TextGeometry` so a font is handled similar to textures or shapes. That means `Font` needs a `uuid` property that can be used by `TextGeometry` for referencing. A font would be saved independently in the JSON file. When `ObjectLoader` loads a JSON with `TextGeometry`, the loader would parse the fonts first and then pass the result to `parseGeometries()`. Something like:
```js
var shapes = this.parseShape( json.shapes );
var fonts = this.parseFonts( json.fonts );
var geometries = this.parseGeometries( json.geometries, shapes, fonts );
```
More information about this approach in #9905.

PR #16054, opened 2019-03-25
[WIP] JSM: Move nodes/*

donmccurdy:
All files in `examples/js/nodes/*` were already ES modules. This PR moves them to the `jsm/` folder, adds imports from the main threejs library, and fixes a few stray `THREE.*Node` references that needed to be imports. With that, it should be possible to import them in the same way as other `jsm/` examples:

```js
import { Scene, Mesh, BufferGeometry } from 'three';
import { StandardNodeMaterial, FloatNode, ... } from 'three/examples/jsm/nodes/Node.js';

var scene = new Scene();
var mesh = new Mesh( new BufferGeometry(), new StandardNodeMaterial() );
scene.add( mesh );
```

I've only fixed the first two examples, and will fix the remaining five if we want to proceed with this PR.  Note that I've _removed_ the original `js/` files, because this change was not fully automated, and it probably isn't worth the effort to do so.

/cc @sunag


sunag:
@donmccurdy Very niceee, thanks!

Leveraging the subject my doubts about ES modules... Is there any demarcation today that would prevent we advance in `class` instead of `prototype`?

sunag:
I think that `THREE.Nodes.js` can be removed since it has lost its function.

donmccurdy:
> Is there any demarcation today that would prevent we advance in class instead of prototype?

I'd be fine with using ES6 Classes in the `nodes/` folder, unless others think it's important to keep the whole codebase more consistent for now?

sunag:
Because this PR is still a WIP? I think is important merge for add futures updates.

donmccurdy:
I will finish this PR soon. 👍 

After that, whatever you prefer on ES6 Classes is OK with me.

donmccurdy:
Hm, I'm not sure I can convert `postprocessing / nodes / pass` without also converting `postprocessing/*` to ES modules. Those files are heavily dependent on one another, and as a result we need to either:

(a) modify `modularize.js` to handle JSM-to-JSM dependencies
(b) do a one-time conversion of `postprocessing/*` to `jsm/` and start generating the `js/` version from that
(c) ???

I think I prefer (b), but it requires #15599 or something similar. Maybe that can be simplified a bit; I'll comment there.

Mugen87:
> Hm, I'm not sure I can convert postprocessing / nodes / pass without also converting postprocessing/* to ES modules. 

The postprocessing classes are now available as modules 😇 .

Mugen87:
@donmccurdy Do you think you are able to finish this PR in the next time 😇? The node material files and the related examples are one of the last todos in our JSM/module task list, see #16688

BTW: It could make sense to start from scratch since many things changed in the last weeks. If you are busy with other stuff, I can do this, too.

donmccurdy:
@Mugen87 the next thing I want to finish up is https://github.com/mrdoob/three.js/pull/15249, and it will be at least late July before I could take another look at this PR unfortunately. I'm not sure how much has changed that would affect the PR, but if you are able to have a look please do! 🙂

PR #16058, opened 2019-03-25
fix missing import at source

aardgoose:
THREE. namespace missing


mrdoob:
That must be why... 😁

mrdoob:
Thanks!

PR #16063, opened 2019-03-26
Add renderOrder for Materials

### LINKED ISSUE #16060: Unity compatible per-material render order support
Already discussed a bit on #14415 

We already have `renderOrder` for meshes and groups,
but not for materials, which is supported by Unity (as `renderQueue`).
It would help sorting rendering order of multi-material meshes in certain condition.

I have no idea how much the idea makes good impact for us but I need this at least (🙄),
Since I have to display Unity-compatible stuff (I mean, VRM) on https://hub.vroid.com ...

demo below
https://codepen.io/FMS-Cat/pen/pYBwzw

0b5vr:
will resolve #16060 

New property for `Material` : `renderOrder` .
Behaves like `Object3D` 's [`renderOrder`](https://threejs.org/docs/#api/en/core/Object3D.renderOrder) .

`WebGLRenderLists` sees the property, then sorts the rendering order.

### DIFF
--- a/src/materials/Material.d.ts
+++ b/src/materials/Material.d.ts
@@ -47,8 +47,8 @@ export interface MaterialParameters {
   transparent?: boolean;
   vertexColors?: Colors;
   vertexTangents?: boolean;
-	visible?: boolean;
-	renderOrder?: number;
+  visible?: boolean;
+  renderOrder?: number;
 }
 
 /**
--- a/src/materials/Material.d.ts
+++ b/src/materials/Material.d.ts
@@ -47,7 +47,8 @@ export interface MaterialParameters {
   transparent?: boolean;
   vertexColors?: Colors;
   vertexTangents?: boolean;
-  visible?: boolean;
+	visible?: boolean;
+	renderOrder?: number;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


FMS-Cat:
https://gist.github.com/FMS-Cat/2b822c74c4918a826b31d60a0d1c486a

This one should work properly

src/renderers/webgl/WebGLRenderLists.js
```diff
@@ -12,6 +12,10 @@ function painterSortStable( a, b ) {
 
 		return a.renderOrder - b.renderOrder;
 
+	} else if ( a.materialOrder !== b.materialOrder ) {
```
FMS-Cat: Unsure whether the materialOrder should come after Group / Object's renderOrder or not

src/materials/Material.d.ts
```diff
@@ -247,6 +248,11 @@ export class Material extends EventDispatcher {
    */
   visible: boolean;
 
+  /**
+   * Overrides the default rendering order of scene graph objects, from lowest to highest renderOrder. Opaque and transparent objects remain sorted independently though. renderOrder of Objects and Groups takes priority than this property.
```
FMS-Cat: Review my English

Mugen87:
Closing, see https://github.com/mrdoob/three.js/issues/16060#issuecomment-522933870

PR #16076, opened 2019-03-27
CameraHelper: Fix copy()/clone()

### LINKED ISSUE #13565: CameraHelper: .clone() is broken

`CameraHelper::constructor` expects a `Camera` argument, but since `CameraHelper::clone` is not overwritten it will forward to `LineSegments::constructor` which expects a geometry and a material.

As a result `cameraHelper.camera` won't be set and `cameraHelper.clone()` throws via `.update()` (see https://jsfiddle.net/bbnk59af/).

Overwriting `CameraHelper::clone` to account for its own constructor solves the issue (see https://jsfiddle.net/bbnk59af/1/):

```
THREE.CameraHelper.prototype.clone = function () {
	var clone = new this.constructor(this.camera).copy(this);
	return clone;
}

```

I am not sure what the intended behaviour is in the case that `cameraHelper.clone()` gets called as a child of `scene.clone()`. The cameras and their helpers would be cloned individually, but the cloned helpers would point to the original cameras instead of the cloned ones.

##### Three.js version

- [X] Dev
- [X] r90

##### Browser

- [x] All of them

##### OS

- [x] All of them

Mugen87:
Fixed #13565

The idea is to allow `CameraHelper` to be created without parameters (like `GridHelper`). In this way, it's possible to implement `clone()` in this fashion:

```js
```
which is the ideal implementation of a clone method.

BTW: `pointMap` is actually unique for all helpers so a copy of this property is not necessary.

### DIFF
--- a/src/helpers/CameraHelper.js
+++ b/src/helpers/CameraHelper.js
@@ -110,101 +110,127 @@ function CameraHelper( camera ) {
 	LineSegments.call( this, geometry, material );
 
 	this.camera = camera;
-	if ( this.camera.updateProjectionMatrix ) this.camera.updateProjectionMatrix();
-
-	this.matrix = camera.matrixWorld;
 	this.matrixAutoUpdate = false;
-
 	this.pointMap = pointMap;
 
-	this.update();
+	if ( camera ) {
+
+		if ( camera.updateProjectionMatrix ) camera.updateProjectionMatrix();
+
+		this.matrix = camera.matrixWorld;
+
+		this.update();
+
+	}
 
 }
 
-CameraHelper.prototype = Object.create( LineSegments.prototype );
-CameraHelper.prototype.constructor = CameraHelper;
+CameraHelper.prototype = Object.assign( Object.create( LineSegments.prototype ), {
 
-CameraHelper.prototype.update = function () {
+	constructor: CameraHelper,
 
-	var geometry, pointMap;
+	update: function () {
 
-	var vector = new Vector3();
-	var camera = new Camera();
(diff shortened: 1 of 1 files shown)


WestLangley:
>The idea is to allow CameraHelper to be created without parameters (like GridHelper).

Instantiating a `CameraHelper` without a camera does not seem to me to be a good idea because it can't be rendered without a camera.

Also, why does a user need to clone a camera helper?

Similarly for the other renderable helpers that depend on another object.

Related: #16078.

Mugen87:
> Instantiating a CameraHelper without a camera does not seem to me to be a good idea because it can't be rendered without a camera.

That's true. I've only added this possibility in order to implement the `clone()` as posted above.

WestLangley:
>Instantiating a CameraHelper without a camera does not seem to me to be a good idea because it can't be rendered without a camera.
>>That's true. I've only added this possibility in order to implement the clone() as posted above.

I understand. It is the cloning that is the problem with our design. I am trying to see if we can maybe fix our design.

Mugen87:
> Also, why does a user need to clone a camera helper?

Yes, the reasonableness of this is questionable. TBH, I'm doing this PR for technical reasons. So the methods are present and produce an understandable result.

But yeah,  there are scenarios where the clone does not work as expected. For example if you clone an entire scene, a clone of `CameraHelper` would still refer to the "old" camera of the source scene. Not to the new one.

Mugen87:
> I am trying to see if we can maybe fix our design.

Closing for now then.

PR #16082, opened 2019-03-28
Add support for SVG Groups

pgodwin:
For discussion more than anything else. 
Adds support for passing through the groups parsed out of the SVG file. This makes it easier to do things like highlighting on groups from the SVG, or rendering different SVG groups with different parameters.

### DIFF
--- a/examples/js/loaders/SVGLoader.js
+++ b/examples/js/loaders/SVGLoader.js
@@ -4,6 +4,51 @@
  * @author yomboprime / https://yombo.org
  */
 
+
+/**
+ * Represents a group of items in from SVG File
+ */
+THREE.SVGGroup = function(node) {
+	this.node = node;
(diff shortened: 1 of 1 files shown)


yomboprime:
Now that the SVGLoader exports the root node, the user application can do the groups management.

See [This example](https://yomboprime.github.io/three_example_svg_groups/webgl_loader_svg_groups.html) on how to do it.

In the example, each path is assigned a random color depending on which group the path belongs to.

WestLangley:
@yomboprime Thanks for your example.

Tip: if you use the un-minified `three.js`, it is easier for users to debug.

WestLangley:
Closing. This issue appears to be resolved.

pgodwin:
Thanks for the example, that does the trick. So this is certainly resolved.

PR #16088, opened 2019-03-28
Fixed a bug with loading KTX cube textures

### LINKED ISSUE #16089: KTX Cube Textures Fail
Loading a compressed cube KTX texture causes THREE.js to crash. The bug appears to be that CompressedImageTextures with an array as their image are still treated as individual images. This bug is related to a PR that I submitted earlier this morning.

There is an additional bug with the KTX Texture loader, in that the supplied mipmaps are loaded in the level order [0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, etc] when CompressedTexture expected them to be in order [0, 1, 2, 3, 4, 5, 0, 1, 2, 3, 4, etc], but I believe this is a greater issue that needs to be addressed first.

The place that THREE.js crashes is https://github.com/mrdoob/three.js/blob/640f046fc6f3918778e9dfb59fcc9ab3fac9cb05/src/renderers/webgl/WebGLTextures.js#L676, because it's attempting to read the mipmaps from the base image container when the mipmaps are all stored in an array inside of the image parameter.

John-DeCorato-HZ:
Fixed #16089. Moved code accounting for moving past the data field for the size of the texture outside of the loop loading them.

For reference, see the loader this was ported from in Babylon for the correct implementation:

https://github.com/BabylonJS/Babylon.js/blob/master/src/Misc/khronosTextureContainer.ts#L169

### DIFF
--- a/examples/js/loaders/KTXLoader.js
+++ b/examples/js/loaders/KTXLoader.js
@@ -137,13 +137,14 @@ var KhronosTextureContainer = ( function () {
 		for ( var level = 0; level < mipmapCount; level ++ ) {
 
 			var imageSize = new Int32Array( this.arrayBuffer, dataOffset, 1 )[ 0 ]; // size per face, since not supporting array cubemaps
+			dataOffset += imageSize + 4; // size of the image + 4 for the imageSize field
+			
 			for ( var face = 0; face < this.numberOfFaces; face ++ ) {
 
 				var byteArray = new Uint8Array( this.arrayBuffer, dataOffset + 4, imageSize );
 
 				mipmaps.push( { "data": byteArray, "width": width, "height": height } );
 
-				dataOffset += imageSize + 4; // size of the image + 4 for the imageSize field
 				dataOffset += 3 - ( ( imageSize + 3 ) % 4 ); // add padding for odd sized image
 
 			}
--- a/examples/js/loaders/KTXLoader.js
+++ b/examples/js/loaders/KTXLoader.js
@@ -137,14 +137,15 @@ var KhronosTextureContainer = ( function () {
 		for ( var level = 0; level < mipmapCount; level ++ ) {
 
 			var imageSize = new Int32Array( this.arrayBuffer, dataOffset, 1 )[ 0 ]; // size per face, since not supporting array cubemaps
-			dataOffset += imageSize + 4; // size of the image + 4 for the imageSize field
+			dataOffset += 4; // size of the image + 4 for the imageSize field
 			
 			for ( var face = 0; face < this.numberOfFaces; face ++ ) {
 
-				var byteArray = new Uint8Array( this.arrayBuffer, dataOffset + 4, imageSize );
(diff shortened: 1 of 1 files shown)


examples/js/loaders/KTXLoader.js
```diff
@@ -137,13 +137,14 @@ var KhronosTextureContainer = ( function () {
 		for ( var level = 0; level < mipmapCount; level ++ ) {
 
 			var imageSize = new Int32Array( this.arrayBuffer, dataOffset, 1 )[ 0 ]; // size per face, since not supporting array cubemaps
+			dataOffset += imageSize + 4; // size of the image + 4 for the imageSize field
+			
 			for ( var face = 0; face < this.numberOfFaces; face ++ ) {
 
 				var byteArray = new Uint8Array( this.arrayBuffer, dataOffset + 4, imageSize );
```
Mugen87: The new code produces currently a runtime error (`Uncaught RangeError: Invalid typed array length)` at this line, see https://raw.githack.com/John-DeCorato-HZ/three.js/dev/examples/webgl_loader_texture_ktx.html

Maybe like so:
```js
KhronosTextureContainer.prototype.mipmaps = function ( loadMipmaps ) {

		var mipmaps = [];

		// initialize width & height for level 1
		var dataOffset = KhronosTextureContainer.HEADER_LEN + this.bytesOfKeyValueData;
		var width = this.pixelWidth;
		var height = this.pixelHeight;
		var mipmapCount = loadMipmaps ? this.numberOfMipmapLevels : 1;

		for ( var level = 0; level < mipmapCount; level ++ ) {

			var imageSize = new Int32Array( this.arrayBuffer, dataOffset, 1 )[ 0 ]; // size per face, since not supporting array cubemaps
			dataOffset += 4; // image data starts from next multiple of 4 offset. Each face refers to same imagesize field above.
			for ( var face = 0; face < this.numberOfFaces; face ++ ) {

				var byteArray = new Uint8Array( this.arrayBuffer, dataOffset, imageSize );

				mipmaps.push( { "data": byteArray, "width": width, "height": height } );

				dataOffset += imageSize; // add size of the image for the next face/mipmap
				dataOffset += 3 - ( ( imageSize + 3 ) % 4 ); // add padding for odd sized image

			}
			width = Math.max( 1.0, width * 0.5 );
			height = Math.max( 1.0, height * 0.5 );

		}

		return mipmaps;

	};
```

Mugen87: @John-DeCorato-HZ Are you still interested in doing the change and add the suggested code? Right now, the PR can't be merged.

John-DeCorato-HZ: Yes, apologies. I'll check this out.

John-DeCorato-HZ: Fixed.

Mugen87 (approved):

mrdoob:
Thanks!

PR #16094, opened 2019-03-29
Update OrbitControls.js

mtsee:
Add a seek event, in the execution of drag, zoom, trigger this seek method, eg: do react performance optimization, animating do update() is very bad. 
use:  this.orbitControls.on('seek', () => {  do somthing });

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -85,6 +85,18 @@ THREE.OrbitControls = function ( object, domElement ) {
 	//
 	// public methods
 	//
+	
+	/**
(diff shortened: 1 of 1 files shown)


gkjohnson:
What's your goal with "seek"? Does the "change" event let you achieve what you're looking for?

```js
controls.addEventListener('change', () => {
  // controls have been moved
});
```

mtsee:
When used on react or vue, the controller needs to render when zooming or dragging, but currently only render when animate does, which can be very expensive.

gkjohnson:
If I'm understanding correctly and you only want to rerender the scene or your application when the controls are dragged or zoomed then you can achieve that with the `change` event I listed above:

```js
controls.addEventListener('change', () => {
  // rerender scene or application component
});
```

mtsee:
thanks! just this `change`

PR #16104, opened 2019-03-30
WebXR: Remove deprecated origin trial token.

Mugen87:
The trial ended at March 06 and there is no other/new trial you could register to.


mrdoob:
Thanks!

PR #16118, opened 2019-04-01
Address setFramebuffer not having any effect

dmarcos:
It might short sighted or I might be misusing API. It seems current method does not have any effect. Problem uncovered while working on https://github.com/mrdoob/three.js/pull/15604

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -189,7 +189,8 @@ function WebGLRenderer( parameters ) {
 			antialias: _antialias,
 			premultipliedAlpha: _premultipliedAlpha,
 			preserveDrawingBuffer: _preserveDrawingBuffer,
-			powerPreference: _powerPreference
+			powerPreference: _powerPreference,
+			xrCompatible: true
 		};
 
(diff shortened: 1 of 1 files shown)


src/renderers/WebGLRenderer.js
```diff
@@ -2421,6 +2421,8 @@ function WebGLRenderer( parameters ) {
 
 	this.setFramebuffer = function ( value ) {
 
+		if (_framebuffer !== value ) _gl.bindFramebuffer( _gl.FRAMEBUFFER, value );
+
```
takahirox: According to the behavior in `.setRenderTarget()`

https://github.com/mrdoob/three.js/blob/r103/src/renderers/WebGLRenderer.js#L2444-L2483

`_framebuffer` is used if `renderTarget` is `null`. So binding framebuffer here only if  `_currentRenderTarget` is `null`?

```javascript
if ( _framebuffer !== value && _currentRenderTarget === null ) _gl.bindFramebuffer( _gl.FRAMEBUFFER, value );
```

dmarcos: Thanks. Request incorporated in the PR

Mugen87:
see #17521

PR #16120, opened 2019-04-01
Editor: Add bounding sphere check

### LINKED ISSUE #16081: Editor: Problem loading some scenes "TypeError: geometry.boundingSphere is null"

Since r91, JSON files exported from r91 and previous versions of the editor will throw an error on the console and not load properly in more recent versions of the editor (threejs.org/editor). I've also encountered the error with some files exported from the editor as recently as r102.

The problem seems to be here: 
https://github.com/mrdoob/three.js/blob/dev/editor/js/Sidebar.Geometry.js#L182

Where `geometry.boundingSphere.radius` is used without checking whether `geometry.boundingSphere` is `null`. On the console I can put a breakpoint there, then call `geometry.computeBoundingSphere()`, then continue.

It looks like at least version 4.5 and above of the JSON scene format is expected to have a `boundingSphere` defined in its `geometry`. I found at least once instance where a scene exported from the live editor at threejs.org/editor (r102) doesn't have `boundingSphere` defined for one object. Perhaps that was because the object visibility was set to false. I've removed all the other objects and materials from the file and included it below for reference.

I think the easiest resolution would be something like
`if(geometry.boundingSphere === null) geometry.computeBoundingSphere()`

##### Three.js version

- [x] Dev
- [x] r103
- [x] all versions back to r91

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

```
{
	"metadata": {
		"version": 4.5,
		"type": "Object",
		"generator": "Object3D.toJSON"
	},
	"geometries": [
		{
			"uuid": "4DDF2CF9-2E72-4316-874F-28856BE99869",
			"type": "PlaneBufferGeometry",
			"width": 1.1,
			"height": 1.712598,
			"widthSegments": 1,
			"heightSegments": 1
		}],
	"materials": [
		{
			"uuid": "7426DD54-1B52-47C3-BFD4-3AC8FC80AC08",
			"type": "MeshBasicMaterial",
			"color": 16777215,
			"depthFunc": 3,
			"depthTest": true,
			"depthWrite": true,
			"wireframe": true
		}],
	"object": {
		"uuid": "85B71DC5-A721-4908-A8E3-E3852A075FB2",
		"type": "Scene",
		"name": "Scene",
		"layers": 1,
		"matrix": [1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1],
		"children": [
			{
				"uuid": "8C4782C0-8A22-4452-912A-87E789214E52",
				"type": "Mesh",
				"name": "Reference Plane",
				"visible": false,
				"layers": 1,
				"matrix": [1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1],
				"geometry": "4DDF2CF9-2E72-4316-874F-28856BE99869",
				"material": "7426DD54-1B52-47C3-BFD4-3AC8FC80AC08"
			}],
		"background": 11184810
	}
}
```

Temdog007:
Fixes #16081


mrdoob:
Thanks!

PR #16130, opened 2019-04-02
Camera: Introduce .dispose()

Mugen87:
Solves https://github.com/mrdoob/three.js/pull/12464#issuecomment-465563120

Analogous to `Scene.dispose()`. Allows more control about the disposal of internal cached objects.

### DIFF
--- a/docs/api/en/cameras/Camera.html
+++ b/docs/api/en/cameras/Camera.html
@@ -74,7 +74,8 @@
 
 		<h3>[method:null dispose]()</h3>
 		<p>
-		Clears camera related data internally cached by [page:WebGLRenderer].
+			Clears camera-related data structures cached by [page:WebGLRenderer].
+			You should call this method if you don't need the camera for rendering anymore.
 		</p>
 
 		<h3>[method:Vector3 getWorldDirection]( [param:Vector3 target] )</h3>
--- a/docs/api/en/cameras/Camera.html
+++ b/docs/api/en/cameras/Camera.html
@@ -72,6 +72,11 @@
 		Copy the properties from the source camera into this one.
 		</p>
 
+		<h3>[method:null dispose]()</h3>
+		<p>
+		Clears camera related data internally cached by [page:WebGLRenderer].
+		</p>
+
 		<h3>[method:Vector3 getWorldDirection]( [param:Vector3 target] )</h3>
 		<p>
 		[page:Vector3 target] — the result will be copied into this Vector3. <br /><br />
--- a/docs/manual/en/introduction/How-to-dispose-of-objects.html
+++ b/docs/manual/en/introduction/How-to-dispose-of-objects.html
@@ -54,11 +54,11 @@
 		for realizing custom rendering destinations. These objects are only deallocated by executing [page:WebGLRenderTarget.dispose]().
 	</p>
 
-	<h2>Scenes</h2>
+	<h2>Scenes and Cameras</h2>
 
 	<p>
-		The renderer maintains for scenes special data structures for sorting and rendering. If for some reasons a scene object becomes obsolete in an application,
-		call [page:Scene.dispose]() in order to free these resources.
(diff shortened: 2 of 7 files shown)


docs/api/en/cameras/Camera.html
```diff
@@ -72,6 +72,11 @@ <h3>[method:Camera copy]( [param:Camera source], [param:Boolean recursive] )</h3
 		Copy the properties from the source camera into this one.
 		</p>
 
+		<h3>[method:null dispose]()</h3>
+		<p>
+		Clears camera related data internally cached by [page:WebGLRenderer].
```
WestLangley: Clears camera-related data cached by [page:WebGLRenderer].
    You should call this method when you ___________.

docs/manual/en/introduction/How-to-dispose-of-objects.html
```diff
@@ -54,11 +54,11 @@ <h2>Render Targets</h2>
 		for realizing custom rendering destinations. These objects are only deallocated by executing [page:WebGLRenderTarget.dispose]().
 	</p>
 
-	<h2>Scenes</h2>
+	<h2>Scenes and Cameras</h2>
 
 	<p>
-		The renderer maintains for scenes special data structures for sorting and rendering. If for some reasons a scene object becomes obsolete in an application,
-		call [page:Scene.dispose]() in order to free these resources.
+		The renderer maintains for scenes and cameras special data structures for sorting and rendering. If for some reasons a camera or a scene becomes obsolete in an
```
WestLangley: >If for some reasons a camera or a scene becomes obsolete

What does that mean? Is there a better way to say this?

>The renderer maintains for scenes and cameras special data structures for sorting and rendering.

I do not think we need to explain what the data structures are for.

WestLangley:
Added some minor suggestions... :-)

docs/manual/en/introduction/How-to-dispose-of-objects.html
Mugen87: > What does that mean? Is there a better way to say this?

TBH, I don't know. `Scene.dispose()` and `Camera.dispose()` are indented for the case you don't need a scene or camera object anymore. Is the term "obsolete" not right in this context?

Mugen87: > I do not think we need to explain what the data structures are for.

A few users asked in the forum what's happening under the hood. So what is exactly cached and why. Since it's a guide and not a doc page, I thought adding this info is okay...

docs/api/en/cameras/Camera.html
Mugen87: "You should call this method when you don't need the camera for rendering anymore".

How about this? We can use a similar wording for `Scene.dispose()`?

WestLangley: maybe this:  

    Clears camera-related data structures cached by [page:WebGLRenderer].
    You should call this method if you don't need the camera for rendering anymore.

docs/manual/en/introduction/How-to-dispose-of-objects.html
WestLangley: >Is the term "obsolete" not right in this context?

I don't think so.

>A few users asked in the forum what's happening under the hood.

Then, tell them in the forum.

>So what is exactly cached and why.

What is written there is not descriptive enough to be useful.

The title is "How to dispose of objects". So you can say:

    If you delete a camera, or no longer need the camera for rendering, call camera.dispose() to clear camera-related data structures cached by WebGLRenderer.

Mugen87: Okay, sounds good.

Mugen87:
Update the PR with your suggestion :+1:

WestLangley:
@mrdoob I am happy with the docs. I'll defer review of the source to you.

errorx666:
Please add `Camera.dispose` to the TypeScript definitions.

Mugen87:
@errorx666 Okay, I'm doing this if @mrdoob approves the change in general.

PR #16131, opened 2019-04-02
Removal of the wrongly put @deprecated annotation

Zielon:
In the typings for `BufferGeometry` is wrongly put annotation of `deprecation` which causes problems.


mrdoob:
Thanks!

PR #16137, opened 2019-04-03
Editor: Set scene when loaded entirely

### LINKED ISSUE #15180: Editor: Texture map doesn't render when refresh web page in Chrome, but this problem doesn't reappear in Firefox.
##### Three.js version

- [x] Dev
- [x] r97
- [ ] ...

##### Browser

- [ ] All of them
- [x] Chrome 70
- [ ] Firefox 63

##### OS

- [x] Windows

---

image
image

Temdog007:
Fixes #15180

### DIFF
--- a/editor/index.html
+++ b/editor/index.html
@@ -209,9 +209,17 @@
 
 					if ( state !== undefined ) {
 
-						editor.fromJSON( state );
+						editor.fromJSON( state, setSelected );
+
+					} else {
+
+						setSelected();
 
 					}
+
+				} );
+
+				function setSelected() {
 
 					var selected = editor.config.getKey( 'selected' );
 
@@ -221,7 +229,8 @@
 
 					}
 
-				} );
+				}
+
 
 				//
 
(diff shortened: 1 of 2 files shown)


Mugen87:
This is a valid solution in order to fix the issue. However, `Editor.fromJSON()` is not synchronous anymore. Do you think this could be a problem for existing code?

Temdog007:
> Do you think this could be a problem for existing code?

The editor uses a series of signals to update so it being asynchronous shouldn't be and issue for the existing code. The only scenario I can think of that would be an issue is if there is a long delay before the editor updates and shows the scene because there are a lot of images it must parse.

 A solution to that would be to show a "loading" text or icon when `Editor.fromJSON()` is called and disable all of the GUI controls. Once the scene is loaded, it would be removed and the controls would be re-enabled. I think that would be a separate PR though.

Temdog007:
I think all "loading" code needs to be asynchronous in order for there to be consistency in the Editor. Closing for now in order to re-implement later.

PR #6405, opened 2015-04-15
Initial untested spherical harmonics implementation.

bhouston:
I've made this PR at @WestLangley's request.

This is an initial implementation of SphericalHarmonics, both the JS class and a shader.  I haven't tested it, but I think that this implementation form is pretty close to what we want as it is based on the general standards.

Thus this PR is for others to build upon.

PR #13115, opened 2018-01-15
Added SphericalHarmonicsLight.

atteneder:
SphericalHarmonicsLight is an ambient lighting term based on 3rd order spherical harmonics.

See the example I included to see its effect.

Motivation:
At the moment you can only add ambient lighting by one flat color value (THREE.AmbientLight). I was investigating how to get a more sophisticated visual result. Spherical Harmonics seem to be state of the art in other engines, so I tried to introduce it to three.js

Thoughts:
Only tested with one spherical harmonics light per scene. Having more than one doesn't really make sense I guess.

A follow up feature could be to retrieve/calculate the SH values from a cubemap directly. For now this can be done by tools like Lys ( https://www.knaldtech.com/lys/ ) or using this little Unity3d helper script i created:
https://gist.github.com/atteneder/d4fe33368afaf176413ec09d14ef41db

PR #16152, opened 2019-04-04
GPU cubemap to spherical harmonics

richardmonette:
Related https://github.com/mrdoob/three.js/issues/6251

Screen Shot 2019-04-04 at 4 20 02 PM

Spherical harmonics (GPU based) generator and example. Pulled in some of the code from https://github.com/mrdoob/three.js/pull/6405

Similar pipeline where we load an HDR cubemap, but instead of converting to PMREM, we convert to SH.

SH has the advantage of being represented, at 2nd order, by only 27 float values (9 * 3 channel RGB values). This impressively compact representation comes at the expense of only being able to capture low-mid frequency lighting information. 

I noticed that `envMap` actually supports several different `USE_ENVMAP_TYPE_XXX`s, so it was quite natural adding in one more representation of environment map lighting data.

SH_v_PMREM

SH on the left, PMREM on the right. Note how SH is plausibly consistent, but also that this only holds for rough surfaces. It's possible to approximate smooth reflections, but as significant data is omitted in the low-mid frequency representation mirror reflections are not possible. Still, this compact size makes this an attractive mapping type for multiple light probes. 

Also, a few performance notes:

SHGenerator uses one step: ~18.63ms
PMREMGenerator+PMREMCubeUVPacker: ~22.09ms + ~49.59ms

_Note: not really a fair performance comparison because PMREM is able to capture high frequency reflections, which SH does not capture. Still, interesting to see they are both in the same ballpark._

Per https://github.com/mrdoob/three.js/issues/6251, we still need to introduce a LightProbe meta system, where we pick the closest LightProbe and set it onto the material. This PR moves towards that goal by adding support for SH env map representation.

### DIFF
--- a/examples/js/loaders/SHGenerator.js
+++ b/examples/js/loaders/SHGenerator.js
@@ -1,0 +1,132 @@
+/**
+* @author Richard M. / https://github.com/richardmonette
+*/
+
+THREE.SHGenerator = function( sourceTexture ) {
+
+	this.sourceTexture = sourceTexture;
+
+	this.camera = new THREE.Camera();
+
+	this.shader = this.getShader();
+
+	this.planeMesh = new THREE.Mesh( new THREE.PlaneBufferGeometry( 2, 2 ), this.shader );
+	this.scene = new THREE.Scene();
+	this.scene.add( this.planeMesh );
+
+	var params = {
+		format: THREE.RGBFormat,
+		magFilter: THREE.NearestFilter,
+		minFilter: THREE.NearestFilter,
+		type: THREE.HalfFloatType,
+	};
+
+	this.renderTarget = new THREE.WebGLRenderTarget( 9, 1, params );
+	this.renderTarget.texture.name = "SHTexture";
+	this.renderTarget.texture.mapping = THREE.SphericalHarmonicReflectionMapping;
+
+};
+
+THREE.SHGenerator.prototype = {
+
+	constructor : THREE.SHGenerator,
+
+	update: function( renderer ) {
+
+		var currentRenderTarget = renderer.getRenderTarget();
+
+		renderer.setRenderTarget(this.renderTarget);
+		renderer.clear();
+		renderer.render( this.scene, this.camera );
+
+		renderer.setRenderTarget( currentRenderTarget );
+
+	},
+
+	dispose: function () {
+
+		this.shader.dispose();
+
+	},
+
+	getShader: function() {
+
+		return new THREE.ShaderMaterial( {
+
+			uniforms: {
+				"cubeMap": { value: this.sourceTexture },
+			},
+
+			vertexShader:
+				`
+				void main() {
+					gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
+				}
+				`,
+
+			fragmentShader:
+				`
+				#include <common>
+				uniform samplerCube cubeMap;
+
+				vec3 calcSH(int id, vec3 dir, vec3 color) {
+					float p_0_0 = 0.282094791773878140;
+					float p_1_0 = 0.488602511902919920 * dir.z;
+					float p_1_1 = -0.488602511902919920;
+					float p_2_0 = 0.946174695757560080 * dir.z * dir.z - 0.315391565252520050;
+					float p_2_1 = -1.092548430592079200 * dir.z;
+					float p_2_2 = 0.546274215296039590;
+
+					if (id == 0) {
+						return color * p_0_0;
+					} else if (id == 1) {
+						return color * p_1_1 * dir.y;
+					} else if (id == 2) {
+						return color * p_1_0;
+					} else if (id == 3) {
+						return color * p_1_1 * dir.x;
+					} else if (id == 4) {
+						return color * p_2_2 * (dir.x * dir.y + dir.y * dir.x);
+					} else if (id == 5) {
+						return color * p_2_1 * dir.y;
+					} else if (id == 6) {
+						return color * p_2_0;
+					} else if (id == 7) {
+						return color * p_2_1 * dir.x;
+					} else if (id == 8) {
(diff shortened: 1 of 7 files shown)


WestLangley:
Your code is based on r.91. You need to upgrade to r.103, in which case I will look at this more carefully.

> it was quite natural adding in one more representation of environment map lighting data.

An environment map models radiance; SH models irradiance.

SH cannot represent what three.js refers to as an environment map; SH coefficients _can_ be used as an alternative to an irradiance map, which three.js does not currently support.

SH is more appropriate for modeling global illumination, or as a generalization of `AmbientLight` or `HemisphereLight` -- light that is reflected diffusely.

My guess is we should add `SHLightProbe` and treat it as an indirect light source to complement the existing indirect light sources.

It would be interesting to hear how other libraries have incorporated SH into their API, and we can do something similar.

richardmonette:
> Your code is based on r.91. You need to upgrade to r.103, in which case I will look at this more carefully.

😬I guess it's been longer than I thought since I last submitted a PR. @WestLangley Rebased, thank you for your patience! 😂

> SH cannot represent what three.js refers to as an environment map; SH coefficients can be used as an alternative to an irradiance map, which three.js does not currently support.

I was looking for the right place to include SH information into the rendering pipeline and found `getLightProbeIndirectIrradiance`. 

As an aside, am I correct in understanding that in the cases of say `ENVMAP_TYPE_CUBE` and `ENVMAP_TYPE_CUBE_UV` we are using the first (i.e. blurriest, lowest resolution) mipmap level? 

At any rate, at least for the indirect irradiance, it seemed correct to plug in the SH data.

Then when it came to `getLightProbeIndirectRadiance` I realized that the SH data wasn't exactly going to be correct, but that it could be re-used as an approximation. When moving towards metalness 1.0 the surface does indeed appear more metallic, albeit with an apparently glossy reflection regardless of the roughness setting, as a result of having no different mipmap levels to sample. Agreed this probably isn't entirely correct, but if the only lighting data available is SH, maybe it's still worth using it?

When I was looking at @bhouston 's https://github.com/mrdoob/three.js/pull/6405/files#diff-601668f06145d1927138551eb971303c example code, the two sampling functions seemed to suggest one was for radiance and the other for irradiance. 

> My guess is we should add SHLightProbe and treat it as an indirect light source to complement the existing indirect light sources.

So that seems along the lines of https://github.com/mrdoob/three.js/pull/13115 

What I found interesting about that old PR is that here https://github.com/mrdoob/three.js/pull/13115/files#diff-a63071d948a349b79bddc096ce9f35f4R165 we would keep overwriting the uniforms if there was more than one SH light (whereas it would probably be better to select the closest probe.) Also, you can see how the SH data has been incorporated much the same as I did https://github.com/mrdoob/three.js/pull/13115/files#diff-7fd66278c32321da2cb1439758bf8f42R19 however, they did not touch `getLightProbeIndirectRadiance` 

Of course, the other major contribution in this new PR is the run-time SHGenerator in the first place. 

Keen to continue the discussion and hopefully we can find the right direction to proceed with this feature. 😄

bhouston:
> When I was looking at @bhouston 's https://github.com/mrdoob/three.js/pull/6405/files#diff-601668f06145d1927138551eb971303c example code, the two sampling functions seemed to suggest one was for radiance and the other for irradiance.

True.

One should either have a single global SH in which case you pass in the one set of values.  Or one should have probably 4 SHs each with a weight value to them.  And then you interpolate them.  This is similar to a skinning approach where you have to pick the N-closest vertices and weight them to find the skin value.  4 SHs works for both sparse radial basis functions (if using basic spare light probes) or tetrahedral-based bary centric coordinates (if using a tetrahedral mesh of light probes.)  One could normalize the weights for the radial basis approach.

This should be relatively easy to do.

Now one could either have a data texture with all weights for the SHs in the scene and the renderer could pass in the indices and weights to us (similar to skin approach.)  Or one could have 4 SHs + weights defined in the fragment parameters and just set them directly.

The renderer itself (or a class of the renderer) should be responsible for picking which which light probes to use for a particular object and their relative weights.  I suggest picking only a single set and weights of light probes per renderable object, otherwise it gets really complex.

Maybe start with a sparse set of points for the light probe field.  Pick the 4 closest and then figure out some type of radial basis weighting scheme with normalization.  This should be may 30 lines of code.

bhouston:
There seems to be different scene settings between the left and right examples.  I think the tone mapping type or exposure or gamma is different?  Probably gamma?

I can not properly compare left and right with these different settings, but they should not be as different as they appear in the example images you posted.  I say they are different because the orange of the texture on the plane in the right doesn't seem to be the same orange as the one on the left in the background.

richardmonette:
@WestLangley if you have a min, please checkout https://github.com/mrdoob/three.js/pull/16199 I've started working on implementing SH via ProbeLight, with basic blending between the ProbeLights in the scene. Is this more along the lines of what you are thinking?

p.s. I am travelling this week, so may bit a little slower to respond, but I'll work on this as fast as I can manage 🚀

WestLangley:
@richardmonette Will do. Glad you are working on this. There are several parts to the problem, and I see you are focusing on how to handle multiple probes.

Last weekend, I recovered some code that @bhouston and I worked on awhile ago, and I filed #16187 and #16191. Have a look. :-)

I am focusing more on the modeling part of the problem.

donmccurdy:
I'm not sure how to assess if there's anything in the PR that https://github.com/mrdoob/three.js/pull/17790 didn't cover — should this be closed?

WestLangley:
>should this be closed?

At this point, I would say 'yes'.

However, I think an interesting contribution of this PR is the use of the GPU to estimate SH coefficients. I have not verified it, though.

Mugen87:
Closing for now. The code is still available so one might reuse it in future PRs if necessary.

PR #16183, opened 2019-04-07
Add Back Face Culling Extension Support to LDrawLoader

gkjohnson:
This PR adds back face culling directive support to the `LDrawLoader`. The behavior of the extension is described in detail here: https://www.ldraw.org/article/415.

To handle double-sided polygons the triangles are now added to the model twice (facing two directions) but it doesn't look like any of the models in the examples folder require that and I'm not sure how common this is in general.

I'm going to be looking into adding geometry smoothing and vertex merging in a subsequent PR.

PS I love that there's a Lego loader in the library. Awesome work!

@yomboprime @mrdoob

### DIFF
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -475,7 +475,15 @@ THREE.LDrawLoader = ( function () {
 					subobject.url = subobjectURL;
 
 					// Load the subobject
-					subobjectLoad( subobjectURL, onSubobjectLoaded, undefined, onSubobjectError, subobject );
+					// Use another file loader here so we can keep track of the subobject information
+					// and use it when processing the next model.
+					var fileLoader = new THREE.FileLoader( scope.manager );
+					fileLoader.setPath( scope.path );
+					fileLoader.load( subobjectURL, function ( text ) {
+
+						processObject( text, onSubobjectLoaded, subobject );
+
+					}, undefined, onSubobjectError );
 
 				}
 
@@ -1162,6 +1170,7 @@ THREE.LDrawLoader = ( function () {
 
 								case 'BFC':
 
+									// Changes to the backface culling state
 									while ( ! lp.isAtTheEnd() ) {
 
 										var token = lp.getToken();
@@ -1265,6 +1274,8 @@ THREE.LDrawLoader = ( function () {
 
 						}
 
+						// If the scale of the object is negated then the triangle winding order
+						// needs to be flipped.
 						if ( scope.separateObjects === false && matrix.determinant() < 0 ) {
 
 							bfcInverted = ! bfcInverted;
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -1164,13 +1164,14 @@ THREE.LDrawLoader = ( function () {
 									while ( ! lp.isAtTheEnd() ) {
 
 										var token = lp.getToken();
-										bfcCCW = true;
+
 										switch ( token ) {
 
 											case 'CERTIFY':
 											case 'NOCERTIFY':
 
 												bfcEnabled = token === 'CERTIFY';
+												bfcCCW = true;
 
 												break;
 
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -1265,7 +1265,7 @@ THREE.LDrawLoader = ( function () {
 
 						}
 
-						if ( matrix.determinant() < 0 ) {
+						if ( scope.separateObjects === false && matrix.determinant() < 0 ) {
 
 							bfcInverted = ! bfcInverted;
 
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -1305,7 +1305,7 @@ THREE.LDrawLoader = ( function () {
 						var material = parseColourCode( lp );
 
 						var inverted = currentParseScope.inverted;
-						var ccw = ! bfcEnabled || bfcEnabled && bfcCCW && ! inverted;
+						var ccw = ! bfcEnabled || ( bfcCCW !== inverted );
 						var v0, v1, v2;
 
 						if ( ccw === true ) {
@@ -1338,7 +1338,7 @@ THREE.LDrawLoader = ( function () {
 						var material = parseColourCode( lp );
 
 						var inverted = currentParseScope.inverted;
-						var ccw = ! bfcEnabled || bfcEnabled && bfcCCW && ! inverted;
+						var ccw = ! bfcEnabled || ( bfcCCW !== inverted );
 						var v0, v1, v2, v3;
 
 						if ( ccw === true ) {
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -271,7 +271,19 @@ THREE.LDrawLoader = ( function () {
 
 			}, onProgress, onError );
 
-			function processObject( text, onProcessed ) {
+			function subobjectLoad( url, onLoad, onProgress, onError, subobject ) {
+
+				var fileLoader = new THREE.FileLoader( scope.manager );
+				fileLoader.setPath( scope.path );
+				fileLoader.load( url, function ( text ) {
+
+					processObject( text, onLoad, subobject );
+
+				}, onProgress, onError );
+
+			}
+
+			function processObject( text, onProcessed, subobject ) {
 
 				var parseScope = scope.newParseScopeLevel();
 				parseScope.url = url;
@@ -284,8 +296,9 @@ THREE.LDrawLoader = ( function () {
 
 					scope.subobjectCache[ currentFileName ] = text;
 
-
 				}
+
+				parseScope.inverted = subobject !== undefined ? subobject.inverted : false;
 
 				// Parse the object (returns a THREE.Group)
 				var objGroup = scope.parse( text );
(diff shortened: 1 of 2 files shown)


yomboprime:
Have you tried if it still works with the official repository of parts? I did not include it for testing because it is so many files. Please try to load some parts which references other files in the library. Just `setPath()` to the library folder.

Model downloads: http://omr.ldraw.org/
Official library needed for the models: http://www.ldraw.org/article/13.html

Otherwise the changes look right to me. Thanks!

Live demo: https://raw.githack.com/gkjohnson/three.js/ldraw-bfc/examples/webgl_loader_ldraw.html

gkjohnson:
@yomboprime 
I did test with some small cases (individual parts) but I'm having problems loading full sets even with the unaltered version on `dev`. Do you have a specific model you've tested with previously?

Here are the errors I'm getting:
- The loader can't find a color that the model is trying to use. [This](https://www.ldraw.org/article/547.html) article seems to imply that there's a default set of colors to use but I haven't looked at it too in depth. I got around this by just copy / pasting the color definitions into the model file.
- [This](http://omr.ldraw.org/file/778) model references internal models with uppercase names but the files are defined as lowercase. Maybe the loader should ignore case?
- With a lot of models I'm getting stackoverflow errors because of how the sub objects are being recursively loaded. If both model references are fixed in the above truck model you'll see this error.
- [This](http://omr.ldraw.org/file/84) model partially loads but there are pieces that are definitely missing. When the model is loaded with `separateObjects = true` the whole model loads correctly:

image

Having said that I'm seeing the same results between dev and this branch but if you have a good model you want me to test I'm happy to give it a go!

gkjohnson:
@yomboprime 

I fixed a few of those issues in another branch. I'm happy to make another PR with those changes or commit them into this one. You can see the diff between that branch and this one here: https://github.com/gkjohnson/three.js/compare/ldraw-bfc...gkjohnson:ldraw-fixes.

I haven't fixed the material colors issue but I'm curious as to what your thoughts are on it. Should it be expected that the user loads a material LDR first before loading the model to "seed" the material definitions?

The Willis Tower now works and here are some images of other models I've been able to load with the new fixes:

_10156-1 - LEGO Truck.mpd:_

image

_10182-1 - Cafe Corner.mpd:_

image

_10030-1 - Imperial Star Destroyer - UCS.mpd:_

image

yomboprime:
> @yomboprime
> I did test with some small cases (individual parts) but I'm having problems loading full sets even with the unaltered version on `dev`. Do you have a specific model you've tested with previously?
> 
> Here are the errors I'm getting:
> 
>     * The loader can't find a color that the model is trying to use. [This](https://www.ldraw.org/article/547.html) article seems to imply that there's a default set of colors to use but I haven't looked at it too in depth. I got around this by just copy / pasting the color definitions into the model file.
> 
>     * [This](http://omr.ldraw.org/file/778) model references internal models with uppercase names but the files are defined as lowercase. Maybe the loader should ignore case?
> 
>     * With a lot of models I'm getting stackoverflow errors because of how the sub objects are being recursively loaded. If both model references are fixed in the above truck model you'll see this error.
> 
>     * [This](http://omr.ldraw.org/file/84) model partially loads but there are pieces that are definitely missing. When the model is loaded with `separateObjects = true` the whole model loads correctly:
> 
> 
> Having said that I'm seeing the same results between dev and this branch but if you have a good model you want me to test I'm happy to give it a go!

@gkjohnson 

About colors: Yes, the file to load is `LDCONFIG.LDR`, which is in the root folder of the library and contains all official color definitions. You can do this by just loading the mentioned file and after that, loading the model file (with the same `LDrawLoader`, because the member `materials` is kept)

About stack overflow errors: Yes, that's why I ended up making the packing utility.

About upper/lower case: As you've seen, when loading from the library the loader has to try different file paths because there are standard folder locations where a file could be, and the loader can't know in advance in which folder it is because the previous file that makes the reference only gives the file name (usually). The trial and errors of paths is done in the function `loadSubobject`, and it also tries to lower case the file name, so this should work. The retry itself is in the function `onSubobjectLoaded`.

I don't know if the function you added, `subobjectLoad` (and the rest of the code in your branch `ldraw-fixes`), will keep this trial of paths and lowercase working. But I feel it is all right.

Some notes:
 - The two functions `subobjectLoad` and `loadSubobject` make confusion in my opinion. Since one of them is called only once from the other, could you merge these two functions?
- What does this do?: `Promise.resolve().then( function () {`

I don't have additional models but in the library you can load parts as if they were models. There are tons of them, and they reference other simpler parts. The file extension doesn't matter.

gkjohnson:
> About colors: Yes, the file to load is `LDCONFIG.LDR` ...

Great! Easy enough.

> About stack overflow errors: Yes, that's why I ended up making the packing utility.

My impression is that the stackoverflow errors are being caused by the models being recursively processed synchronously in a single file or grabbed from the cache so I'm not sure if packing would fix it. When fetching models from the server the stack is more or less "reset" because it's async.

> The trial and errors of paths is done in the function `loadSubobject`, and it also tries to lower case the file name, so this should work.

Does this happen for internally defined files, as well, that are just expected to be pulled from the cache? It looks like the cached path uses the `originalFileName` to check the cache while the `fileName` field is made lowercase. In the case of the truck model the minifigure is referenced as `MINIFIG.LDR` while the internal file definition is `minifig.ldr`. I'm pretty sure I saw this problem in `dev`, too, and the code in my branches will still continue to call `onSubobjectError` which tries to load the model again until all options fail so it should work the same but I can dig into this a bit more to see if I broke something.

> The two functions `subobjectLoad` and `loadSubobject` make confusion in my opinion. Since one of them is called only once from the other, could you merge these two functions?

Definitely agree -- I noticed this while working on the other branch.

> What does this do?: `Promise.resolve().then( function () {`

I should have explained this more. This is to fix the stack overflow issue by making all file processing (even internal and cached files) asynchronous. `Promise.resolve().then` will run _before_ this frame is rendered while `requestAnimationFrame` will run _after_, which means the loading will take much longer if `rAF` is used. I admit the use of `Promise.resolve()` is a bit hacky here but it's a minimal way to fix the stack overflow problem. It might be best to use a depth first stack data structure to process the files and prevent too much recursion but that would require a larger change.

yomboprime:
> > The trial and errors of paths is done in the function `loadSubobject`, and it also tries to lower case the file name, so this should work.
> 
> Does this happen for internally defined files, as well, that are just expected to be pulled from the cache?

No, objects in cache were succesfully loaded previously, or cached from the packed file, so no try&error.

> It looks like the cached path uses the `originalFileName` to check the cache while the `fileName` field is made lowercase. In the case of the truck model the minifigure is referenced as `MINIFIG.LDR` while the internal file definition is `minifig.ldr`. I'm pretty sure I saw this problem in `dev`, too, and the code in my branches will still continue to call `onSubobjectError` which tries to load the model again until all options fail so it should work the same but I can dig into this a bit more to see if I broke something.

All right, if you can load the hotel and the imperial destructor, then you've got farther than me and your code is better :-) (I tried those models without luck)

> > What does this do?: `Promise.resolve().then( function () {`
> 
> I should have explained this more. This is to fix the stack overflow issue by making all file processing (even internal and cached files) asynchronous. `Promise.resolve().then` will run _before_ this frame is rendered while `requestAnimationFrame` will run _after_, which means the loading will take much longer if `rAF` is used. I admit the use of `Promise.resolve()` is a bit hacky here but it's a minimal way to fix the stack overflow problem. It might be best to use a depth first stack data structure to process the files and prevent too much recursion but that would require a larger change.

Thanks for the explanation.

gkjohnson:
Ha I really wanted to get the Imperial Star Destroyer to load D:

> No, objects in cache were succesfully loaded previously, or cached from the packed file, so no try&error.

I see -- it looks like it's possible for models to refer to the same file in different ways throughout so maybe it's best if the cached file names are just made lowercase to ensure consistency.

Okay so my plan now is to fix the `subobjectLoad` function naming and then I'll make the other changes in another PR.

Thanks!

gkjohnson:
@mrdoob @yomboprime 

I removed the confusingly named `subobjectLoad` function and just moved the logic into `loadSubobject` because it was only being used there so I think this PR is ready to merge if everything else looks good!

I'll make a follow up PR with the other changes.

> But, won't this make the cache fail always? Different files can reference common files in upper or in lower case indistinctively.

I think it should be okay if we just lowercase every key before putting it into the cache so there's no ambiguity.

yomboprime:
> I removed the confusingly named `subobjectLoad` function and just moved the logic into `loadSubobject` because it was only being used there so I think this PR is ready to merge if everything else looks good!
> 
> I'll make a follow up PR with the other changes.
Ok!

> > But, won't this make the cache fail always? Different files can reference common files in upper or in lower case indistinctively.
> 
> I think it should be okay if we just lowercase every key before putting it into the cache so there's no ambiguity.

Yes, sorry, I mean while loading from different files from the library (not when loading from a packed model)

gkjohnson:
> Yes, sorry, I mean while loading from different files from the library (not when loading from a packed model)

It should be the same, right? The server-loaded models can still be fetched using the same fallback strategies but once they're found they're placed in the cache indexed by the lowercase filename -- that's what happens in the other branch I linked to. Sorry if I'm misunderstanding!

yomboprime:
> > Yes, sorry, I mean while loading from different files from the library (not when loading from a packed model)
> 
> It should be the same, right? The server-loaded models can still be fetched using the same fallback strategies but once they're found they're placed in the cache indexed by the lowercase filename -- that's what happens in the other branch I linked to. Sorry if I'm misunderstanding!

Ok, I understand. I just wanted to make sure. Thanks.

mrdoob:
@yomboprime looks good?

yomboprime:
> @yomboprime looks good?

@mrdoob Yes. The only thing I'm not sure about is the `Promise.resolve().then(` call which has replaced the recursive call to the main load() function. But I think it is ok. It is only that I didn't understand it.

gkjohnson:
@mrdoob @yomboprime 

> The only thing I'm not sure about is the Promise.resolve().then( call which has replaced the recursive call

Just to clarify this change is not in this PR -- I'll add that in a subsequent one. This PR only adds the BFC extension support.

yomboprime:
> @mrdoob @yomboprime
> 
> > The only thing I'm not sure about is the Promise.resolve().then( call which has replaced the recursive call
> 
> Just to clarify this change is not in this PR -- I'll add that in a subsequent one. This PR only adds the BFC extension support.

Yes, sorry, that is correct.

mrdoob:
Thanks!

PR #16188, opened 2019-04-08
Editor: Ensure viewport is valid

Temdog007:
Issue was that when a camera is removed, it wasn't removed from the list of viewport cameras. This issue also occurred when the editor was cleared.


Temdog007:
Closing due to lack of interest.

PR #16191, opened 2019-04-08
Added LightProbe class

WestLangley:
We will have to continue to discuss this, and what it represents. To me, it models indirect, ambient light, and in that sense, is a generalization of `AmbientLight` or `HemisphereLight`.

I've included a method `LightProbe.setFromCubeTexture()` that computes the SH coefficients. From testing, it works as one would expect. The method can be replaced with a faster GPU-based one later.

This PR relies on #16187.

### DIFF
--- a/src/lights/LightProbe.js
+++ b/src/lights/LightProbe.js
@@ -1,0 +1,162 @@
+import { _Math } from '../math/Math.js';
+import { Vector3 } from '../math/Vector3.js';
+import { Color } from '../math/Color.js';
+import { SphericalHarmonics3 } from '../math/SphericalHarmonics3.js';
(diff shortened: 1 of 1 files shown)


WestLangley:
WIP. Sphere illuminated by LightProbe only. Probe SH coefficients computed from environment map. 
Screen Shot 2019-04-07 at 11 54 44 PM

richardmonette:
This looks awesome! 

> The method can be replaced with a faster GPU-based one later.

I've been working on a GPU based version here https://github.com/mrdoob/three.js/pull/16152 which I think I can make a bit more mathematically correct referencing this CPU implementation!

mrdoob:
Thanks!

PR #16201, opened 2019-04-08
Refactored PointerLockControls

### LINKED ISSUE #16160: PointerLockControls needs better implementation

arodic:
This fixes #16160 

With this new implementation, `PointerLockControls` does not change camera hierarchy and it does not require camera rotation to be 0,0,0.

In addition, this implementation exposes `reverseY`, `sensitivity`, `minPolarAngle` and `maxPolarAngle` similar to `OrbitControls`, and it emits `change` event like other controls.

This is a breaking change because it requires some code to be removed from applications using it. Most notably, there is no need to do `scene.add( controls.getObject() );`.


examples/js/controls/PointerLockControls.js
```diff
@@ -83,28 +116,33 @@ THREE.PointerLockControls = function ( camera, domElement ) {
...
 
-		return yawObject;
+		console.warn( 'THREE.PointerLockControls: getObject() has been deprecated.' );
+
+		return scope.object;
 
 	};
 
-	this.getDirection = function () {
+	this.getDirection = function ( v ) {
```
arodic: I noticed that getDirection function is not used anywhere. Perhaps it is safe to remove?

WestLangley:
I think #16171 is a cleaner implementation.

examples/js/controls/PointerLockControls.js
WestLangley: Other users may use it.

looeee:
I don't have any opinion on which implementation is better, but I do like the idea of exposing `minPolarAngle` and `maxPolarAngle`, and adding `change` events - whichever implementation we choose should include these.

arodic:
I'm gonna close this PR in hopes to prevent the stall on #16171. Once it is merged, I will open another PR with the rest of the fixes.

PR #16205, opened 2019-04-09
Docs: Improve material browser.

Mugen87:
- Removed `grasslight-thin.jpg`. Introduce `brick_diffuse.jpg` instead.
- Added an alpha map. Removed option for setting a specular or light map (currently the diffuse textures are used for both).
- Fixed some minor `dat.gui` issues (slider and default values)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev28/docs/index.html

PR #16206, opened 2019-04-09
Introduce .autoUpdate to LOD

takahirox:
Currently `LOD` requires manually calling `.update()` after updating world matrices in a scene as `webgl_lod` example does in render loop. Users also need to manually call `scene.updateMatrixWorld()` so that it implicitly requires `scene.autoUpdate = false` for efficiency, otherwise `scene.updateMatrixWorld()` is called again in renderer.

```javascript

scene.autoUpdate = false;

function render() {

	scene.updateMatrixWorld();

	scene.traverse( function ( object ) {

		if ( object instanceof THREE.LOD ) {

			object.update( camera );

		}

	} );

}
```

This seems a bit complicated API. I want to automatically update.

I'd like to suggest introducing `.autoUpdate` to `LOD`. If it's `true` (default), renderer calls `LOD.update( camera )` in `projectObject()` which renderer calls after updating world matrices in a scene.

If a user thinks updating every frame is costly, one can set `.autoUpdate = false` and manually update.

This PR is still draft because I'm wondering if there is any special reasons why currently we don't do update automatically. Feedback is welcome. If this change looks ok, I'll update the document.


takahirox:
Updated the documents. I think ready.

docs/api/en/objects/LOD.html
```diff
@@ -60,6 +53,13 @@ <h3>[name]( )</h3>
 		<h2>Properties</h2>
 		<p>See the base [page:Object3D] class for common properties.</p>
 
+		<h3>[property:boolean autoUpdate]</h3>
+		<p>
+		Default is true. If set, then the renderer calls [page:.update] every frame.
+		When it isn't, then you have to manually call by yourself in the render loop
+		for levels of detail to be updated dynamically.
```
Mugen87: Maybe like so:
```
Whether the LOD object is updated automatically by the renderer per frame or not. If set to false, you have to call [page:LOD.update]() in the render loop by yourself. Default is true.
```

Mugen87:
BTW: This change needs to be noted in the migration guide. With not change on app level, LOD objects will be updated twice (by the app and the renderer).

docs/api/en/objects/LOD.html
takahirox: Updated, thanks.

takahirox:
Yup that's the plan if merged.

Mugen87 (approved):

takahirox:
Friendly ping @mrdoob

mrdoob:
Thanks!

PR #16217, opened 2019-04-11
Make Box3.expandByObject use geometry.boundingBox

mqp:
This seems preferable to the existing implementation for three reasons:

1. It reuses the geometry's vertex iteration code, so it's more concise and doesn't need to distinguish between varieties of geometry representations.
2. It is polite to re-use (or populate if not present) the geometry's cached bounding box, rather than recomputing it every time.
3. It should be much more efficient to compute each child bounding box in local space and only thereafter apply the object's transform to the maximum and minimum of the child box and expand the root's bounds, rather than to apply the object's transform to each vertex on the child geometry and expand the root's bounds one vertex at a time.

### DIFF
--- a/src/math/Box3.js
+++ b/src/math/Box3.js
@@ -220,9 +220,9 @@ Object.assign( Box3.prototype, {
 		// Computes the world-axis-aligned bounding box of an object (including its children),
 		// accounting for both the object's, and children's, world transforms
 
-		var scope, i, l;
+		var scope;
 
-		var v1 = new Vector3();
+		var bounds = new Box3();
 
 		function traverse( node ) {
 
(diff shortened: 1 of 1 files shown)


mqp:
By the way, I also noticed that `BufferGeometry.computeBoundingSphere` seems like it could reuse or populate the `boundingBox` as [part of its operation](https://github.com/mrdoob/three.js/blob/dev/src/core/BufferGeometry.js#L672), but it does not. Should it be changed to do so?

WestLangley:
Can you try this test case and compare your result with the current method?

```js
var object = new THREE.Mesh( new THREE.SphereBufferGeometry( 1, 32, 16 ) );

object.rotation.z = Math.PI / 4;

console.log( new THREE.Box3().expandByObject( object ) );
```

mqp:
I see your point -- unfortunately my number 3 is not a valid optimization, because "compute bounding box" and "apply rotation" aren't associative.

I can't think of any way to reuse the geometry's bounding box any more given this (unless the object's rotation is zero -- but I don't particularly feel like writing this special case) so closing this unless I come up with something.

Issue #16132, opened 2019-04-02
Suggestion: Making BufferGeometry.attributes/index properties private or immutable


takahirox:
##### Description of the problem

Currently there are two ways to access `attributes` and `index` properties of `BufferGeometry`. 

1. Directly accessing like `geometry.attributes/index`
2. Using accessing methods `geometry.get/add/removeAttribute()` and `geometry.get/setIndex()`.

I'd like to suggest letting 1 be deprecated by making `.attributes/index` properties private.

**Purposes**

1. For clearer API.

The existence of two ways to access geometry's `attributes/index` may be a bit confusing to users. When should they use accessing methods and when should they directly access? I think the only one way using accessing method is clearer API.

Actually, in `BufferGeometry` document we already encourage using accessing methods rather than directly accessing. So I don't think making them private will be so big trouble.

https://threejs.org/docs/#api/en/core/BufferGeometry

> you must call .getAttribute to access the 'position' attribute

> .attributes : Object
> This hashmap has as id the name of the attribute to be set and as value the buffer to set it to. Rather than accessing this property directly, use .addAttribute and .getAttribute to access attributes of this geometry.

> .addAttribute ( name : String, attribute : BufferAttribute ) : BufferGeometry
> Adds an attribute to this geometry. Use this rather than the attributes property, because an internal hashmap of .attributes is maintained to speed up iterating over attributes.

2. For VAO implementation.

We've been working on VAO implementation in #14600. We proposed assigning one VAO to one geometry which manages the set of attributes and index. VAO is for performance, so we need to quickly detect whether we need to update VAO or not. Taking time each frame to judge if we need the update doesn't make sense.

So I proposed attributes/index addition and removal to/from geometry observation. But currently Three.js API is too flexible that users can directly access `attributes/index` properties and add/remove to/from them. This makes the observation hard. Making the properties private can have it easy because we only need to observe accessing methods.

**Disadvantage**

1. Breaking core/examples/user code.

I see a lot of codes directly accessing `geometry.attributes/index`. If we make the properties private, such codes need to be updated.

2. We need to add `.getAttributeNames()` method (or something)

I sometimes see this type of code.

```javascript
for ( var attributeName in geometry.attributes ) {
    var attribute = geometry.attributes[ attributeName ];
    ....
}
```

It can't be handled with the current attribute accessing methods. Probably we will need to add a new method `.getAttributeNames()` which returns attribute names as an array.

Feedback is welcome.

##### Three.js version

[...]

PR #16232, opened 2019-04-12
Deprecate BufferGeometry.index/attributes direct access.

takahirox:
From #16132. To show how the change will look like and how big the change will be.

This PR includes `core`, `tests`, `examples/*.html`, `examples/js/loaders/GLTFLoader.js`. I'd like to make another PR for `examples/js/*.js` if we go with this change.

I added private (more precisely friend in C++) properties `._attributes` and `._index` to `BufferGeometry`. Core codes still can directly access them, not via getter for performance and less change.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -539,9 +539,13 @@ THREE.GLTFLoader = ( function () {
 
 				dracoLoader.decodeDracoFile( bufferView, function ( geometry ) {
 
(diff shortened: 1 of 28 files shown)


WestLangley:
Whether this is a "cleaner" API is a matter of opinion.

So, ignoring that reason for now, could you please summarize the consequences of _not_ making this proposed change.

takahirox:
Closing for now because I made VAO without any breaking existing APIs PR #16287 We can revisit later if needed.

PR #16249, opened 2019-04-14
Editor: Make `#toolbar` centered correctly.

### LINKED ISSUE #16248: Editor: #toolbar dom should not set width in css
image
image

linbingquan:
This PR fix #16248.

[WIP] I try to descript this PR in Chinglish. I don't know how to make `#toolbar` center in `#viewport`(canvas DOM). I make `#toolbar` right and bottom in `#viewpot`(canvas DOM).

  @mrdoob @Mugen87 Please look at this screenshot, how do you think.

PC screenshot:
image
image

Mobile screeshot:
image
image

### DIFF
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -207,7 +207,7 @@ select {
 	bottom: 10px;
 	background-color: #111;
 	color: #333;
-	white-space: nowrap; /* for zh */
+	white-space: nowrap; /* for button */
 }
 
 	#toolbar * {
@@ -215,17 +215,8 @@ select {
 	}
 
 	#toolbar .Panel {
-		display: flex;
-		justify-content: center;
-		align-items: center;
 		padding: 4px;
 		color: #888;
-	}
-
-	#toolbar .Panel span {
-		display: flex;
-		justify-content: center;
-		align-items: center;
 	}
 
 	#toolbar button {
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -203,7 +203,6 @@ select {
 #toolbar {
 	position: absolute;
 	left: calc(50% - 290px); /* ( ( 100% - 300px ) / 2.0 ) - 140px */
-	width: 280px;
 	bottom: 16px;
 	height: 32px;
 	background-color: #111;
@@ -293,7 +292,6 @@ select {
 
 	#toolbar {
 		left: calc(50% - 140px);
-		width: 280px;
 		top: 68px;
 	}
 
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -202,12 +202,12 @@ select {
 
 #toolbar {
 	position: absolute;
-	left: 50%;
+	left: calc( 50% - 150px ); /* 150px: #sidebar width / 2 */
 	transform: translateX(-50%);
 	bottom: 10px;
 	background-color: #111;
 	color: #333;
-	white-space: nowrap;
+	white-space: nowrap; /* for zh */
 }
 
 	#toolbar * {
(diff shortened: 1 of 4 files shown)


Mugen87:
> I make #toolbar right and bottom in #viewpot(canvas DOM).

Why do you think it's better to position the toolbar like that. I've always felt the center/bottom toolbar is nicely clear since the focus object is normally in your screen's center.

Mugen87:
http://raw.githack.com/linbingquan/three.js/dev-editor/editor/index.html

editor/css/dark.css
```diff
@@ -203,7 +203,6 @@ select {
 #toolbar {
 	position: absolute;
 	left: calc(50% - 290px); /* ( ( 100% - 300px ) / 2.0 ) - 140px */
```
Mugen87: Removing the `width` definition is critical since the `left` computation is now not valid anymore. It seems you need a different solution if you want to correctly center a toolbar with dynamic width.

linbingquan: I remove `width` to fix #16248.

Mugen87: But even if this PR solves one style issue, it introduces another one.

linbingquan:
@Mugen87 How about this solution, I make `#toolbar` position center with dynamic width.

http://raw.githack.com/linbingquan/three.js/dev-editor/editor/index.html

Mugen87:
That looks indeed better!

linbingquan:
> That looks indeed better! However, there still seems some issue when using the editor on a mobile device.

TBH, this PR is not very good, I also found `Local` element not in a line this issue,  but I am not sure I found this issue as you say.

---

I tried to make it like that (that PR with `new UI.Panel()` I wasn't commit it), how do you think.

iPad screenshot with :
image

linbingquan:
@Mugen87 I made `#toolbar` position correct in mobile phone, but I found click event issue, when I clicked `#toolbar button` after the `SelectedObj` not be selected.

linbingquan:
@Mugen87 I made `#toolbar` position correct for mobile phone.

Mugen87:
Looks better! I just wonder if @mrdoob is okay with using `display: flex;` since the flexbox support in IE11 is poor.

linbingquan:
> Looks better! I just wonder if @mrdoob is okay with using `display: flex;` since the flexbox support in IE11 is poor.

I tried to remove `display: flex;`, the `#toolbar` also looks correct in PC and mobile phone.

Mugen87:
Just one tiny addition: The toolbar is not centered correctly when using Chinese language setting. However, it looks still better than the current implementation so I guess its okay.

linbingquan:
> The toolbar is not centered correctly when using Chinese language setting

@Mugen87 I am sorry, I don't get your point. I think this PR is OK when I tested in PC and mobile. Please tell me, maybe I don't know or forget some thing in this `#toolbar`'s position.

Mugen87:
Never mind^^. The behavior of your version is similar to the current one. I've just meant that ideally the toolbar is automatically centered at the center and not via a left offset. This can be changed at a later point, too.

Mugen87 (approved):

linbingquan:
@mrdoob How do you feel about this PR.

Mugen87:
Not sure this is a improvement after all. The status quo seems good enough...

mrdoob:
The toolbar needs a bit of redesign.

I think having it be vertical with icons on the top left (like blender 2.8) may be better, plus it won't need this center alignment.

linbingquan:
I am not sure blender 2.8 icons is free, and not sure which icons is better(use other icons not use blender 2.8). @mrdoob Can you do me a favor.

I like blender 2.8.
图片

Issue #16234, opened 2019-04-13
Editor should have most of THREE's features


Temdog007:
##### Description of the problem

There are several features of `THREE` that aren't accessible from the Editor. When I say accessible, I mean combination of:
* properties are not visible for editing
* objects not being able to be created
* objects not being able to be imported or exported. 

Ideally, the Editor should be capable of creating and editing most objects and features that the API contains. This issue is really more to keep track of what the Editor can and can't do currently. I'm listing features that I think the Editor would be capable of handling.

#### Cameras
- [x] `PerspectiveCamera`
- [ ] `OrthographicCamera` (#16008)
- [ ] `ArrayCamera`
- [ ] `CubeCamera`
- [ ] `StereoCamera`
#### Core
- [x] `BufferAttribute`
- [x] `BufferGeometry`
- [ ] `Layers`
- [ ] `InstancedBufferAttribute` *
- [ ] `InstancedBufferGeometry` *
- [ ] `InstancedInterleavedBuffer` *
- [ ] `InterleavedBufferAttribute` *
#### Geometry
- [x] `BoxGeometry`
- [x] `CircleGeometry`
- [x] `ConeGeometry`
- [x] `CylinderGeometry`
- [x] `DodecahedronGeometry`
- [ ] `EdgesGeometry` **
- [x] `ExtrudeGeometry`
- [x] `IcosahedronGeometry`
- [x] `LatheGeometry`
- [ ] `OctahedronGeometry`
- [ ] `ParametricGeometry`
- [x] `PlaneGeometry`
- [ ] `RingGeometry`
- [ ] `ShapeGeometry`
- [x] `SphereGeometry`
- [ ] `TetrahedronGeometry`
- [ ] `TextGeometry` **
- [x] `TorusGeometry`
- [x] `TorusKnotGemetry`
- [x] `TubeGeometry`
- [ ] `WireframeGeometry` **
#### Lights
- [x] `AmbientLight`
- [x] `DirectionalLight`
- [x] `HemisphereLight`
- [x] `PointLight`
- [ ] `RectAreaLight`
- [x] `SpotLight`
- [ ] `LightProbe`
#### Materials
- [x] `LineBasicMaterial`
- [x] `LineDashedMaterial`
- [x] `MeshBasicMaterial`
- [x] `MeshDepthMaterial`
- [x] `MeshLambertMaterial`
- [x] `MeshMatcapMaterial`
- [x] `MeshNormalMaterial`
[...]

PR #16251, opened 2019-04-14
Editor: Add RectAreaLight

Temdog007:
See #16234


Mugen87:
It seems the PR has a merge conflict...

Mugen87:
To get `RectAreaLight`s work properly, you have to include `RectAreaLightUniformsLib` and call the respective `init()` method.

editor/js/Sidebar.Object.js
```diff
@@ -433,6 +453,18 @@ Sidebar.Object = function ( editor ) {
 
 			}
 
+			if ( object.width !== undefined && Math.abs( object.width - objectWidth.getValue() ) >= 0.01 ) {
+
+				editor.execute( new SetValueCommand( object, 'width', objectWidth.getValue() ) );
```
Mugen87: You now have to pass the `editor` variable to all `Command` constructors e.g.
```js
new SetValueCommand( editor, object, 'width', objectWidth.getValue()
```

editor/js/Editor.js
```diff
@@ -619,3 +623,40 @@ Editor.prototype = {
 	}
 
 };
+
+// RectAreaLightHelper was designed to work only if it is a child of its light.
+// Since that cannot be done in the edtior, the updateMatrixWorld function is overwritten to apply the matrix calculations to the light.
+THREE.RectAreaLightHelper.prototype.updateMatrixWorld = function () {
```
Mugen87: > RectAreaLightHelper was designed to work only if it is a child of its light.

@WestLangley Do you know the reasons for this? I've checked the history of `RectAreaLightHelper` but found nothing.

Other users have already run into issues because of how `RectAreaLightHelper` works (see https://discourse.threejs.org/t/rectarealight-remaining-at-0-0-0/10057). I think it would be less confusing if adding the helper to the light is not required (similar to e.g. `PointLightHelper`). This modification here would not be necessary then.

WestLangley: >Do you know the reasons for this? I've checked the history of RectAreaLightHelper but found nothing.

#14935

I expect all light helpers will be changed to match the behavior of `RectAreaLightHelper` once lights are controlled by the quaternion.

Mugen87: As long as we try to keep the helpers consistent, I'm fine with it. Thanks for the link 👍

Mugen87:
Since the editor is now module based (#18201), a lot of code has changed. It's better to implement this feature from scratch.

PR #16005, opened 2019-03-17
Editor: Allow Tube Geometry to edited by dragging points on screen

Temdog007:
Tube Edit Demo

If this PR goes through, my plan is add support for `Points` and `Lines` (and maybe `Paths`?) to the editor.

PR #16252, opened 2019-04-14
Editor: Added UI.Points

Temdog007:
Both `Sidebar.Geometry.LatheGeometry` and `Sidebar.Geometry.TubeGeometry` use a series of UI elements to represent a list of points. Since this code will be useful for when `THREE.Line`, `THREE.Points` and `THREE.Shapes` eventually get added to the editor, I refactored the code into one UI element called `UI.Points`.

Change was originally done in #16005 but I decided it should be its own PR.


mrdoob:
Thanks!

PR #16256, opened 2019-04-15
LoaderUtils: Added parseData() function

### LINKED ISSUE #16218: ColladaLoader: parsing optimisation
Hi there!
It's about ColladaLoader (I didn't check others).
If you replace the code `text.trim().split( /\s+/ )` by `text.replace(/\n/g, " ").replace(/ \s+/g, " ").trim().split(" ")` (see parseFloat and parseInt functions) the parser will work much faster.

bbmv:
Fixes #16218

### DIFF
--- a/src/loaders/LoaderUtils.js
+++ b/src/loaders/LoaderUtils.js
@@ -45,7 +45,7 @@ var LoaderUtils = {
 		return text.replace( /\n/g, " " ).
 			replace( /\r/g, " " ).
 			replace( /\t/g, " " ).
-			replace( /\s\s+/g, " " ).
+			replace( /  +/g, " " ). // double space
 			trim().
 			split( " " );
 
--- a/src/loaders/LoaderUtils.js
+++ b/src/loaders/LoaderUtils.js
@@ -46,7 +46,7 @@ var LoaderUtils = {
 			replace( /\r/g, " " ).
 			replace( /\t/g, " " ).
 			replace( / +/g, " " ).
-			trim( ).
+			trim().
 			split( " " );
 
 	}
--- a/src/loaders/LoaderUtils.js
+++ b/src/loaders/LoaderUtils.js
@@ -55,7 +55,7 @@ var LoaderUtils = {
 
 function isSafari() {
 
-	var browser = ( navigator ) ? navigator.userAgent.match( /chrome|opera|safari|firefox|msie|trident/i ) : null;
+	var browser = ( typeof( navigator ) !== "undefined" ) ? navigator.userAgent.match( /chrome|opera|safari|firefox|msie|trident/i ) : null;
 
 	return ( browser !== null ) ? /safari/i.test( browser[ 0 ] ) : false;
 
--- a/src/loaders/LoaderUtils.js
+++ b/src/loaders/LoaderUtils.js
@@ -42,7 +42,7 @@ var LoaderUtils = {
 
 	splitByWhitespace: function ( text ) {
 
-		return text.replace( /\n/g, " " ).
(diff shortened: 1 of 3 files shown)


Mugen87:
How about using this function now in `ColladaLoader`? Besides, it would be also great if you add an unit test right here: https://github.com/mrdoob/three.js/blob/dev/test/unit/src/loaders/LoaderUtils.tests.js

src/loaders/LoaderUtils.js
```diff
@@ -37,6 +38,17 @@ var LoaderUtils = {
...
 
+	},
+
+	parseNumbers: function ( text ) {
+
+		return text.replace( /\n/g, " " ).
+			replace( /\r/g, " " ).
+			replace( /\t/g, " " ).
+			replace( / +/g, " " ).
+			trim( ).
```
Mugen87: `trim().` is better (without the space)

bbmv:
I renamed the function to parseData.

Mugen87 (dismissed):

donmccurdy:
Both 'parse' and 'data' are somewhat vague terms for what this function does I think, especially in the context of a common utility class like LoaderUtils, which is shared by other formats' loaders, not all of which are whitespace delimited. Perhaps consider naming this `splitByWhitespace` instead?

Also note that regex has a dedicated character for whitespace, `\s`. With that, the function could be simplified considerably:

```js
return text.trim().split( /\s+/ );
```

mrdoob:
>  Perhaps consider naming this `splitByWhitespace` instead?

Agreed.

> Also note that regex has a dedicated character for whitespace, `\s`.

See #16218 for context.

bbmv:
I renamed the function and replaced /  +/ with /\s\s+/. I guess, that's more descriptive.
https://jsperf.com/loader-speed-test

mrdoob:
Seems like version 1 is faster here (ChromeOS, Pixelbook).

Screenshot 2019-04-16 at 04 34 19

bbmv:
Yes, but the version 1 doesn't check CRs and tabs. It may be important if anybody loves to edit xml files manually. The versions 2 and 3 have the same results on windows 10.
win10

bbmv:
Perhaps, will be better to return back to the version 2.

mrdoob:
Some more testing:

Screen Shot 2019-04-16 at 3 11 41 PM

Screen Shot 2019-04-16 at 3 12 51 PM

Seems like the proposed code will make Safari run slower.

bbmv:
well, then something like this - https://jsperf.com/loader-speed-test (version 2.1)
version_2 1

mrdoob:
Are you proposing checking `navigator.userAgent` every time we parse a data string?
This PR is getting a bit out of hand...

bbmv:
Probably, that's not a good idea. And I'm not sure that it should be inside the LoaderUtils module at all.

mrdoob:
These kind of improvements need to be throroughly tested in as many browser, os and device configurations. Otherwise we may be investing too much time in what could be a micro-optimisation that improves one configuration but make other configurations worse.

src/loaders/LoaderUtils.js
```diff
@@ -37,8 +38,27 @@ var LoaderUtils = {
...
+			replace( /\t/g, " " ).
+			replace( /  +/g, " " ). // double space
+			trim().
+			split( " " );
+
 	}
 
 };
 
+function isSafari() {
```
Mugen87: Please remove this function. We need a solution that is appropriate for all browsers.

mrdoob:
I'm going to close this until there is more data that shows this is beneficial in all modern browsers/os.

PR #16275, opened 2019-04-17
Added LightProbeHelper

WestLangley:
This should be helpful to someone.

There is a bit of repeated shader code, but we can clean this up when things clarify.

```js
var helper = new THREE.LightProbeHelper( lightProbe, size );
scene.add( helper );
```

### DIFF
--- a/src/helpers/LightProbeHelper.js
+++ b/src/helpers/LightProbeHelper.js
@@ -1,0 +1,152 @@
+import { Mesh } from '../objects/Mesh.js';
+import { ShaderMaterial } from '../materials/ShaderMaterial.js';
+import { SphereBufferGeometry } from '../geometries/SphereGeometry.js';
+
+/**
+ * @author WestLangley / http://github.com/WestLangley
+ */
+
+function LightProbeHelper( lightProbe, size ) {
+
+	this.lightProbe = lightProbe;
+
+	this.size = size;
+
+	var defines = {};
+	defines [ 'GAMMA_OUTPUT' ] = "";
+
+	// material
+	var material = new ShaderMaterial( {
+
+		defines: defines,
+
+		uniforms: {
+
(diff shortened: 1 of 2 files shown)


WestLangley:
The helper is the little guy.

Screen Shot 2019-04-17 at 1 40 50 PM

donmccurdy:
I like this, although the outcome of https://github.com/mrdoob/three.js/pull/16223 discussion may affect it. `LightProbeVolumeHelper`?

bhouston:
I think it is a light probe helper.  It visualizes a single light probe. :)  It is not a volume.  The volume helper would probably show the tetrahedral structure as lines with multiple little spheres.

donmccurdy:
It does that now, but if individual light probes were no longer Object3D instances, should they still have individual helpers?

But LightProbe exists today, and LightProbeVolume is still a topic of active discussion, so I'm fine with merging this. :)

mrdoob:
Thanks!

src/helpers/LightProbeHelper.js
```diff
@@ -0,0 +1,152 @@
...
+
+	this.lightProbe = lightProbe;
+
+	this.size = size;
+
+	var defines = {};
+	defines[ 'GAMMA_OUTPUT' ] = "";
+
+	// material
+	var material = new ShaderMaterial( {
```
Mugen87, line 19: In the meanwhile, it would be preferable to have `LightProbeHelper` without a dependency to `ShaderMaterial` so the helper is directly usable in `WebGPURenderer`.

Instead of directly evaluate the probe in a shader, how about converting the probe to an (irradiance) environment map and apply that to the sphere. The texture itself could be generated via JS as a`CanvasTexture`.

Since instances of `LightProbe` should present baked lighting, updating the texture does not have to be a fast operation (so generating it in JS is fine).

PR #16283, opened 2019-04-19
Added change event to PointerLockControls.

### LINKED ISSUE #16160: PointerLockControls needs better implementation

arodic:
Fixes #16160

### DIFF
--- a/examples/js/controls/PointerLockControls.js
+++ b/examples/js/controls/PointerLockControls.js
@@ -5,10 +5,18 @@
 
 THREE.PointerLockControls = function ( camera, domElement ) {
 
-	var scope = this;
-
 	this.domElement = domElement || document.body;
 	this.isLocked = false;
+
+	//
+	// internals
+	//
+
+	var scope = this;
+
+	var changeEvent = { type: 'change' };
+	var lockEvent = { type: 'lock' };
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

PR #16286, opened 2019-04-19
WebGLRenderer: Update TS file.

Mugen87:
see https://github.com/mrdoob/three.js/pull/16269#issuecomment-484535072


mrdoob:
Thanks!

PR #16264, opened 2019-04-16
Readonly Attribute properties for VAO.

takahirox:
Draft PR to show how the change looks like for https://github.com/mrdoob/three.js/pull/14600#issuecomment-478768913

Readonly
- BufferAttribute: .normalized, .itemSize, .array
- InterleavedBufferAttribute: .normalized, .itemSize, .offset, .data(InterleavedBuffer)
- InterleavedBuffer: .stride, .array
- InstancedBufferAttribute/InstancedInterleavedBuffer: .meshPerAttribute

Remove
- BufferAttribute: .setArray()
- InterleavedBuffer: .setArray()

Private
- BufferAttribute: .copy()
- InterleavedBuffer: .copy()
- InstancedBufferAttribute: .copy()
- InstancedInterleavedBuffer: .copy()

PR #16287, opened 2019-04-19
VAO Support without breaking API

takahirox:
I sent VAO PR #14600 a while ago. But it requires some API changes then was not easy to make decision.

So I made VAO PR without breaking any APIs. It's still faster than non-VAO. We can optimize with breaking APIs later if we need.

**Benefit**

Performance improvement.

`setupVertexAttributes()` in `renderer` sets up VBOs for a geometry. This is one of the CPU hot spot functions and is called every frame for each geometry.

With VAO we can skip `setupVertexAttributes()` except for initialization. I assign one VAO to one geometry. VAO saves VBOs settings so we just need to switch VAO associated with an geometry every frame other than the first frame instead of calling `setupVertexAttributes()` every frame.

**Changes**

- Add `WebGLBindingStates` which manages VAO and vertex attributes. It observes the addition/removal of an attribute of a geometry. It it detects the addition/removal, it updates VAO.
- Move the following functions to `WebGLBindingStates`. Now managing vertex attributes per VAO.
    - `initAttributes()`, `disableUnusedAttributes()`, `enableAttribute()`, and `enableAttributeAndDivisor()` of `WebGLBindingStates`.
    - `setupVertexAttributes()` of `WebGLRenderer`
- ~Add `.version2` (better name idea is welcome) property to `BufferAttribute`s and `InterleavedBuffer`s. It's counting up if their properties are changed. It's used as hint to detect if VAO update is necessary in `WebGLBindingStates`~
- Assume attribute properties are readonly https://github.com/mrdoob/three.js/pull/16287#pullrequestreview-425739920
- Move index buffer update into `WebGLBindingStates`
- Add/Remove morph attributes only when necessary in `WebGLMorphtargets`
- Dispose VAOs in `renderer.dispose()`
- Release VAO when geometry/program using it is released

**Performance Test**

With VAO
https://raw.githack.com/takahirox/three.js/NewVAOSupportTest/examples/webgl_vao_test.html

Without VAO (three.js in dev branch)
https://raw.githack.com/takahirox/three.js/NewVAOSupportTest/examples/webgl_non_vao_test.html

You can control the number of objects in a scene by adding `?num` to URL like `...html?2000`. In my Windows 10 both Chrome and FireFox become 10-15% faster overall.

| Browser | VAO | non-VAO |
| ---- | ---- | ---- |
| Chrome (num=2000) | 40fps | 36fps |
| FireFox (num=1000) | 40fps | 33fps |

**Further optimization**

VAO update necessity detection can be more efficiently if we accept the following API changes. I think we can do later if we need.

- Prohibit direct change of `geometry.attributes` #16232
- Readonly `BufferAttribute`s and `InterleavedBuffer`s properties #16264


src/renderers/webgl/WebGLBindingStates.js
```diff
@@ -0,0 +1,532 @@
...
+			if ( cachedAttribute.version !== geometryAttribute.version2 ) return true;
+
+			if ( cachedAttribute.data.buffer !== geometryAttribute.data ) return true;
+
+			if ( geometryAttribute.data &&
+				cachedAttribute.data.version !== geometryAttribute.data.version2 ) return true;
+
+		}
+
+		return false;
```
takahirox: Here is the part where checking if `geometry.attributes` has an update. If we accept the API changes I wrote in PR comment, here will be much more efficient just like checking `geometry.version`.

src/core/BufferAttribute.js
```diff
@@ -17,23 +17,79 @@ function BufferAttribute( array, itemSize, normalized ) {
...
+	this._itemSize = itemSize;
 	this.count = array !== undefined ? array.length / itemSize : 0;
-	this.normalized = normalized === true;
+	this._normalized = normalized === true;
 
 	this.dynamic = false;
 	this.updateRange = { offset: 0, count: - 1 };
 
 	this.version = 0;
+	this.version2 = 0;
```
Mugen87: Maybe `versionVAO`? This will indicate that the property is used in `VAO` context.

src/renderers/webgl/WebGLBindingStates.js
Mugen87: Doing all these checks is a bit unfortunate and feels like a hack. However, it's still better do it this way than having no VAO support at all. I hope we can make `BufferAttribute` a bit more strict in the future so parts of `needsUpdate()` won't be necessary anymore.

Mugen87:
I've downloaded your branch and tested with various examples. Nothing breaks and the performance of the mentioned demo is better even with simple scenes. At least on my iMac tested with latest Chrome and FF.

I think this performance benefit justifies the changes to `BufferAttribute`, `InstancedBufferAttribute`, `InstancedInterleavedBuffer`, `InterleavedBuffer` and `InterleavedBufferAttribute`.

src/core/BufferAttribute.js
takahirox: Renamed. Thanks

src/renderers/webgl/WebGLBindingStates.js
takahirox: Agreed with that. Added comment as a reminder.

takahirox:
@mrdoob Could you add this PR to milestone instead of #14600?

takahirox:
Let me now If you folks think this PR is a bit too huge and it's hard to review. I could try breaking up this PR into some PRs and we could review step by step.

takahirox:
Please let me know anything blocking this PR merged if exists. I'm willing to try to resolve.

takahirox:
@mrdoob @Mugen87 Friendly ping

Mugen87 (approved):

Usnul:
hey guys, are there any blockers to merging this? I would really love to see this in next release for the performance benefit it brings.

Samsy:
@mrdoob Let's have a look at this one !

toji:
Gentle ping on this. VAO use is a good performance improvement for WebGL even when the driver-side benefit is negligible simply because it allows most of the vertex state binding to happen off the main javascript thread and with far fewer jumps through the JS/Native bindings layer. It would be great to see Three making use of it!

takahirox:
I'll try to resolve the conflicts if you folks are still interested in

Mugen87:
After #17654, `BufferAttribute` can be considered as fixed size. According to my understanding, properties like `array`, `count` and `itemSize` should not be changed anymore after creating a buffer attribute. Because of this, it should be possible to simplify the PR at a few places.

takahirox:
My target in my mind is getting VAO in the beginning of r116dev (or r117dev) because it may be kind of big change so I hope we review and test much enough by the official release.

toji:
This would still be really nice to see. Is there a reason it stalled?

takahirox:
Sorry for delaying. I'll resolve the conflicts soon so far.

src/core/BufferAttribute.js
```diff
@@ -20,23 +20,79 @@ function BufferAttribute( array, itemSize, normalized ) {
...
-		if ( value === true ) this.version ++;
+		set: function ( value ) {
+
+			if ( value === true ) this.version ++;
+
+		}
+
+	},
+
+	array: {
```
Mugen87, line 48: @mrdoob If we consider properties like `array`, `itemSize`, `normalized` as read-only properties, this PR could be simplified at certain places. Is there a use case for changing `itemSize` or `normalized` of an existing buffer attribute after initial creation? Besides, the engine does not provide `setArray()` anymore so it assume buffer attributes to be fixed size. Meaning `array` could be declared as read-only, too.

src/renderers/webgl/WebGLBindingStates.js
```diff
@@ -0,0 +1,573 @@
+﻿/**
+ * @author Mugen87 / https://github.com/Mugen87
+ * @author Takahiro / https://github.com/takahirox
+ */
+
+function WebGLBindingStates( gl, extensions, attributes, capabilities ) {
+
+	var maxVertexAttributes = gl.getParameter( gl.MAX_VERTEX_ATTRIBS );
```
Mugen87, line 8: @takahirox Can you please remove all usage of `var` from the PR? The core uses now `let` and `const`.

takahirox: Yes, that's the plan.

src/core/BufferAttribute.js
takahirox: Similarly can we consider `meshPerAttribute` of `InstancedBufferAttribute` and `InstancedInterleavedBufferAttribute` as read-only? If we can, we can further simplify the code.

I don't think users want to change it on the fly (in the practical use). And `WebGLRenderer` seems to already consider it as immutable, changing `meshPerAttribute` doesn't have an effect to `geometry._maxInstanceCount` in the following code

https://github.com/mrdoob/three.js/blob/dev/src/renderers/WebGLRenderer.js#L938-L942

```javascript
if ( geometry._maxInstanceCount === undefined ) {

	geometry._maxInstanceCount = data.meshPerAttribute * data.count;

}
```

src/renderers/webgl/WebGLBindingStates.js
takahirox: Done

src/core/BufferAttribute.js
Mugen87: I think all properties of buffer attributes which values are set via ctor parameters should be read-only. I'm not aware of a compelling use case for changing them.

takahirox:
From the discussion in https://github.com/mrdoob/three.js/pull/16287#pullrequestreview-425739920 I simplified the code by assuming `array`, `itemSize`, `normalized`, and `meshPerAttribute` are read-only. Now no change are necessary in `src/core`.

Mugen87:
That looks way more manageable now 👍 .

Mugen87:
Even if the mentioned properties are not changed, we should still test the case where a buffer attribute is copied to another one. A copy operation actually sets a new array and other properties. Not sure there is an example that makes a copy during rendering.

takahirox:
In this PR, I don't explicitly make the properties read-only. Then changing the properties before calling the first `renderer.render()` call has an effect. But after the first call, changing them doesn't have an effect (unless attribute is added/removed to/from geometry).

So `.copy()` should work before the first `renderer.render()` call. And question. Do we really need to support the copy during the rendering? If we consider the properties as read-only (because they are not changed in practical use),  perhaps similarly we could consider that `.copy()` during the rendering should not be used.

Mugen87:
Okay, let's make it more concrete. In the following fiddle, the geometry is changed after 1 second by a copy operation. Does this code still work with this PR?

https://jsfiddle.net/2pndf0cb/1/

>  If we consider the properties as read-only (because they are not changed in practical use), perhaps similarly we could consider that .copy() during the rendering should not be used.

The copy actually creates a new attribute and index buffers which is the bit that makes it work now.

mrdoob:
Okay, lets give this a go!

mrdoob:
Many thanks @takahirox! 🙏

mrdoob:
Considering [how much support does `OES_vertex_array_object` have](http://webglstats.com/webgl/extension/OES_vertex_array_object), I think we may want to remove the non-VAO code.

takahirox:
Thanks for merging!

This is kinda big change in renderer. I confirmed the examples keep working but I'd be pleased if you folks would test well your applications (if you have), too.

Mugen87:
> Considering how much support does OES_vertex_array_object have, I think we may want to remove the non-VAO code.

Sounds good! If there are too many complains, it should be no deal to add the code back.

mrdoob:
@takahirox Any chance you can do another PR removing the non-VAO code?

arpu:
@takahirox  awesome!  i tested the dev118 on vrland.io local installation and all works just fine!

takahirox:
@mrdoob I'll try, may be next week

mrdoob:
Hmm, next week is too close to release.
Don't worry then. I'll give it a go.

mrdoob:
Seems like IE11 does not support `OES_vertex_array_object`.

@elalish If we removed the non-VAO code path, `<model-viewer>` would no longer be able to support IE11. What do you think...?

arpu:
i update @takahirox  TestSample  to latest three dev118 and compare with 117.1
https://jsfiddle.net/d30we7nc/16/


Bildschirmfoto von 2020-06-22 14-45-50
Bildschirmfoto von 2020-06-22 14-42-53

elalish:
@mrdoob Please don't break IE11 support; that would definitely block us from updating. We've done some polls and people still care about IE11 more than you'd think. If you do want to remove support for IE11, it might be nice to agree to a plan about it, since I imagine we'll be making that call eventually too.

mrdoob:
Okay! 👍

src/renderers/webgl/WebGLBindingStates.js
gchoqueux: A bug is appeared, in our project iTowns/itowns/pull/1444, after that this PR was merged and seem to come from that function `needsUpdate`. 

The error is 
> WebGL: INVALID_OPERATION: drawElements: no buffer is bound to enabled attribute

`needsUpdate` doesn't check if the **program attributes** are identical to the **cached attributes**.  The **program attributes** can change depending on the conditional statements then the bound buffers aren't correct and it returns this error.

**fix proposal** (I'm not sure it's all right)
```js
function needsUpdate( geometry, program ) {
	const cachedAttributes = currentState.attributes;
	const geometryAttributes = geometry.attributes;
	const programAttributes = program.getAttributes();

	const cachedAttributesCount = Object.keys( cachedAttributes ).length;

	if ( cachedAttributesCount !== Object.keys( geometryAttributes ).length || 
		cachedAttributesCount !== Object.keys( programAttributes ).length
	 ) return true;

	for ( const key in cachedAttributes ) {

		const cachedAttribute = cachedAttributes[ key ];
		const geometryAttribute = geometryAttributes[ key ];
		const programAttribute = programAttributes[ key ];

		if ( cachedAttribute.attribute !== geometryAttribute ) return true;

		if ( cachedAttribute.data !== geometryAttribute.data ) return true;

		if ( programAttribute == undefined ) return true;

	}
	return false;
}
```

takahirox: Thanks for the bug report!

> The program attributes can change depending on the conditional statements

I'm investigating and found that [program attributes are initialized when `program.getAttributes()` is first called and they are cached.](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLProgram.js#L810-L820) So wondering how they can change. Would you tell me the condition where they can change if you already figured out? Or can you share the code snippet which can reproduce the error?

gchoqueux: If the material preprocessor instructions change (`#define`, `#if`, etc.) then the `WebGLProgram` instance won't be the same so the program attributes and cache attributes may be different.

takahirox: My understanding is, if shader code or preprocessor instruction changes then a new `rogram` is created or switching to another existing `program`

https://github.com/mrdoob/three.js/blob/r119/src/renderers/WebGLRenderer.js#L1311-L1353
https://github.com/mrdoob/three.js/blob/r119/src/renderers/webgl/WebGLPrograms.js#L345-L374

If a new `program` is created, a new binding state is created and `needsUpdate` returns `true` because new `cachedAttributes` of the new binding state is an empty object.

https://github.com/mrdoob/three.js/blob/r119/src/renderers/webgl/WebGLBindingStates.js#L100-L133
https://github.com/mrdoob/three.js/blob/r119/src/renderers/webgl/WebGLBindingStates.js#L166-L186

If switching to another existing `program`, an existing binding state corresponding to the `program` would be selected by `getBindingState()` and its cache should be valid to the `program`.

So I'm wondering how to reproduce the error. Could you share the code snippet which can reproduce this error?

gchoqueux: I found this webGL error origin :
**WebGL: INVALID_OPERATION: drawElements: no buffer is bound to enabled attribute**

It's **`geometry.dispose()`** method calling,  whose attributes are **shared** with other geometries. 

This bug occurred with **`VAOs`** in **THREE v118**. 
Even if it didn't appear in the previous versions, but it's really a bug? 

In your library, we fixed this error.  Your garbage collector avoids to dispose shared attributes.

I have created 2 JsFidlle examples that simulate asynchronous disposes by your garbage collector:

* [JsFiddle example three v117.1 without webgl error](https://jsfiddle.net/iTownsIGN/9ax61ebL/9/) 
* [JsFiddle example three v118.0 with webgl error](https://jsfiddle.net/iTownsIGN/borcfx0p/3/)

The error is still present in THREE 121.

takahirox: Thanks for the report. I confirmed the error happens.

I think this is a bug in `WebGLGeometries` (or `WebGLAttributes`), rather than in `VAO(WebGLBindingStates)`, and the bug have existed even since before we adopted `VAO`.

As you speculated, the root issue seems that `geometry.dispose()` deletes webgl buffer of shared attribute even if it's still used by other geometry.

https://github.com/mrdoob/three.js/blob/r121/src/renderers/webgl/WebGLAttributes.js#L107-L121
https://github.com/mrdoob/three.js/blob/r121/src/renderers/webgl/WebGLGeometries.js#L15-L25

You seem to have already fixed this problem in your side. So would you like to make PR?

gchoqueux: I avoid the problem by deleting the shared attribute properties before calling `geometry.dispose()`,
because I know that the index and position attributes are shared.

```js
geometry.index = null;
delete geometry.attributes.position;
geometry.dispose();
```
But I avoid the problem, I don't fix it.

How would you solve the problem? 

You would need to know the number of times the attribute is shared and delete it when it's 0.
This to be a property of the attribute?

takahirox: Yes, counting the reference is the solution I thought of.

PR #16292, opened 2019-04-20
Object3D: use updateWorldMatrix() instead updateMatrixWorld()

WestLangley:
`updateMatrixWorld()` does not do what its name says it does. Certainly not correctly.

Maybe we can someday replace its use with `updateWorldMatrix()`.

For now, this is a step in that direction.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -475,7 +475,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 		}
 
-		this.updateMatrixWorld( true );
+		this.updateWorldMatrix( true, false );
 
 		return target.setFromMatrixPosition( this.matrixWorld );
 
@@ -495,7 +495,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 			}
 
-			this.updateMatrixWorld( true );
+			this.updateWorldMatrix( true, false );
 
 			this.matrixWorld.decompose( position, target, scale );
 
@@ -519,7 +519,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 			}
 
-			this.updateMatrixWorld( true );
+			this.updateWorldMatrix( true, false );
 
 			this.matrixWorld.decompose( position, quaternion, target );
 
@@ -538,7 +538,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 		}
 
-		this.updateMatrixWorld( true );
+		this.updateWorldMatrix( true, false );
 
 		var e = this.matrixWorld.elements;
 


titansoftime:
With updateMatrixWorld a user must set `scene.matrixAutoUpdate = false` in order to not update the world matrices of the children of objects that are set to `matrixAutoUpdate = false;` which is crazy.

If this new method avoids bizarre hidden performance hits such as that, I am 100% on board.

gkjohnson:
I am massively in favor of this function and have had to build it a few times for different applications. `updateMatrixWorld` is unfortunately lacking in that it doesn't guarantee the parents are up to date and that the child node is correctly updated in world space.

Some of the models in my applications have several thousand nodes and are in both broad and deep transform hierarchies so we turn off `matrixAutoUpdate` and `scene.autoUpdate` for the sake of performance. The model transforms animate so as they move their local matrices are updated. To keep performance up the world matrices are selectively updated as needed (before rendering a specific piece of geometry or running a custom raycast traversal). For this case we _never_ update the whole hierarchy of transforms at once because it's just too expensive.

@WestLangley Our implementation looked a little different, though. Here are a few thoughts on the function itself:

- The `updateWorldMatrix` function should early out if `matrixWorldNeedsUpdate === false`. If it doesn't early out here then if you update two or more sibling objects you'll wind up unnecessarily updating the whole chain of parents multiple times which can get expensive. If this optimization wasn't available then I think it limits the utility pretty heavily from a performance perspective.
- There should be a `setWorldMatrixNeedsUpdate()` or similar function that traverses all the children and marks their world matrices as needing an update. Of course you can afford an early out here, too, by stopping the traversal if a childs `matrixWorldNeedsUpdate` field is `true`. Maybe `updateMatrix()` could call this, as well.

Would it make sense to roll the functionality into the existing `updateMatrixWorld` function in a backwards compatible way so it's at least more exposed? The signature could change to `updateMatrixWorld(force = false, updateChildren = true, updateParents = false)`. With the above-mentioned optimizations it's not necessarily my preferred set of defaults but it's a start.

@titansoftime This isn't a magic bullet for solving that problem. The issue there is that THREE has no way of knowing whether or not a local transform has changed so by default the renderer just updates everything just to be safe.

Let me know if you wan to know anything more about my implementation or use cases!

arodic:
If we're changing matrix update function names, may I suggest we use more intuitive naming scheme?

I'll [quote Inigo Quilez](https://twitter.com/iquilezles/status/1107522177341902849):

> I'd expect others to find useful too: I've been for a while naming all my transforms such that they actually encode their purpose - ObjectToWorld, WorldToObject, EyeToHead, NodeToParent, RootToNode, ObjectToNDC. Because "WorldMatrix" alone doesn't tell me or my clients anything.

WestLangley:
@gkjohnson Feel free to recommend a specific change to `updateWorldMatrix()` in a separate post if you feel the method can be improved. I wouldn't worry too much about the performance of edge cases, however.

WestLangley:
I would like to see PRs merged more quickly.

If there is a reason why a PR is not merged, then the OP needs to be provided feedback so deficiencies can be corrected. If the PR is deemed unacceptable, then it should be closed.

PR #16294, opened 2019-04-20
use optimal image loader if available for textures

aardgoose:
Use ImageBitmap loader in preference if available in TextureLoader

Reduces frame time from 500ms+ to around 300ms for texture loading for the "Battle Scared Helmet" GLTF model.

### DIFF
--- a/src/loaders/TextureLoader.js
+++ b/src/loaders/TextureLoader.js
@@ -4,6 +4,7 @@
 
 import { RGBAFormat, RGBFormat } from '../constants.js';
 import { ImageLoader } from './ImageLoader.js';
+import { ImageBitmapLoader } from './ImageBitmapLoader.js';
 import { Texture } from '../textures/Texture.js';
 import { DefaultLoadingManager } from './LoadingManager.js';
 
@@ -22,7 +23,8 @@ Object.assign( TextureLoader.prototype, {
 
 		var texture = new Texture();
 
-		var loader = new ImageLoader( this.manager );
+		var loader = ( typeof createImageBitmap === undefined ) ? new ImageLoader( this.manager ) : new ImageBitmapLoader( this.manager );
+
 		loader.setCrossOrigin( this.crossOrigin );
 		loader.setPath( this.path );
 


donmccurdy:
Note that this can be done without modification to the file. From https://github.com/mrdoob/three.js/pull/12456 –


```js
class CustomTextureLoader {

	constructor () {

		this.loader = new THREE.ImageBitmapLoader();

	}

	load ( url, onLoad, onProgress, onError ) {

		this.loader.load( url, ( imageBitmap ) => {

			onLoad( new THREE.CanvasTexture( imageBitmap ) );

		}, onProgress, onError );

	}

}

```

Unfortunately the existence of `createImageBitmap` does not mean it is optimal... in Firefox, the implementation is incomplete and not really helpful for anything in my experience. 😥

/cc @takahirox unless https://github.com/mrdoob/three.js/pull/15910 solved that? It would be helpful to see a working demo using ImageBitmap + GLTFLoader in Firefox to verify this.

aardgoose:
The test case worked in FF ok, but other things break, and Y flipping seems inconsistent.
I'll close for now.

aardgoose:
@donmccurdy I have been looking at the source for Firefox and can see the following problems with createImageBitmap().

1. gl.UNPACK_FLIP_Y_WEBGL is acted on for ImageBitmaps, this per spec should be ignored.
I have identified the code path where this is implemented, fairly easy to fix)

2. The options {} parameter is not accepted, specifically imageOrientation: 'none' | 'flipY'.

I have made mods so this dictionary is passed into the code as required, but not yet implemented any of the options, apart from prototyping image flipping.

Are there any other issues you have identified?

fernandojsg:
@takahirox do you have any feedback on this? as I know you have been looking at the same issue with ImageBitmap on Firefox too 

> @donmccurdy I have been looking at the source for Firefox and can see the following problems with createImageBitmap().
> 
> 1. gl.UNPACK_FLIP_Y_WEBGL is acted on for ImageBitmaps, this per spec should be ignored.
>    I have identified the code path where this is implemented, fairly easy to fix)
> 2. The options {} parameter is not accepted, specifically imageOrientation: 'none' | 'flipY'.
> 
> I have made mods so this dictionary is passed into the code as required, but not yet implemented any of the options, apart from prototyping image flipping.
> 
> Are there any other issues you have identified?

donmccurdy:
@aardgoose I don't know of any other issues. The two you mentioned had seemed, in the past, to prevent practical use with e.g. GLTFLoader.

aardgoose:
Update.

Patches to implement imageOrientation and ignore UNPACK_FLIP_Y_WEBGL submitted and awaiting review.  

Also had basic tests for imageOrientation added to the web platform test suite. 

Is the ability to control behaviour of alpha premultiply etc useful?

takahirox:
Personally I prefer more aggressively using `ImageBitmap` for less blocking texture upload. But one of the concerns is 

https://threejs.org/docs/#api/en/loaders/ImageBitmapLoader

> Note that Texture.flipY and Texture.premultiplyAlpha with ImageBitmap are ignored. ImageBitmap needs these configuration on bitmap creation unlike regular images need them on uploading to GPU. You need to set the equivalent options via ImageBitmapLoader.setOptions instead.

So if we automatically switch from `ImageLoader` to `ImageBitmapLoader` in `TextureLoader`, texture with `.flipY = true` (default is `true`) will be upside down. We may need to add an way for `TextureLoader` users to control `ImageBitmapOptions` `TextureLoader` instance uses and also need to update `examples/js/loaders/*` using `TextureLoader` to set proper `ImageBitmapOptions`. Plus, we may need to update `Texture` spec, for example making `Texture.flipY` and `Texture.premultiplyAlpha` immutable.

PR #16306, opened 2019-04-22
Docs: Added note about importable examples.

### LINKED ISSUE #16220: JSM: Documentation for importable examples

Mugen87:
Fixed #16220

The PR updates `Import-via-modules.html` by adding useful notes about JSM. It also lists all examples files which are currently available as modules.

### DIFF
--- a/docs/manual/en/introduction/Import-via-modules.html
+++ b/docs/manual/en/introduction/Import-via-modules.html
@@ -62,11 +62,64 @@
 		...
 		</code>
 
-		<h2>Caveats</h2>
-
+		<h2>Importable Examples</h2>
 		<p>
-			Currently it's not possible to import the files within the "examples/js" directory in this way.
-			This is due to some of the files relying on global namespace pollution of THREE. For more information see <a href="https://github.com/mrdoob/three.js/issues/9562" target="_blank">Transform `examples/js` to support modules #9562</a>.
+			The core of three.js is focused on the most important components of a 3D engine. Many other components like loaders or controls are part of the
+			examples directory. three.js ensures to keep these files in sync with the core but users have to import them separately if they are required
+			for their project. However, most of these files are no modules which makes their usage in certain cases inconvenient. In order to address this issue,
+			three.js is going to provide more and more files as modules in the [link:https://github.com/mrdoob/three.js/tree/master/examples/jsm examples/jsm] directory.
+			If you install three.js via npm, you can import them like so:
+		</p>
+		<code>
+		import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
+		</code>
+		<p>
+			The following examples files are already available as modules:
+			<ul>
+				<li>controls
+					<ul>
+						<li>MapControls</li>
+						<li>OrbitControls</li>
+						<li>TrackballControls</li>
+					</ul>
+				</li>
+				<li>exporters
+					<ul>
+						<li>ColladaExporter</li>
+						<li>GLTFExporter</li>
+						<li>MMDExporter</li>
+						<li>OBJExporter</li>
+						<li>PLYExporter</li>
+						<li>STLExporter</li>
+						<li>TypedGeometryExporter</li>
+					</ul>
+				</li>
+				<li>loaders
+					<ul>
+						<li>GLTFLoader</li>
+						<li>MTLLoader</li>
+						<li>OBJLoader</li>
+					</ul>
+				</li>
+				<li>pmrem
+					<ul>
+						<li>PMREMCubeUVPacker</li>
+						<li>PMREMGenerator</li>
+					</ul>
+				</li>
+				<li>utils
+					<ul>
+						<li>BufferGeometryUtils</li>
+						<li>GeometryUtils</li>
+						<li>MathUtils</li>
+						<li>SceneUtils</li>
+						<li>ShadowMapViewer</li>
+						<li>SkeletonUtils</li>
+						<li>TypedArrayUtils</li>
+						<li>UVsDebug</li>
+					</ul>
+				</li>
+			</ul>
 		</p>
 	</body>
 </html>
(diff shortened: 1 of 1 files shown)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev28/docs/index.html#manual/en/introduction/Import-via-modules

docs/manual/en/introduction/Import-via-modules.html
```diff
@@ -62,11 +62,64 @@ <h2>Importing the module</h2>
...
 
-		<h2>Caveats</h2>
-
+		<h2>Importable Examples</h2>
 		<p>
-			Currently it's not possible to import the files within the "examples/js" directory in this way.
-			This is due to some of the files relying on global namespace pollution of THREE. For more information see <a href="https://github.com/mrdoob/three.js/issues/9562" target="_blank">Transform `examples/js` to support modules #9562</a>.
+			The core of three.js is focused on the most important components of a 3D engine. Many other components like loaders or controls are part of the
+			examples directory. three.js ensures to keep these files in sync with the core but users have to import them separately if they are required
+			for their project. However, most of these files are no modules which makes their usage in certain cases inconvenient. In order to address this issue,
```
gkjohnson: > no modules

To "not modules"?

docs/manual/en/introduction/Import-via-modules.html
```diff
@@ -62,11 +62,64 @@ <h2>Importing the module</h2>
...
+			three.js is going to provide more and more files as modules in the [link:https://github.com/mrdoob/three.js/tree/master/examples/jsm examples/jsm] directory.
+			If you install three.js via npm, you can import them like so:
+		</p>
+		<code>
+		import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
```
gkjohnson: I know it's mentioned in the previous section so maybe it's not worth noting again but this type of "bare module" syntax will only work with build processes like webpack or parcel.

docs/manual/en/introduction/Import-via-modules.html
```diff
@@ -62,11 +62,64 @@ <h2>Importing the module</h2>
...
 		</code>
 
-		<h2>Caveats</h2>
-
+		<h2>Importable Examples</h2>
 		<p>
-			Currently it's not possible to import the files within the "examples/js" directory in this way.
-			This is due to some of the files relying on global namespace pollution of THREE. For more information see <a href="https://github.com/mrdoob/three.js/issues/9562" target="_blank">Transform `examples/js` to support modules #9562</a>.
+			The core of three.js is focused on the most important components of a 3D engine. Many other components like loaders or controls are part of the
+			examples directory. three.js ensures to keep these files in sync with the core but users have to import them separately if they are required
```
gkjohnson: > directory. three.js ensures

I'm not sure if there's a precedent set elsewhere in the documentation but `three.js` not being capitalized at the beginning of a sentence here stuck out to me while I was reading it.

gkjohnson:
Awesome! I'm excited to see this the modularized examples start to make their way into the advertised set of features for package. I'm looking forward to more of the examples packages moved into modules.

docs/manual/en/introduction/Import-via-modules.html
WestLangley: three.js is not capitalized. :-)

gkjohnson: Fair enough! I guess it just looked odd because it was at the beginning of a sentence.

Mugen87: TBH, I always write it `three.js`^^

docs/manual/en/introduction/Import-via-modules.html
Mugen87: Um I think this should be already clear from the existing document. If we notice this section is unclear for users, we can still add an additional note.

looeee (approved):

docs/manual/en/introduction/Import-via-modules.html
```diff
@@ -62,11 +62,64 @@ <h2>Importing the module</h2>
...
+			for their project. However, most of these files are not modules which makes their usage in certain cases inconvenient. In order to address this issue,
+			three.js is going to provide more and more files as modules in the [link:https://github.com/mrdoob/three.js/tree/master/examples/jsm examples/jsm] directory.
```
looeee: > three.js is going to provide more and more files as modules

It's kind of unusual to say that "three.js" is doing this - rather say "we are working to provide all the examples as modules..."

docs/manual/en/introduction/Import-via-modules.html
looeee: > three.js ensures to keep these files in sync with the core 

This a little incorrect. 

> three.js ensures that these files are kept in sync with the core

looeee: > but users have to import them separately if they are required

Could we also add a note here saying that users need to make sure they are using the examples that match the three.js version they are using? This might seem obvious, but it seems like this is a common problem.

looeee: Sorry, guess I hit the "approve" button to early 😅  I can't see any way to undo that.

Mugen87: > Could we also add a note here saying that users need to make sure they are using the examples that match the three.js version they are using?

I've tried to address your suggestion with a new section. I hope this makes things more clear.

Mugen87: >  I can't see any way to undo that.

It's actually possible to dismiss a review^^.

image

looeee: Oh, not sure how I missed that! Thanks 😁

looeee (approved):

mrdoob:
Thanks!

PR #16309, opened 2019-04-22
Examples: Replace "DRACOExporter" with "DracoExporter"

moraxy:
https://github.com/mrdoob/three.js/blob/0828c0b4671443dc55dce05c4b1cc7d0ed4a8a3f/examples/misc_exporter_draco.html#L42

Error 404 for `js/exporters/DRACOExporter.js`: The file is named `DracoExporter.js`, only the class is `DRACOExporter`.

### DIFF
--- a/examples/misc_exporter_draco.html
+++ b/examples/misc_exporter_draco.html
@@ -39,7 +39,7 @@
 		<script src="js/libs/draco/draco_encoder.js"></script>
 
 		<script src="js/controls/OrbitControls.js"></script>
-		<script src="js/exporters/DRACOExporter.js"></script>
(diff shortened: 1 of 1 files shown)


donmccurdy:
The names of DracoExporter and DRACOLoader are inconsistent – i think one of the two should be renamed to match, probably DracoExporter -> DRACOExporter.

(related: https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git)

Mugen87:
The JSM version is already named as `DRACOExporter`. So when converting the example, the HTTP 404 was automatically fixed. Renaming `DracoExporter` -> `DRACOExporter` right here: #16872

PR #16310, opened 2019-04-22
Update WebGLState.d.ts

sttz:
The typing for WebGLState includes removed methods, misses new methods/parameters and has some wrong or unclear parameter types.

This PR:
- Removes old methods that have moved to the `buffers` property
- Remove missing `init` and `getScissorTest` methods
- Adds new `useProgram`, `setMaterial` and `texImage3D` methods
- Adds new parameters, makes parameter types more specific (avoid `any`, use existing three types)


mrdoob:
Thanks!

PR #16316, opened 2019-04-23
Multiview support using OVR_multiview2

fernandojsg:
## Intro
This PR introduces multiview support using the current `OVR_multiview2` API (spec https://www.khronos.org/registry/webgl/extensions/OVR_multiview2) as all the previous ones (`WebVR multiview`, `WEBGL_multiview`, `OVR_multiview`) are deprecated and should not be implemented even if they are still supported by some browsers.

The first implementations of multiview were attached to the WebVR API but this is not the case anymore and we could use this API to paint several views using a drawcall on the screen without need to have a VR device, eg for different views for a CAD editor. 
With that in mind I tried to detach the implementation from VR details and rely on `ArrayCamera` instead to determine when to render multiple views.

The key point for me here is that the expected benefit for this feature when using 2 views (as currently in VR) is around **40%** without any modification on the user's codebase.

Capture

## Test
Currently this extension is just implement on [Chrome Canary](https://www.google.com/chrome/canary/) and must be run with the following command line arguments:

```
--use-cmd-decoder=passthrough --enable-webgl-draft-extensions
```

## Implementation
- Added `multiview: boolean` attribute to the `WebGLRenderer` to detect if you want to enable it if available. It's used by default if it's available and `multiview !== false`. But probably we could just enable it by default if `vr.enabled = true`
- Introduced `WebGLMultiview.js` that should take care of all the binding/unbinding of the textures and framebuffers
- Introduced `WebGLMultiviewRenderTarget` that will include the  `numViews` that render target has.
- Modify preffix on the shaders to inject an array of matrices that will be used based on the `gl_ViewID_OVR` attribute.
- On `setProgram` switch from setting single matrices to array of matrices when using multiview and `ArrayCamera`

## TO-DO / Open questions
This is still a WIP implementation, there are some issues and decisions that need to be addressed before merging it:

- [x] Enable by default just when extension is available and `vr.enabled = true` ?
- [x] Currently the code is open enough to support more than 2 views, but when creating the shader we should know the number of views that will be used to define the size of the matrices arrays. Currently I have set 2 by default, but not sure if it could be nice to try setting it to `MAX_VIEWS_OVR` even if we use less views in our app.
- [x] Split the `renderObjects` function into two, to call multiview vs single. >> *It's been simplified enough like to keep both together on the same function*
- [x] Remove the code to bind fbo, set viewport and clear it from the multiview block on `renderObjects` and inject it on `setRenderTarget` instead.
- [x] Create a temporary structure to hold the array of matrices instead of allocating them on every call to `setProgram` when using multiview.
- [x] Currently just `viewMatrix` and `projectionMatrix` are injected as an array, I'll include the rest of matrices affected by them per view.
- [x] Support for shadows, currently broken
- [x] Move creation of multiview fbo to `WebGLTextures::setupRenderTarget()` ?


takahirox:
Please add ShadowMap support to the TODO list. Currently this implementation doesn't work with Shadow.

fernandojsg:
> Please add ShadowMap support to the TODO list. Currently this implementation doesn't work with Shadow.

Added

mrdoob:
Looking good!

fernandojsg:
Regarding the **shadows** issue: `ShadowMap` will use a custom render target and a `MeshDepthMaterial`. If multiview is enabled all the materials will get injected the multiview specific directives and uniforms in the shaders: https://github.com/mrdoob/three.js/pull/16316/files#diff-58e8247d21441c70f4f66e068eac5bfaR578 but as the shadowmap won't use the multiview fbo it will throw a warning because of the incompatible number of views on the shader vs the render target used.

To fix (https://github.com/mrdoob/three.js/pull/16316/commits/1cd69c9df880f01019be42b6fb091f30795f6b5b) this I've introduced a new material attribute `supportsMultiview` (open to suggestions) that will be set to `true` by default for all the materials and currently just `false` for the `MeshDepthMaterial` used on the `ShadowMap` class.

What do you think @takahirox @mrdoob ?

PS: I have added temporarily a copy of the `dragging` demo using multiview too to check the shadows working.

fernandojsg:
[I added support](https://github.com/mrdoob/three.js/pull/16316/commits/08ded2c0b55919b02d691dd0fa3b8806123bc5f3) for `modelViewMatrix` and `normalMatrix`. I created a method `WebGLMultiview::computeObjectMatrices()` that will create the array of matrices to avoid allocating them on every call, and update them as needed.

fernandojsg:
I removed the array allocation and `map` calls for camera's `viewMatrices` and `projectionMatrices` by introducing a similar method to the object's `WebGLMultiview::computeObjectMatrices()` -> `WebGLMultiview::computeCameraMatrices`

src/renderers/webgl/WebGLMultiview.js
```diff
@@ -0,0 +1,244 @@
...
+			} else {
+
+				for ( var i = 0; i < numViews; i ++ ) {
+
+					camera.projectionMatrices[ i ].copy( camera.projectionMatrix );
+					camera.viewMatrices[ i ].copy( camera.matrixWorldInverse );
+
+				}
+
+			}
```
takahirox: If I'm right, here L64-L82 should be out of `if ( ! camera.projectionMatrices ) {...}`. Otherwise, `camera.projectionMatrices/viewMatrices` won't be updated.

fernandojsg: @takahirox yeah! good catch, thanks! :) Updated!

fernandojsg:
Use `multiview.getNumViews()` everywhere so if we define a different (default = 2) number of views on the `WebGLMultiview` the number of matrices, shaders and so will get the correct value.

fernandojsg:
Moved the creation of the rendertarget out from `WebGLMultiview` into `WebGLMultiviewRenderTarget`, and changed the `update*Matrices` to use aux matrices instead of creating one on each object and camera.
Kudos to @takahirox for the help refactoring it

fernandojsg:
Currently the only item that I haven't addressed from the list is:
```
Enable by default just when extension is available and vr.enabled = true ?
```
As the `WebGLMultiview` object will be created on the `WebGLRenderer` constructor, so currently we will just do that when the attribute `multiview` is set true on the constructor's options.

I have noticed that there are a huge impact in performance when updating uniforms that consist on array of matrices, but I'll like to move the discussion to https://github.com/mrdoob/three.js/issues/16355.
In the meantime I'll be working on a alternative solution by setting the multiview uniforms directly from `WebGLMultiview` without using any cache.

Another performance impact is that currently once you request multiview, you "should" use, as requesting it for rendering just 1 view doesn't make much sense, as you will still be rendering to a N-dimensional FBO (depending on the number of views) even if you just copy one of these layers to the final canvas. I created an issue to discuss it here https://github.com/mrdoob/three.js/issues/16356

Still I believe neither of these issues should be a blocking for this as the performance improvement that we will get specially on WebXR is going to be really good (Benchmarks soon)

Feedback? :)

fernandojsg:
Added a new example using multiple views + multiview using just array cameras, currently using `bounds` to define the viewports, although I've opened an issue regarding it https://github.com/mrdoob/three.js/issues/16361

Introducing some proposals by @takahirox adding `setNumViews()` on the `WebGLMultiview` and automatically detect the array camera size.

Currently the user doesn't need to do anything else than just request `multiview` on the renderer and use `WebVR` or render with a `ArrayCamera` and it should just work.

fernandojsg:
I have created a different branch to try to fix the performance issue introduced because of the cache of uniforms (Described in https://github.com/mrdoob/three.js/issues/16355)
https://github.com/fernandojsg/three.js/commit/d35d296ee2cf2b49f3f36515d1c6be5c0c9d020f

The performance difference between setting the uniforms "manually" and doing the flatten versus the standard version of calling `uniforms.setValue` results in the `webgl_multiple_views_multiview` example goes from `55ms` measuring `renderer.render()` to `22ms`...

examples/webvr_ballshooter_multiview.html
```diff
@@ -0,0 +1,307 @@
...
+			init();
+			animate();
+
+			function init() {
+
+				container = document.createElement( 'div' );
+				document.body.appendChild( container );
+
+        const urlParams = new URLSearchParams(window.location.search);
+        const multiview = urlParams.has('enableMultiview');
```
takahirox: Wrong indentations.

fernandojsg: Fixed

examples/webgl_multiple_views_multiview.html
```diff
@@ -0,0 +1,294 @@
...
+				var SIZE = 1 / AMOUNT;
+				var WIDTH = ( window.innerWidth / AMOUNT ) * window.devicePixelRatio;
+				var HEIGHT = ( window.innerHeight / AMOUNT ) * window.devicePixelRatio;
+
+				for ( var y = 0; y < AMOUNT; y ++ ) {
+
+					for ( var x = 0; x < AMOUNT; x ++ ) {
+
+						var subcamera = new THREE.PerspectiveCamera( 40 + (x + y) * 20, ASPECT_RATIO, 0.25, 100 );
+						subcamera.bounds = new THREE.Vector4( x / AMOUNT, y / AMOUNT, SIZE, SIZE );
```
mrdoob: https://github.com/mrdoob/three.js/issues/16361#issuecomment-488054088

takahirox: How about merging this PR as is so far (if looking good) and then considering to switch from `ArrayCamera` bounds to `ArrayCamera` viewport (if needed) in another PR? That'd be easier to discuss and review maybe.

fernandojsg: Addressed on https://github.com/mrdoob/three.js/pull/16367 once that will get merged I'll push the changed to fix this PR

fernandojsg: Updated it https://github.com/mrdoob/three.js/pull/16316/commits/3c58278708626269ef64c2db549e12ad4ca2a606#diff-c0e88b98497597a015ecf238e91ac3a0 /cc @mrdoob @takahirox

fernandojsg:
@mrdoob @takahirox friendly ping :)

takahirox:
Looking good to me now. (I reviewed only core code. I didn't example code tho.)

examples/webgl_multiple_views_multiview.html
```diff
@@ -0,0 +1,308 @@
...
+				options.innerHTML =`Number of balls:
+						${addNumLink(200)}
+						${addNumLink(500)}
+						${addNumLink(1000)}
+						${addNumLink(2500)}
+						${addNumLink(5000)}`;
+
+
+				var ASPECT_RATIO = window.innerWidth / window.innerHeight;
+				var SIZE = 1 / AMOUNT;
```
Mugen87: It seems this variable is unused.

examples/webvr_dragging_multiview.html
```diff
@@ -0,0 +1,287 @@
...
+					object.castShadow = true;
+					object.receiveShadow = true;
+
+					group.add( object );
+
+				}
+
+				//
+
+        var canvas = document.createElement( 'canvas' );
```
Mugen87: Minor formatting issue.

examples/webvr_ballshooter_multiview.html
```diff
@@ -0,0 +1,307 @@
...
+					object.userData.velocity.y -= 9.8 * delta;
+
+				}
+
+				var ts = performance.now();
+				renderer.render( scene, camera );
+				var trender = performance.now() - ts;
+
+				// Compute average of the last frames
+				for (var i = 0; i < avgCounter.length-1; ++i) avgCounter[i] = avgCounter[i+1];
```
Mugen87: Minor code style issues. There are missing some spaces in the for loop and the upcoming two lines.

src/renderers/webgl/WebGLMultiview.js
```diff
@@ -0,0 +1,259 @@
...
+
+		currentRenderTarget = renderer.getRenderTarget();
+		resizeRenderTarget( camera );
+		renderer.setRenderTarget( renderTarget );
+
+	}
+
+	function detachRenderTarget( camera ) {
+
+		if ( renderTarget !== renderer.getRenderTarget() ) return false;
```
takahirox: No need to return false here.

examples/webgl_multiple_views_multiview.html
fernandojsg: fixed

examples/webvr_dragging_multiview.html
fernandojsg: Fixed, although I don't expect this example to be merged but just to test the PR, and then we could add multiview enabled by default if available on the WebXR demos

examples/webvr_ballshooter_multiview.html
fernandojsg: Fixed, thanks!

src/renderers/webgl/WebGLMultiview.js
fernandojsg: Fixed thanks

examples/files.js
```diff
@@ -203,6 +203,7 @@ var files = {
 		"webgl_multiple_renderers",
 		"webgl_multiple_scenes_comparison",
 		"webgl_multiple_views",
+		"webvr_multiple_views_multiview",
```
mrdoob: It should be:

```js
"webgl_multiple_views_multiview.html",
```

mrdoob:
> Fixed, although I don't expect this example to be merged but just to test the PR, and then we could add multiview enabled by default if available on the WebXR demos

Yes, could this PR be simplified? As far as I understand this is something that could just be enabled by default when available and we do not need additional examples. Right?

fernandojsg:
I'll remove the specific dragging and ballshooter _multiview examples and leave just the multiple_views_multiview as I believe that's different from what we have already unless you feel we should remove it too

mrdoob:
That'd definitely help me see the changes better 🙏

fernandojsg:
@mrdoob done!
also TL;DR how to use the extension is basically request `multiview: true` when creating the renderer, if it's not available it will just not use it but it won't break anything. The only requirement is that the WebGL context should be WebGL2 as the `OVR_multiview2` extension is Webgl2 only

```javascript
var context = canvas.getContext( 'webgl2', { antialias: false } );
renderer = new THREE.WebGLRenderer( { context: context, canvas: canvas, antialias: true, multiview: true } );
```

mrdoob:
But, wouldn't it be better to use the extension by default when available instead of requiring the user to pass `multiview: true`? Is there a side effect when enabling the extension?

fernandojsg:
I agree, that would be the expected usage of the extension. But currently with multiview enabled we need to compile the shader with a specific number of views. It means that even if you don't enter VR and are just rendering to the screen you will be doing both passes, although just one of the two layers will get copied to the render target. Although the CPU won't get impacted that much as with multiview we will be doing the same number of drawcalls anyway independently of the number of views, the GPU overhead will be bigger.
One of the solutions will be to have something like shader group profiles or so, that will let you compile both version of the same shader, with `views=1` and `views=2` and use it depending on your state application. @takahirox has been working on a solution for that, but to avoid adding more complexity to this PR we decided to flag the requirement for multiview and move that feature to another PR once this will get reviewed.

examples/files.js
fernandojsg: fixed

takahirox:
Yes, currently this PR has GPU overhead in non-immersive mode. And I've locally confirmed that compiling multiple programs and switching program associated with num views can solve (#16356). And if this PR includes it the PR can be very big and hard to review. It may take further time to merge.

So I'd like to suggest we go with this PR as is, and later think removing `multiview: true` flag when we optimize.

mrdoob:
How much is the GPU overhead? Also, the overhead is only while in the headset, right?

takahirox:
2x GPU slower in non-immersive mode because it unnecessarily renders 2x rendering size (two views).

In immersive mode, no overhead. More precisely, copying two views framebuffers to screen at last can't be said overhead but it's small. I think the benefit from half drawcall number is bigger.

(I assumed "in the headset" meant "in immersive mode", correct?)

fernandojsg:
@takahirox can you elaborate the 2x GPU slower?
I have been testing the ballshooter example removing the physics to avoid impacting the draw calls performance, and painting 15.000 balls. The results are:
* No multiview enabled:
 * Non immersive mode: 48fps
 * Immersive mode: 33fps
* Multiview enabled:
 * Non immersive mode: 41fps
 * Immersive mode: 47fps

So on that specific example we have that for non-immersive mode using multiview rely on a worse performance by around -15%, while on immersive mode the benefits are >40% as we have been expecting in all our previous tests.

I believe we all agree that this is a feature that it's better to enable by default so users just get the benefits of it. Still we could just try to enable multiview if an array camera is used or if `vr.enabled = true`.
So if we do so and enable multiivew by default on these situations we will end up with the following:
- Most of the browsers won't have this feature in the short term or at least not enabled by default -> Users won't have any overhead or benefit on any mode.
- Some browsers could start enabling it by default (Firefox Reality) -> It will impact negatively to the users that are browsing immersive experiences without entering VR, but it won't affect the non-VR content made with three.

My bet is to enable it by default at least when `vr.enabled = true` and address the issues on non-immersive mode with multiview in other issues/PRs

takahirox:
> 2x GPU overhead

More precisely, rendering itself cost. Rendering double size so rendering cost can be double.

What I tried was just rendering a few complex (many polygon) objects to focus on rendering cost. A few objects, so that a few uniform update, a few drawcall, and a few CPU processing (for example matrices update). And I saw half fps number.

So, what I mentioned was just double rendering cost. The overall performance regression of course very depends on application. For example, CPU cost of painting 15,000 balls may be big. So not 50% drop but 15% drop on that example isn't strange to me.

mrdoob:
It's okay that it's 15% slower in non immersive mode. We can try to optimise that afterwards.

What I'm trying to avoid is making users having to learn more APIs. It would be great if this was something transparent that users didn't have to know about.

mrdoob:
Moving to next cycle as it looks like the implementation is still not solved plus we'll need time for testing.

takahirox:
Likely when will r105 release be?

mrdoob:
Wednesday

takahirox:
> It would be great if this was something transparent that users didn't have to know about.

Right now, `OVR_multiview2` works only with WebGL2 and doesn't work with antialias. And we request to users to create WebGL context in user side to use WebGL2.

```javascript
var canvas = document.createElement('canvas');
var context = canvas.getContext('webgl2', {antialias: boolean});
var renderer = new WebGLRenderer({canvas: canvas, context: context});
``` 

I think we can automatically enable multiview if WebGL context is WebGL2, antialias is false, and the browser supports `OVR_multiview2`. With this change, users don't need to learn any new Multiview API.

```javascript
// in WebGLRenderer
var enableMultiview = ('WebGL2RenderingContext' in window) &&
    (_gl instanceof WebGL2RenderingContext) &&
    (_gl.getContextAttributes().antialias === false) &&
    (_gl.getExtension('OVR_multiview2') !== null);
```

If multiview is turned off, no any performance/functional effect. Currently `OVR_multiview2` extension isn't enabled by default on browsers. So the case where `enableMultiview` will be `true` may be very rare. That means I can say users don't see non-immersive performance regression on existing application in general.

I know a lot of developers are interested in multiview performance. So I hope we can put this PR in r105 by removing `multiview: true` flag and adding automatic multiview enabling instead of wait for r106.

mrdoob:
I usually have the bad habit of merging things last minute, but the fact that [WebVR support was broken for a month](https://github.com/mrdoob/three.js/pull/16561) during this cycle makes me extremely hesitant.

mrdoob:
Also, this `antialias` caveat is something I was not aware of.
Is support going to change eventually or is it a limitation of the feature?

fernandojsg:
Regarding `antialias` the solution seems to be adding `OVR_multiview_multisampled_render_to_texture` extension to WebGL (https://github.com/immersive-web/webxr/issues/317#issuecomment-472243817) but it's not something that is going to happen on the short term, so we should just need to assume that this initial version of multiview won't support antialias.
So just to make it sure the `OVR_multiview2` is an official spec, but just doesn't have AA support we need the next extension, but I don't think we should hold on that, if you are really struggling with performance you probably have already AA disabled? or in any case you can balance between better performance vs noAA, (also is still aframe setting antialias:false by default or at least on mobile?).

So I believe we have two ways to unblock this PR:
- Keep the `multiview` attribute so people could use it knowing the limitations
- Remove the attribute and check automatically `WebGL2 && non-antialias && multiview extension`

I understand your concerns about last minute merging of a sensitive PR that is modifying `WebGLRenderer` so let's just move this one to the next release although please let us know what you expect so we can unblock it and merge asap after this release.

takahirox:
We're in r106 dev cycle now and I'd like to go forward. We can merge, and test it for the rest of the cycle.

fernandojsg:
Refactored and change the API so this functionality is hidden for the user and it will get activated automatically if `OVR_multiview2` is supported.
This extension is only available on `webgl2` contexts and currently just on context created with `antialias: false`, so it will be detected and enabled if the conditions are met (`webgl2 && !antialias`).
I added a new example that it's basically drawing many objects on the scene that just collide with the wall, to avoid adding a lot of CPU overhead so we can compare the benefits on the drawcalls impact having this extension enabled or not.
If the extension is enabled the background grid will be green otherwise it will be red.
I did some benchmarks:
- on the oculus quest with Firefox Reality:
  * multiview: **51fps**
  * no multiview: **33fps**
- on my laptop with a 1080 and chrome canary and with 20k objects:
  * multiview: **29fps**
  * no multiview: **18fps**

image

TL;DR if the extension can be use, it will be used by default. As soon as we will get `OVR_multiview_multisampled` extension support (https://www.khronos.org/registry/OpenGL/extensions/OVR/OVR_multiview_multisampled_render_to_texture.txt) I'll update the code and it will just work for both multisampled or not render targets.

examples/webgl_multiple_views_multiview.html
```diff
@@ -0,0 +1,325 @@
...
+
+				var info = document.createElement( 'div' );
+				info.style.position = 'absolute';
+				info.style.top = '10px';
+				info.style.width = '100%';
+				info.style.textAlign = 'center';
+				container.appendChild(info);
+
+				info.innerHTML = `<a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> webgl - multiple views using multiview extension - by <a href="https://twitter.com/fernandojsg">fernandojsg</a><br/>`;
+				info.innerHTML += 'enabled: ' + colorize( renderer.multiview.isEnabled() ) + ` <a href="?num=${ numObjects }&${ multiview ? '' : 'enableMultiview' }">toggle</a>`
```
modulesio: Should this be `isAvailable`? It throws for me.

modulesio:
In testing this I ran into:

- https://github.com/mrdoob/three.js/pull/16316#pullrequestreview-262103836
- [`WebXRManager` initialization crash](https://github.com/mrdoob/three.js/issues/17034)

However, with those changes it worked in Exokit desktop.

examples/webgl_multiple_views_multiview.html
fernandojsg: Yeah, although after speaking with @mrdoob I'm going to remove this example as using array camera for multiple viewports is not recommended, at least in this use case because of the frustum culling issues.

src/renderers/WebGLRenderer.js
```diff
@@ -84,7 +84,7 @@ function WebGLRenderer( parameters ) {
 		 * Enables error checking and reporting when shader programs are being compiled
 		 * @type {boolean}
 		 */
-		checkShaderErrors: true
+		checkShaderErrors: false
```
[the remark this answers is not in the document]
fernandojsg: @mrdoob ops, fixed

fernandojsg:
Updated the example to use modules and removed the previous one with multiple viewports. 
It should be good to go /cc @takahirox @modulesio @mrdoob

examples/webvr_multiview.html
```diff
@@ -0,0 +1,182 @@
...
+
+			function init() {
+
+				container = document.createElement( 'div' );
+				document.body.appendChild( container );
+
+				//
+
+				var canvas = document.createElement( 'canvas' );
+				var context = canvas.getContext( 'webgl2', { antialias: false } );
```
takahirox: I think adding comment here about why these two lines are necessary would be helpful to users.

fernandojsg: Yep, what about something like:
```js
// Currently OVR_multiview2 is supported just on WebGL2 non-multisampled contexts
// so we must create the context manually ensuring `antialias` is set to false
//
// There is an ongoing discussion on how to add multisampled multiview support
// on the Khronos Group's WebGL repo: https://github.com/KhronosGroup/WebGL/issues/2912
// as soon as that will get solved we will update the code so it will be
// transparent for the user and you could use this extension with or without multisampled
// contexts
var context = canvas.getContext( 'webgl2', { antialias: false } );
```

examples/webvr_multiview.html
```diff
@@ -0,0 +1,182 @@
...
+				container.appendChild( info );
+
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color( 0x505050 );
+
+				camera = new THREE.PerspectiveCamera( 70, window.innerWidth / window.innerHeight, 0.1, 10 );
+
+				room = new THREE.LineSegments(
+					new BoxLineGeometry( 6, 6, 6, 10, 10, 10 ),
+					new THREE.LineBasicMaterial( { color: renderer.multiview.isAvailable() ? 0x99ff99 : 0xff3333 } )
```
takahirox: I think message in info like "OVR_multiview2 extension doesn't seem to be supported on your browser" is more straightforward to users rather than grid color. What do you think?

takahirox:
Added some comments inline on example code but I think basically looking good.

examples/webvr_multiview.html
takahirox: Looking good!

examples/webvr_multiview.html
fernandojsg: I have added a text indicating if it's available or not:
image

But I'd like to keep the background color because if you auto enter VR on quest or go and you don't see HTML you won't be able to detect if it's available or not without exiting

fernandojsg:
Thanks @takahirox , fixed the issues

examples/webvr_multiview.html
fernandojsg: thanks

fernandojsg:
fwiw the OVR_multiview2 extension is going to be moved to community approved  (https://github.com/KhronosGroup/WebGL/pull/2916)
And there's an issue to discuss around the multisampled multiview support (https://github.com/KhronosGroup/WebGL/issues/2912). As soon as we will have a proposal we will work on supporting it and it should be just a matter of changing one or two lines.

examples/webvr_multiview.html
```diff
@@ -0,0 +1,195 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webvr - ball shooter</title>
```
takahirox: I think better to change the title. This example doesn't shoot ball.

Mugen87:
I have tested the PR on my local machine. Currently, no matter what example you execute the following warning is logged in the browser console.

> THREE.WebGLRenderer: OVR_multiview2 extension not supported.

fernandojsg:
@Mugen87 that's part of the `WebGLCapabilities` detection: https://github.com/mrdoob/three.js/pull/16316/files#diff-c286888722879bf7927784081e925eb4R89
as long as you will query for that, you will get that warn message, I believe we had similar warning with other extensions in the past.
The only way to get rid of this is to query for that extension directly on the `WebGLMultiview` instead of getting it through `WebGCapabilities`, or modifying the `extension.get(...)` to support a second parameter such as `optional` so it won't throw the warning if the extension is not supported (https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLExtensions.js#L44-L48)?

Mugen87:
Not sure how @mrdoob thinks about this by putting the logic into `WebGLMultiview` seems appropriate to me.

I'm sure we will get a lot of user complains if we don't fix this warning for non-VR code^^.

fernandojsg:
Yep, that could works. I just mentioned the `optional` parameter on `extension.get(..)` so we can keep `WebGLCapabilities` as the main place where we can check the extensions available instead of spreading that kind of code around other features in the future

fernandojsg:
fwiw the extension is already Community Approved so it means it could be already enabled by default on browsers, as we do on Firefox Reality, and soon on Firefox desktop: https://github.com/KhronosGroup/WebGL/pull/2916

src/renderers/WebGLRenderer.js
```diff
@@ -315,6 +315,12 @@ function WebGLRenderer( parameters ) {
 
 	this.vr = vr;
 
+	// Multiview
+
+	var multiview = new WebGLMultiview( _this, _gl, _gl.getContextAttributes() );
+
+	this.multiview = multiview;
```
mrdoob: I don't think we should expose this.

examples/webvr_multiview.html
```diff
@@ -0,0 +1,195 @@
...
+				renderer.vr.enabled = true;
+				container.appendChild( renderer.domElement );
+
+				var info = document.createElement( 'div' );
+				info.style.position = 'absolute';
+				info.style.top = '10px';
+				info.style.width = '100%';
+				info.style.textAlign = 'center';
+				info.innerHTML = '<a href="https://threejs.org" target="_blank" rel="noopener">three.js</a> webvr - multiview<br/>';
+				info.innerHTML += renderer.multiview.isAvailable() ? `<span style="color: #33ff33"><b>OVR_multiview2</b> is supported in your browser</span>` :
```
mrdoob: `renderer.capabilities.multiview` would be better here I think.

examples/webvr_multiview.html
```diff
@@ -0,0 +1,195 @@
...
+				container.appendChild( info );
+
+				scene = new THREE.Scene();
+				scene.background = new THREE.Color( 0x505050 );
+
+				camera = new THREE.PerspectiveCamera( 70, window.innerWidth / window.innerHeight, 0.1, 10 );
+
+				room = new THREE.LineSegments(
+					new BoxLineGeometry( 6, 6, 6, 10, 10, 10 ),
+					new THREE.LineBasicMaterial( { color: renderer.multiview.isAvailable() ? 0x99ff99 : 0xff3333 } )
```
mrdoob: `renderer.capabilities.multiview`

src/renderers/webgl/WebGLMultiview.js
```diff
@@ -0,0 +1,235 @@
...
+
+		return cameraArray;
+
+	}
+
+	//
+
+	function isAvailable() {
+
+		return capabilities.multiview && ! contextAttributes.antialias;
```
mrdoob: This seems like should go in `WebGLCapabilities`.

src/renderers/WebGLRenderer.js
fernandojsg: Yep, fixed. It was used to access the number of views on the WebGLProgram, but I just changed it to get the number of views from the rendertarget directly https://github.com/mrdoob/three.js/pull/16316/files?file-filters%5B%5D=.js#diff-58e8247d21441c70f4f66e068eac5bfaR334

examples/webvr_multiview.html
fernandojsg: yep good idea. done

examples/webvr_multiview.html
fernandojsg: done

src/renderers/webgl/WebGLMultiview.js
fernandojsg: done

fernandojsg:
@mrdoob thanks for the review, I fixed the issues you mentioned. The only one left is the warning when checking the `OVR_multiview2` extension. As I mentioned previously I see two ways to solve it, but that's your call:
1. Add an optional parameter to `extension.get(name, silent)` everything should works as expected if `silent` is not defined, but it won't throw any warning if `silent = true`. So in our case we could call `extensions.get(OVR_multiview2, true)`

2. In the `WebGLCapabilities` code that checks for multiview, instead of doing:
```javascript
var multiviewExt = extensions.get( 'OVR_multiview2' );
```
we will just do
```javascript
var multiviewExt = gl.getExtension( 'OVR_multiview2' );
```

I tend to prefer the first one as I haven't seen any other example using the later and it help us centralize everything on the `WebGLExtensions`, but up to you :)

src/renderers/WebGLRenderer.js
```diff
@@ -1371,7 +1387,13 @@ function WebGLRenderer( parameters ) {
 			var material = overrideMaterial === undefined ? renderItem.material : overrideMaterial;
 			var group = renderItem.group;
 
-			if ( camera.isArrayCamera ) {
+			if ( capabilities.multiview ) {
```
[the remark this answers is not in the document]
fernandojsg: Good point yep, I addressed that issue already https://github.com/mrdoob/three.js/pull/16316/commits/23550627a74a01ed82105dd64885331e7cb965bb

src/renderers/webgl/WebGLMultiview.js
```diff
@@ -0,0 +1,215 @@
...
+
+	if ( renderer.capabilities.multiview ) {
+
+		renderTarget = new WebGLMultiviewRenderTarget( 0, 0, DEFAULT_NUMVIEWS );
+
+		renderSize = new Vector2();
+		mat4 = [];
+		mat3 = [];
+		cameraArray = [];
+
```
mrdoob: Seems like this is the only place where `getMaxViews()` is being called?

```js
var maxViews = capabilities.maxMultiviewViews;

for ( var i = 0; i < maxViews; i ++ ) {
```

src/renderers/WebGLRenderer.js
```diff
@@ -1175,6 +1179,12 @@ function WebGLRenderer( parameters ) {
 
 		}
 
+		if ( capabilities.multiview ) {
+
+			multiview.attachRenderTarget( camera );
```
mrdoob: Should this method be called `multiview.attachCamera()` instead?

src/renderers/webgl/WebGLProgram.js
```diff
@@ -340,6 +341,9 @@ function WebGLProgram( renderer, extensions, code, material, shader, parameters,
 
 	var prefixVertex, prefixFragment;
 
+	var renderTarget = renderer.getRenderTarget();
+	var numMultiviewViews = renderTarget instanceof WebGLMultiviewRenderTarget ? renderTarget.numViews : 0;
```
mrdoob: We try not to use `instanceof`:

```js
var numMultiviewViews = renderTarget.isWebGLMultiviewRenderTarget ? renderTarget.numViews : 0;
```

src/renderers/webgl/WebGLMultiview.js
fernandojsg: Done!

src/renderers/WebGLRenderer.js
fernandojsg: Yep that sounds cool, changed

src/renderers/webgl/WebGLProgram.js
fernandojsg: Yep I forgot that one, I was using `isWebGL..` on the rest of the code already. Thanks

fernandojsg:
@mrdoob fixed issues and added suggestions, thanks!

src/materials/Material.js
```diff
@@ -71,7 +71,7 @@ function Material() {
 	this.visible = true;
 
 	this.toneMapped = true;
-
+	this.supportsMultiview = true;
```
mrdoob: Hmm?

fernandojsg: Ouch, nice catch 👼, fixed

src/renderers/webgl/WebGLMultiview.js
```diff
@@ -0,0 +1,211 @@
...
+		if ( ! isMultiviewCompatible( camera ) ) return;
+
+		currentRenderTarget = renderer.getRenderTarget();
+		resizeRenderTarget( camera );
+		renderer.setRenderTarget( renderTarget );
+
+	}
+
+	function detachRenderTarget( camera ) {
```
mrdoob: Should this be `detachCamera`?

src/materials/Material.js
```diff
@@ -71,7 +71,6 @@ function Material() {
 	this.visible = true;
 
 	this.toneMapped = true;
-
```
mrdoob: Leave `Material.js` alone 😁

src/renderers/webgl/WebGLMultiview.js
fernandojsg: It's been so long that I've been just updating the issues as you spot them X) I'll take a deep look again and review the whole pr to see if I see something weird and I'll ping you back, thanks :)

src/materials/Material.js
fernandojsg: That new line totally fits there! :P I'll remove 👼

fernandojsg:
@mrdoob Ok fixed all the minor issues I've found as linter warnings and the changes on the feedback you provided to me. I tested it already on desktop and quest.

mrdoob:
Thanks for your patience! 🙏

takahirox:
Yay, finally!

fernandojsg:
It was worth it! Is looking much better than at the beginning, thanks for the exhaustive review! ;)

arpu:
is it right this only works with webgl2 and  alpha: false ?
like
canvas.getContext( 'webgl2', { alpha: false } )

arpu:
ah i see in the example canvas.getContext( 'webgl2', { antialias: false } );

fernandojsg:
@arpu currently the multiview extension is just available on WebGL2 contexts because it needs glsl3 and also texture arrays. And currently the extension doesn't support multisampling so it should be `antialias: false`. The WebGL WG is working on a spec for a multisampled multiview extension that should come up soon ;)

carstenschwede:
@fernandojsg Congrats! I was looking forward to this merge. Could the OVR_multiview2 extension be used to increase performance for CubeCamera (i.e. "WebGLMultiViewRenderTargetCube")?

mrdoob:
I think `OVR_multiview2` is limited to 4 views max?

mrdoob:
> Yep I agree, it could be nice to get rid of these warnings. I'm happy to create a new PR with the preferred way by @mrdoob as discussed on #16316 (comment)

I'll look at this today.

fernandojsg:
> I think `OVR_multiview2` is limited to 4 views max?

Exactly, currently the maximum is 4, although is recommended to check for it using (As in `WebGLCapabilities`: https://github.com/mrdoob/three.js/pull/16316/files#diff-c286888722879bf7927784081e925eb4R91): 	
```var maxMultiviewViews = multiview ? gl.getParameter( multiviewExt.MAX_VIEWS_OVR ) : 0;``` 

> I'll look at this today.

Cool

fernandojsg:
> @fernandojsg Congrats! I was looking forward to this merge. Could the OVR_multiview2 extension be used to increase performance for CubeCamera (i.e. "WebGLMultiViewRenderTargetCube")?

I'm not familiar enough with how is implemented the cube camera, but because of the maximum number of views it won't be possible to render it in one drawcall but in 2

carstenschwede:
Don't want to hijack the PR, so just as as sidenote: I couldn't find the Chrome commit that implemented multiview, but the [Khronos spec](https://www.khronos.org/registry/OpenGL/extensions/OVR/OVR_multiview.txt) suggests

> **What is the minimum values permitted for MAX_VIEWS_OVR?**
>    **2. But at least 6 is recommended for consistency with Vulkan.**
    Six views is desireable as it can be used to render all faces of a cube map
    in a single pass.

Might be interesting for a future use case, once MAX_VIEWS_OVR has increased. Even a max. value of 4 views could reduce drawcalls from 6 to 2.

fernandojsg:
@carstenschwede the `MAX_VIEWS_OVR` in the browser should be populated with the `GL_MAX_VIEWS_OVR` value and so far I have seen just 4 as maximum on all the devices I have.
As you mention, depending on that max the logic of specific tasks could be divided into smaller number of drawcalls automatically, so currently from 6 to 2 and eventually if we get 6 as max, it will be done just in 1 drawcall.
Feel free to open another issue with this topic and we could keep the discussion there publicly as many people won't be reading this because is already closed

fernandojsg:
fwiw I wrote a short blog post around multiview support, focused on webxr though https://blog.mozvr.com/multiview-on-webxr/

RemusMar:
@fernandojsg 

Fernando,
On the latest Firefox 69 (Windows) I get
"THREE.WebGLRenderer: OVR_multiview2 extension not supported."
on various machines.

fernandojsg:
@RemusMar it will just be enabled by default on Firefox 71. Give it a try with Firefox Nightly and go to `about:config` and set `webgl.enable-draft-extensions` to  `true`

mrdoob:
@Mugen87 

>> Like mentioned [here](https://github.com/mrdoob/three.js/pull/16316#issuecomment-515413796), many examples log now the following warning:
>
> THREE.WebGLRenderer: OVR_multiview2 extension not supported.

Fixed #17575

mrdoob:
@fernandojsg 

> Yep I agree, it could be nice to get rid of these warnings. I'm happy to create a new PR with the preferred way by @mrdoob as discussed on https://github.com/mrdoob/three.js/pull/16316#issuecomment-523577544

Sorry. After studying the problem, the proposed solutions seemed a bit like "hiding the dust under the carpet". I think the refactored code is a bit more robust and easier to follow.

mrdoob:
Additional PRs: #17596 #17599

yorkie:
This has been removed from code at #18750.

From #20368:

> We used to support `OVR_multiview2` but I decided to remove the code because the lack of antialias support.

PR #16322, opened 2019-04-23
Update SVGLoader.js

OmarJay1:
These changes parse SVG transforms without basing them on spaces.  For instance translate(200, 200) would split improperly because it has a space between the comma and the 2.

The following example appears differently in Windows Edge than in other browsers without the changes.  Edge inserts white spaces after commas in transforms for some reason.
https://threejs.org/examples/webgl_loader_svg.html


yomboprime (approved):

mrdoob:
Thanks!

PR #16342, opened 2019-04-27
Add ts typing for CSS2DRenderer in example/jsm/renderer ( JS file not converted yet )

### LINKED ISSUE #16337: Export CSS2DRenderer in jsm
Currently it can:
```js
import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls';
```
But can't
```js
import { CSS2DRenderer, CSS2DObject } from 'three/examples/js/renderers/CSS2DRenderer';
```

Can you export `/renderer` to `/jsm/`?

linonetwo:
I just want to help `/jsm` to complete soon

This fixes https://github.com/mrdoob/three.js/issues/16337

Tested locally.


gkjohnson:
Awesome!

Hand converting examples to jsm isn't currently sustainable because it requires maintaining both the original and new files as updates are made. For the moment I think it would be preferred to modify the [modularize.js](https://github.com/mrdoob/three.js/blob/dev/utils/modularize.js#L10) script to automatically convert the CSS2DRenderer files, instead.

PR #15599 would enable a path forward for this type of manual conversion but it's unclear whether that will be merged.

/cc @donmccurdy

Mugen87:
Um, when using `modularize.js`, it seems the script is not able to export `CSS2DObject` and `CSS2DRenderer` in a single file. Looks like the script needs an enhancement...

linonetwo:
Oh, I'm new to threejs, didn't know there is an auto converter.

I'll remove the `.js` file, and only keep `.d.ts`, seems your converter can't generate typing.
And I'm going to put two class into a single `.d.ts` file, since you are going to export CSS2DObject and CSS2DRenderer in a single file.

gkjohnson:
@Mugen87 I may sound a bit like a broken record but I would be concerned that updating and maintaining that script will be error prone if we try to support every case needed to convert every file.

Mugen87:
Yeah, makes sense. I just thought if it's possible to convert a given file with `modularize.js` by implementing an appropriate enhancement, that should be the preferred way.

donmccurdy:
I agree we're going to need https://github.com/mrdoob/three.js/pull/15599 or something like it. I tried converting this file a while ago and hit the same problems.

We started with everything in the global namespace, and we want to end up in a state where everything is an ES module and the legacy version is generated from that. Trying to create a middle step where every ES module can be generated from the global version automatically is a very tough – and avoidable – obstacle to the end goal.

/cc @mrdoob see https://github.com/mrdoob/three.js/pull/15599#issuecomment-477215326 😇

Mugen87:
#16472 added `CSS2DRenderer` as a module and the respective TS file. 

BTW: The code base does not the keywords `public` and `private` so far. Besides, module internal functions like `getDistanceToSquared` or `zOrder` are usually not respected by the declaration file. Just the public API.

linonetwo:
Glad to see those modules added, closing this PR.

Issue #16307, opened 2019-04-22
OBJLoader2 _throwError not a function, onError not called


zacaj:
##### Description of the problem

OBJLoader2 is crashing on certain invalid files.  I get an error:

`Uncaught TypeError: this._throwError is not a function
    at THREE.OBJLoader2.Parser.processCompletedMesh (OBJLoader2.js:1209)
    at THREE.OBJLoader2.Parser.checkFaceType (OBJLoader2.js:1034)
    at THREE.OBJLoader2.Parser.processLine (OBJLoader2.js:961)
    at THREE.OBJLoader2.Parser.parse (OBJLoader2.js:778)
    at THREE.OBJLoader2.parse (OBJLoader2.js:378)
    at Object.fileLoaderOnLoad [as onLoad] (OBJLoader2.js:241)
    at XMLHttpRequest.<anonymous> (three.js:34525)`

Seems like Parser doesn't have this function, just OBJLoader2 itself?

The bigger issue though is that, when this happens, my onError callback I passed to OBJLoader2.load isn't called, and neither is onLoad.  It seems like this error just vanishes.  Since I'm wrapping my OBJLoader2.load call in a promise, this ends up causing problems since I can't resolve/reject the promise, and my application can't finish loading

##### Three.js version

latest

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
> OBJLoader2 is crashing on certain invalid files

Can you please share one of those files for debugging?

zacaj:
I can't sadly, they're client's private files, which is also why I can't really work around this issue by just passing in valid objs.  The issue should be reproducible by just putting an || true in the if statement around the line I noted?

kaisalmen:
@zacaj and @Mugen87 I will investigate...

kaisalmen:
Ok, problem is understood. Function is not defined at that point. I am backporting the solution from OBJLoader2 V3.0.0-Beta. Will Be fixed in referenced PR soon...

kaisalmen:
Problem is fixed. onError callback function from `OBJLoader2` is now properly passed on (with correct scope) if defined.

PR #16343, opened 2019-04-27
OBJLoader2 V2.5.1

kaisalmen:
This fixes the following issues:
#15219: Materials are initialised as objects.
#15468. OBJLoader2: Reduced log level from warn to info when defaultMaterial is used when material name was not resolvable.
#16307: Backported onError function usage from OBJLoader2/Parser V3. Unified callback naming

From now on versions are only printed once in the logs.


kaisalmen:
This is good to go now!

mrdoob:
Thanks!

PR #16368, opened 2019-05-02
Added SH3-based AmbientLight

WestLangley:
A subtle change is `AmbientLight` no longer extends `Light`, but `Object3D` directly. It no longer has a `color` property; the color is encoded in the SH3 coefficients.

Adding this enhancement, while striving to maintaining backwards compatibility, has been a challenge.

This change will have an effect on the loaders.


looeee:
We can still set the color in the constructor, right? So the behaviour of this code should be the same:

``` js
const ambientLight = new AmbientLight( 0x191919, 1.0 );
scene.add( ambientLight );
```

If that's the case then the `FBXLoader` at least should be unaffected.

WestLangley:
@looeee I sure hope you are right. :-)

looeee:
BTW I'm getting an error when trying to [run your code](https://raw.githack.com/WestLangley/three.js/dev-ambient_sh3/examples/webgl_lightprobe.html): 

`Uncaught TypeError: Cannot read property 'coefficients' of undefined` 

at line 20 in `SHUtils`:

`var shCoefficients = sh.coefficients;`

WestLangley:
Did you rebuild?

looeee:
Ah, of course. No I was trying to run this on raw.githack.

src/lights/AmbientLight.js
```diff
@@ -1,26 +1,96 @@
...
+
+		// without extra factor of PI in the shader, would be 2 / Math.sqrt( Math.PI );
+
+		this.sh.coefficients[ 0 ].set( color1.r, color1.g, color1.b ).multiplyScalar( 2 * Math.sqrt( Math.PI ) );
+
+		return this;
+
+	},
+
+	setFromHemisphereColors: function ( skyColor, groundColor, /* up */ ) { // todo: support tilt
```
Mugen87: The comma behind `groundColor` seems to be problematic...

Mugen87:
It seems unit test breaks with this PR:

```
not ok 693 Source > Lights > ArrowHelper > Standard light tests
  ---
  message: "Died on test #4     at Object.<anonymous> (/home/travis/build/mrdoob/three.js/test/unit/three.source.unit.js:60362:10): Cannot read property 'setHex' of undefined"
  severity: failed
  actual: null
  expected: undefined
  stack: TypeError: Cannot read property 'setHex' of undefined
    at checkLightCopyClone (/home/travis/build/mrdoob/three.js/test/unit/three.source.unit.js:27599:18)
    at runStdLightTests (/home/travis/build/mrdoob/three.js/test/unit/three.source.unit.js:27573:4)
    at Object.<anonymous> (/home/travis/build/mrdoob/three.js/test/unit/three.source.unit.js:60364:5)
```

WestLangley:
>It seems unit test breaks with this PR:

Yes, `AmbientLight` no longer has a color property; the color is encoded in the SH3 coefficients.

This is a core change, and it is going to require support from project managers if it is to be implemented.

In previous iterations of the PR, I added a new light type, which was non-breaking. However, enhancing `AmbientLight` was preferred.

Perhaps there is a refactoring preferable to the one I have here...

donmccurdy:
> Perhaps there is a refactoring preferable to the one I have here...

I like the direction of this, and having AmbientLight be SH-based. Couple thoughts:

- (1) Is lack of a `color` property the only reason AmbientLight doesn't extend Light? If so, I wonder if we could imitate HemisphereLight (which uses one of its colors for the property) and have the AmbientLight pass its first three coefficients. It's a little arbitrary, but less so than a light that doesn't extend Light, IMO.
- (2) Do you feel strongly that LightProbe should extend AmbientLight, or is that a temporary thing to keep the codebase in a reasonable state while we work on the other aspects of light probes? I'm unsure about it.

WestLangley:
(1) Believe me, I have tried for days to shoehorn this new feature into the existing API. This was the best I have been able to do so far. Suggestions welcome.

(2)  To review, a "reflection probe" can be used as a source of light, so I found it natural to use "light probe" as a source of light, too. Thus in my first PRs, a "light probe" served two purposes: to store irradiance in a probe volume, and in addition, to be a source of light in a scene graph.... This was not well-received, so now "ambient light" is the source of light, and "light probe" just records irradiance. Behind-the-scenes, they share the same code, except adding an "ambient light" to the scene will brighten the scene, while adding a "light probe" to the scene will not. That was my reasoning. And, no, I do not feel strongly about it.

I am open to any suggestions.

donmccurdy:
> Believe me, I have tried for days to shoehorn this new feature into the existing API. This was the best I have been able to do so far. Suggestions welcome.

I may have missed something then... couldn't AmbientLight continue to extend Light, and just store its first 3 coefficients in the `.color` property, similar to what HemisphereLight does? I don't see any obvious problem with that.

> ... "ambient light" is the source of light, and "light probe" just records irradiance. Behind-the-scenes, they share the same code, except adding an "ambient light" to the scene will brighten the scene, while adding a "light probe" to the scene will not.

With this change, and assuming LightProbeVolume / baking will be left out in `examples/js[m]` for now (which I think is right), I think we could move LightProbe out as well. But this PR doesn't need to worry about that, looks fine.

Mugen87:
The PR should at least clean up the unit test, otherwise all subsequent builds are broken.

WestLangley:
TLDR;

@WestLangley said:

>so [in this PR] "ambient light" is [a] source of light, and "light probe" just records irradiance.

Right. That is how it should be. The probe measures light, it does not emit it.

Also, a light has both a color (unit-less), and an intensity parameter (which contains units specific to the light type.).

The probe does not require an intensity parameter. The units are implied by what is being measured.

An exception is if the probe is normalized. In that case, a scale factor containing the proper units is required to map normalized units to actual units.

PR #16374, opened 2019-05-02
OrbitControls enhancement: Support target in same local space as camera / controlled object

EliasHasle:
Works well in my own, non-public, example, contrary to the current version. Now I have OrbitControls that follow a camera attached to a moving object.

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -133,6 +133,8 @@ THREE.OrbitControls = function ( object, domElement ) {
 		var lastPosition = new THREE.Vector3();
 		var lastQuaternion = new THREE.Quaternion();
 
+		var globalTarget = new THREE.Vector3();
+		
 		return function update() {
 
 			var position = scope.object.position;
@@ -178,7 +180,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 			position.copy( scope.target ).add( offset );
 
-			var globalTarget = scope.target.clone();
+			globalTarget.copy( scope.target );
 			if ( scope.object.parent ) {
 			 
 				scope.object.parent.localToWorld( globalTarget );
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -178,7 +178,13 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 			position.copy( scope.target ).add( offset );
 
-			scope.object.lookAt( scope.target );
+			var globalTarget = scope.target.clone();
+			if ( scope.object.parent ) {
+			 
+				scope.object.parent.localToWorld( globalTarget );
+			
+			}
+			scope.object.lookAt( globalTarget );
 
 			if ( scope.enableDamping === true ) {
 
--- a/examples/misc_controls_orbit_transformed.html
+++ b/examples/misc_controls_orbit_transformed.html
@@ -1,0 +1,162 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - orbit controls, transformed</title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+		<style>
+			body {
+				color: #000;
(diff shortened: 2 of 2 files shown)


EliasHasle:
Note that the default global target behavior is conserved when camera is in local space.

Note also that instead merely updating the controls with a moving global target does not work when the camera is attached to a moving object, because the other calculations in OrbitControls assume that the camera is in the same space as the target.

My solution is the smallest sufficient modification that solves the problem. Another solution that conserves the target as strictly global would have to modify all the other calculations (polar angle etc.).

examples/js/controls/OrbitControls.js
```diff
@@ -178,7 +178,13 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 			position.copy( scope.target ).add( offset );
 
-			scope.object.lookAt( scope.target );
+			var globalTarget = scope.target.clone();
```
WestLangley: Avoid calling `clone()` in a tight loop.

EliasHasle: Is it better to preallocate globalTarget and use copy?

WestLangley:
`OrbitControls` was not designed to support a camera that is a child of a moving or rotated object.

Plus, targets in three.js are by convention in global space, not relative space.

As I have said previously, I am reluctant to support additional changes to `OrbitControls` at this point. `OrbitControls` is included in the examples so users can modify it as they see fit.

I hope that is considered reasonable.

EliasHasle:
It is not unreasonable. I understand your concerns to some extent. :-)

However, I question the value of the tradition that targets are in global space. There are enough issues and confusion around `lookAt`, as demonstrated by this query (several involving OrbitControls): https://github.com/mrdoob/three.js/issues?utf8=%E2%9C%93&q=lookat

My change breaks no code that follows the intended usage, but provides a simple solution for a slightly different usage that I believe is interesting to many users. In fact, I believe many users may encounter the default behavior as a hard-to-explain bug, when they, for instance, apply OrbitControls to a z-up camera in a rotated z-up scene container. (For that particular case, there exists a workaround which is to instead use a y-up camera outside the container, but first the user must understand what went wrong.)

arodic:
> OrbitControls was not designed to support a camera that is a child of a moving or rotated object.

One can argue that almost any bug (or missing feature) exists because X was not designed to support Y. I notice this argument quite too often here. The question here is "why" shouldn't X be redesigned to support Y? Is it because there is are other ways to achieve Y on the application side using one of many possible workarounds?

> As I have said previously, I am reluctant to support additional changes to OrbitControls at this point. OrbitControls is included in the examples so users can modify it as they see fit.

Any piece of code in this repository can be modified by users as they see fit, so examples are not special in that regard.

EliasHasle:
I made a simple working example available [here ](https://eliashasle.github.io/misc_controls_orbit_transformed.html) (don't know if it is online yet when a reader clicks the link, but it was not ready when I posted this.)

However, after fiddling with a more advanced application [here](https://stackoverflow.com/questions/53292145/forcing-orbitcontrols-to-navigate-around-a-moving-object-almost-working/56012175#56012175), I realize that my solution is far from universal. The problem appears to be that lookAt orients the camera to the world up axis (I have no idea why that is considered a good thing). The question asker's rather simple solution of applying OrbitControls on a global camera and copying the resulting transform to the rendering camera notably does not have this problem.

Closing my PR. Thank you anyway.

WestLangley:
>The problem appears to be that lookAt orients the camera according to the world up axis

No, `lookAt()` orients the camera according to the world-space `.up` direction. That is, according to the direction `object.up`. `object.up` is a world-space direction.

EliasHasle:
@WestLangley Thank you for correcting me. So object.up (or, often in the case of OrbitControls, camera.up) is a world-space direction that determines the rotation around the pointing axis. object.up could have been in local space, but is in world space.

levilovelock:
Just wanted to say thankjs @EliasHasle for this change - I'm using it locally and it's awesome. I banged my head against OrbitControls for nearly 2 evenings trying to make it do what I wanted.

DVLP:
Is there already a built-in way to use this fix without this workaround? It takes ages to find what's causing the problem and then luck in finding this thread.

PR #16378, opened 2019-05-03
Replacement for Cache

akoww:
Good day!
I have implemented a replacement for the existing caching system based on IndexedDB. Now it is possible to store and load data locally on the client side. This might reduce the loading time of 3d objects and all kinds of textures and images. I hope it's a meaningful improvement. :D

### DIFF
--- a/src/loaders/FileLoader.js
+++ b/src/loaders/FileLoader.js
@@ -2,7 +2,7 @@
  * @author mrdoob / http://mrdoob.com/
  */
 
-import { Cache } from './Cache.js';
+import { IndexedCache } from './IndexedCache.js';
 import { DefaultLoadingManager } from './LoadingManager.js';
 
 var loading = {};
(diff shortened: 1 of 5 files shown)


Mugen87:
Is it possible to implement your cache without changing the loaders? In this way, you could offer it as an alternative cache implementation that can be used by just doing this:
```js
THREE.Cache = THREE.IndexedCache;
```

akoww:
> Is it possible to implement your cache without changing the loaders? In this way, you could offer it as an alternative cache implementation that can be used by just doing this:
> 
> ```js
> THREE.Cache = THREE.IndexedCache;
> ```

Would really like to do it this way, but i have no idea if it's possible. I had two major problems with the current THREE.Cache implementation:

**1.** THREE.Cache is implemented synchronous with a single object that stores all the data. IndexedDB is on the other hand is an async API. At least THREE.Cache.get() needs a callback function and not a simple return to make it work.  
It would be much better, if all the THREE.Cache functions had a callback function and not a return. This would allow us to implement different caching systems using all kinds of backends.

It's possible to deal with this problem by using async/await but it wont work with older browser versions.  [async/await - Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility).

**2.** THREE.ImageLoader caches HTMLImage​Elements. This form of images can't be simply insert into indexedDB or any other kind of caching system expect the current THREE.Cache. This images should be converted into serializable form before caching it.

This is a problem that needs to be solved within THREE.ImageLoader. Pushing HTMLElements into the cache is not a good idea.

**Possible Solution**
For compatibility reasons it would be a good idea to rework the THREE.Cache in multiple steps.
**1.** Extend the all the THREE.Cache functions with callbacks.
**2.** Rewrite all loader to use callbacks instead of return values.
**3.** Mark THREE.Cache function calls without any callback parameter as deprecated
**4.** Remove return from the functions to guarantee compatibility with all upcoming async caching systems.

Usnul:
I'm quite interested in replacement for a cache too. I use my own asynchronous cache and i've been fighting with THREE.js for a while now on that front. I would like to see a pattern analogous to Guava's LoadingCache, where you can have the 2 tiers, one : a simple lookup table (i.e. how Cache currently works) and second level, with loaders and async API.

Mugen87:
Related #16798

Mugen87:
Closing for now. It should be clear that we can't merge the PR in its current from. However, it provides valuable input for #16798.

@akoww Thanks for sharing your work here 👍

PR #16385, opened 2019-05-05
Allow for serialization of custom interpolants

Temdog007:
Changes from #15819

### DIFF
--- a/examples/js/exporters/GLTFExporter.js
+++ b/examples/js/exporters/GLTFExporter.js
@@ -1555,7 +1555,7 @@ THREE.GLTFExporter.prototype = {
 				// Detecting glTF cubic spline interpolant by checking the method's type parameter.
 				// GLTFCubicSplineInterpolant is a custom interpolant and track doesn't return
 				// valid value from .getInterpolation().
-				if ( track.createInterpolant !== undefined && track.createInterpolant.type == "GLTFCubicSplineInterpolant" ) {
+				if ( track.createInterpolant !== undefined && track.createInterpolant.isInterpolantFactoryMethodGLTFCubicSpline === true ) {
 
 					interpolation = 'CUBICSPLINE';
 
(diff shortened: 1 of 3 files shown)


donmccurdy:
Any reason for replacing the `isInterpolantFactoryMethodGLTFCubicSpline=true` property with a `type` property? My assumption had been that the `isFoo` pattern was somewhat preferred in the three.js codebase, but I'm OK with either.

Have you been able to test this already, or need help with that? I'd suggest https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/InterpolationTest as an example.

Temdog007:
> Any reason for replacing the isInterpolantFactoryMethodGLTFCubicSpline=true property with a type property? 

No, I'll add it back.

> Have you been able to test this already, or need help with that?

I've tested it with the model you suggested. No issues found.

donmccurdy:
Sorry for the long delay here! If you (or another contributor) would like to have this change and could rebase it, I think we can merge the PR.

Mugen87:
Closing. The code base has significantly changed since the PR was opened. It's best to create a fresh PR on latest `dev`.

BTW: A round trip will only work with updating the core (e.g. `getInterpolation()` and `setInterpolation()` in `KeyframeTrack`). I'm not sure this is possible when custom keyframe tracks are not "injected" in some way before the deserialization runs. It's probably less complex to just move the glTF based cubic spline interpolant to the core and don't support custom interpolants.

Mugen87:
@donmccurdy Do we have a glTF asset in the repository with cubic spline interpolation?

Mugen87:
> It's probably less complex to just move the glTF based cubic spline interpolant to the core and don't support custom interpolants.

https://github.com/mrdoob/three.js/blob/ebf6d68e08377b9918dce6508ce1b2f28f194770/src/animation/AnimationUtils.js#L258-L271

Since `AnimationUtils` already refers to `GLTFCubicSplineInterpolant`, it makes even more sense^^.

donmccurdy:
> Do we have a glTF asset in the repository with cubic spline interpolation?

It seems not — they're all linear. This would be a good test case:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/InterpolationTest

Mugen87:
Yeah, importing this asset in the editor and reloading breaks the cubic spline animations :neutral_face:

Issue #11943, opened 2017-08-13
THREE.AnimationMixer:The finished Event causes an infinite loop[BUG]


xjindf:
I used the code to create an animation, but when I wanted to re-execute the last one after the execution of the last animation, the THREE.AnimationMixer's finished event produced an infinite loop and was in great need for your help.
This is my code:
**function createAnimation() {
        mixer = new THREE.AnimationMixer(model);
        mixer.addEventListener('finished', function (e)
        {
            var curAction = e.action;
            var clip = curAction.getClip();

            var len = clip.tracks.length;
            while (len--) {
                var kf = clip.tracks[len];
                var vc = kf.values.length;
                while (vc--) {
                    var value = kf.values[vc];
                    if (value > 0) {
                        kf.values[vc] += 10;
                    }
                }
            }
            curAction.stop();//Is there a need to restore other parameters to stop the current animation? please help me
            //When an animation is executed again, the state should not be completed
curAction.play();//The finished event was fired immediately when playing again
        });
        var positionGo = new THREE.VectorKeyframeTrack('.position', [0, 3], [0, 0, 0, 0, 0, 50]);
        var moveClipGo = new THREE.AnimationClip('Move_Car_Go', -1, [positionGo]);
        var moveActionGo = mixer.clipAction(moveClipGo, car);
        moveActionGo.clampWhenFinished = true;
        moveActionGo.loop = THREE.LoopOnce;
        moveActionGo.play();
    }**

ehs035:
> I used the code to create an animation, but when I wanted to re-execute the last one after the execution of the last animation, the THREE.AnimationMixer's finished event produced an infinite loop and was in great need for your help.
> 
> This is my code:
> 
> ```js
>     function createAnimation() {
>         mixer = new THREE.AnimationMixer(model);
>         mixer.addEventListener('finished', function (e)
>         {
>             var curAction = e.action;
>             var clip = curAction.getClip();
> 
>             var len = clip.tracks.length;
>             while (len--) {
>                 var kf = clip.tracks[len];
>                 var vc = kf.values.length;
>                 while (vc--) {
>                     var value = kf.values[vc];
>                     if (value > 0) {
>                         kf.values[vc] += 10;
>                     }
>                 }
>             }
>             curAction.stop();//Is there a need to restore other parameters to stop the current animation? please help me
>             //When an animation is executed again, the state should not be completed
> curAction.play();//The finished event was fired immediately when playing again
>         });
[...]

PR #16407, opened 2019-05-09
AnimationAction: Update .time before firing events.

### LINKED ISSUE #9857: AnimationMixer finished event fires prematurely

As far as I understand it, the `finished` event on `AnimationMixer` should only fire when an action has finished.

In practice, it seems to often fire just before I'd expect.

For a 1000ms clip, at the point in time the `finished` fires, `action.time` will often be 983.4ms. This looks a lot like a single frame gap (ie 16.6ms) from the end. In addition, the property being manipulated will be just short (say, by `16.6 / 1000`) of the lerp to the terminal value in a track associated with the clip.

Is this intended behavior? If so, can an event be added that only fires when the action is truly finished? If not, is this a matter of moving the location of the logic that fires the event to be a bit later in its code path?
##### Three.js version
- [x] Dev
- [ ] r81
- [ ] ...
##### Browser
- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer
##### OS
- [x] All of them
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

Mugen87:
Fixes #9857

This PR adds the suggestion from here https://github.com/mrdoob/three.js/issues/9857#issuecomment-253872545 to ensure `AnimtionAction.time` is updated before the events `loop` or `finished` fire. This change should also solve @ehs035's issue in https://github.com/mrdoob/three.js/issues/11943#issuecomment-490621178.

### DIFF
--- a/src/animation/AnimationAction.js
+++ b/src/animation/AnimationAction.js
@@ -503,10 +503,18 @@ Object.assign( AnimationAction.prototype, {
 
 					time = 0;
 
-				} else break handle_stop;
+				} else {
+
+					this.time = time;
+
+					break handle_stop;
+
+				}
 
 				if ( this.clampWhenFinished ) this.paused = true;
 				else this.enabled = false;
+
+				this.time = time;
 
 				this._mixer.dispatchEvent( {
 					type: 'finished', action: this,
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

PR #16412, opened 2019-05-10
GLTFLoader - store original node name

Kalacione:
Create a new property `originalName` on node.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1299,7 +1299,7 @@ THREE.GLTFLoader = ( function () {
 
 			if ( typeof gltfDef.extras === 'object' ) {
 
-				object.userData = Object.assign(object.userData, gltfDef.extras);
+				Object.assign( object.userData, gltfDef.extras );
 
 			} else {
 
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -2996,6 +2996,7 @@ THREE.GLTFLoader = ( function () {
 
 			if ( nodeDef.name !== undefined ) {
 
+				node.userData.name = nodeDef.name;
 				node.name = THREE.PropertyBinding.sanitizeNodeName( nodeDef.name );
 
 			}
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1299,7 +1299,7 @@ THREE.GLTFLoader = ( function () {
 
 			if ( typeof gltfDef.extras === 'object' ) {
 
-				object.userData = gltfDef.extras;
+				object.userData = Object.assign(object.userData, gltfDef.extras);
 
 			} else {
 
--- a/examples/jsm/loaders/GLTFLoader.js
+++ b/examples/jsm/loaders/GLTFLoader.js
@@ -3088,7 +3088,6 @@ var GLTFLoader = ( function () {
 
 			if ( nodeDef.name !== undefined ) {
 
-				node.userData.name = nodeDef.name;
 				node.name = PropertyBinding.sanitizeNodeName( nodeDef.name );
 
 			}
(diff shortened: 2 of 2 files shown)


examples/jsm/loaders/GLTFLoader.js
```diff
@@ -3088,6 +3088,7 @@ var GLTFLoader = ( function () {
 
 			if ( nodeDef.name !== undefined ) {
 
+				node.originalName =  nodeDef.name ;
 				node.name = PropertyBinding.sanitizeNodeName( nodeDef.name );
```
donmccurdy: I'd rather not attach new properties to native three.js objects. Perhaps `node.userData.name = nodeDef.name`?

/cc @takahirox

takahirox: Yeah, if we store an original name, I prefer `.userData`.

Kalacione: Yep, you're absolutely right.

donmccurdy:
Two remaining issues:

(1) We should make this change to the version in `examples/js/`. Changing the version in `examples/jsm/` isn't necessary, it is automatically generated.
(2) If the node has other extras, `assignExtrasToUserData` will currently overwrite the name because it replaces the whole `userData` object.

Kalacione:
Thank you for your precious comments 🥇 

Maybe in `assignExtrasToUserData`, we can replace
`object.userData = gltfDef.extras;`
by
`object.userData = Object.assign(object.userData, gltfDef.extras);`
so all existing properties in `userData` will be kept ?

donmccurdy:
Thanks for the PR! :)

Let's just iterate `gltdDef.extras` and assign each property to `object.userData` in a loop – `Object.assign` would be cleaner but it's one more thing that devs supporting IE11 have to polyfill (currently they only need a Promise polyfill).

takahirox:
I think `Object.assign` is in Three.js core. See the top of https://raw.githubusercontent.com/mrdoob/three.js/dev/build/three.js

donmccurdy:
Oh! Even better. 🙂

examples/js/loaders/GLTFLoader.js
```diff
@@ -1299,7 +1299,7 @@ THREE.GLTFLoader = ( function () {
 
 			if ( typeof gltfDef.extras === 'object' ) {
 
-				object.userData = gltfDef.extras;
+				object.userData = Object.assign(object.userData, gltfDef.extras);
```
donmccurdy: Object.assign modifies its first argument, and threejs' styleguide uses spaces inside parens. So instead:

```js
Object.assign( object.userData, gltfDef.extras );
```

donmccurdy (approved):

mrdoob:
Thanks!

Kalacione:
I just can't believe that my first pull request of all time is now part of three.js
You guys are awesome.

takahirox:
@Kalacione BTW curious to know what do you use original name for?

Kalacione:
> 
> 
> @Kalacione BTW curious to know what do you use original name for?

We use THREE to build a CAD file viewer/editor that imports any format and outputs optimised GLTF and USDZ. 
Import process and export with materials assignation is done server-side.
As we assign materials to part names, they have to match on both sides.

Quick [demo](http://product-tour.05296988ec5e4f16b02b.westeurope.aksapp.io/).

looeee:
I just noticed this today and was wondering where `userData.name` had come from. I expected that `userData` would be left empty, reserved for use at the app level. 

If we do keep this, it would probably be better if it was given a less generic name, something like `userData.gltf.originalNodeName` so that people can tell why it's there.

donmccurdy:
Yeah, unfortunately I'm thinking we probably should revert this and leave it for users to do themselves – it would be very easy with the extension mechanism @takahirox is proposing. This need seems too specific, and too easy to work around, to be putting data into `.userData` for all users.

Issue #15351, opened 2018-11-30
GLTFExporter: Support spec/gloss PBR materials


tiancaipipi110:
Hello, I'm using GLTFLoader to import the .gltf models, then use gltf exporter to export them. 

However some materials are only able to be displayed, but unable to be exported.  It seems like three.js has a couple of built-in materials+ shader materials. gltf loader seems to create shader materials as far as three.js built-in materials do not cover what' s in gltf file. now gltf exporter does not support exportinng shader materials. and from here we have a problem.

Can anyone help with this? Thanks.

makc:
I imagine that for th proper fix gltfloader would need to create its own material type, instead of ShaderMaterial. Then exporter could easily export it.

looeee:
> glTF has three material modes:

> * unlit
> * metal/rough PBR
> * spec/gloss PBR

Perhaps we should support spec/gloss PBR directly in three.js?

makc:
@looeee which is what I said, basically, but `material.isGLTFSpecularGlossinessMaterial` could work, too

donmccurdy:
I'm certainly not opposed to adding a spec/gloss PBR material to the core materials. We've discussed that a bit in https://github.com/mrdoob/three.js/issues/10985 and https://github.com/mrdoob/three.js/pull/11038, and that would solve the fact that GLTFSpecularGlossinessMaterial can't be cloned (although https://github.com/mrdoob/three.js/pull/14099 might also achieve that?). But the difference between these two PBR materials, and MeshPhysicalMaterial for that matter, is not likely to be obvious to most users. So there is an argument for keeping it 'tucked away' in GLTFLoader, too.

I believe GLTFExporter can be fixed to support spec/gloss either way.

looeee:
> But the difference between these two PBR materials, and MeshPhysicalMaterial for that matter, is not likely to be obvious to most users. 

As a counterargument, many users will come to three.js already familiar with PBR, or want to bring in assets authored in Substance Painter or other programs using spec/gloss, and find it frustrating that only one of the standard PBR pathways is supported. 

It's not trivial to [convert spec/gloss maps to metal/roughness](https://marmoset.co/posts/pbr-texture-conversion/), meaning that many PBR assets cannot be (easily) three.js at the moment.

RemusMar:
There is no perfect workflow, but this one is the probably the best:
https://github.com/BabylonJS/Exporters/issues/269

donmccurdy:
[...]

PR #16413, opened 2019-05-10
Added support for "KHR_materials_pbrSpecularGlossiness" extension in GLTFExporter.js

ted537:


donmccurdy:
It would be nice to support spec/gloss in GLTFExporter I agree. Was there an issue with this? If you want to reopen it feel free, I would suggest using `npm run lint` and the threejs styleguide for whitespace though. Thanks.

pawlusmall:
isn't this related with https://github.com/mrdoob/three.js/issues/15351 ?

donmccurdy:
This PR would need a bit of linting to match the styleguide, but yes it would fix #15351.

PR #16421, opened 2019-05-13
Design update

yuinchien:
TODO Suggestions:
- Switching btw docs & examples is now one click away, but we should default to the first article/example instead of blank content.
- Landing page needs to be updated to adopt the new fonts. 
- The example list in left panel has wide space btw '/', consider wrapping a span tag around it and adjust letter-spacing.
- Consider default closing or hide the Dat.gui panel for mobile.


mrdoob:
Thanks!

PR #16427, opened 2019-05-13
3MFLoader: Added basic material support.

Mugen87:
The loader is now able to parse material definitions. Added a new example with a more complex `3MF` asset that demonstrates this feature and component parsing.

image


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev29/examples/webgl_loader_3mf_materials.html

Mugen87:
BTW: This is the actual thumbnail of the asset:

image

So the colors of the model are not too bright^^.

mrdoob:
Nice!

mrdoob:
Thanks!

PR #16429, opened 2019-05-13
LDrawLoader: Add correct optional line rendering, better transparent highlights

gkjohnson:
Depends on #16424. (Once that PR is merged I'll update this one)

This PR adds correctly rendered optional lines using the algorithm described [on the LDRaw Spec page](https://www.ldraw.org/article/218.html). A custom shader is instantiated that discards a line if the control points that are passed in via attribute are on the same side of the line. I've noticed that some of the lines get lost inside of or partially obscured by the 3d geometry -- I wonder if it's worth offsetting them a little bit toward the camera more?

_AT-ST_

image

_Lunar Vehicle_

image

This also enables "premultipliedAlpha=true" to transparent materials which makes their highlights brighter as expected:

image

For testing:
https://raw.githack.com/gkjohnson/three.js/add-optional-lines-shader/examples/webgl_loader_ldraw.html

@yomboprime


WestLangley:
>I've noticed that some of the lines get lost inside of or partially obscured by the 3d geometry -- I wonder if it's worth offsetting them a little bit toward the camera more?

If the mesh and lines share vertex positions, you can try setting [polygonOffset](https://stackoverflow.com/a/31541369/1461008) on the mesh's material.

gkjohnson:
@WestLangley That works great! I'd tried it before but on the lines and it seems that `polygonOffset` doesn't work with line rendering.

Here's code to run in the console to show the difference:
```js
model.traverse(c => {

  if ( c.isMesh ) {

    c.material.forEach( m => {
      m.polygonOffset = true;
      m.polygonOffsetFactor = 1;
      m.polygonOffsetUnits = 1;

    } );
   
  }

} );
```

And a before / after image:

image

@yomboprime How do you feel about adding a polygon offset to the mesh materials to the loader?

yomboprime:
> @yomboprime How do you feel about adding a polygon offset to the mesh materials to the loader?

Sounds good. I wonder if it will interact with shadowmaps though.

yomboprime:
I've tried the offset in the console. Really nice! Good work.

gkjohnson:
@yomboprime
> Sounds good. I wonder if it will interact with shadowmaps though.

It looks like the polygon offset doesn't affect the shadows -- here's an image.

image

Because it affects the depth map, though, it will change the depth overlap with other models that don't use the polygon offset _slightly_ but not all that much.

examples/js/loaders/LDrawLoader.js
```diff
@@ -337,18 +599,161 @@ THREE.LDrawLoader = ( function () {
...
+							objGroup.add( createObject( parseScope.lineSegments, 2 ) );
+
+						}
+
+						if ( parseScope.conditionalSegments.length > 0 ) {
+
+							var conditionalSegments = parseScope.conditionalSegments;
+							var lines = createObject( conditionalSegments, 2 );
+							lines.isConditionalLine = true;
+							lines.visible = false;
```
yomboprime: I would leave the optional lines visible by default as you mentioned in the previous PR.

gkjohnson: Oh oops! Thanks I'll do that -- also once the other PRs are merged this one will be easier to take a look at. I know there's a lot of noise right now!

yomboprime:
Looking very good!

Besides the comment I've put in the code, I would do an optimization in the shader: The discard calculations can be performed in the vertex shader instead of the fragment one. Also, instead of passing all the vectors as varyings, just one float varying (`discardFlag`) with the result of the calculations can be passed. If it is `!= 0.0`, just discard the fragment. This way the calculations are done once per line, not once per fragment.

gkjohnson:
@yomboprime 
> The discard calculations can be performed in the vertex shader instead of the fragment one.

You're right! I don't know why I didn't do that in the first place -- I just made both of your suggested changes and merged `dev` in so things should be easier to take a look at now.

examples/js/loaders/LDrawLoader.js
```diff
@@ -566,9 +644,66 @@ THREE.LDrawLoader = ( function () {
...
+
+							}
+
+							lines.geometry.addAttribute( 'control0', new THREE.BufferAttribute( controlArray0, 3, false ) );
+							lines.geometry.addAttribute( 'control1', new THREE.BufferAttribute( controlArray1, 3, false ) );
+							lines.geometry.addAttribute( 'direction', new THREE.BufferAttribute( directionArray, 3, false ) );
+
+							lines.material = lines.material.map( mat => {
+
+								return new THREE.ShaderMaterial( {
```
gkjohnson: @yomboprime Do you have any suggested ways for how to handle this unique type of material? I suppose the right way would be for the material libraries to be refactored to support two different types of line materials.

yomboprime: Yes, the library should provide optional lines materials, otherwise you can end up with lots of different custom material shaders with the same colour.

I would replicate `material.userData.edgeMaterial` into another `material.userData.optionalEdgeMaterial` variable. Both are created at the same time with the same colour; one is `LineBasicMaterial` and the other is your custom shader. And wether one or the other is used depends on wether we're currently creating normal lines or optional lines.

Or even better, I would optimize further and create the line materials in a lazy way: Create LineBasic or custom shader line material the first time it is needed if it is null while parsing elements, and set it in the `material.userData`. Probably you will need to store a material's `userdata.edgeColour` when creating it, for the lazy creation.

yomboprime: @gkjohnson I can do it myself in another PR if you find it confusing.

gkjohnson: @yomboprime Thanks! If you already have an idea of how you want to implement the material cache I'd be happy to pass it off. Otherwise I may be able to get to it later in the week / next weekend!

yomboprime: All right. This can be merged then.

yomboprime (approved):

mrdoob:
Thanks!

PR #16431, opened 2019-05-14
Fix the file path in Import-via-modules.html.

ddingchen:
There's a small mistake with the js path.


mrdoob:
I think that path is correct...

ddingchen:
Sorry, it;s my mistake.

PR #16437, opened 2019-05-14
OBJLoader: Fix parsing of vertex colors.

Mugen87:
Fixes #15171


Mugen87:
see https://github.com/mrdoob/three.js/pull/15171#issuecomment-492308392

PR #16449, opened 2019-05-15
JSM: Clean up for MapControls.d.ts

linbingquan:


gkjohnson:
Is there a consistent code style for the `d.ts` files? It seems like there are a couple style variants in the repo, now. I feel like it should stay in line with the style used in the main codebase which might mean using single quotes for strings.

Is there any way to get the eslint rules to apply to the `d.ts` files?

Mugen87:
The `d.ts` files use all soft tabs. I would prefer to clean these files up all at once and not individually. I bet certain editors/IDEs might be able to auto-change this with a search and replace function.

gkjohnson:
@Mugen87 It looks like it might be possible to configure eslint to also lint the `d.ts` files so all of the existing lint rules can be applied and tested:

https://www.npmjs.com/package/@typescript-eslint/parser

Maybe it's worth looking into so we can enforce more consistency among the typsecript definitions?

Mugen87:
Sounds good!

mrdoob:
Sounds good to me too!

Mugen87:
Closing. Let's try it with the eslint approach first before start manually converting files.

PR #16452, opened 2019-05-15
ColladaLoader: Report xml parse errors

### LINKED ISSUE #16443: ColladaLoader: Throw error if xml fails to parse correctly
If a Collada file fails to parse as XML correctly the Collada loader tries to parse it anyway and gives less than helpful errors. Here is one of the models I was having trouble loading:

https://bitbucket.org/DataspeedInc/velodyne_simulator/src/master/velodyne_description/meshes/HDL32E_base.dae.

With the problematic lines being the following because arrow characters are used in an attribute:

```xml
<library_visual_scenes>
    <visual_scene id="Scene" name="Scene">
      <node id="<STL_BINARY>" name="<STL_BINARY>">
        <matrix>1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1</matrix>
```

The result is a partially-parsed xml document with a `<parsererror>` tag as the first child node in the `COLLADA` tag. In this case the partially-formed xml was invalid and caused the loader to fail but it could have failed silently and only loaded part of the model. This file does not load in Meshlab, either, and it looks like Assimp is the source of the invalid Collada file conversion: https://github.com/assimp/assimp/issues/823.

My suggestion would be to throw / log an error if a `<parsererror>` tag is present and return a `null` model. Thoughts?

gkjohnson:
Fix #16443.

Logs an error in ColladaLoader if the browsers xml parser fails to parse the text. I'm not sure if there's a standard way to report errors in the loaders. Is logging an error good enough? Should it throw an error? It doesn't look like there's an `onError` function to call.

Each browser seems to format the parsererror differently so here's how it looks in different browsers:

#### Firefox
image

#### Chrome
image

#### Edge

A still extremely unhelpful error...

image

### DIFF
--- a/examples/js/loaders/ColladaLoader.js
+++ b/examples/js/loaders/ColladaLoader.js
@@ -3824,6 +3824,14 @@ THREE.ColladaLoader.prototype = {
 
 		var collada = getElementsByTagName( xml, 'COLLADA' )[ 0 ];
 
+		var parserError = xml.getElementsByTagName( 'parsererror' )[ 0 ];
+		if ( parserError !== undefined ) {
+
+			console.error( 'ColladaLoader: Failed to parse collada file.', parserError );
+			return null;
+
+		}
+
 		// metadata
 
 		var version = collada.getAttribute( 'version' );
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #16456, opened 2019-05-16
Homepage design update

yuinchien:
TODO:
- I updated the default selection for docs & examples to be the first item in the list. However, we should also update the selected title in the left panel for docs. (clicking on "docs" in landing page)

- Please review the left panel in landing, I tried listing all the books & online courses since the list isn't that long. Feel free to change as you see fit.

Preview link: https://raw.githack.com/mrdoob/three.js/design-update/

Extra large screen layout:
Screen Shot 2019-05-15 at 4 48 14 PM


mrdoob:
Looking good!

mrdoob:
Oh, seems like it's missing an image in mobile mode.

Screenshot 2019-05-15 at 17 07 24

yuinchien:
aha, the menu icon. Will fix.

docs/page.css
```diff
@@ -159,32 +177,36 @@ strong {
 	font-weight: normal;
 }
 
+
+/* TODO: Duplicate styles in main.css. Needed here cause button is inside the iframe */
```
yuinchien: This bit of CSS is duplicated in main.css. Since the floating button anchored at the bottom right corner is a reused component, it makes sense to have the styles in main.css.

However, the button for docs seems to live inside iframe, is it easy to move it outside of iframe and remove the css here? 

@mrdoob

mrdoob:
@yuinchien unrelated to homepage but I just noticed that `<br>`s do not seem to work inside `<p>`s in the docs?

Before:
Screenshot 2019-05-15 at 17 20 04

After:
Screenshot 2019-05-15 at 17 20 17

https://raw.githack.com/mrdoob/three.js/design-update/docs/index.html#api/en/geometries/BoxGeometry

yuinchien:
I broke it... i set br to display none, cause they exist randomly in the article, i needed to accurately adjust spacing. will revert it.

mrdoob:
> I broke it... i set br to display none

Haha! Living dangerously 😁
Don't worry about the random `<br>`s, we'll clean them over time.

yuinchien:
So now we see the <br> white space below Creating a scene. 
Should do a close look at the docs and make formatting consistent. Otherwise, page will look off.

Screen Shot 2019-05-15 at 5 30 41 PM

Mugen87:
I've tested the [preview](https://raw.githack.com/mrdoob/three.js/design-update/) on my iMac 5K and it seems I'm not getting the large screen layout:

image

Unfortunately, the screenshot does not really convey how oversized the entire website looks on my screen.

yuinchien:
I think the extra large screen scale could make sense for TV size display viewing from a distance, but doesn't work for iMac display. I pushed change to postponed the layout for extra large screen. Will update when i can access a real display. Thanks!

mrdoob:
Thanks!

PR #16472, opened 2019-05-17
JSM: More modules and TS files.

Mugen87:
This PR adds the frequently requested `CSS2DRenderer` and `CSS3DRenderer` to the module list. Also `TransformControls` is added which means all controls are now available as modules (🎉). Of course all modules were individually tested and updated with TS declaration files.

The PR does also some minor style clean up in `TransformControls` and `MTLLoader` and the respective modules.


mrdoob:
Many thanks!

PR #16477, opened 2019-05-17
modularize.js: Enable example dependencies.

Mugen87:
With this PR, it's now possible to convert modules like `ColladaLoader` that have dependencies to other modules. Interestingly, the new code revealed a bug in the module of `ShadowMapViewer` since `UnpackDepthRGBAShader` is no part of the core. 

The PR implements the suggestion mentioned by @mrdoob here: https://github.com/mrdoob/three.js/pull/16400/files#r284492725

It introduces per module an array that represents the dependencies to other example files. Per entry there is a key/value mapping that looks like: `classname`=>`path to module`. In this way, it's very flexibel to define dependencies since you not have a 1:1 relationship between class name and module name. E.g. it's now easy to define that `Pass` should be loaded from `EffectComposer.js` which would be not possible to derive only from code.

### DIFF
--- a/examples/jsm/shaders/UnpackDepthRGBAShader.js
+++ b/examples/jsm/shaders/UnpackDepthRGBAShader.js
@@ -1,0 +1,53 @@
+/**
+ * @author alteredq / http://alteredqualia.com/
+ *
+ * Unpack RGBA depth shader
+ * - show RGBA encoded depth as monochrome color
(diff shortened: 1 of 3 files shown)


donmccurdy (approved):

mrdoob:
You're the best! 😀

Issue #15584, opened 2019-01-15
GLTFLoader onProgress total is always 0


santacrab:
When I load a model using the GLTFLoader, when I try to calculate the progress % in the onProgress function, the "total" value of the ProgressEvent is always 0. I tried with different models and I always get the same result.

Here my onProgress function:
`function (xhr) {
        console.log((xhr.loaded / xhr.total * 100) + '% loaded');
      }`

xhr.total is always 0 so the result is always "Infinity% loaded". I saw the same thing on many examples online.
Attached you can also find a screen of the ProgressEvent debugged.
<img width="642" alt="schermata 2019-01-15 alle 10 03 48" src="https://user-images.githubusercontent.com/1327320/51169808-17d8be00-18ad-11e9-94d0-8630d8a400ae.png">

##### Three.js version

- [ ] Dev
- [X] r100
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [ ] Windows
- [X] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
In general, using the `onProgress` callback of `THREE.LoadingManager` makes more sense when loading a `gltf` file, see #13895

donmccurdy:
From the issue, I can't tell if the OP is loading multiple files. If there's only one file, progress events aren't much help...

@santacrab note that onProgress callbacks for all loaders track the number of network requests completed vs the number of network requests made. They don't provide smooth incremental progress while a single file is parsed, for example. Are you loading multiple files?

However, I don't know why `.total` should ever be less than `.loaded` ... could you share a demo?

Mugen87:
>note that onProgress callbacks for all loaders track the number of network requests completed vs the number of network requests made.

I don't think this is true. In the following example `onProgress()` actually tracks how much of the `OBJ` file is already downloaded.


https://threejs.org/examples/webgl_loader_obj

donmccurdy:
Oh good point. GLTFLoader just passes the `onProgress` event into its FileLoader for the initial request, and not for additional resources... I can't think of a better way to handle that at the moment, but open to ideas. In general LoadingManager is probably more helpful, yes.

RemusMar:
Come on Don ...
In the GLTF loader "onProgress" tracks how much of the GLB/GLTF file is already downloaded.
[...]

PR #16485, opened 2019-05-19
GLTFLoader - onProgress 

eXponenta:
Saving onProgress event and pass to another loaders in GLTFParser.parse.

Default GLTFLoader call onProgress only for *.gltf files and ignore it for other files (textures, mesh-bundle).

I fixed it.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -218,7 +218,7 @@ THREE.GLTFLoader = ( function () {
 
 				path: path || this.resourcePath || '',
 				crossOrigin: this.crossOrigin,
-				manager: this.manager
+				manager: this.manager,
 				onProgressHandler: this.onProgressHandler
 
 			} );
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -4,6 +4,7 @@
  * @author Tony Parisi / http://www.tonyparisi.com/
  * @author Takahiro / https://github.com/takahirox
  * @author Don McCurdy / https://www.donmccurdy.com
+ * @author Timoshenko Konstantin / https://github.com/eXponenta
  */
 
 THREE.GLTFLoader = ( function () {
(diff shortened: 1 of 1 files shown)


examples/js/loaders/GLTFLoader.js
```diff
@@ -4,6 +4,7 @@
  * @author Tony Parisi / http://www.tonyparisi.com/
  * @author Takahiro / https://github.com/takahirox
  * @author Don McCurdy / https://www.donmccurdy.com
+ * @author Timoshenko Konstantin / https://github.com/eXponenta
```
Mugen87: The authorship of files is normally not changed if you just make a minor addition to a file.

examples/js/loaders/GLTFLoader.js
```diff
@@ -1843,7 +1848,7 @@ THREE.GLTFLoader = ( function () {
 
 		return new Promise( function ( resolve, reject ) {
 
-			loader.load( resolveURL( bufferDef.uri, options.path ), resolve, undefined, function () {
+			loader.load( resolveURL( bufferDef.uri, options.path ), resolve, options.onProgressHandler, function () {
```
Mugen87: I don't think it makes sense to use the same `onProgress` callback for multiple requests. Consider to use `LoadingManager` instead, see https://github.com/mrdoob/three.js/issues/15584#issuecomment-454515571

eXponenta: LoadingManager works only with files, but onProgress in loader must return loading progress for current task.
I have gltf file about 30mb of binary bundle. Current implementation works fine, but loading manager return always 1.

silvainSayduck:
This pull requests passes progress events from multiple sources (so with different `total` and `loaded` values for each resource - which prevents from really computing the glTF+resources total progress), and does not take into account textures.

I made a fork of three-gltf-loader for which I implemented an overall onProgressHandler + that loads Texture with xhr to workaround the fact that Three.js currently doesn't support onProgress for textures:

https://github.com/silvainSayduck/three-gltf-loader/tree/v1.100.0

For other reasons, I am using r100 for now, so my implementation is built on top of v1.100.0, but it can easily be adapted to the latest Three.js.

Issue #9103, opened 2016-06-08
Paint on model example


VinnyVicious:
A very helpful addition to the list of examples would be an example that allowed you to paint on the texture of a mesh by clicking on a part of the mesh. It's a good example on how to transfer coordinates from screen and world view into texture projection.

WestLangley:
See http://threejs.org/examples/webgl_raycast_texture.html.

makc:
Btw http://jsdo.it/makc/zXNX

VinnyVicious:
Amazing! Thank you.

AbhinavPelapudi:
How to do free-hand painting(i Can choose any color to paint except white) and erasing(i want to use white color to erase the painting on rendered scene) on three js scene. i also want to extract the coordinates. can I know how to do this? are there any examples ?

WestLangley:
@AbhinavPelapudi Sorry, this is not a help site. Please use the three.js [forum](https://discourse.threejs.org), and show what you have tried.

sciecode:
I noticed issue has been pushed over and over again. So I decided to give it try.
I've used a technique similar to what you can find Blender's Texture Painter.
Here is a ['working' example](https://rawcdn.githack.com/sciecode/three.js/f4e363a8e0cf6c496f4191192d7eb15110442a7c/examples/webgl_paint_texture.html).

This is, by no means, finished. I would still need to implement/fix the following things:
- face clipping artifacts
- normal opacity falloff
- geometry occlusion
- BufferGeometry support
- public interface and variables

I just wanted to receive some feedback if this is inline with what's expected from this example, and if I should continue working on it.

sciecode:
Updated [example](https://rawcdn.githack.com/sciecode/three.js/392661079fce6413e2e7ff6b35d0700ceee209e7/examples/webgl_paint_texture.html).

This is still missing Occlusion Culling, you can check that by switching to the 'Torus' mesh and painting it sideways. I believe that sorting a high number of intersecting faces would slow this to a crawl. 

Changing the texture update procedure from canvas-based to gpu-based would probably allow for Occlusion Culling within 16ms constraint, but that would require a great deal of time. Time that I probably don't have, at the moment.

 [THREE.TexturePainter](https://github.com/sciecode/three.js/blob/dev8/examples/js/TexturePainter.js)
I don't know, if this is good enough for example, I can prepare this for PR. I don't believe there's too much to be done. Also considered switching name to MeshPainter, open to suggestions.

[...]

PR #16498, opened 2019-05-21
Add texture painter example

sciecode:
#9103 - [example](https://rawcdn.githack.com/sciecode/three.js/3c60dbd6bb26896ba3d488ac73d479dfca7540bd/examples/webgl_paint_texture.html)

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -310,6 +310,7 @@ var files = {
 		"webgl_lightningstrike",
 		"webgl_lightshafts",
 		"webgl_materials_modified",
+		"webgl_paint_texture",
 		"webgl_raymarching_reflect",
 		"webgl_shadowmap_pcss",
 		"webgl_simple_gi",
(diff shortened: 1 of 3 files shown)


looeee:
This is cool! 

But I think the test UV texture is wrong, (0,0) should be in the bottom left. 

uv_test_bw_1024

WestLangley:
Thank you for your contribution, but the code will need significant improvement. You are cloning or instantiating _tens-of-thousands_ of objects per frame when interacting with the mouse. Avoid calling `clone()` and `new` in tight loops. This must be fixed.

You likely do not need an animation loop, since your scene is static.

Also, `Geometry` will soon not be renderable -- only `BufferGeometry` will be. I expect that will require a major rewrite.

But to be honest, the purpose of the examples is to demonstrate three.js features. This demo appears to be more of a three.js-based application -- a bit too complicated to be useful as an example,  IMO.

sciecode:
Appreciate the feedback.

sciecode:
@looeee just a heads up, I agree with UV coordinates being wrong. 
I'm simply using the uv grid texture on dev: [UV_Grid_Sm.jpg](https://github.com/mrdoob/three.js/blob/dev/examples/textures/UV_Grid_Sm.jpg)

PR #16499, opened 2019-05-21
Matrix4: support .setPosition( x, y, z )

WestLangley:
Previously, only `.setPosition( v )` was supported.

This is now consistent with how we handle the args for `Object3D.lookAt()`.

### DIFF
--- a/docs/api/en/math/Matrix4.html
+++ b/docs/api/en/math/Matrix4.html
@@ -375,6 +375,7 @@ x, y, 1, 0,
 		</p>
 
 		<h3>[method:this setPosition]( [param:Vector3 v] )</h3>
+		<h3>[method:this setPosition]( [param:Float x], [param:Float y], [param:Float z] ) // optional API</h3>
 		<p>
 			Sets the position component for this matrix from vector [page:Vector3 v], without affecting the
 			rest of the matrix - i.e. if the matrix is currently:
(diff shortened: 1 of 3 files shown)


WestLangley:
There is a docs formatting issue that perhaps someone else can help with.

Here is how I handled the alternate API with `Matrix4.setPosition()`:
https://raw.githack.com/westlangley/three.js/dev-matrix4_setPosition/docs/#api/en/math/Matrix4.setPosition

Compare with the buggy docs for `Object3D.lookAt()`
https://raw.githack.com/westlangley/three.js/dev-matrix4_setPosition/docs/#api/en/core/Object3D.lookAt

Mugen87:
It seems this formatting issue is related to the design update, see #16456. There should be a line break after the first method signature similar to the prod version (/cc @yuinchien).

image

yuinchien:
@Mugen87 I don't remember removing`<br>` tag or changing the format of three API docs. What happened previously was that, I set `display: none;` to `<br>` in order to properly adjust spacing. The css was removed and pushed days ago. 

Looking at the html markup, `Object3D.lookAt()` is wrapped in one `<h3>` versus `Matrix4.setPosition()` is wrapped in separate `<h3>`. Which is the reason why they render differently. I think it'd be good to have a *style guide* for formatting content.

WestLangley:
@mrdoob This PR does not need to wait for the docs bug to be fixed.

Mugen87 (approved):

mrdoob:
Thanks!

PR #16506, opened 2019-05-21
Fix orbit control: look at target position correction

but0n:
Seems like the spherical here is base on local space:
```js
offset.copy( position ).sub( scope.target );
```

**LookAt** receive a position in world space.


WestLangley:
`OrbitControls` was not designed to support a camera that is a child of another object.

See discussion in #15268 and #16374.

but0n:
I think it is necessarily to point out this limitation in [document](https://threejs.org/docs/index.html#examples/controls/OrbitControls).

WestLangley:
>I think it is necessarily to point out this limitation in document.

Done!

but0n:
@WestLangley Thanks!

PR #16507, opened 2019-05-21
CubeTexture rotation

### LINKED ISSUE #16328: CubeTexture orientation
In my app, I would like to rotate the CubeTexture that defines the scene background and the environment map reflections by 180° along the vertical y-axis.
My rotation is defined by the VRML and X3D specifications about the background fields.

I tried to modify several parameters like the `mapping`, `flipY` and `rotation` Texture fields. However I didn't found any good way to do so.

- Is there some obvious way to do so that I missed?
- Any chance to see this coming in `three.js`?
- If I would like to implement it, what parameters would you like to see (for example `Texture.flipX` and `CubeTexture.flipZ`)? Would you be interested in this feature?

fabienrohrer:
Fix https://github.com/mrdoob/three.js/issues/16328

- [x] Implement "CubeTexture.rotation" property (inherited from Texture.rotation)
- [x] Support this property for the Material.envMap
- [x] Support this property for the Background
- [x] Document the modification

### DIFF
--- a/src/renderers/shaders/ShaderChunk/envmap_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/envmap_fragment.glsl.js
@@ -26,7 +26,7 @@ export default /* glsl */`
 
 	#ifdef ENVMAP_TYPE_CUBE
 
-		vec4 envColor = textureCube( envMap, vec3( flipEnvMap * reflectVec.x, reflectVec.yz ) );
+		vec4 envColor = textureCube( envMap, envMapRotation * vec3( flipEnvMap * reflectVec.x, reflectVec.yz ) );
 
 	#elif defined( ENVMAP_TYPE_EQUIREC )
 
--- a/docs/api/en/textures/CubeTexture.html
+++ b/docs/api/en/textures/CubeTexture.html
@@ -9,7 +9,7 @@
 	</head>
 	<body>
 		[page:Texture] &rarr;
-		
+
 		<h1>[name]</h1>
 
 		<p class="desc">Creates a cube texture made up of six images.</p>
@@ -19,7 +19,7 @@
 		<code>
 		var loader = new THREE.CubeTextureLoader();
 		loader.setPath( 'textures/cube/pisa/' );
-		
+
 		var textureCube = loader.load( [
 			'px.png', 'nx.png',
 			'py.png', 'ny.png',
@@ -33,7 +33,7 @@
 
 
 		<h3>[name]( images, mapping, wrapS, wrapT, magFilter, minFilter, format, type, anisotropy )</h3>
-		
+
 		<p>
 		CubeTexture is almost equivalent in functionality and usage to [page:Texture]. The only differences are that the
 		images are an array of 6 images as opposed to a single image, and the mapping options are
@@ -43,13 +43,23 @@
 
 		<h2>Properties</h2>
 
-		<h3>See [page:Texture]</h3>
+		<p>
+    See the base [page:Texture Texture] class for common properties.
+    </p>
+
+		<h3>[property:Matrix3 rotation]</h3>
+
+		<p>
+			Rotate the texture.
+			Especially useful to rotate the background and the environment maps.
+			Default is the identity matrix.
+		</p>
 
 		<h2>Methods</h2>
-		
+
 
 		<h3>See [page:Texture]</h3>
-		
+
 		<h2>Source</h2>
 
 		[link:https://github.com/mrdoob/three.js/blob/master/src/[path].js src/[path].js]
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -2004,6 +2004,7 @@ function WebGLRenderer( parameters ) {
 		if ( material.envMap ) {
 
 			uniforms.envMap.value = material.envMap;
+			uniforms.envMapRotation.value = material.envMap.rotation;
 
 			// don't flip CubeTexture envMaps, flip everything else:
 			//  WebGLRenderTargetCube will be flipped for backwards compatibility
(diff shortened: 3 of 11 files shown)


fabienrohrer:
This is my first PR in threejs; tell me if I did things correctly 😄

src/renderers/shaders/ShaderChunk/envmap_fragment.glsl.js
```diff
@@ -26,7 +26,7 @@ export default /* glsl */`
 
 	#ifdef ENVMAP_TYPE_CUBE
 
-		vec4 envColor = textureCube( envMap, vec3( flipEnvMap * reflectVec.x, reflectVec.yz ) );
+		vec4 envColor = textureCube( envMap, envMapRotation * vec3( flipEnvMap * reflectVec.x, reflectVec.yz ) );
```
WestLangley: Every pixel, for every frame? I would not support this -- even if it were wrapped in a #DEFINE.

Correct your envmap once before you use it.

fabienrohrer: > Every pixel, for every frame? I would not support this

I agree that doing this multiplication so often is probably overkilled. Rotating the background at any time is certainly not very useful.

> even if it were wrapped in a #DEFINE.

I don't understand why you would not accept this in a DEFINE. Could you explain me your concern?

> Correct your envmap once before you use it.

I implemented this workaround since a while and it's working.
But I'm not satisfied by this because it increases the complexity of my application and it increases the loading time, either on the server or on the client side. Both are annoying.

I cannot change the background assets in my application because it would break our support of the VRML and X3D standard, and/or the huge user backward compatibility we are trying to support.

Reading at the issue, I'm far to be the only guy having this issue ^^ I think everybody who would like to support some background standard specification which doesn't match with the threejs rigid CubeTexture orientation will have this issue.

---

I would recommend you to accept a DEFINE mechanism at least, and I can still spend time to implement it. But if you don't agree, I don't want to use more time in this direction ;-) Please keep me in touch!

WestLangley: Even with a DEFINE you would be rotating the reflection vector every pixel, every frame.

>Correct your envmap once before you use it.
>>I implemented this workaround since a while and it's working.

Great!

fabienrohrer:
Unfortunately, @WestLangley doesn't want to merge this PR.

WestLangley:
@fabienrohrer I've changed my mind on this.

I still think swapping the cube faces makes the most sense, but if you are interested in pursuing this again, I won't voice objections -- especially if it can be implemented in a manner that doesn't burden everyone to accommodate a few.

Apologies that I didn't come to this conclusion sooner -- especially given that it was your first three.js PR.

fabienrohrer:
No need to apologize, I easily understand your choice, I also maintain several open source projects ;-)

However, I'm not sure when I will be able to dig again in this topic.
Maybe that having a review of an expert at this stage would be welcome, in order to know if I didn't miss something.

Mugen87:
@fabienrohrer Since the original branch is deleted, I think it's not possible the reopen the PR. Anyway, it's probably better to make a new PR based on the latest `dev` branch. We can then checkout your repo and validate the changes locally. 

> However, I'm not sure when I will be able to dig again in this topic.

No problem. Since there is no time pressure, feel to do this as soon as you have the time for. The PR was closed for several month so I don't think time matters here.

looeee:
@fabienrohrer do you intend to reopen this PR? If not, do you mind if I do it? I would like to see this feature added.

fabienrohrer:
Please do it. My interest in this topic is now close to 0, unfortunately.

looeee:
Hmm, I realised when implementing this that it's not going to work with PMREM generation, or at least it would be a lot more work. Since I always use a PMREM now my interest in this has also dropped off. Unless there's a way to rotate a PMREM...

unphased:
Messing in the shader seems to me like a poor way to go about this. The most common use case for this might be to mitigate the problem when one tries to use a cube map background when you have Z-up cubemap textures and want to use it in "Y-up" three.js... We might be able to rotate the camera and set Object3D's default up to 0,0,1, but cube maps come up rotated 90 degrees on the X axis. (I'm stating all this context, in case I'm misunderstanding this PR & discussion) 

Wouldn't we want to just do the rotation in the CubeTextureLoader? the API could allow you to specify being in this Z-up cube texture situation, and it could also allow you to specify a specific rotation (provide a multiple of 90 degrees and an axis) to perform. It'll do the required manipulations and then the cubetex you get back will be what you want. 

Ah, ok I think I know what's going on here. Maybe the above is what @WestLangley had in mind and why the PR was originally rejected. Then, like I just realized, allowing for an arbitrary uniform rotation settable on the background could be a useful effect, as an alternative technique to rotating the camera, and you could achieve effects with this that you couldn't easily do otherwise so dynamic rotation would have totally legit uses after all... And that path runs into PMREM support issues. 

I do have one actual comment though. Why would we need to touch the fragment shader at all? Why can't a rotation like this on the environment be done with the vertex shader only, and then there'd be very little compute overhead.

src/renderers/shaders/ShaderChunk/envmap_fragment.glsl.js
skrat: @WestLangley yeah, every pixel every frame with a DEFINE to make it a user's choice, not yours, or propose more efficient solution, or just don't comment.

mqnc:
> Messing in the shader seems to me like a poor way to go about this. The most common use case for this might be to mitigate the problem when one tries to use a cube map background when you have Z-up cubemap textures and want to use it in "Y-up" three.js... We might be able to rotate the camera and set Object3D's default up to 0,0,1, but cube maps come up rotated 90 degrees on the X axis. (I'm stating all this context, in case I'm misunderstanding this PR & discussion)

This is exactly my use case... I am working with a 2d physics engine and I want to map xy to xy, not to xz. Is there any update on this?

Being able to rotate the background in real-time would also be interesting...

nkallen:
Just FYI -- being able to rotate an HDRI at any point is a common feature of 3d modeling and texturing programs (e.g., quixel mixer) and is used to check the shape of objects. It's definitely a useful feature, independent of dealing with the y-up / z-up camera issues

PR #16511, opened 2019-05-21
Update webgl_lines_sphere.html

melissaanzelmonano:


WestLangley:
I assume this PR was filed in error. It has no effect unless the renderer is instantiated with `alpha: true`.

PR #16519, opened 2019-05-22
JSM: Added modules and TS files for most shaders.

Mugen87:
That one was easy, just a bit time-consuming^^.


mrdoob:
Awesome job! Thanks!

PR #16528, opened 2019-05-23
Object3D: added attach() method

### LINKED ISSUE #6426: THREE Group and Object3D add child BUG.

WestLangley:
`Object3D.attach( object )` adds object as a child of this, while maintaining the object's world transform.

Fixes #6426 and replaces #16380.

This method alleviates the need for `THREE.SceneUtils.attach/detach()`, which can be deprecated in a later PR.
```js
newParent.attach( object ); // detaches from oldParent...

oldParent.attach( object ); // ...now it's back where it started
```

### DIFF
--- a/docs/api/en/core/Object3D.html
+++ b/docs/api/en/core/Object3D.html
@@ -217,6 +217,9 @@
 		<h3>[method:this applyQuaternion]( [param:Quaternion quaternion] )</h3>
 		<p>Applies the rotation represented by the quaternion to the object.</p>
 
+		<h3>[method:this attach]( [param:Object3D object] )</h3>
+		<p>Adds *object* as a child of this, while maintaining the object's world transform.</p>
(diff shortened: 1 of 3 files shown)


mrdoob:
Thanks!

PR #16529, opened 2019-05-23
Editor: Add Points

Temdog007:
See #16234


Mugen87:
Since the editor is now module based (#18201), a lot of code has changed. It's better to implement this feature from scratch.

Issue #16524, opened 2019-05-22
BasisTextureLoader: Next steps


donmccurdy:
Initial support for .basis textures added in https://github.com/mrdoob/three.js/pull/16522. This issue tracks remaining cleanup and planned enhancements. I'm not working on these yet, and will update this issue when I begin, so PRs are welcome in the meantime:

- [x] Clean up remaining TODOs in the code
- [x] Apply eslint fixes
- [x] Add documentation
- [x] Add example
- [x] Add `setMaxWorkers()` method
- [x] Support mipmaps
- [x] Fix mipmap support in iOS (working at https://github.com/mrdoob/three.js/pull/21131)
- [x] Return texture synchronously from `load()` (added at https://github.com/mrdoob/three.js/pull/21131)
- [x] Support alpha
- [x] Add ES module
- [ ] ~Recompile Basis transcoder with bootstrapping suggested by @austinEng~
- [ ] ~Support user-configurable transcode output format~

takahirox:
I haven't looked into the code in detail yet but could we synchronously return `texture` from `load` as other texture loaders do?

I think it's good for the consistency. And if we don't synchronously return texture user needs to call `material.needsUpdate = true` in callback (if already started animation loop).

```javascript
var material = new XXXMaterial();
textureLoader.load(..., function (texture) {
  material.map = texture;
   // .map is from null to non-null.
   // User needs to call material.needsUpdate = true here if already started animation loop
   // because whether material.map is null or not affects the final shader code.
  material.needsUpdate = true;
});
```

donmccurdy:
I haven't checked yet to confirm that works with `THREE.CompressedTexture`, but if so I agree that would be best. 👍

donmccurdy:
Other cleanup: the properties assigned to the texture are a little arbitrary (left over from Basis demo), like `flipY=false`. And there's an unused `startTime` variable in the worker.

takahirox:
If I'm right mipmaps isn't supported. Transcoder doesn't support? Or loader hasn't just implemented yet?

donmccurdy:
A `.basis` file can contain multiple mipmap levels, yes. I think the transcoder supports it already, but I'm not sure. But BasisTextureLoader doesn't yet.

We should update to the new/smaller version of the transcoder, too: https://github.com/BinomialLLC/basis_universal/pull/7

austinEng:
[...]

PR #16553, opened 2019-05-24
Add BasisTextureLoader example

takahirox:
This PR adds `BasisTextureLoader` example.

Live demo: https://raw.githack.com/takahirox/three.js/BasisTextureLoaderExample/examples/index.html#webgl_loader_texture_basis

`kodim20.basis` is from https://github.com/BinomialLLC/basis_universal/tree/master/webgl/texture/assets

I think this example may be too simple. But better than no example. Users can see the loader works and know the API from the example. Please feel free to replace with a cool and/or interesting one later.

#16524


examples/webgl_loader_texture_basis.html
```diff
@@ -0,0 +1,109 @@
...
+
+			init();
+			animate();
+
+			function init() {
+
+				renderer = new THREE.WebGLRenderer( { antialias: true } );
+				renderer.setPixelRatio( window.devicePixelRatio );
+				renderer.setSize( window.innerWidth, window.innerHeight );
+				document.body.appendChild( renderer.domElement );
```
donmccurdy: BasisTextureLoader currently marks all textures as sRGB:

https://github.com/mrdoob/three.js/blob/b39de785004e155e744fc47604de5e973135746b/examples/js/loaders/BasisTextureLoader.js#L144-L151

We should probably remove that, and the `flipY=false` preset, which both made sense in the original Basis example but not so much in a general-purpose texture loader.

My suggestion for this PR would be:

- [ ] Remove flipY=false and encoding=sRGBEncoding from the loader
- [ ] Add encoding=sRGBEncoding after loading the texture (in the example itself)
- [ ] set renderer.gammaOutput=true, renderer.gammaFactor=2.2 as best practice.

donmccurdy:
Good to have simple examples, I think!

WestLangley:
Personally, I would just copy the other `webgl_loader_texture_xxx` examples.

https://raw.githack.com/takahirox/three.js/BasisTextureLoaderExample/examples/index.html?q=hdr#webgl_loader_texture_hdr

takahirox:
Thanks for the reviews.

> Personally, I would just copy the other webgl_loader_texture_xxx examples.

The reason why I didn't copy but made a new example rotating cube is I wanted to show transcode doesn't block main thread.

Reply to the other comments will be tomorrow (or later).

examples/webgl_loader_texture_basis.html
takahirox: > Remove flipY = false from the loader

`flipY = true` doesn't seem to work for compressed texture and default `CompressedTexture.flipY` is `false`. Removing from the loader looks right direction. Updated.

> Move encoding = sRGBEncoding into example

Updated.

> renderer.gammaOutput = true, renderer.gammaFactor=2.2

Thanks, I missed them. Updated.

mrdoob:
Screen Shot 2019-05-27 at 7 37 46 PM

Is this expected?

takahirox:
No. And I can't reproduce here... The root issue seems the server 504 error? Does it still happen even if you reload?

mrdoob:
Weird, I can no longer reproduce...

mrdoob:
Thanks!

Issue #14683, opened 2018-08-09
DirectionalLightHelper and HemisphereLightHelper's orientations are incorrect


robertlong:
The DirectionalLightHelper and HemisphereLightHelper do not take the light's orientation into account.

DirectionalLight:
In this example I have set the DirectionalLight's target to a child of the light and set the light's position and rotation. I have added an AxesHelper to visualize the orientation of the light. The DirectionalLightHelper should be facing the same direction as the light's target.

https://jsfiddle.net/3foLr7sn/80/

HemisphereLight:
In this example I have set the HemisphereLight's position and rotation. I expect the HemisphereLightHelper to reflect the effect of the hemisphere light. So I think I'd expect the yellow side to be facing the sphere. The hemisphere light faces the origin, regardless of rotation, and the HemisphereLightHelper should do the same.

https://jsfiddle.net/3foLr7sn/90/

##### Three.js version

- [x] Dev

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

WestLangley:
`DirectionalLight` and `HemisphereLight` do not use their rotation property.

robertlong:
Right, but the helpers shouldn't either. Take a look at my examples, the helpers do not reflect the direction of the lights.

WestLangley:
Please do not set the `rotation` property of your lights.

Add `light.target` as a child of the scene -- not the light itself.

PR #16555, opened 2019-05-24
Change HemisphereLight orientation to respect quaternion instead of facing origin

DanielSturk:
Currently, HemisphereLights face the origin, but it would be more intuitive/logical for them to simply be oriented using their quaternion. This is also how practically any other software works (orient lights by rotating them, not moving them).

I'm guessing the reason why they face the origin is because the DirectionalLight and SpotLight face their targets, which is a similar concept, but HemisphereLights don't support targets and environment lighting should not be calculated as such.

In practice, the HemisphereLight.quaternion has had no reason to be changed up until now, so I assume most projects leave theirs' unchanged, allowing for the groundColor and [sky]color to still be oriented correctly after this change.


WestLangley:
>Currently, HemisphereLights face the origin

Hemisphere lights do not "face the origin". The position property is used to set the axis of each hemisphere. Granted, it would have been better to use the `.up` property for that, instead.

Mugen87:
In any event, the overall idea of the change does make sense to me. It should be irrelevant where you position an instance of `HemisphereLight`. Its direction should be derived from its orientation and thus `quaternion`. 

@DanielSturk Do you mind resolving the merge conflicts? Please do not include build files into the PR.

Mugen87:
Related #14683

DanielSturk:
@Mugen87 conflicts resolved, build files removed

WestLangley:
Sorry, I do not think the proposal here is the correct API. I appreciate the PR, however, because it gives us the chance to revisit this. :-)

First, an aside...

An _ambient light probe_ is not a _source_ of light -- it is a _probe_ of ambient light; it _measures_ the amount of ambient light -- or indirect light -- at a particular location.

A probe aligns with the coordinate axes, maintaining a fixed frame of reference.

A probe is not rotated. It has a position, however.

// 

A three.js `AmbientLight` is is a special case of a `LightProbe` that has a constant value in every direction. It is not rotated. It has a position. (Currently three.js ignores the position, and assumes the level of ambient light is the same at every location.)

We call it a "light", but that is really not what it is. It is not like a "light bulb".

A `HemisphereLight` is also a special case of a `LightProbe`. It has a simple, two-color model for the level of ambient light. It has a position, and it is not rotated.

Since you want to be able to "tilt" the hemisphere light, then was can use the "up" direction for that purpose, or we can add a new property.

Only directional lights should be controlled by their quaternion -- that is, lights for which `lookAt()` makes sense.

I am hopeful that at some point "targets" will be removed from three.js directional lights in favor of quaternion control.

//

Currently, we use the hemisphere light's `position` property for a purpose for which it was not intended. 

Let's not make the same mistake with the `quaternion` property.

/ping @bhouston

PR #16565, opened 2019-05-26
Remove preventDefault() and console.log from mousedown handlers.

arodic:
It turns out that calling `preventDefault()` on `mousedown` event prevents capture of `mouseup` event on window/document if the pointer is outside the iframe hosting the document. This causes many examples (inside iframe) to break if mouse pointer is released outside the frame/window. For example, examples that use `OrbitControls` turn sticky if you drag and release outside the frame.

I am not sure why so many examples call `preventDefault()` on `mousedown`. In my quick tests everything appears to work fine once I removed the lines with `preventDefault()`.

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -429,15 +429,11 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	function handleMouseDownRotate( event ) {
 
-		//console.log( 'handleMouseDownRotate' );
-
 		rotateStart.set( event.clientX, event.clientY );
 
 	}
 
 	function handleMouseDownDolly( event ) {
-
-		//console.log( 'handleMouseDownDolly' );
 
 		dollyStart.set( event.clientX, event.clientY );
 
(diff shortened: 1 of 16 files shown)


examples/js/controls/OrbitControls.js
```diff
@@ -429,32 +429,24 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	function handleMouseDownRotate( event ) {
 
-		//console.log( 'handleMouseDownRotate' );
```
WestLangley: If you want to propose changes unrelated to the topic of this PR, then please do so in a separate PR. ~12 instances. Thanks!

Mugen87:
Sorry, for the late reply. But in the meantime, I definitely support this change^^. I think we should review the usage of `event.preventDefault()` and `event.stopPropagation()` in context of all other event listeners: https://github.com/mrdoob/three.js/pull/13706#issuecomment-391633631. 

Since I've seen multiple user complains about the status quo, I think this is a valid reason for changing the controls.

> I am not sure why so many examples call preventDefault() on mousedown. 

@arodic It's probably a pattern which was just copied over and over again. Are you willing to resolve the merge conflicts? The controls were somewhat refactored in the last months: `OrthographicTrackballControls` does not exist anymore (`TrackballControls` now supports orthographic cameras)  and `MapControls` is now integrated in `OrbitControls.js`.

Mugen87:
Closing. Let's have a go at this topic with another PR.

PR #16567, opened 2019-05-26
ColladaLoader & TransformControls Typescript Definitions fixes

amiorenstx:
found some bugs with the typescript definitions
here is a fix
keep being awesome :)

### DIFF
--- a/examples/jsm/controls/TransformControls.d.ts
+++ b/examples/jsm/controls/TransformControls.d.ts
@@ -28,7 +28,7 @@ export class TransformControls extends Object3D {
   isTransformControls: boolean;
   visible: boolean;
 
-  attach(object: Object3D): void;
+  attach(object: Object3D): this;
   detach(): void;
   pointerHover(pointer: Object): void;
   pointerDown(pointer: Object): void;
--- a/examples/jsm/loaders/ColladaLoader.d.ts
+++ b/examples/jsm/loaders/ColladaLoader.d.ts
@@ -12,7 +12,7 @@ export interface Collada {
   scene: Scene;
 }
 
-export class Collada {
+export class ColladaLoader {
   constructor(manager?: LoadingManager);
   manager: LoadingManager;
   crossOrigin: string;


examples/jsm/controls/TransformControls.d.ts
```diff
@@ -28,7 +28,7 @@ export class TransformControls extends Object3D {
   isTransformControls: boolean;
   visible: boolean;
 
-  attach(object: Object3D): void;
+  attach(object: Object3D): this;
```
donmccurdy: Arguably it should, but the method doesn't appear to return `this` today:

https://github.com/mrdoob/three.js/blob/93e72ba7b24958ddb0652bd33171edd14ed2d693/examples/js/controls/TransformControls.js#L140-L155

amiorenstx: please look at dev branch 
at TransformControls in the jsm directory (not js)
it is different :/

amiorenstx: i will provide a link

amiorenstx: Ok. The real issue is that Transform Controls extends Object3d
and Object3d returns "this" in that method:
https://github.com/mrdoob/three.js/blob/f86fe1a8ec00c56fd5ffe88a6af3357e30b78fa5/src/core/Object3D.d.ts#L293

Mugen87: This method in `Object3D` was added lately ,see https://github.com/mrdoob/three.js/pull/16528.

Mugen87: It would be bad if this issue blocks the other change for `ColladaLoader`. How about making two PRs? This one focuses just on `TransformControls` and a new one for `ColladaLoader`(which can be immediately merged).

amiorenstx:
I will split issue into 2 different issues as suggested by Mugen87

PR #11349, opened 2017-05-16
Added linewidth support

WestLangley:
This PR adds support for "fat" lines. The lines are rendered using instancing, and line segments are rendered in screen space with rounded ends.

`LineSegmentsGeometry` is a new class which extends `InstancedBufferGeometry`. 

Regardless of the complexity of the line, only 6 faces are pushed to the GPU. The additional data pushed to the GPU is an array of start-end xyz-pairs -- similar to how line segments are specified currently. Instancing makes implementation straight-forward.

`LineMaterial` is also added. It renders lines in screen-space. The relevant parameter is `lineWidth`, which is specified in _pixels_. A rendered line segment looks like this.

screen-space line segment

The difficult part in implementing `LineMaterial` was getting the screen-space rendering correct. But once this was done, segments chain together quite nicely.

I did not implement perspective size attenuation. With that feature, lines close to the camera can be as wide as the screen, and they look terrible. IMHO, `linewidth` should be specified in pixels -- not in world units -- so perspective size attenuation does not make sense.

Line-joins are round. I think this is preferable to mitering the line-joins; miters are problematic when line segments are nearly parallel.

I have stubbed out several new classes. We can see if they make sense. Due to name conflicts, I added "2" to some of the names for now. This entire API is up for debate.

New classes:
```
THREE.LineSegments2 // extends Mesh
THREE.Line2	    // extends LineSegments2
THREE.Wireframe     // extends LineSegments2
```
New geometries:
```
THREE.LineSegmentsGeometry // extends InstancedBufferGeometry
THREE.LineGeometry         // extends LineSegmentsGeometry
THREE.WireframeGeometry2   // extends LineSegmentsGeometry
```
New material:
```
THREE.LineMaterial // extends ShaderMaterial
```
`THREE.LineMaterial` has not been "wired-in" as a built-in material yet.

fat line

In the demo, I have added an inset viewport to study the effects of the `resolution` parameter and the effects of antialiasing.

---

edit: live demo removed

PR #16570, opened 2019-05-27
Specify opacity in LineMaterial object properties

EmperorJack:
**Change**

When using `LineMaterial` I would like to specify an opacity value to create transparent lines. Currently, you must dive into the uniforms to set the opacity like this:
```
// To initialise a LineMaterial with an opacity value
var matLine = new THREE.LineMaterial({
  color: 0xff0000,
  transparent: true,
  linewidth: 5
});
matLine.uniforms.opacity.value = 0.5;

// To update a LineMaterial's opacity
matLine.uniforms.opacity.value = 0.25;
```

This pull request changes `LineMaterial` so that a more direct approach can be used to set the opacity:
```
// To initialise a LineMaterial with an opacity value
var matLine = new THREE.LineMaterial({
  color: 0xff0000,
  transparent: true,
  opacity: 0.5,
  linewidth: 5
});

// To update a LineMaterial's opacity
matLine.opacity = 0.25;
```

This change makes `LineMaterial` more consistent with existing materials, despite not being properly "wired-in" as a built-in material yet (https://github.com/mrdoob/three.js/pull/11349#issue-120934417).

**Testing**

To test this change you can modify the `webgl_lines_fat_wireframe.html` example. E.g: Add `transparent: true` and `opacity: 0.1` to `matLine` which is defined on line 102.

### DIFF
--- a/examples/js/lines/LineMaterial.js
+++ b/examples/js/lines/LineMaterial.js
@@ -3,6 +3,7 @@
  *
  * parameters = {
  *  color: <hex>,
+ *  opacity: <float>,
  *  linewidth: <float>,
  *  dashed: <boolean>,
  *  dashScale: <float>,
@@ -267,6 +268,24 @@ THREE.LineMaterial = function ( parameters ) {
 			set: function ( value ) {
 
 				this.uniforms.diffuse.value = value;
+
+			}
+
+		},
+
+		opacity: {
+
+			enumerable: true,
+
+			get: function () {
+
+				return this.uniforms.opacity.value;
+
+			},
+
+			set: function ( value ) {
+
+				this.uniforms.opacity.value = value;
 
 			}
 


WestLangley:
Thank you for your contribution. I chose not to support opacity for fat lines because they have artifacts when rendered with transparency -- and that invariably leads to user complaints.

Screen Shot 2019-05-27 at 10 12 20 AM

It can be improved somewhat by not writing to the depth buffer.

Screen Shot 2019-05-27 at 10 35 12 AM

WestLangley:
In other words, as it stands now, I would not be inclined to add opacity support.

Can you please explain your use-case?

EmperorJack:
Thank you for your reply. I understand your concern with the artifacts. My use-case is a data visualisation composed of many flat shaded or wireframe polygons. The visualisation must work on different platforms/browsers and looks best when using lines that are both fat and transparent (better readability). The artifacts don't affect my use-case much as the joins are obfuscated by other geometry, though, disabling `depthWrite` does improve things a little.

usecase-transparent

Does this use-case make sense or is it clasping at the edges of what `LineMaterial` was intended for? Is there any other solutions for creating fat, transparent lines that work consistently across platforms/browsers. For example, I can't use `LineBasicMaterial` because of this issue: https://github.com/mrdoob/three.js/issues/10357.

Regardless, I can keep using `LineMaterial` using the `uniforms` trick, but it's not as nice to work with, e.g: I can't treat a list of `MeshPhongMaterial` and `LineMaterial`'s in a generic manner if they don't both support opacity.

Also, what is the future plan for `LineMaterial`, will it eventually be built-in? If you opted to support opacity, is it possible that the documentation could mention how `depthWrite` improves the quality?

WestLangley:
Try [THREE.MeshLlne](https://www.npmjs.com/package/three.meshline). It bevels corners, instead.

EmperorJack:
I have attempted to use `MeshLine` but unfortunately it doesn't support `BufferGeometry`, which I need for translating a polygon mesh (e.g: `ConvexGeometry`) into a wireframe. I tried converting my mesh through `EdgesGeometry` and then into regular geometry (using `fromBufferGeometry`) but no luck. I also tried converting through  `WireframeGeometry2` but that didn't work either.

My use-case requires that I can draw the wireframes of meshes, of which `LineMaterial` works nicely with `WireframeGeometry2` and `Wireframe`. I suppose another option would be to generate individual line objects for each edge given I don't care about the corners.

WestLangley:
Are you drawing only the wireframes of meshes?

Mugen87:
Closing since https://github.com/mrdoob/three.js/pull/16570#issuecomment-496235146 explained why transparency is not supported for wide lines. The alternative solutions mentioned in https://github.com/mrdoob/three.js/pull/16570#issuecomment-498485420 are reasonable workarounds.

PR #16585, opened 2019-05-28
Triangle: include isFrontFacing

sciecode:
Includes `isFrontFacing` method to math Triangle. 
Can be a useful test for cutting number of iterations on faces related processes
```
direction.subVectors( target, camera.position );
// ...
if ( triangle.isFrontFacing( direction ) ) process();
```
[live example](https://rawcdn.githack.com/sciecode/three.js/0daa97cd4bfba628176dfd2fc3a070e1cbe58ed5/examples/misc_example.html)

### DIFF
--- a/docs/api/en/math/Triangle.html
+++ b/docs/api/en/math/Triangle.html
@@ -121,6 +121,13 @@
 		Determines whether or not this triangle intersects [page:Box3 box].
 		</p>
 
+		<h3>[method:Boolean isFrontFacing]( [param:Vector3 direction] )</h3>
+		<p>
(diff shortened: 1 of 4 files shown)


WestLangley:
@mrdoob This seems to me to be a reasonable addition to the library.

Mugen87:
It would be super-cool if you could also add a unit test here:

https://github.com/mrdoob/three.js/blob/dev/test/unit/src/math/Triangle.tests.js

sciecode:
There's something weird going on after #16588 changes. It's saying there are conflicts when there's not. Just gonna close this PR and make a new one.

PR #16593, opened 2019-05-28
LDrawLoader: Use renderOrder instead of depthWrite for transparency.

yomboprime:
This sets `Object3D.renderOrder` to a lower value for lines and higher for meshes, so they are rendered in that order, and lines are correctly seen through translucent meshes.

This avoids the use of `depthWrite=true` which is hacky and can lead to artifacts or side effects. Note the rear stop and blinker lights of the car.

Temporary live link: https://raw.githack.com/yomboprime/three.js/ldraw_opacity_order/examples/webgl_loader_ldraw.html

### DIFF
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -469,6 +469,9 @@ THREE.LDrawLoader = ( function () {
 
 		}
 
+		// Draw lines before transparent meshes
+		object3d.renderOrder = elementSize;
+
 		if ( isConditionalSegments ) {
 
 			object3d.isConditionalLine = true;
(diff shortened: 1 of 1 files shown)


WestLangley:
How come the bulldozer cab( behind the windscreen ) looks so different? Which version is most-correct?

gkjohnson:
> How come the bulldozer cab( behind the windscreen ) looks so different?

image image

_left: depthWrite=false, right: depthWrite=true_

This is one of the artifacts of writing depth for transparent objects -- in this case I would assume that the triangles on the front of the part are drawn first and then the triangles associated with the cylinder in the middle of the part are drawn second and are discarded due to failing the depth test. If the cylinder triangles had been drawn first before the front ones then they would be visible. This can change depending on view direction because of the different order in which triangles could be drawn. If depth write is false then they will still be drawn but blended in the wrong order.

I'm not sure if I see one approach as more correct -- I feel it's just different pros and cons.

yomboprime:
Yes, both of you are right. I created more artifacts than I solved.

Closing.

PR #16598, opened 2019-05-29
LDrawLoader: Let use parse()

yomboprime:
In b08bdc14a8ca74b88af71e6f0235ccf831563e8f the possibility of using `LDrawLoader.parse()` directly as in other loaders was removed (the use case is where file content is provided by the user via other mechanisms).

This PR brings back the use of `parse()`, though asynchronously, not sync as in other loaders.

It only makes sense for packed LDraw files.

The big chunk of code moved is the function `processObject()`, which I've extracted from inside `load()`

### DIFF
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -593,312 +593,17 @@ THREE.LDrawLoader = ( function () {
 			fileLoader.setPath( this.path );
 			fileLoader.load( url, function ( text ) {
 
-				processObject( text, onLoad );
+				scope.processObject( text, onLoad, null, url );
 
 			}, onProgress, onError );
 
-			function processObject( text, onProcessed, subobject ) {
+		},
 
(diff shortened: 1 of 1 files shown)


gkjohnson:
It might be possible to match the `GLTFLoader.parse` signature considering it's also async and loads other files from the network so it takes another path:

https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L121

```js
		parse: function ( data, path, onLoad, onError ) {
```

yomboprime:
> It might be possible to match the `GLTFLoader.parse` signature considering it's also async and loads other files from the network so it takes a path:
> 
> https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js#L121
> 
> ```js
> 		parse: function ( data, path, onLoad, onError ) {
> ```

Done. 👌

mrdoob:
Thanks!

PR #16605, opened 2019-05-29
Wrong default value

luruke:
The doc says the value by default is `true`, but it's actually `false`.


Mugen87:
It was not changed because of #16371

PR #16607, opened 2019-05-29
LDrawLoader: Implement construction STEP directive

yomboprime:
The STEP directive of the LDraw format is used to separate the objects into consecutive 'drawing steps', or 'construction steps', which are stages in the manual construction of the model with real pieces. This PR adds implementation of construction steps.

For this to work, the `Separate` option must be checked in the gui.

Each `THREE.Group` inside the model is assigned a `.userData.constructionStep`, with a value from 0 to `numConstructionSteps - 1`
The total number of steps is stored in `rootGroup.userData.numConstructionSteps`.

The example has a slider in the gui (when `Separate` is enabled) that lets you change the displayed construction step. It simply changes the visibility of the model parts.

Roughly half of the model files have STEP directives. As Separate option is slow, you can disable Smooth option to speed up the loading, specially with the big AT-ST.

Edit: If  #16598 is merged before this, I will redo this PR.

/ping @gkjohnson

### DIFF
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -593,7 +593,13 @@ THREE.LDrawLoader = ( function () {
 			fileLoader.setPath( this.path );
 			fileLoader.load( url, function ( text ) {
 
-				processObject( text, onLoad );
+				processObject( text, function onMainObjectProcessed( model ) {
+
+					scope.computeConstructionSteps( model );
+
+					onLoad( model );
+
+				} );
 
 			}, onProgress, onError );
 
@@ -610,7 +616,7 @@ THREE.LDrawLoader = ( function () {
 					parseScope.currentMatrix.multiplyMatrices( parentParseScope.currentMatrix, subobject.matrix );
 					parseScope.matrix.copy( subobject.matrix );
 					parseScope.inverted = subobject.inverted;
-					parseScope.constructionStep = subobject.constructionStep;
+					parseScope.startingConstructionStep = subobject.startingConstructionStep;
 
 				}
 
@@ -980,7 +986,8 @@ THREE.LDrawLoader = ( function () {
 				lineSegments: null,
 				conditionalSegments: null,
 
-				constructionStep: false
+				// If true, this object is the start of a construction step
+				startingConstructionStep: false
 
 			};
 
@@ -1331,6 +1338,33 @@ THREE.LDrawLoader = ( function () {
 
 		},
 
+		computeConstructionSteps: function ( model ) {
+
+			// Sets userdata.constructionStep number for Group objects and userData.numConstructionSteps number for the root Group object.
+
+			var stepNumber = 0;
+
+			model.traverse( c => {
+
+				if ( c.isGroup ) {
+
+					if ( c.userData.startingConstructionStep ) {
+
+						stepNumber ++;
+
+					}
+
+					c.userData.constructionStep = stepNumber;
+
+				}
+
+			} );
+
+			model.userData.numConstructionSteps = stepNumber + 1;
+
+		},
+
+
 		//
 
 		parse: function ( text ) {
@@ -1381,7 +1415,7 @@ THREE.LDrawLoader = ( function () {
 			var bfcCull = true;
 			var type = '';
 
-			var currentConstructionStep = false;
+			var startingConstructionStep = false;
 
 			var scope = this;
 			function parseColourCode( lineParser, forEdge ) {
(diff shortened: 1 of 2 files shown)


gkjohnson:
Love it! It seems like there are some steps missing, though -- the last step of the bulldozer model causes the whole top of the vehicle to pop in and the AT-ST model only has 35 steps with large chunks popping in at once.

Logging the step numbers of all the parts it looks like there are multiple of the same step number (particularly a lot of step `0`):

```js
model.traverse(c => console.log(c.userData.constructionStep))
```

I think what needs to happen is the construction steps should to be tracked globally instead of per subobject / file. And because all the sub files are parsed sequentially it should be a bit easier to do.

Another option would be to iterate over the full model once everything is loaded and "correct" the step numbers by traversing the hierarchy so every part has a globally-adjusted step number. That would make it robust to changes later if modify the loader so multiple parts can be loaded and parsed at once to speed things up when loading from the network.

yomboprime:
> Another option would be to iterate over the full model once everything is loaded and "correct" the step numbers

Thanks a lot! I've implemented this suggestion and the steps are perfect now. Even better, now almost all of the models have constructions steps, including the X-Wing.

Doing the commit right now...

examples/webgl_loader_ldraw.html
```diff
@@ -112,68 +113,58 @@
...
+				// load materials and then the model
 
-				} );
+				reloadObject( true );
 
-				gui.add( guiData, 'separateObjects' ).name( 'Separate Objects' ).onChange( function ( value ) {
+			}
 
-					reloadObject( false );
+			function processConstructionSteps( model ) {
```
gkjohnson: I like this approach but I wonder if processing the construction steps is better done inside the loader right after the last part is finished loading? And then the total number of steps can go on the root nodes `userData` field?

examples/js/loaders/LDrawLoader.js
```diff
@@ -1375,6 +1381,8 @@ THREE.LDrawLoader = ( function () {
 			var bfcCull = true;
 			var type = '';
 
+			var currentConstructionStep = false;
```
gkjohnson: Would this variable better be called `completedConstructionStep` to indicate that the next subobject represents the beginning of a new "step"? It seems that `0 STEP` indicates the "end" of a construction step spec. Just a thought! I was trying to figure out exactly what this meant when reading through the code.

gkjohnson:
It's beautiful!

examples/webgl_loader_ldraw.html
yomboprime: Yes, that seems reasonable. Thanks.

examples/js/loaders/LDrawLoader.js
yomboprime: It is more like a separator. For example, it can be optionally put at the start and at the end of the geometry of an object, and those two STEP are ignored.

The boolean flag on an object indicates that this object is the start of a step.

So instead of `completedConstructionStep`, what about `startingConstructionStep`? Or something similar.

examples/webgl_loader_ldraw.html
yomboprime: I think I'll expose the function `processConstructionSteps` and put the call to it  in `LDrawLoader.load`.

Have you seen my other PR about using `LDrawLoader.parse()` standalone?

yomboprime:
> It's beautiful!

Yes, it is pleasant for the eyes. I suggest you guys to try the live link, activate Separate option and move the Construction slider:

 https://raw.githack.com/yomboprime/three.js/ldraw_step/examples/webgl_loader_ldraw.html

examples/js/loaders/LDrawLoader.js
yomboprime: I've made a commit renaming to `startingConstructionStep`. Let me know.

examples/webgl_loader_ldraw.html
yomboprime: I've made a commit with your suggestion and renamed `processConstructionSteps` to `computeConstructionSteps`.

gkjohnson: Looks good to me! Presumably the call to `computeConstructionSteps` would happen in `parse`, as well, once your other PR is merged?

examples/js/loaders/LDrawLoader.js
gkjohnson: Looks good!

examples/webgl_loader_ldraw.html
yomboprime: You're right! I've made a commit to move the call to `finalizeObject` so it is done in only one place.

yomboprime:
Continues in #16624, closing this one.

Issue #9606, opened 2016-09-01
Suggestion: Support animating multiple SkinnedMeshes using a single Skeleton


fallenoak:
##### Description of the problem 

According to this PR, https://github.com/mrdoob/three.js/pull/4812#issuecomment-43788375, there is ostensibly a way to support animating multiple SkinnedMeshes using a single Skeleton:

> Please see the most recent code. I have re-introduced THREE.Skeleton and decoupled it from THREE.SkinnedMesh. This will allow us to animate multiple meshes with a single skeleton. The mesh-specific bind matrices are stored in THREE.SkinnedMesh and used by the vertex shader to transform to/from "bind space" (world space in the current implementation).

Unfortunately, no docs or examples appear to have been contributed to demonstrate how this might be accomplished, and try as I might, I can't seem to come up with a way to achieve this on my own.

I'd like to suggest either adding an example, updating the docs, or both. Alternatively, in the event that it's not currently possible, does anyone have a suggestion on an approach to make it possible in the future?

##### Three.js version

- [ ] Dev
- [x] r80
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

titansoftime:
I'm right there with you. I am about to start posting some detailed inquiries regarding how to  most efficiently animate multiple meshes (some share geometry, some do not).

satori99:
@titansoftime / @fallenoak I have achieved this (multiple meshes sharing same skeleton) using the current dev branch, but its a bit of a pain to do with `SkinnedMesh` and `BufferGeometry` currently.  First, I had to modify the `ObjectLoader` to load `SkinnedMesh`s without bones from JSON properly.

Then, I used a Group instance as a stand-in for an armature root, with multiple skinned mesh children. And I stored the shared bones in a custom `userData` property in the JSON, so it survives the loading process.

You will have to handle the `Bone` and `Skeleton` creation manually in your code. These can be attached to the `Group` root too. Then binding that `Skeleton` to each skinned mesh is straightforward. The bone and skeleton creation code is more or less the same as what the `SkinnedMesh` constructor usually does when bones are present on the mesh geometry.

satori99:
[...]

PR #16608, opened 2019-05-29
Added shared skeleton example

funwithtriangles:
Multiple instances of SkinnedMesh all controlled by a single Skeleton. Use the controls to compare performance with a multi skeleton approach.

Related issue: https://github.com/mrdoob/three.js/issues/9606

### DIFF
--- a/examples/files.js
+++ b/examples/files.js
@@ -2,9 +2,10 @@ var files = {
 	"webgl": [
 		"webgl_animation_cloth",
 		"webgl_animation_keyframes",
+		"webgl_animation_multiple",
+		"webgl_animation_sharedskeleton",
 		"webgl_animation_skinning_blending",
 		"webgl_animation_skinning_morph",
-		"webgl_animation_multiple",
 		"webgl_camera",
 		"webgl_camera_array",
 		"webgl_camera_cinematic",
--- a/examples/webgl_animation_sharedskeleton.html
+++ b/examples/webgl_animation_sharedskeleton.html
@@ -44,7 +44,7 @@
 		</div>
 
 		<div id="explanation">
-			Multiple instances of SkinnedMesh all controlled by a single Skeleton. Use the controls to compare performance with a multi skeleton approach.
+			Multiple instances of SkinnedMesh all controlled by a single Skeleton. Use the controls to compare performance with a multi skeleton approach. You may need to increase the rows before the performance gains are noticeable.
 		</div>
 
 		<script src="../build/three.js"></script>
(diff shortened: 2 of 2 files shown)


WestLangley:
Temporary live link:

https://raw.githack.com/funwithtriangles/three.js/dev/examples/webgl_animation_sharedskeleton.html

titansoftime:
Very impressive. Now since each mesh has its own mixer, is it possible for the animations to be out of sync or even play different animations simultaneously?

Even if not, your approach would be a huge help in a variety of scenarios. Nice.

funwithtriangles:
Where you see each mesh having its own mixer is just for the multi skeleton example to show the performance difference. With the shared skeleton it is just one mixer. The only way I can see to have a bit of variation is to increase the number of shared skeletons and assign them randomly throughout the meshes.

funwithtriangles:
Wondering if the example should be slightly simpler in terms of the GPU side of things. It runs nice and smooth on my machine but I fear it's going to be quite hellish for slower machines. Perhaps start with less dancers and maybe simpler materials?

looeee:
This is cool, but the example should start with less rows (probaly just one). Attempting to run it on my Moto G6, Firefox just crashes and Chrome only displays a black screen.

funwithtriangles:
Yep, I think I'm a little spoilt running on my high end gaming laptop. 😅 Will get around to doing some tweaks to this.

titansoftime:
> Where you see each mesh having its own mixer is just for the multi skeleton example to show the performance difference.

Ah. Gotcha.

> The only way I can see to have a bit of variation is to increase the number of shared skeletons and assign them randomly throughout the meshes.

Makes sense.

funwithtriangles:
@looeee I've massively simplified the demo. Only using one part of the model to reduce geometries and have swapped out the material for `MeshNormalMaterial`. I think this makes sense anyway as the performance gain is really only a CPU one so it's good to keep the GPU part simple. Also the default setting is now just one row of dancers.

It now works on mobile but interestingly I can't see any performance gains at all between having the skeleton shared or duplicated. I'm guessing the bottleneck for most mobile devices is still on the GPU so we don't see any gains. I guess finding a model with the same number of bones but a simpler mesh would be a good way to test this theory.

FlorentMasson:
Thanks, this example was really useful and should be merged 😉

jpweeks:
I've found the use of `detached` mode to be pretty confusing and needing an example like this.

Mugen87:
Closing in favor of #26995.

PR #16614, opened 2019-05-30
JSM: Added module and TS files for fat lines.

Mugen87:
Also resolved some minor linter issues in the original JS files.


WestLangley (approved):
Thanks, @Mugen87  !

mrdoob:
Thanks!

PR #16637, opened 2019-05-31
refined large screen layout & type scale

yuinchien:
Thanks for all the feedback on the design update. I understand people have different preferences and habit of using the docs, this is the latest update to address the responses from users. Changes:

– Refined type scale (switched back to px from rem to make sure numbers are always integers for diff screen sizes)
– Refined large screen layout to improve readability

Change takes time, I hope people can give the new design some time to get used to. PR is welcome down the line to continue improvements!

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -39,9 +39,7 @@
 
 		</div>
 
-		<div id="iframe-wrapper">
-			<iframe name="viewer"></iframe>
-		</div>
+		<iframe name="viewer"></iframe>
 
 		<script src="list.js"></script>
 
@@ -428,8 +426,9 @@
 
 				}
 
-				document.getElementById('iframe-wrapper').replaceChild( iframe, oldIframe );
+				document.body.replaceChild( iframe, oldIframe );
 				document.title = subtitle + 'three.js docs';
+
 
 			}
 
--- a/docs/index.html
+++ b/docs/index.html
@@ -39,7 +39,9 @@
 
 		</div>
 
-		<iframe name="viewer"></iframe>
+		<div id="iframe-wrapper">
+			<iframe name="viewer"></iframe>
+		</div>
 
 		<script src="list.js"></script>
 
@@ -426,9 +428,8 @@
 
 				}
 
-				document.body.replaceChild( iframe, oldIframe );
+				document.getElementById('iframe-wrapper').replaceChild( iframe, oldIframe );
 				document.title = subtitle + 'three.js docs';
-
 
 			}
 
--- a/docs/page.css
+++ b/docs/page.css
@@ -2,18 +2,14 @@
 	--color-blue: #049EF4;
 	--text-color: #444;
 
-	--font-size: 14px;
+	--font-size: 16px;
 	--line-height: 24px;
-	--font-size-sanserif: 1.14285714286rem;
-	--line-height-sanserif: 1.71428571rem;
 
 	--border-style: 1px solid #E8E8E8;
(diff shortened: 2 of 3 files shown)


gkjohnson:
https://raw.githack.com/mrdoob/three.js/refine-design/docs/#api/en/core/Object3D

looeee:
Getting there! 

I think that this is perfect when viewed on my phone. Unfortunately, I never use the three.js docs on my phone, and it's still not quite right, to me at least, on bigger screens. Mainly there's not enough differentiation between heading sizes/fonts and the spacing is still a bit off, making it hard to scan quickly. 

**Left**: this PR. 

**Right**: Browser defaults for `margin-top`, `margin-bottom`, `font-size`, `line-height`, `font-weight`, with a bit of extra `<p.margin-bottom>` . 

comp

The main differences that make the defaults easier to scan, are the reduced spacing in the menu, the bold fonts in the `<h1>`, and the more consistent decrease in size and spacing between heading levels. 

Things I'd change from browser defaults: Heading could be increased in size a bit, perhaps a bit less space between `<h3>` and content, and slightly increase `<p.line-height>`.

looeee:
By the way, I just opened the current [Object3D doc page](https://threejs.org/docs/#api/en/core/Object3D) fullscreen, and found myself instinctively stepping back from my screen so that I could see it properly. 

After this PR I no longer feel the need to do that, so this is already a big improvement.

yuinchien:
There's an issue on iOS for setting height for iframe, which results in the header doesn't lock at the top. https://stackoverflow.com/questions/34320046/iframe-height-issues-on-ios-mobile-safari

An *iframe-wrapper* was added to fix the height issue. You can test the latest link on iOS. 
https://raw.githack.com/mrdoob/three.js/refine-design/docs/#manual/en/introduction/Creating-a-scene

Scrolling the panel is now smooth on iOS too. The new main.css includes smooth scrolling for the landing page as well.

mrdoob:
In my system...

Before:

Screen Shot 2019-06-20 at 11 20 00 AM

After:

Screen Shot 2019-06-20 at 11 20 20 AM

Notice the double scrollbar.

After many years on MacOS and continuously being bitten by the issue of MacOS blindfolding me of scrollbars being there with its auto-hide feature I decided to force-enable scrollbars always.

Screen Shot 2019-06-20 at 11 24 01 AM

yuinchien:
@mrdoob Does the double scrollbar occur because of this change https://github.com/mrdoob/three.js/pull/16637/commits/d1474e8794eacbf6e2ad6194b55dabd3d5d56629? Because on iOS setting height for iFrame is buggy, i added a wrapper in order to fix the header at the top. The double scrollbar is not good, we can revert this commit.

WestLangley:
For a separate PR, in the examples tab, flipping from portrait to landscape on an iPhone after the page has loaded results in all sorts of problems.

mrdoob:
@yuinchien that seems to be the case:

https://raw.githack.com/mrdoob/three.js/40783c3447203b876b317519bf675355ca44f884/docs/index.html#api/en/core/Object3D

Screen Shot 2019-06-20 at 1 06 34 PM

mrdoob:
@WestLangley Uh oh!

WestLangley:
>@WestLangley Uh oh!

I forgot to mention... it is looking much better. :-)

mrdoob:
Double scrollbar is now gone! 🙌

mrdoob:
Thanks!

PR #16655, opened 2019-06-02
THREE.MathUtils: added seededRandom() method

WestLangley:
This is useful in generating repeatable results in an existing example.

```js
THREE.Math.prng( 123 );

Math.random = THREE.Math.prng;
```

### DIFF
--- a/docs/api/en/math/Math.html
+++ b/docs/api/en/math/Math.html
@@ -83,6 +83,9 @@
 		<h3>[method:Integer randInt]( [param:Integer low], [param:Integer high] )</h3>
 		<p>Random integer in the interval [page:Float low] to [page:Float high].</p>
 
+		<h3>[method:Float prng]( [param:Integer seed] )</h3>
+		<p>Deterministic pseudo-random float in the interval [ 0, 1 ]. The integer [page:Integer seed] is optional.</p>
+
 		<h3>[method:Float smoothstep]( [param:Float x], [param:Float min], [param:Float max] )</h3>
 		<p>
 		[page:Float x] - The value to evaluate based on its position between min and max. <br />
--- a/docs/api/en/math/MathUtils.html
+++ b/docs/api/en/math/MathUtils.html
@@ -83,6 +83,9 @@
 		<h3>[method:Integer randInt]( [param:Integer low], [param:Integer high] )</h3>
 		<p>Random integer in the interval [page:Float low] to [page:Float high].</p>
 
+		<h3>[method:Float prng]( [param:Integer seed] )</h3>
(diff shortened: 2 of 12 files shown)


Mugen87:
This code was removed from the examples here: #16576

And now you want to add it to the core?

WestLangley:
@Mugen87 If you don't want users of the library to have access to this method, I'll close this.

Mugen87:
No offense, I'm just irritated by the PR. I think we made a decision in #16576 to not add such code to the lib...

WestLangley:
>No offense, I'm just irritated by the PR

Please don't be. I was not intending to annoy you. I thought I was doing what you suggested when you said:

>Okay, if @mrdoob wants to keep PRNG, I suggest to move it into the THREE namespace and the file to the math directory.

But I guess you were suggesting to move it to `math/js/examples`...

We can do that, merge this, or close this. I don't care.

Mugen87:
Yeah, in this sentence I was referring to `examples/js/math`^^. Anyway, since `PRNG.js` is now gone, I would only add it back if there is an actual user request.

mrdoob:
Seems like this was a misunderstanding. @Mugen87 suggested adding PRNG in the `THREE` namespace so we can generate the [JSM](https://github.com/mrdoob/three.js/tree/dev/examples/jsm) version automatically.

Mugen87:
I actually wanted to remove `PRNG.js` in the first place since it was not used in the repo^^. So I think the status quo is oka ynow.

WestLangley:
Trying again...

mrdoob:
We just need a good use case.

munrocket:
```
var x = crypto.getRandomValues( new Uint32Array(1) )[0] / 0xFFFFFFFF;
```
Should work almost in any browser. Performance only twice worse, because it’s cryptography safe PRNG and supports an arrays.

WestLangley:
Closing again...

mrdoob:
@WestLangley can we call this `MahtUtils.deterministicRandom()`?

mrdoob:
Oh, can't reopen the PR because the branch is gone.

WestLangley:
>can we call this MathUtils.deterministicRandom()?

@mrdoob Maybe `MathUtils.seededRandom()` or `.seedRandom()`?

Similar nomenclature is used here: https://github.com/davidbau/seedrandom

mrdoob:
Hmm, `seedRandom()` is shorter... I have no preference, as long as is not an acronym.

mrdoob:
Actually, I think `seededRandom()` sounds better.

mrdoob:
Thanks!

wrightwriter:
This function expects big numbers, otherwise it results in a very biased distribution.
I think this warrants a comment, explaining this, no?
image
image

PR #16657, opened 2019-06-03
Implement mipmap support in BasisTextureLoader

zeux:
This change makes sure that we load all mipmaps present in the .basis
file. This is important for multiple reasons, one of them being that to
use .basis files in glTF out of the box, the texture has to have a
complete mipchain (otherwise the texture is incomplete as per WebGL spec
and sampling this texture returns (0,0,0)).

Contributes to #16524.


donmccurdy (approved):
Thank you! 🎊

mrdoob:
Thanks!

PR #16674, opened 2019-06-04
DRAFT: getProgramCode to getProgramHashCode

sunag:
This should be a reduction in memory usage since the whole code is stored in a string as a hash.

Still, a benchmark is required to check the memory usage reduction and additional processing used to generate the checksum.

**It is a WIP - not ready for merger**

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1449,7 +1449,7 @@ function WebGLRenderer( parameters ) {
 		var parameters = programCache.getParameters(
 			material, lights.state, shadowsArray, fog, _clipping.numPlanes, _clipping.numIntersection, object );
 
-		var code = programCache.getProgramCode( material, parameters );
+		var checksum = programCache.getProgramChecksum( material, parameters );
 
 		var program = materialProperties.program;
 		var programChange = true;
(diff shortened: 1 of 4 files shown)


Mugen87:
> It is a WIP - not ready for merger

If so, consider to make Draft PRs in the future (e.g. #16678)

src/renderers/WebGLRenderer.js
```diff
@@ -1449,7 +1449,7 @@ function WebGLRenderer( parameters ) {
 		var parameters = programCache.getParameters(
 			material, lights.state, shadowsArray, fog, _clipping.numPlanes, _clipping.numIntersection, object );
 
-		var code = programCache.getProgramCode( material, parameters );
+		var checksum = programCache.getProgramChecksum( material, parameters );
```
WestLangley: The crc is used as a hash, so maybe use this terminology, instead?
```js
var hash = programCache.getProgramHashCode( material, parameters );
```

Mugen87:
@sunag The idea is great but we need a performance test to verify the consequences on memory consumption and processing speed. The performance of `initMaterial()` does matter and should ideally not be worse than before.

Closing for now. Please make a new (non-WIP) PR if you can confirm no performance degradation in `initMaterial()`.

PR #16686, opened 2019-06-05
BasisTextureLoader: Fix example on iOS

donmccurdy:
Transcoding to PVRTC1 requires square power-of-two textures. The transcoder may be able to upscale automatically in the future, but for now we should just use a different texture. I'm not particularly attached to this texture / other suggestions welcome.

Screen Shot 2019-06-04 at 8 25 23 PM

Tested on:
- macOS / Chrome
- macOS / Firefox
- macOS / Safari
- iOS / Safari (iPhone SE, iPad)
- Android / Chrome (Nexus 5X)

I tried a version of the texture including mipmaps, but for some reason that still fails on iOS with the error:

> INVALID_VALUE: compressedTexImage2D: length of ArrayBufferView is not correct for dimensions.


mrdoob:
Thanks!

PR #16696, opened 2019-06-05
Add capability to disable shadow maps per light type

Oletus:
This will save texture uniform slots in case shadow maps are only used for some types of lights but not others.

Right now this only toggles shadow maps in the shader code, same as WebGLShadowMap.enabled. Shadow map related uniforms are kept in the uniform data structures regardless of the settings. To optimize this further in the future, the uniform uploads could be completely disabled as well in case shadow maps are not in use.

A typical use case would be in a scene that has one primary directional light with shadows enabled and a bunch of smaller local point and spot lights without shadows.

### DIFF
--- a/docs/api/en/renderers/WebGLRenderer.html
+++ b/docs/api/en/renderers/WebGLRenderer.html
@@ -248,6 +248,15 @@
 		<h3>[property:Boolean shadowMap.enabled]</h3>
 		<p>If set, use shadow maps in the scene. Default is *false*.</p>
 
+		<h3>[property:Boolean shadowMap.pointLightEnabled]</h3>
+		<p>If set and shadow maps are enabled in general, use shadow maps for point lights. Setting this to *false* makes it possible to have a lot of point lights in the scene while still using shadow maps for other types of lights. Default is *true*.</p>
+
+		<h3>[property:Boolean shadowMap.spotLightEnabled]</h3>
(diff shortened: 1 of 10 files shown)


titansoftime:
This would be nice. I had to implement a texture atlas system to get around it in certain scenarios.

gkjohnson:
It seems heavy handed to turn off the shadow maps for _all_ lights of a certain type to save texture slots and to ask users to disable that manually. What if there are four directional lights and you want just one to cast shadows?

@titansoftime @Oletus can you describe your use cases a bit more and the problems you were running in to?

I would think this could happen automatically and with more precision by doing something like sorting the light uniforms such that the first lights in the array cast shadows and ones at the end do not. Then you might add a new define called `NUM_DIR_LIGHT_SHADOWS` which could be different from `NUM_DIR_LIGHTS`.

Oletus:
@gkjohnson The use case is having a scene with lots of lights, of which just a few directional lights are casting shadows. In this case texture units can run out even if most lights don't have shadow maps enabled.

We thought of controlling this via explicit configuration at first since it was the simplest alternative, but I see that automatic configuration would be superior for usability. So I'm happy to move to a solution where the number of shadow-casting lights in the shader can be different from the number of total lights, and light uniforms are sorted putting shadow-casting lights first. Should I update this PR or make a new one with that kind of a change?

Sometime after that I'd like to see more explicit configuration of lights in shaders though, because it can prevent janks caused by recompiles when the number of lights changes in the scene. But it has to come in a form of specifying the max number of lights in shaders (and possibly shadow maps), instead of disabling features entirely.

Oletus:
Replaced by: https://github.com/mrdoob/three.js/pull/17074

PR #16702, opened 2019-06-06
Add module and TS file for Ocean and OceanShader

looeee:
`OceanShader.js` was adding shaders to `THREE.ShaderLib` which won't work for a module, so I moved the shaders to `THREE.OceanShaders` (similar to how other files in the `shaders/` directory are defined). 

Moved `examples/Ocean.js` to `examples/misc/Ocean.js`.

Also, it seems like the modularize script doesn't work unless the file being converted has a header of the form:

``` js
/*
 Some text here
*/
```

`Ocean.js` had no header, so I added some simple text to make the script happy, and converted the comment style in the `OceanShader.js` header. 

[Ocean2 example after this update](https://raw.githack.com/looeee/three.js/dev/examples/webgl_shaders_ocean2.html)


mrdoob:
Thanks!

PR #16718, opened 2019-06-07
HDRCubeTextureLoader: add setType() method

WestLangley:
With this change, the signature of of the `load()` method now is consistent with that of the other texture loaders.

```js
new THREE.HDRCubeTextureLoader()
	.setType( THREE.UnsignedByteType )
	.load( hdrUrls, function ( hdrCubeMap ) { ...
```

Also, `HDRCubeTextureLoader` now leverages the decoding capabilities of `RGBELoader`. `UnsignedByteType`, `FloatType`, and `HalfFloatType` are supported.

The examples are updated to accommodate the API change.


mrdoob:
Thanks!

PR #16731, opened 2019-06-08
Update Quaternion.html

trusktr:


Mugen87 (approved):

WestLangley:
`onChangeCallback` is not intended to be part of the public API, and for that reason, should not be modified by the user.

>By default [onChangeCallback] is the empty function, however you can change it if needed using onChange( onChangeCallback ).

If needed?

Mugen87:
@WestLangley The PR only improves a single line in the doc page. I think other improvements can go into a different one.

trusktr:
The only value that is passable to `update` that makes any sense is `false`. Otherwise, you can pass `null`, `""`, `NaN`, etc, and it will still call the changed callback.

I think it might be easier to read if it defaults to `true`, and the conditional becomes `if (update)` instead of `if (update !== false)`.

https://github.com/mrdoob/three.js/blob/dev/src/math/Quaternion.js#L268

Mugen87:
> I think it might be easier to read if update defaults to true, and the conditional becomes if (update) instead of if (update !== false). Is there a specific reason for !== false?

Please address this in a different issue/PR 👍

PR #16734, opened 2019-06-08
update incorrect Quaternion type declaration

trusktr:


src/math/Quaternion.d.ts
```diff
@@ -117,15 +117,10 @@ export class Quaternion {
...
 		src1: number[],
-		stcOffset1: number,
+		srcOffset1: number,
 		t: number
-	): Quaternion;
-
-	/**
-   * @deprecated Use {@link Vector#applyQuaternion vector.applyQuaternion( quaternion )} instead.
-   */
-	multiplyVector3( v: any ): any;
```
Mugen87: Why have you removed the declaration of `multiplyVector3()`?

trusktr: It doesn't seem to exist in the implementation: https://github.com/mrdoob/three.js/blob/dev/src/math/Quaternion.js

Mugen87: It's defined in `THREE.Legacy.js`. So the method is still accessible.

Mugen87:
Closing due merge conflicts.

Issue #16688, opened 2019-06-05
Example files that still need to be converted to modules


looeee:
Thanks to the work of @Mugen87 over the last while most of the examples have been made available as modules in the **jsm/** folder.

Here's a list of all the examples that are still not available in the **jsm/** folder. I've included _everything_ for completeness, but we may not have to convert all of them (e.g. files in the **lib/** folder). 

Each file needs to be converted to a module and put in the corresponding directory in the `examples/jsm/` folder. If possible, conversion should be done via the [modularize.js](https://github.com/mrdoob/three.js/blob/dev/utils/modularize.js) script (the example may need to be refactored for this to work). Modules also need a [Typescript definition file](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/OrbitControls.d.ts) to be added in the **jsm/** 
 folder alongside them. 

**-- libs/** 
- [ ]  ammo.js _[repo](https://github.com/kripken/ammo.js), [open issue on ES6 module support](https://github.com/kripken/ammo.js/issues/146), used in `MMDPhysics`, physics examples_
- [ ] ctm.js (`CTMLoader`)
- [ ] jszip.min.js _[repo](http://stuartk.com/jszip), used by `3MFLoader`, `AMFLoader`, `KMZLoader`, `webgl2_materials_texture2darray.html`_
- [ ] lzma.js (`CTMLoader`)
- [ ] opentype.min.js _[repo](https://github.com/opentypejs/opentype.js), used by `TTFLoader`_
- [ ] timeliner_gui.min.js _[repo](https://github.com/tschw/timeliner_gui), used by `TimeLineController`

looeee:
Since the node materials files are already modules, they should probably be moved out of the **js/** folder and into the **jsm/** folder. Then we just need to create typescript files for them.

Mugen87:
> files in the lib/ folder

I looking for a way to provide `inflate` as a module. I'm afraid issue like #16602 will frequently pop up with certain loaders like `FBXLoader`. Unfortunately, it seems there is no module version of [zlib.js](https://github.com/imaya/zlib.js). Could we maybe use another library?

looeee:
Yes, I was thinking we'd have to do that, unfortunately. It's just a standard zip inflate function I think so that shouldn't be too hard.

Mugen87:
[...]

PR #16754, opened 2019-06-10
JSM: Added dat.gui module version.

Mugen87:
And adjusted one of the examples so it uses the new code.

### DIFF
--- a/examples/jsm/libs/dat.gui.module.js
+++ b/examples/jsm/libs/dat.gui.module.js
@@ -1,0 +1,3575 @@
+/**
+ * dat-gui JavaScript Controller Library
+ * http://code.google.com/p/dat-gui
+ *
+ * Copyright 2011 Data Arts Team, Google Creative Lab
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+function ___$insertStyle( css ) {
+
+	if ( ! css ) {
+
+		return;
+
+	}
+	if ( typeof window === 'undefined' ) {
+
+		return;
+
+	}
+
+	var style = document.createElement( 'style' );
+
+	style.setAttribute( 'type', 'text/css' );
+	style.innerHTML = css;
+	document.head.appendChild( style );
+
+	return css;
+
+}
+
+function colorToString( color, forceCSSHex ) {
+
+	var colorFormat = color.__state.conversionName.toString();
+	var r = Math.round( color.r );
+	var g = Math.round( color.g );
+	var b = Math.round( color.b );
+	var a = color.a;
+	var h = Math.round( color.h );
+	var s = color.s.toFixed( 1 );
+	var v = color.v.toFixed( 1 );
+	if ( forceCSSHex || colorFormat === 'THREE_CHAR_HEX' || colorFormat === 'SIX_CHAR_HEX' ) {
+
+		var str = color.hex.toString( 16 );
+		while ( str.length < 6 ) {
+
+			str = '0' + str;
+
+		}
+		return '#' + str;
+
+	} else if ( colorFormat === 'CSS_RGB' ) {
+
+		return 'rgb(' + r + ',' + g + ',' + b + ')';
+
+	} else if ( colorFormat === 'CSS_RGBA' ) {
+
+		return 'rgba(' + r + ',' + g + ',' + b + ',' + a + ')';
+
+	} else if ( colorFormat === 'HEX' ) {
+
+		return '0x' + color.hex.toString( 16 );
+
+	} else if ( colorFormat === 'RGB_ARRAY' ) {
+
+		return '[' + r + ',' + g + ',' + b + ']';
+
+	} else if ( colorFormat === 'RGBA_ARRAY' ) {
+
+		return '[' + r + ',' + g + ',' + b + ',' + a + ']';
+
+	} else if ( colorFormat === 'RGB_OBJ' ) {
+
+		return '{r:' + r + ',g:' + g + ',b:' + b + '}';
+
+	} else if ( colorFormat === 'RGBA_OBJ' ) {
+
+		return '{r:' + r + ',g:' + g + ',b:' + b + ',a:' + a + '}';
+
+	} else if ( colorFormat === 'HSV_OBJ' ) {
+
+		return '{h:' + h + ',s:' + s + ',v:' + v + '}';
+
+	} else if ( colorFormat === 'HSVA_OBJ' ) {
+
+		return '{h:' + h + ',s:' + s + ',v:' + v + ',a:' + a + '}';
+
+	}
+	return 'unknown format';
+
+}
+
+var ARR_EACH = Array.prototype.forEach;
+var ARR_SLICE = Array.prototype.slice;
+var Common = {
+	BREAK: {},
+	extend: function extend( target ) {
+
+		this.each( ARR_SLICE.call( arguments, 1 ), function ( obj ) {
+
+			var keys = this.isObject( obj ) ? Object.keys( obj ) : [];
+			keys.forEach( function ( key ) {
+
+				if ( ! this.isUndefined( obj[ key ] ) ) {
+
+					target[ key ] = obj[ key ];
+
+				}
+
+			}.bind( this ) );
+
+		}, this );
+		return target;
+
+	},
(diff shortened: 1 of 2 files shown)


donmccurdy:
Do we want to have ES module copies of `libs/*` code? Since that isn't maintained here, we'd really never be generating UMD from ES modules, in the way that we plan to with source files. Also note that the dat.gui code in this PR is probably a newer release than the existing copy.

My first reaction would be that maybe `libs/` should remain as-is, and we can consider ES versions once the source files are all converted and stable. other pros/cons to that?

Mugen87:
> Do we want to have ES module copies of libs/* code?

I think it makes definitely sense for stuff like `inflate` since the usage of certain modules like `FBXLoader` becomes easier. However, I don't feel strong about `dat.gui` or `stats`. I just thought it's more clean if the examples import only modules if possible.

Mugen87:
> Also note that the dat.gui code in this PR is probably a newer release than the existing copy.

That's a good point. The current used version is `0.6.2`. However, there is no module version for this relaese, see https://github.com/dataarts/dat.gui/tree/v0.6.2/build

It seems ES6 support was added with [0.7.0](https://github.com/dataarts/dat.gui/tree/v0.7.0/build). The one I have added here is `0.7.6`. Well, it can't hurt to upgrade the library^^. We could do this when changing the HTML files by using the newer module version.

donmccurdy:
>I think it makes definitely sense for stuff like inflate since the usage of certain modules like FBXLoader becomes easier. However, I don't feel strong about dat.gui or stats. 

Ah, yeah that's a good distinction. I have no objection to converting inflate.js. Once we're generating UMD from ES modules, we should remove the old version of inflate.js.

> I just thought it's more clean if the examples import only modules if possible.

I don't think we should be creating ES versions of external libraries that don't already have them (like Draco and Basis), so we can't go 100% on this... not having any more duplicates than necessary would be nice, too.

looeee:
> Ah, yeah that's a good distinction. I have no objection to converting inflate.js. Once we're generating UMD from ES modules, we should remove the old version of inflate.js.

All the libs, except `stats.js`, `dat.GUI`, `tween.js` are consumed by other modules, so they fall in the same category as `inflate.js`. I've added some (possibly incomplete) data to the lib entries in #16688.

Converting `inflate.js` and `gunzip.js` by hand was fine since they are unlikely to ever need to be updated (last update on their repo was 2012). That's not likely to be the case for the other files though. 

 Perhaps we can reach out to the library owners and see what the status of ES6 support is?

Mugen87:
> Perhaps we can reach out to the library owners and see what the status of ES6 support is?

Um, this sounds like a time consuming task. I mean to wait until all relevant projects provide ES6 builds. Besides, they will also provide the latest build as a module and no the version used in `three.js`. Hence, an upgrade would be necessary similar to `dat.gui`.

In the meantime, I think it's more "target-aimed" to manually convert the modules similar to `inflate.js`. Users won't notice them anyway since they are only internally used by the example files.

looeee:
> In the meantime, I think it's more "target-aimed" to manually convert the modules similar to inflate.js. 

I'm fine with doing that, and it's generally fairly easy to do. We should add a comment to the start of each file that we edit this way though.

donmccurdy:
> All the libs, except stats.js, dat.GUI, tween.js are consumed by other modules, so they fall in the same category as inflate.js...

This has considerably more side effects than just converting our source files to ES modules. Consider JSZip (74KB) or Ammo (1.6MB). I don't think either of these should be consumed by other modules, or not as relative dependencies anyway. MMDLoader depends on Ammo, but not all MMD files need it (as far as I know) and certainly not all users want that 1.6MB added to their build. And even for much smaller dependencies, it seems worth thinking twice before we locally edit and republish modules that are already available on NPM.

We can certainly work through all of this, there are several options. But... this seems much lower priority to me. Could we leave unmodified global scripts for `examples/js/libs/*` until we've finished the current conversions, and then figure out what to do with them?

looeee:
> Consider JSZip (74KB) 

There are probably much smaller decompression utilities that we could replace this with, such as [zip.js](https://gildas-lormeau.github.io/zip.js/) or files from [zlib.js](https://github.com/imaya/zlib.js/tree/develop/bin). Performance may be different though.

Ammo.js I agree on, I don't think we should convert that to a module manually. Perhaps we can remove the dependency from `MMDLoader`, or even just postpone dealing with this particular case for the time being.

> And even for much smaller dependencies, it seems worth thinking twice before we locally edit and republish modules that are already available on NPM.

That's a fair point, it could cause complications for users or cause them to have some libs included twice by mistake if they are already using the NPM version. 

> But... this seems much lower priority to me. Could we leave unmodified global scripts for examples/js/libs/* until we've finished the current conversions, and then figure out what to do with them?

That's fine by me, but that's probably going to be in just a few days at the rate @Mugen87 has been converting the files 😅

Aside from that, I already find it awkward in my workflow when I import say, the `FBXLoader` module from the NPM package, and then have to include `inflate.min.js` manually. I don't think we can consider the modularization of a file compete until we can just `import FBXLoader from 'three/examples/jsm/loaders/FBXLoader'` and everything just works. The simplest way to do that is include modular versions of the libs in `jsm/libs`.

takahirox:
Sorry I haven't read though the discussion in this thread but one comment for MMD

> MMDLoader depends on Ammo, but not all MMD files need it (as far as I know) and certainly not all users want that 1.6MB added to their build.

Most of MMD model files include Physics settings and most of MMD users expect Physics works fine because Physics is one of the big features in MMD.

donmccurdy:
> Most of MMD model files include Physics settings and most of MMD users expect Physics works fine because Physics is one of the big features in MMD.

Still, though... if any users are not relying on physics, surely they'd rather not have a 1.6MB physics engine bundled into the loader? Or even if they are, keep the option of lazy-loading the physics part? I don't think AmmoJS should be required directly by the loader.

Or to take another example, GLTFLoader depends on DRACOLoader, which depends on the 300KB Draco decoder module with JS and WASM variants. Those files should not be direct dependencies of GLTFLoader: (1) it's pointless to bundle both the JS and WASM decoders, (2) there's no bundler-agnostic way to load only the appropriate decoder, and (3) not all glTF models need Draco. In this case, providing the dependency is better: `loader.setDRACOLoader( dracoLoader )`.

> That's fine by me, but that's probably going to be in just a few days at the rate @Mugen87 has been converting the files ... Aside from that, I already find it awkward in my workflow when I import say, the FBXLoader module from the NPM package, and then have to include inflate.min.js manually.  ... The simplest way to do that is include modular versions of the libs in jsm/libs.

I don't think it is that easy or simple, unfortunately. We haven't even gotten to the more important part of compiling ES modules to backward-compatible UMD files. I agree we need to solve the dependencies too (without relying on global names) but my hunch is that several existing modules would benefit from the `setFoo( fooModule )` pattern.

looeee:
> We haven't even gotten to the more important part of compiling ES modules to backward-compatible UMD files

True, but isn't figuring out how to handle the dependencies something we need to do before we proceed to that stage?

> my hunch is that several existing modules would benefit from the setFoo( fooModule ) pattern

Yeah, I was thinking the same. That's possibly the least bad solution here, and if we do that then I see why you're saying we don't need to prioritize the libs yet. 

In the long term though, it's not great to have modules with dependencies that are not available as modules, so we should not ignore them completely. 

In the case of `gunzip` and `inflate` it's unlikely that these will ever be available as modules, and they are just a few KB so I think it's fine to keep the hand converted versions of these, and possibly even bundle then into the UMD files to keep things simple for users. If you don't think that's a good solution then we should look into replacing them with alternatives.

takahirox:
Just in case, to clarify. As I said I haven't read and won't jump in the discussion, and no opinion. But I just found the sentences about MMD which don't look right and I just wanted to correct. So please ignore if it isn't related to the center core of the discussion. And one more nit, the file which has dependency to Ammo is not MMDLoader, but MMDPhysics. Loading MMD doesn't require Ammo. Ammo is required if users want to do animation with Physics. (Please ignore too if it isn't important for the discussion).

Mugen87:
I vote to create a module version for mandatory dependencies like `mmdparser` or `inflate`  and pack them into the generated UMD.

In context of optional dependencies I agree with @donmccurdy. Injecting them seems the better solution. Otherwise we end up with unnecessary overhead.

This means we might want to introduce `setTGALoader()` for certain loaders. Currently, `FBXLoader` or `ColladaLoader` would import `TGALoader` even when no TGA files are defined (which is probably true for most assets).

Mugen87:
Let's shift further module related discussions to #16688.

mrdoob:
Thanks!

Issue #16763, opened 2019-06-11
Cannot serialize scene with WebGLRenderTargets


jsantell:
##### Description of the problem

Previously, serializing a WebGLRenderTarget texture was a noop. Introduced in #16651, the serializer attempts to turn a simple object with width/height into a data URL, which fails: 

```
three.js:2740 Uncaught TypeError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The provided value is not of type '(CSSImageValue or HTMLImageElement or SVGImageElement or HTMLVideoElement or HTMLCanvasElement or ImageBitmap or OffscreenCanvas)'
 ```

Some solutions are:
* Properly serialize render targets (https://github.com/mrdoob/three.js/issues/16762) (:+1:)
* Go back to stubbing out render targets in the serialized form
* Move image info added in #16651 elsewhere (:-1:) 

As of now, this prevents serialization of any scene with a render target texture in use anywhere.

Reproducible here via `scene.toJSON()`: https://threejs.org/examples/webgl_loader_gltf.html Requires patching #16764 to get around scene.background issue, but the environment map used here throws

Mugen87:
This was fixed via #17745. 

`serializeImage()` in the `Texture` class now reports a warning when a texture can't be serialized. A runtime error does not occur anymore.

Because of https://github.com/mrdoob/three.js/issues/16762#issuecomment-504718551, `WebGLRenderTarget`'s `texture` is not supposed to be serialized.

PR #16764, opened 2019-06-11
Serialize render target's texture when used as a scene background

jsantell:
If using a render target for a scene's background, serialize the texture, not the render target which is not currently serializable. Requires #16763 fixed to not throw on serializing the render target's texture.

Reproducible here via `scene.toJSON()`: https://threejs.org/examples/webgl_loader_gltf.html

### DIFF
--- a/src/scenes/Scene.js
+++ b/src/scenes/Scene.js
@@ -49,7 +49,20 @@ Scene.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 		var data = Object3D.prototype.toJSON.call( this, meta );
 
-		if ( this.background !== null ) data.object.background = this.background.toJSON( meta );
+		if ( this.background !== null ) {
+
+			if ( this.background.texture ) {
+
+				data.object.background = this.background.texture.toJSON( meta );
+
+			} else {
+
+				data.object.background = this.background.toJSON( meta );
+
+			}
+
+		}
+
 		if ( this.fog !== null ) data.object.fog = this.fog.toJSON();
 
 		return data;


jsantell:
> So this PR avoids the runtime error `this.background.toJSON is not a function` since `WebGLRenderTargetCube` has no `toJSON()` method, right?

Correct

> Just for clarification: A correct serialization/deserialization is still not possible because...
> 
>     * The serialized texture is the texture of a render target. Hence, it has no valid `image` object and requires #16762 to be fixed.
> 
>     * `ObjectLoader` can only deserialize background with colors so far.

Correct; I haven't thought about the rehydrating via ObjectLoader, but should be possible if we encode all the relevant information as a static texture at time of serialization (with maybe extra meta indicating it was a render target at some point) -- FWIW I'm more interested in the JSON data and at the very least ignoring scene.background, because currently this breaks serializing of any scene with a background

Mugen87:
Okay. Against this backdrop the PR seems fine to me.

Mugen87 (approved):

src/scenes/Scene.js
```diff
@@ -49,7 +49,19 @@ Scene.prototype = Object.assign( Object.create( Object3D.prototype ), {
...
+
+				data.object.background = this.background.texture.toJSON( meta );
+
+			} else {
+
+				data.object.background = this.background.toJSON( meta );
+
+			}
+
+		}
```
WestLangley: nit: insert blank line

jsantell: :+1: updating!

mrdoob:
https://github.com/mrdoob/three.js/issues/16762#issuecomment-504718551

jsantell:
Is there anything else needed for this PR? This PR, together with https://github.com/mrdoob/three.js/pull/16816 allows the serialization of a scene that uses a render target. They do not address the serialization of a render target, just fixes errors thrown at attempts to serialize, but returns an empty string.

Mugen87:
Because of #21360, we don't have to bother with render targets anymore when serializing `Scene.background`.

There is another PR without merge conflicts that implements serialization and deserialization of `Scene.background`. Hence, closing this PR in favor of #18834.

PR #16767, opened 2019-06-12
Vector4: Added width/height aliases

WestLangley:
A viewport is defined by `x`, `y`, `width`, `height`.

```js
var viewport = new THREE.Vector4();

renderer.getViewport( viewport );

var aspect = viewport.width / viewport.height; // instead of viewport.z / viewport.w
```

These new aliases are akin to the existing `Vector2.width` and `Vector2.height`.

### DIFF
--- a/docs/api/en/math/Vector4.html
+++ b/docs/api/en/math/Vector4.html
@@ -78,6 +78,12 @@ var d = a.dot( b );
 
 		<h3>[property:Float w]</h3>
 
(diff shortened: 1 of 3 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

mrdoob:
I thought I added this in the past already? 🤔

WestLangley:
You do something similar in `WebXRManager`...

PR #16788, opened 2019-06-14
Fixed TextureCubeNode

DanielSturk:
TextureCubeNode references a non-existing function for decoding (`NodeBuilder.getTexelDecodingFunctionFromTexture`), making it unusable at the moment

Issue #16375, opened 2019-05-02
NodeMaterial: Incorrect RectAreaLight specular highlight


MEBoo:
I know that RectAreaLight can't cast shadows, and that's ok... but I have this problem

A scene with a single spotlight and a single area light (you can see its highlight on the floor)
![image](https://user-images.githubusercontent.com/39593253/57103577-7889ea80-6d26-11e9-9384-7915e029ad52.png)

The same scene with shadow casting enabled for the spotlight
![image](https://user-images.githubusercontent.com/39593253/57103723-d6b6cd80-6d26-11e9-879d-ac7588a77f8a.png)

what's wrong?

##### Three.js version

- [x]  Dev

WestLangley:
It is best to redirect your question to the three.js [forum](https://discourse.threejs.org).

If you can provide a _live example_ demonstrating a bug in the _current three.js revision_, you may reopen this issue.

MEBoo:
I've planned to update Three in a month, it's a bit complex since there are many changes I've to take care.

But if you are redirecting me to the forum, you say it isn't a bug.. maybe with the update the problem is already resolved.

WestLangley:
>But if you are redirecting me to the forum, you say it isn't a bug.

I did not say that. I don't know if it is a bug.

I do know it is something I was not able to reproduce.

>Sorry, could you confirm me that the RectAreaLight can't and should not conflict with another light even if that one is casting shadows?

I would not expect such a conflict to occur.

MEBoo:
Hi @WestLangley 
Here your Fiddle!

https://jsfiddle.net/9zkrbj2m/6/

The problem is when the specular highlight is on a NodeMaterial! 
Check the floor plane on my demo.
It works when If you use a normal material

WestLangley:
@MEBoo Thanks for the fiddle.

It seems to me the `RectAreaLight` specular highlight is incorrect with `NodeMaterial` regardless of whether another light casts a shadow; toggling the shadow can amplify the error.

Data is being corrupted somewhere, apparently. I am not sure what is going on...

sunag:
sorry delay... `RectAreaLightUniformsLib.js` modify native uniforms but nodematerial not use native uniforms... this bug is because of this `ltc_1` and `ltc_2` is not sended to uniforms. 
This example above fix this:

https://jsfiddle.net/8nhk975v/

```javascript
 material.onBeforeCompile2 = material.onBeforeCompile;
[...]

PR #16796, opened 2019-06-16
NodeMaterial revision

sunag:
This includes fixes and example of @DanielSturk, irradiance feature and better interface for Color Space using nodes.

Related:
https://github.com/mrdoob/three.js/pull/16788
https://github.com/mrdoob/three.js/pull/16668
https://github.com/mrdoob/three.js/issues/16375

## Changes
- unify `Math1Node`, `Math2Node`, `Math3Node` to `MathNode`
- new `ColorSpaceNode`
- new example `webgl_materials_envmaps_hdr_nodes.html`
- rename `Node.parse` to `Node.analyze`
- rename `Node.buildCode` to `Node.flow`
- rename `Node.eval` to `Node.parse`
- move `NodeBuilder.optimize = false` to `NodeBuilder.context.ignoreCache = true`
- fix irradiance `LDR`, `HDR` and `RGBM16`
- fix `RectAreaLight`

https://rawgit.com/sunag/three.js/dev-fix-TextureCubeNode/examples/index.html?q=nodes#webgl_materials_envmaps_hdr_nodes

image


sunag:
How should we call the `vector2`?
This is the indirect(irradiance) cube coords vector.
/ping @WestLangley @bhouston @mrdoob 

Select `vector2` in `scope` option:
https://rawgit.com/sunag/three.js/dev-fix-TextureCubeNode/examples/webgl_materials_nodes.html?e=node-reflect

sunag:
Here the original code:
https://github.com/mrdoob/three.js/blob/2ad018c7c7ad103c9da9d7f22b3154537384ab26/src/renderers/shaders/ShaderChunk/envmap_physical_pars_fragment.glsl.js#L30

sunag:
Opts it is a `world normal` with `flipEnvMap`!?

```javascript
new THREE.NormalNode( THREE.NormalNode.WORLD )
```

DanielSturk:
Great progress! Looking forward to seeing these changes made

sunag:
Thanks @DanielSturk! 

Done the fixes and update the example. The visual result should be the same for both now`MeshStandardNodeMaterial` and `MeshStandardMaterial`.

https://rawgit.com/sunag/three.js/dev-fix-TextureCubeNode/examples/webgl_materials_envmaps_hdr_nodes.html

sunag:
@donmccurdy following your tip.
https://github.com/mrdoob/three.js/pull/16796/commits/4665931e7b3a5af40066b015dfc2c42ab1c12bcb

examples/js/nodes/materials/NodeMaterial.js
```diff
@@ -18,6 +18,22 @@ function NodeMaterial( vertex, fragment ) {
 
 	this.updaters = [];
 
+	// onBeforeCompile can't be in the prototype because onBeforeCompile.toString varies per material
```
donmccurdy: Where does `onBeforeCompile.toString` get called?

donmccurdy:
@sunag thank you! :)

***

In words, how would you define 'analyze', 'parse', and 'flow' as they relate to NodeMaterial? I'm having some trouble understanding what "analyze" means — the word usually doesn't imply side effects, but in this case `analyze()` calls methods like `build()` and `addFlow()`.

sunag:
### Primary process

1. **analyze()** It is necessary to analyze and validate all nodes and slots(channels), it is will generate info necessary for build the code, share nodes(create temp vars), optimization as well the variable type and the return value.

2. **flow()** It generates codes, includes functions, defines, uniforms, and other codes need to work the shader. It depends on `analyze` be execute before `flow` to generate info necessary to build the code/shader. `flow` returns all date of the slot/channel.

3. **analyzeAndFlow()** Call the two methods `analyze` and after the `flow`. It does not optimize the shader if used many slots/channels, for the reason, it is mostly used in `transform` slot of the `vertex shader`.

### Secundary process

4. **build** returns the final value, generate by `generate` function or the variable generate by `NodeTemp` just like any other optimization node.

5. **generate** create the snippets fundamental to generate the code. Here too, is responsible to include functions, defines, etc. Return only the final value converted to output using `format` function.

The process superficially is similar to this diagram:

Compiler

sunag:
`parse` is used only to parse GLSL code, example is FunctionNode and StructNode

examples/js/nodes/materials/NodeMaterial.js
sunag: It is used to create a hash modification in onBeforeCompile. More details:
https://github.com/mrdoob/three.js/pull/14442
https://github.com/mrdoob/three.js/pull/16668

sunag: https://github.com/mrdoob/three.js/pull/15674

donmccurdy: I don't see anything calling `toString()` in these PRs?

donmccurdy:
Thanks! I'm confused, but I appreciate the detailed writeup. Probably I need to read the code more and see what calls what. 😅

examples/js/nodes/materials/NodeMaterial.js
mvc1006: I'm confused,threejs is a new unfamiliar technology for me ,how to learning it ?

sunag: In 15674 have a commit calling toString
https://github.com/mrdoob/three.js/pull/15674/commits/07236bb765fd2d2231196ca89125b335144241e1
The description of the problem:
https://github.com/mrdoob/three.js/pull/15674#issuecomment-459417579

I think that this approach will be improved in the future...
https://github.com/mrdoob/three.js/blob/7fe9a41cb6930d8600305e8ee85830b011e461bb/src/renderers/webgl/WebGLPrograms.js#L249

mrdoob:
@donmccurdy should I merge this one for now or do you want to take a closer look?

mvc1006:
thanks




| |
陈鹏菲
|
|
邮箱：mvc1006@163.com
|

签名由 网易邮箱大师 定制

On 06/20/2019 10:03, Mr.doob wrote:

@donmccurdy should I merge this one for now or do you want to take a closer look?

donmccurdy (approved):
Good to merge. 👍

mrdoob:
Thanks!

PR #16803, opened 2019-06-17
OrbitControls: customizable touch binds

sciecode:
It's becoming fairly common for me to need to modify the hardcoded number of touches for rotate and dollypan. I would imagine others could benefit from having control over that as well.

The only problem with this approach would be when users incorrectly pick a single touch for the dollypan. Perhaps a warning would suffice?

### DIFF
--- a/docs/examples/en/controls/OrbitControls.html
+++ b/docs/examples/en/controls/OrbitControls.html
@@ -234,6 +234,18 @@ controls.mouseButtons = {
 			the focus of the controls.
 		</p>
 
+		<h3>
+			[property:Object touches]</h3>
+		<p>
+			This object contains references to the number of touches used for the controls.
+			<code>
+controls.touches = {
+	ROTATE: 1,
+	DOLLY_PAN: 2
+}
+			</code>
+		</p>
+
 		<h3>[property:Float zoom0]</h3>
 		<p>
 			Used internally by the [method:saveState] and [method:reset] methods.
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -77,6 +77,9 @@ THREE.OrbitControls = function ( object, domElement ) {
 	// Mouse buttons
 	this.mouseButtons = { LEFT: THREE.MOUSE.LEFT, MIDDLE: THREE.MOUSE.MIDDLE, RIGHT: THREE.MOUSE.RIGHT };
 
+	// Touch fingers
+	this.touches = { ROTATE: 1, DOLLY_PAN: 2 };
+
 	// for reset
 	this.target0 = this.target.clone();
 	this.position0 = this.object.position.clone();
@@ -833,7 +836,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 		switch ( event.touches.length ) {
 
-			case 1:	// one-fingered touch: rotate
+			case scope.touches.ROTATE:
 
 				if ( scope.enableRotate === false ) return;
 
@@ -843,7 +846,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 				break;
 
-			case 2:	// two-fingered touch: dolly-pan
+			case scope.touches.DOLLY_PAN:
 
 				if ( scope.enableZoom === false && scope.enablePan === false ) return;
 
@@ -876,7 +879,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 		switch ( event.touches.length ) {
 
-			case 1: // one-fingered touch: rotate
+			case scope.touches.ROTATE:
 
 				if ( scope.enableRotate === false ) return;
 				if ( state !== STATE.TOUCH_ROTATE ) return; // is this needed?
@@ -885,7 +888,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 				break;
 
-			case 2: // two-fingered touch: dolly-pan
+			case scope.touches.DOLLY_PAN:
 
 				if ( scope.enableZoom === false && scope.enablePan === false ) return;
 				if ( state !== STATE.TOUCH_DOLLY_PAN ) return; // is this needed?
--- a/examples/jsm/controls/OrbitControls.d.ts
+++ b/examples/jsm/controls/OrbitControls.d.ts
@@ -45,6 +45,7 @@ export class OrbitControls {
 	enableKeys: boolean;
 	keys: { LEFT: number; UP: number; RIGHT: number; BOTTOM: number; };
 	mouseButtons: { LEFT: MOUSE; MIDDLE: MOUSE; RIGHT: MOUSE; };
+	touches: { ROTATE: number; DOLLY_PAN: number; };
 
 	rotateLeft(angle?: number): void;
 
--- a/examples/jsm/controls/OrbitControls.js
+++ b/examples/jsm/controls/OrbitControls.js
@@ -86,6 +86,9 @@ var OrbitControls = function ( object, domElement ) {
 	// Mouse buttons
 	this.mouseButtons = { LEFT: MOUSE.LEFT, MIDDLE: MOUSE.MIDDLE, RIGHT: MOUSE.RIGHT };
 
+	// Touch fingers
+	this.touches = { ROTATE: 1, DOLLY_PAN: 2 };
+
 	// for reset
 	this.target0 = this.target.clone();
 	this.position0 = this.object.position.clone();
@@ -842,7 +845,7 @@ var OrbitControls = function ( object, domElement ) {
 
 		switch ( event.touches.length ) {
 
-			case 1:	// one-fingered touch: rotate
+			case scope.touches.ROTATE:
 
 				if ( scope.enableRotate === false ) return;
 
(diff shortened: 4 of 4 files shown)


sciecode:
/ping @mrdoob @Mugen87 any feedback on this change?

WestLangley:
>It's becoming fairly common for me to need to modify the hardcoded number of touches for rotate and dollypan.

What are the values you are using in your app?

sciecode:
> What are the values you are using in your app?

First time I modified, I disabled dollypan and set:
1. custom event
2. rotate

Second time:
1. custom event
2. rotate
3. dollypan

WestLangley:
Your suggestion has merits, but you can't pinch-to-zoom with one finger, and with three fingers it is -- well --  a bit odd. The code relies on two touches for zooming.

I am reluctant to support adding a feature unless it just works, otherwise we get complaints, but as I said, your suggestion is not without merit.

`MapControls` is also hard-coded as far as the number of touches is concerned. I'd like to keep the two codebases similar. It is easier to maintain that way.

I am on the fence...

sciecode:
> Your suggestion has merits, but you can't pinch-to-zoom with one finger, and with three fingers it is -- well -- a bit odd. The code relies on two touches for zooming.

I completely agree. It is a bit awkward.

The most useful case for this change is definitely being able to remap rotate to two fingers and allow one finger for custom event to be honest. 

Let's see how others feel about it, but I'm not strongly attached to this idea as well. Promise I won't get mad if changes don't get merged :)

Mugen87:
> The most useful case for this change is definitely being able to remap rotate to two fingers and allow one finger for custom events, to be honest.

I don't think we need an API change for this rather unusual use case. The current touch interaction seems ideal to me and I would not encourage users to change with a respective interface.

sciecode:
> TBH, I don't think we need an API change for this rather unusual use case. The current touch interaction seems ideal to me and I would not encourage users to change with it a respective interface.

I understand why you say that, but I don't think it's that unusual at all.

Some applications might want to have a pan option as a single touch event.
Others might want just to zoom in or out based on a single touch y coordinate (not currently supported)
You might want to set a multitude of custom events and integrate with controls api. 

I'm not saying my approach in the PR is the one we should go for. Perhaps, rethinking a more customizable api would be the correct thing to do. I think other approaches are only unusual because we might've gotten used to this fixed scheme, not necessarily because there's no need for other interfaces.

I understand that `OrbitControls` is still an example file and it's meant to be modified accordingly to the user needs, but I also believe it's as much of a core file as it gets. Likely being the most used example file in the whole library.

So I can see how adapting or rethinking the API could be an advantage in this scenario.

WestLangley:
>The current touch interaction seems ideal to me and I would not encourage users to change with it a respective interface.

I think @sciecode's use cases are perfectly reasonable -- it is just that the PR proposes an overly-general solution that allows options that do not make sense.

WestLangley:
As I have said previously, I am reluctant to support any further changes to `OrbitControls` -- other than bug fixes.

However, on balance, I think this PR provides a valuable feature. I would request that the change also be applied to `MapControls`, however.

sciecode:
@Mugen87 @WestLangley thoughts on extending this approach by creating independent zoom and pan functionalities and also displaying error for zoom/dollypan on single touch?

default api would look something like this:

```
this.touches = { 
    rotate: 1,
    dollypan: 2, // keeps current implementation
    pan: THREE.INACTIVE,
    zoom: THREE.INACTIVE
}
```

> it is just that the PR proposes an overly-general solution that allows options that do not make sense

It still doesn't solve this problem, but given that documentation is clear about it and that we provide a descriptive error on incorrect use. I don't see this being a big problem.

On the plus side, it would allow for much better control over touch behaviors.

WestLangley:
@sciecode I prefer your current simple proposal. I wouldn't worry about parameter validation.

sciecode:
`MapControls` and `OrbitControls` are very similar right now.

I don't really see the use of having two classes that fulfill almost the same purpose, when we can adjust one to do both things ( and more ). That will require us to implement `touchPan` and `touchDollyRotate` on `OrbitControl`, but I think this is overall a great improvement over our current scheme.

A different approach to this PR, would be to only allow certain touch actions based on the number of fingers. If an invalid configuration is used, we simply treat it as `none`. That way we solve the current problem with this PR and also remove the necessity of `MapControls`.

1-finger - ( none, rotate, pan )
2-fingers - ( none, rotate, pan, dollyPan, dollyRotate )

WestLangley:
@sciecode I think this looks reasonable if you could have `MapControls` extend `OrbitControls` so it could be retained as an API.

Can you specify some example mouse and touch settings we could try? Don't forget that some users just want to be able to swap left and right mouse buttons.

sciecode:
> @sciecode I think this looks reasonable if you could have `MapControls` extend `OrbitControls` so it could be retained as an API.
>
> Can you specify some example mouse and touch settings we could try? Don't forget that some users just want to be able to swap left and right mouse buttons.

On this newly proposed scheme, changing from `OrbitControls` functionalities to `MapControls` would be as simple as modifying the following properties:

	// mapControls properties

	controls.mouseButtons.LEFT = THREE.MOUSE.RIGHT;
	controls.mouseButtons.RIGHT = THREE.MOUSE.LEFT;

	controls.touches.ONE = THREE.TOUCH.PAN;
	controls.touches.TWO = THREE.TOUCH.DOLLY_ROTATE;

[MapControls example](https://rawcdn.githack.com/sciecode/three.js/afa8ba559ca4ad89718001c0d1ba798ec7f33432/examples/misc_controls_orbit.html)

disabling either one or two finger events would be as this:

     controls.touches.ONE = undefined;

I made sure to not modify mouse / keyboard controls API in anyway. I'm just altering touch events in both PRs suggestions.

WestLangley:
>I made sure to not modify mouse / keyboard controls API in anyway. I'm just altering touch events in both PRs suggestions.

We have to fix that, too. It does not appear you are handling `ctrlKey/metaKey/shiftKey` correctly.

With the following changes to your proposed code, `OrbitControls` appears to work correctly for mouse events, and also to be backward-compatible:

In `constants.js`:
```js
// LEFT, MIDDLE, RIGHT will be deprecated in favor of specifying "actions"
export var MOUSE = { LEFT: 0, MIDDLE: 1, RIGHT: 2, ROTATE: 0, DOLLY: 1, PAN: 2 };
```

In `OrbitControls.js`:
```js
// Mouse button actions
this.mouseButtons = { LEFT: THREE.MOUSE.ROTATE, MIDDLE: THREE.MOUSE.DOLLY, RIGHT: THREE.MOUSE.PAN };

//

function onMouseDown( event ) {

	if ( scope.enabled === false ) return;

	// Prevent the browser from scrolling.

	event.preventDefault();

	// Manually set the focus since calling preventDefault above
	// prevents the browser from setting it automatically.

	scope.domElement.focus ? scope.domElement.focus() : window.focus();

	switch ( event.button ) {

		case 0:

			switch ( scope.mouseButtons.LEFT ) {

				case THREE.MOUSE.ROTATE:

					if ( event.ctrlKey || event.metaKey || event.shiftKey ) {

						if ( scope.enablePan === false ) return;

						handleMouseDownPan( event );

						state = STATE.PAN;

					} else {

						if ( scope.enableRotate === false ) return;

						handleMouseDownRotate( event );

						state = STATE.ROTATE;

					}

					break;

				case THREE.MOUSE.PAN:

					if ( scope.enablePan === false ) return;

					handleMouseDownPan( event );

					state = STATE.PAN;

					break;

				default:

					state = STATE.NONE;

			}

			break;


		case 1:

			switch ( scope.mouseButtons.MIDDLE ) {

				case THREE.MOUSE.DOLLY:

					if ( scope.enableZoom === false ) return;

					handleMouseDownDolly( event );

					state = STATE.DOLLY;

					break;


				default:

					state = STATE.NONE;

			}

			break;

		case 2:

			switch ( scope.mouseButtons.RIGHT ) {

				case THREE.MOUSE.ROTATE:

					if ( scope.enableRotate === false ) return;

					handleMouseDownRotate( event );

					state = STATE.ROTATE;

					break;

				case THREE.MOUSE.PAN:

					if ( scope.enablePan === false ) return;

					handleMouseDownPan( event );

					state = STATE.PAN;

					break;

				default:

					state = STATE.NONE;

			}

			break;

	}

	if ( state !== STATE.NONE ) {

		document.addEventListener( 'mousemove', onMouseMove, false );
		document.addEventListener( 'mouseup', onMouseUp, false );

		scope.dispatchEvent( startEvent );

	}

}
```

sciecode:
I really like this, gonna setup a proper PR with these changes.

    // Mouse button actions
    this.mouseButtons = { LEFT: THREE.MOUSE.ROTATE, MIDDLE: THREE.MOUSE.DOLLY, RIGHT: THREE.MOUSE.PAN };

I was gonna suggest that we changed mouse constants to match touch pattern 👍 
Thanks @WestLangley

WestLangley:
If we carry this to its natural completion, `MapControls` properties would now be:

```js
controls.mouseButtons.LEFT = THREE.MOUSE.PAN;
controls.mouseButtons.RIGHT = THREE.MOUSE.ROTATE;

controls.touches.ONE = THREE.TOUCH.PAN;
controls.touches.TWO = THREE.TOUCH.DOLLY_ROTATE;
```

WestLangley:
You will have to handle `ctrlKey/metaKey/shiftKey` for the `MapControls` case, too, but that is straight-forward.

Please do your best to make sure the behavior is not changed. This is a refactoring only.

Consider defining `MapControls` in the same file to retain the API. I don't particularly care one way or the other, but I expect others will.

PR #16817, opened 2019-06-17

Mugen87:
Moving on with material examples...

### DIFF
--- a/examples/webgl_loader_x.html
+++ b/examples/webgl_loader_x.html
@@ -19,7 +19,6 @@
 
 	<div id="canvase3d"></div>
 
-
 	<script type="module">
 		import {
 			AmbientLight,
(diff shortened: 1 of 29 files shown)


Mugen87:
Half of the job is done 😉

mrdoob:
Amazing! Thanks!

WestLangley:
So... these informative commands no longer work in the console.

    > renderer.info

    > scene.children

Mugen87:
Well, using modules means moving away from polluting the global namespace and all related problems. If these commands does not work anymore, we are doing everything right.

WestLangley:
That reply is not helpful. I suppose that means you do not have a solution to the issue I raised?

mrdoob:
@WestLangley the Chrome DevTool is almost ready ^^

PR #16818, opened 2019-06-17
InterleavedAttribute: normalized defaults to false instead of true

vlenhart:
The docs are wrong. From the source: `this.normalized = normalized === true;`


WestLangley:
Thanks!

PR #16827, opened 2019-06-18

Mugen87:
I'm done with the first iteration meaning most of the examples (approx. 340) are now converted, a few are still left. I'm going to update #16688 with a list. 

BTW: The PR removes `webgl2_loader_gltf` since we need no demo that demonstrates the usage of the module version of `GLTFLoader` anymore. The WebGL 2 context did not really make sense anyway.


mrdoob:
Awesome!

mrdoob:
Thanks!

PR #16833, opened 2019-06-18
examples: replaces Date with THREE.Clock.

sciecode:
Replaced all relevant uses of `Date.now` inside examples with `THREE.Clock` functions.
I've double tested all examples and made sure I wasn't messing the timing on anything.

Some examples no longer start at random points of the animation, because of the nature of `Clock.ElapsedTime` compared with `Date.now` pseudo randomness.
No apparent problems were caused by this, aside from a fixed starting point in the animation cycle. 

[branch examples](https://rawcdn.githack.com/sciecode/three.js/aa3cbe04168a458de1f0444ab530ce564d2fff69/examples/)


sciecode:
@mrdoob I know this is a big one to consider and it might take some time to review these changes.
I'll make sure to resolve conflicts as they appear, until we have a definite position on it.

WestLangley:
>I'll make sure to resolve conflicts as they appear, until we have a definite position on it.

@sciecode I doubt @mrdoob will be reviewing this; I expect he will rely on @Mugen87 for a recommendation. He should be able to give you a reply in a timely fashion.

sciecode:
Gotcha, thank you.

sciecode:
I just realized that I can keep the same pseudo randomness start by using `clock.startTime`.
If that is something we would like to keep, I can make the appropriated changes.

Mugen87:
The pattern of the change looks good however I don't think using `Date` or `performance` in the examples is an issue. Because of this I don't want to invest the time to check all single 84 examples for visual correctness. If you have double tested all examples, it's sufficient to me.

sciecode:
Just re-checked all examples again. I guarantee every example is working.

Mugen87 (approved):

sciecode:
updated namespace conflicts.

sciecode:
/ping @mrdoob any feedback on these changes?
I could divide these changes into smaller PRs to make it easier to review and merge.

Mugen87:
Even in brand new examples `Date` is used instead of `THREE.Clock`, see

https://github.com/mrdoob/three.js/blob/946b8d231dd9017c3343944e1d13849d9b3f5a70/examples/webgl_instancing_suzanne.html#L99

So...I don't think there is a need for a change. `THREE.Clock` is already used in many examples and thus sufficiently promoted.

PR #16836, opened 2019-06-19
Closure compiler using ADVANCED_OPTIMIZATIONS

HypnosNova:
See https://developers.google.com/closure/compiler/docs/api-tutorial3
We can set ADVANCED_OPTIMIZATIONS when using google-closure-compiler and in this way, the three.min.js could be less than 500KB.


Mugen87:
Have you actually tested the resulting `three.min.js`?

PR #16837, opened 2019-06-19
Clean

HypnosNova:
image
I found eslint throw a warning in this line. 
"framebufferScaleFactor" is defined but never used. I don't know whether we need this or not so I add this eslint-disable-line

And XRWebGLLayer, I think it would be better to use window.XRWebGLLayer because when I write a project depends on this, if I use closure-compiler to minify the js file, closure-compiler will throw an Error to tell me that XRWebGLLayer is not defined.


Mugen87:
> closure-compiler will throw an Error to tell me that XRWebGLLayer is not defined.

We are not getting any errors when we build `three.min.js` with the closure compiler. All what you have to do is to define such variables in a separate file and then refer to it when executing the compiler like so:
```
--externs utils/build/externs.js
```

Issue #16835, opened 2019-06-19
InterleavedBufferAttribute misses toJSON method


Alchemist0823:
##### Description of the problem

https://github.com/mrdoob/three.js/blob/dev/src/core/InterleavedBufferAttribute.js
InterleavedBufferAttribute misses toJSON

##### Three.js version

- [ ] Dev
- [ x] r105
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
Read the following two issue why a copy/clone operations on attribute levels is difficult. #15395/#12800

Mugen87:
Fixed via #18537. Cloning/serializing interleaved attributes deinterleaves the data now. Ideally, the engine can retain interleaved data in the future when an entire `BufferGeometry` is cloned /serialized.

PR #16838, opened 2019-06-19
add toJSON to interleavedBufferAttribute

Alchemist0823:
Add toJSON method to interleavedBufferAttribute
added unit test


Mugen87:
see https://github.com/mrdoob/three.js/issues/16835#issuecomment-503463202

Mugen87:
Closing. Serializing makes only sense if the deserialization process is able to restore the interleaved buffer. I'm afraid this does not work with this presented approach.

PR #16852, opened 2019-06-19
LDrawLoader: Implement STEP directive

yomboprime:
The STEP directive serves as a separator of the geometry, dividing it in sets of pieces that represent real construction steps of the model.

In the implementation the `model.userData.numConstructionSteps` and `subpartGroup.userData.constructionStep` variables can be used by the application to render construction steps as desired.

In the example, a slider lets the user change the currently visualized step, when `Separate` option is enabled.


mrdoob:
Thanks!

PR #16863, opened 2019-06-20
Utils: Converters use now modules.

Mugen87:
A lot of examples are already using modules. However, other parts of the repository like the converters (`fbx2three.js` and `obj2three.js`) require UMDs. They are node scripts where using ES6 modules is still experimental. Nevertheless, the ultimate goal is to avoid any duplicates in the code base.

After some research, I've realized that using [esm](https://www.npmjs.com/package/esm) works pretty well for this use case. It's possible to use the same coding patterns like in the examples. It's only necessary to install `esm` before executing the scripts and add `-r esm` to the node command.

Feedback and other ideas are welcome since I've not used this approach in production before.

### DIFF
--- a/utils/converters/README.md
+++ b/utils/converters/README.md
@@ -1,4 +1,4 @@
-Utilities for converting model files to the Three.js JSON format. You have to install [esm](https://www.npmjs.com/package/esm) for executing the converters.
+Utilities for converting model files to the Three.js JSON format. You have to install [node.js](https://nodejs.org/en/) and [esm](https://www.npmjs.com/package/esm) for executing the converters.
 
 ## obj2three.js
 
--- a/utils/converters/README.md
+++ b/utils/converters/README.md
@@ -1,11 +1,11 @@
-Utilities for converting model files to the Three.js JSON format.
+Utilities for converting model files to the Three.js JSON format. You have to install [esm](https://www.npmjs.com/package/esm) for executing the converters.
 
 ## obj2three.js
 
 Usage:
 
 ```
-node obj2three.js model.obj
+node -r esm obj2three.js model.obj
 ```
 
 ## fbx2three.js
@@ -13,5 +13,5 @@ node obj2three.js model.obj
 Usage:
 
 ```
-node fbx2three.js model.fbx
+node -r esm fbx2three.js model.fbx
 ```
--- a/utils/converters/fbx2three.js
+++ b/utils/converters/fbx2three.js
@@ -18,11 +18,8 @@ function parseNumber( key, value ) {
 
 }
 
-THREE = require( '../../build/three.js' );
-require( '../../examples/js/curves/NURBSCurve.js' );
-require( '../../examples/js/curves/NURBSUtils.js' );
-require( '../../examples/js/loaders/FBXLoader.js' );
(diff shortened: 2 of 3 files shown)


donmccurdy:
> However, other parts of the repository like the converters (fbx2three.js and obj2three.js) require UMDs. 

To be specific, we don't use UMD modules anywhere yet – the `examples/js` files are not modules, they're just scripts that write to global variables. Real UMD modules can be consumed as CommonJS modules, or can write to browser globals.

I'm hoping that the end goal is still to generate true UMD modules (to `examples/js`) from source ES modules (in `examples/jsm`). Once we get there, it would probably be better for these nodejs scripts to use the UMD versions via CommonJS. Once nodejs gets ES module support we could switch it over of course.

In the meantime I have no problem with merging this PR.

Mugen87:
> To be specific, we don't use UMD modules anywhere yet – the examples/js files are not modules, they're just scripts that write to global variables.

Sorry for not being precise here. I've accidentally used the term `UMD` in the wrong context.

 > I'm hoping that the end goal is still to generate true UMD modules (to examples/js) from source ES modules (in examples/jsm).

But do we need actually need `examples/js` under source control? If the code base has no dependencies to `examples/js` anymore, why not removing it and let the user generate UMDs if necessary.

donmccurdy:
> But do we actually need examples/js under source control? If the code base has no dependencies to examples/js anymore, why not removing it and let the user generate UMDs if necessary.

Maybe not, but if we were planning to delete `examples/js` then I'm not sure why we're going to so much trouble to keep both directories in sync during the transition... Why not delete `examples/js/GLTFLoader.js` today? I'd assumed we wanted backward-compatibility. Or just for a few versions? There may still be enough people using CommonJS (e.g. with build tools like Browserify and Budo) to justify keeping the CommonJS modules – the core library is available as CommonJS, for instance.

Mugen87:
> Why not delete examples/js/GLTFLoader.js today?

Stuff like the editor still depend on these files. That's because I have not suggested a removal so far.

How would a conversion of ES6 to UMD look like? Can we use an existing tool for this?

donmccurdy:
> How would a conversion of an ES6 module to UMD look like? Can we use an existing tool for this?

Yes, going ES->UMD is more straightforward and we can use Rollup directly.

Mugen87:
When performing the conversion, I expect rollup will bundle all dependencies of an ES6 module into the final UMD, right? Let's say you do this for `GLTFLoader`. Does it also include `three.js` into the `UMD` then?

donmccurdy:
Nope! We'll need a Rollup transform to change the paths from `examples/jsm` to `examples/js`, but the imports become requires. https://github.com/mrdoob/three.js/pull/15526 by @gkjohnson implements this.

There could certainly be some edge cases, e.g. I'm not sure `nodes/` and maybe some groups of files that collectively represent a single module should be transformed 1:1... it might be better to bundle a single file with all of the dependencies in those cases.

Mugen87:
I totally missed that PR! In this case, I'm closing this one. We should not wait too long to finish the transition though.

Mugen87:
@donmccurdy Thanks for the information! 😊

PR #16875, opened 2019-06-21
split VRMLLoader.parse into makeTree and parseTree

palortoff:
Split VRMLLoader.parse into makeTree and parseTree and expose those functions.

This allows to convert the VRML data into a JSON object with makeTree and later convert the tree with parseTree into a Scene object.

### DIFF
--- a/examples/js/loaders/VRMLLoader.js
+++ b/examples/js/loaders/VRMLLoader.js
@@ -65,452 +65,9 @@ THREE.VRMLLoader = ( function () {
 
 		},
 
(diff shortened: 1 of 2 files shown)


Mugen87:
The loaders should provide a common, simple interface like presented here: https://github.com/mrdoob/three.js/issues/5524#issuecomment-388775266

Hence, I'm not a fan of this type of refactoring.

Mugen87:
Closing. Apart from my previous comment, the mentioned JSON data is an internal one and it is not intended to expose it to user level code.

PR #16877, opened 2019-06-21
JSM: Update name of external libs.

Mugen87:
see https://github.com/mrdoob/three.js/pull/16862#issuecomment-504352169


mrdoob:
Thanks!

PR #16886, opened 2019-06-22
Example webgl2 fixes

### LINKED ISSUE #16870: example shaders not compatible with WebGL2
Out of curiosity I have tried running the examples using a WebGL2 context, several fail. The most common problem is the use of ShaderMaterials with a uniform 'texture' which conflicts with the GLSL ES 3.0 texture() function (easily fixed by renaming the uniform of course).

I think it would be good if examples avoided this issue, given the confusion that may occur if/when examples are used as the basis for other's code.

Thoughts?

aardgoose:
Bulk of examples fixed to work with WebGL2 contexts

* rename uniforms where name conflicts with GLSL texture() function
* function in shader that is now implemented in GLSL 3.0: tanh(), round()
* added WebGL2 capability detection in GPUComputationalRenderer + lint fixes

Fixes #16870

### DIFF
--- a/examples/js/misc/GPUComputationRenderer.js
+++ b/examples/js/misc/GPUComputationRenderer.js
@@ -97,7 +97,7 @@
  * @param {WebGLRenderer} renderer The renderer
   */
 
-THREE.GPUComputationRenderer = function( sizeX, sizeY, renderer ) {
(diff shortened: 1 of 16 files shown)


mrdoob:
Thanks!

PR #16892, opened 2019-06-24
Shadowmap camera fix

JustASquid:
I noticed this issue when I was splitting out my application into a number of layers and render calls. It seems that the shadowmap creation doesn't respect the layers of the shadowmap camera, where it perhaps ought to.

The ShadowMap render function has no reason for taking a camera as a parameter any, except to do the layer check. I believe the layer check as-is unintuitive - Checking the shadowCamera layer instead seems to make more sense, and will allow multi-layer multi-render pipelines to handle shadows more precisely.


Mugen87:
Are you aware that your change will break all application that rely on the current behavior? Layers are usually configured for 3D objects and the scene's camera like demonstrated in this [example](https://threejs.org/examples/webgl_layers). Decoupling the layer test for shadows is not conform to this pattern.

JustASquid:
@Mugen87 Yes you're right, thinking more carefully about it this would break simple examples like the one you linked.

However I wonder if there would be a way to have more fine-grained control over the shadow map rendering. It appears that there is no way currently to render some specific layers to the shadow map except as part of an actual render with the scene camera set to these layers.

Anyway, apologies for springing too quickly here, appreciate your thoughtful comments.

Mugen87:
> It appears that there is no way currently to render some specific layers to the shadow map except as part of an actual render with the scene camera set to these layers.

I think it makes more sense to make a feature request than continue the discussion in this PR. The current approach is not appropriate and can't be merged.

PR #16894, opened 2019-06-24
Editor: Fix ctor call of Command.

Mugen87:
see https://github.com/mrdoob/three.js/pull/16889#issuecomment-504914287


mrdoob:
Thanks!

PR #16898, opened 2019-06-24
BoxHelper: fix non-optional args

WestLangley:


WestLangley:
I think this is correct. I don't use ts.

/ping @Mugen87

mrdoob:
Thanks!

PR #16900, opened 2019-06-24
Added HDR (arbitrary encoding) support to CubeTexturePass

DanielSturk:
I also added the MeshCubeMaterial to the WebGLBackground so that it wouldn't have to do that weird map injection (https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLBackground.js#L90)

Examples:
webgl_postprocessing_backgrounds.html (for HDR CubeTexturePass)
webgl_materials_envmaps_hdr.html (for WebGLBackground)


Mugen87:
I guess this closes #9163?

DanielSturk:
> I guess this closes #9163?

Yes, thank you for linking that

WestLangley:
>I guess this closes #9163?

After spending almost an hour studying this PR, I was wondering how you came to have the skills to do this, and why you were making so many proposed changes in a single PR.

You used code #9163? What is different here?

bhouston:
@WestLangley btw to clear things up, @DanielSturk is working with me to porting back some of our private ThreeJS changes into ThreeJS mainline.  Thus yeah, some of this is my code.

DanielSturk:
In addition to `webgl_postprocessing_backgrounds.html` I tested it with all of
-`webgl_envmap_*`
-`webgl_materials_variations_standard.html`
-`webgl_materials_standard.html`
And it seems to work fine

DanielSturk:
> You used code #9163? What is different here?

It's mostly a revive of #9163 (I didn't know that existed, #9163 has now been closed), *plus I integrated it into WebGLBackground*

bhouston:
We have checked and it does not appear to break anything, where as the previous PR did break things.

Mugen87:
@DanielSturk If you resolve the merge conflicts, I'd like to approve this change. Introducing `MeshCubeMaterial` is a good idea.

WestLangley:
>@WestLangley btw to clear things up, @DanielSturk is working with me to porting back some of our private ThreeJS changes into ThreeJS mainline

I see.

This PR was filed with little explanation. Is the objective to be able to blur backgrounds? If so, I am generally supportive of this.

I'll add some comments inline.

examples/js/postprocessing/CubeTexturePass.js
```diff
@@ -10,21 +10,17 @@ THREE.CubeTexturePass = function ( camera, envMap, opacity ) {
 
 	this.needsSwap = false;
 
-	this.cubeShader = THREE.ShaderLib[ 'cube' ];
+	this.cubeMaterial = new THREE.MeshCubeMaterial();
```
WestLangley: I would suggest calling it `SkyboxMaterial`.

examples/js/postprocessing/CubeTexturePass.js
```diff
@@ -10,21 +10,17 @@ THREE.CubeTexturePass = function ( camera, envMap, opacity ) {
...
-			depthWrite: false,
-			side: THREE.BackSide
-		} )
+		this.cubeMaterial
 	);
 
 	this.envMap = envMap;
+	this.envMapIntensity = 1.0;
 	this.opacity = ( opacity !== undefined ) ? opacity : 1.0;
+	this.roughness = 0.0;
```
WestLangley: Perhaps `blurriness` is a better property name in this case.

examples/js/postprocessing/CubeTexturePass.js
```diff
@@ -38,19 +34,29 @@ THREE.CubeTexturePass.prototype = Object.assign( Object.create( THREE.Pass.proto
...
-		this.cubeCamera.quaternion.setFromRotationMatrix( this.camera.matrixWorld );
-
-		this.cubeMesh.material.uniforms[ "tCube" ].value = this.envMap;
-		this.cubeMesh.material.uniforms[ "opacity" ].value = this.opacity;
-		this.cubeMesh.material.transparent = ( this.opacity < 1.0 );
-
-		renderer.setRenderTarget( this.renderToScreen ? null : readBuffer );
-		if ( this.clear ) renderer.clear();
-		renderer.render( this.cubeScene, this.cubeCamera );
+		this.cubeCamera.projectionMatrix.copy( camera.projectionMatrix );
```
WestLangley: Be careful copying only `projectionMatrix`, as `projectionMatrixInverse` will be out-of-sync. If you are sure of what you are doing, it is OK.

examples/js/postprocessing/CubeTexturePass.js
```diff
@@ -38,19 +34,29 @@ THREE.CubeTexturePass.prototype = Object.assign( Object.create( THREE.Pass.proto
...
+		if( this.cubeMaterial.envMap != this.envMap ) {
+			this.cubeMaterial.envMap = this.envMap;
+			this.cubeMaterial.needsUpdate = true;
+		}
+		this.cubeMaterial.envMapIntensity = this.envMapIntensity;
+		this.cubeMaterial.roughness = this.roughness;
+		this.cubeMaterial.opacity = this.opacity;
+		this.cubeMaterial.transparent = ( this.opacity < 1.0 );
+
+		renderer.setRenderTarget(this.renderToScreen ? null : readBuffer);
```
WestLangley: whitespace. multiple places in the PR.

src/renderers/WebGLRenderer.js
```diff
@@ -2077,6 +2082,15 @@ function WebGLRenderer( parameters ) {
 
 	}
 
+	function refreshUniformsCube( uniforms, material ) {
```
WestLangley:
```js
refreshUniformsSkybox
```

src/renderers/shaders/ShaderLib.js
```diff
@@ -267,6 +267,29 @@ var ShaderLib = {
 
 };
 
+/* -------------------------------------------------------------------------
+//	Cube map shader
```
WestLangley: these comments are not needed

src/renderers/shaders/ShaderLib.js
```diff
@@ -267,6 +267,29 @@ var ShaderLib = {
...
+ ------------------------------------------------------------------------- */
+
+ShaderLib.cubeEnv = {
```
WestLangley:
```js
ShaderLib.skybox
```

src/renderers/shaders/ShaderLib.js
```diff
@@ -267,6 +267,29 @@ var ShaderLib = {
...
+		{
+			envMap: { value: null },
+			envMapIntensity: { value: 1.0 },
+			tFlip: { value: - 1 },
+			opacity: { value: 1.0 },
+			roughness: { value: 0.0 }
+		}
+	] ),
+
+	vertexShader: ShaderChunk.cubeEnv_vert,
```
WestLangley:
```js
ShaderChunk.skybox_vert
ShaderChunk.skybox_frag
```

src/renderers/shaders/ShaderLib/cubeEnv_frag.glsl.js
```diff
@@ -0,0 +1,52 @@
+export default /* glsl */`
+uniform float tFlip;
+uniform float opacity;
+uniform float roughness;
+
+varying vec3 vWorldPosition;
+
+#include <common>
+#include <logdepthbuf_pars_fragment>
+#include <clipping_planes_pars_fragment>
```
WestLangley: Why do we care about log-depth-buffer or clipping planes for a skybox?

src/renderers/shaders/ShaderLib/cubeEnv_frag.glsl.js
```diff
@@ -0,0 +1,52 @@
...
+#include <envmap_pars_fragment>
+#include <cube_uv_reflection_fragment>
+
+void main() {
+
+	#include <clipping_planes_fragment>
+
+	vec3 queryReflectVec = vec3( tFlip * vWorldPosition.x, vWorldPosition.yz );
```
WestLangley: This is not a reflection vector. Just call it `queryVec`.

src/renderers/shaders/ShaderLib/cube_vert.glsl.js
```diff
@@ -11,6 +11,8 @@ void main() {
 	#include <project_vertex>
 
 	gl_Position.z = gl_Position.w; // set z to camera.far
+	// important because otherwise the cube used for rendering the cubeMap might
+	// get clipped
```
WestLangley: This file has nothing to do with this PR. I think it should be removed from the PR.

WestLangley:
>I also added the MeshCubeMaterial to the WebGLBackground

I would request that you remove that for now since it changes the background behavoir, and the current background code works fine as-is.

Perhaps you can use your new material in WebGLBackground in a follow-up PR if there is agreement.

WestLangley:
@mrdoob, an alternative to adding `THREE.SkyboxMaterial` is to add `THREE.Skybox`, instead.

It may be more handy as it could automate things such as:
```js
skybox.onBeforeRender = function ( renderer, scene, camera ) {

	this.matrixWorld.copyPosition( camera.matrixWorld );
													
};
```

If we went with `THREE.Skybox`, it would still use the shader proposed in this PR -- it just would not be a separate material; the shader would be inlined.

DanielSturk:
Thanks for all the feedback @WestLangley, I'll look into making these changes soon

DanielSturk:
> This PR was filed with little explanation. Is the objective to be able to blur backgrounds? If so, I am generally supportive of this.
The primary object is to add HDR support to CubeTexturePass

src/renderers/shaders/ShaderLib/skybox_vert.glsl.js
```diff
@@ -0,0 +1,26 @@
+export default /* glsl */`
+varying vec3 vWorldPosition;
+
+#include <common>
+
+#if NUM_CLIPPING_PLANES == 0
+	#if ! defined( VARYING_VVIEWPOSITION )
+   		varying vec3 vViewPosition;
```
WestLangley: Can you please explain this? Your code is throwing console warnings that `vViewPosition` is not used by the fragment shader.

WestLangley:
Should the `dat.gui` provide a roughness control? Maybe in a later PR?

In the mean time, this is what is rendered for me. Do you get the same result?

Screen Shot 2019-07-06 at 5 59 48 PM

src/renderers/shaders/ShaderLib/skybox_vert.glsl.js
DanielSturk: Removed vViewPosition

DanielSturk:
> In the mean time, this is what is rendered for me. Do you get the same result?

That looks correct to me. A better demonstration would be turning the exposure way down to really illustrate the HDR 
51ebab13847dee0d41fe6b42c140e190

WestLangley:
>That looks correct to me

Sorry, but there is no way that is correct.

Compare with http://threejs.org/examples/webgl_materials_envmaps_hdr.html.

DanielSturk:
> > That looks correct to me
> 
> Sorry, but there is no way that is correct.
> 
> Compare with http://threejs.org/examples/webgl_materials_envmaps_hdr.html.

It is correct. The other demo you compared with (http://threejs.org/examples/webgl_materials_envmaps_hdr.html) had `renderer.gammaOutput=true`. I added `renderer.gammaOutput=true` to this demo so that you can compare them more easily.

WestLangley:
> I added renderer.gammaOutput=true to this demo so that you can compare them more easily.

You don't add that so @WestLangley can "compare them more easily". You add it because it is required. These are concepts you need to understand.

With the last commit, you have broken the LDR case.

After generously investing several hours helping you with this PR, I think fixing that is the last change required.

WestLangley:
@DanielSturk This is almost ready if you can fix the LDR case... :-)

DanielSturk:
@WestLangley 
Correct me if I'm mistaken, but there was never actually an issue with this demo
60761527-696e3400-a018-11e9-96ad-6ccd94b51d32
The color only appeared off because the newly added 'hdr' box was checked. If you check out my new commit which reverts the `renderer.gammaOutput=true` change, and ensure the box isn't checked, then the demo should appear the same. `webgl_materials_envmaps_hdr.html` should also appear the same now, since it was the `renderer.gammaOutput=true` commit that broke it

examples/webgl_postprocessing_backgrounds.html
```diff
@@ -15,8 +15,20 @@
 		<div id="container"></div>
 
 		<script type="module">
-
-			import * as THREE from '../build/three.module.js';
```
WestLangley: I think @mrdoob prefers this pattern. Compare with the existing examples.

WestLangley:
@DanielSturk I will continue help you fix your code...

First, revert the last commit. You need to covert from linear space to gamma-corrected color space.

Second, properly set the encoding for the loaded LDR texture in the loader callback function.
```js
```

The LDR and HDR use cases should now look similar to each other and similar to the other examples.

I think at that point this can be merged.  :-)

examples/webgl_postprocessing_backgrounds.html
Mugen87: Indeed, we actually moved from named imports to this style 👍

DanielSturk: Sure, done.

DanielSturk:
@WestLangley Thanks for sticking around...
I tried applying your 2 changes to `webgl_postprocessing_backgrounds.html`, and I agree, the demo does appear very similar to the master version. However, applying neither of these 2 changes, still appears the same as master. Neither of these changes exist in master, and I believe they only appear correct when applied together because they are canceling each-other out.

I examined the fragment shaders for the skybox/cube background in both master and in my branch, and found that without the changes, the encoding/decoding is the same. With the 2 changes, the encoding is different, but visually, the 2 changes to the encoding cancel each other out.

master: `Linear`
e369995a3ebbc09af1689b7aa6796ace

without changes: `Linear -> Linear -> Linear`
e03dbfa0253f9a8db9dd7659d8e98b9d
4fcb9fcd7dfd50da471247b588f1b55c

with both changes: `sRGB -> Linear -> Gamma`
9a4734ee16fbcc3019534689a7e8637d
5fe1752b88fbb43a2f2dec43996f3f5d

Of course it's no coincidence that `sRGB->Linear->Gamma` appear very similar to just `Linear`. The `sRGBToLinear()` function raises the color to the power of 2.4, and the `linearToGamma()` function raises the color to the power of 1/2.0

WestLangley:
@bhouston See if you can provide Daniel some references about rendering in linear colorspace so he understands the issue and we can get this PR merged. I have done my best to help.

An alternate solution -- an probably more appropriate solution -- is to add `GammaCorrectionPass` in place of `copyPass` as the last `EffectComposer` pass. You will then remove `renderer.gammaOutput` and post-process in linear space. That is probably the most appropriate workflow, but it is up to you.

@DanielSturk Good luck! :-)

bhouston:
If we want to be fully correct, we should be using "sRGB -> Linear -> Gamma" via:
- envMapToTexelToLinaer
- linearToOutputTexel

Then it will work no matter the input and output encoding.

bhouston:
> An alternate solution -- an probably more appropriate solution -- is to add GammaCorrectionPass in place of copyPass as the last EffectComposer pass. You will then remove renderer.gammaOutput and post-process in linear space. That is probably the most appropriate workflow, but it is up to you.

Be aware that GammaCorrectionPass is generally a bad idea and should never be used if it can be avoided.

The reason is that there is a rounding from the main linear space render results when it writes to the intermediate RGBA8 buffer and then you read in the GammaCorrectionPass and then do another quite subtle transform and it rounds again when it writes to the final RGBA8 buffer.  Because of the nature of the gamma correction transform, it will reduce the possible color intensites from the normal 256 per channel to just 184.

Here is a spreadsheet that shows the rounding that occurs and all the intensities that are no longer achievable:

https://docs.google.com/spreadsheets/d/1sEhW8QeilNtsjcgd0fX1dcoO3L1p4Ow-U9v96I0aYiM/edit#gid=0

Instead of 16,777,216 (256*256*256) possible colors, you can only get 6,229,504 (184*184*184) if you use the GammaCorrectionPAss -- you've reduced the number of total colors that Three.JS can render by 64%.  It will exacerbate banding in gradient areas especially.

Basically no one should ever use GammaCorrectionPass if they can avoid it.

bhouston:
@DanielSturk please consult/ping me if @WestLangley or another veteran Three.JS contributor gives you feedback before responding.  I should properly take the load off of them when possible.

WestLangley:
@bhouston Thanks for that info!

DanielSturk:
Thanks @WestLangley. After speaking with @bhouston I now understand the situation. I've add back changes now.

WestLangley:
@DanielSturk Thanks for this!

WestLangley:
@mrdoob Let's merge this if you approve.

examples/webgl_postprocessing_backgrounds.html
```diff
@@ -73,6 +85,14 @@
 
 				gui.add( params, "renderPass" );
 
+				gui.add( params, "exposure", 0, 2 );
+
+				gui.add( params, "hdr" );
+
+				//gui.add( params, "autoRotate" );
+
+				//gui.add( params, 'camera', [ 'perspective', 'orthographic' ] );
```
jvestolas: Do these comments need to be here?

DanielSturk:
@jvestolas Sure... removed

Mugen87 (approved):

Mugen87:
Closing this PR. `Scene.background` now properly supports HDR and blurriness (via PMREM). The `cube` shader was restored to its original implementation and is not used by `WebGLBackground` anymore. Since `CubeTexturePass` relies on `cube` shader, it provides just a limited feature set compared to `Scene.background`.

PR #16905, opened 2019-06-25
Docs: Clean up BufferGeometry page.

Mugen87:
see https://github.com/mrdoob/three.js/pull/12693#issuecomment-505006137


mrdoob:
Thanks!

PR #16907, opened 2019-06-25
support costume painter sort functions

ranbuch:
As discussed [here](https://github.com/mrdoob/three.js/issues/13889#issuecomment-385225320) I've created a PR that enables support for dynamic sort functions.


Mugen87:
Please remove all build files from the PR.

Mugen87:
Now you have removed the build files from the repo^^

ranbuch:
@Mugen87 I've reverted to my first commit but couldn't see `three.js` and `three.module.js`.
I've un-commited only `three.min.js`.
Let me know if there's still an issue, maybe it's best to delete this PR and I'll create a new one.

Mugen87:
The change list looks still not correct. Maybe it's better to start from scratch...

donmccurdy:
One way to reset the build files would be:

```
git checkout dev -- build/
```

But now I don't see the non-build files so maybe something has gone wrong. 😅

PR #16917, opened 2019-06-26
Examples: Move NodeMaterial to JSM.

Mugen87:
Follow up of #16054

Note: In some example I'm using this pattern similar to the `THREE` import:
```js
import * as Nodes from './jsm/nodes/Nodes.js';
```
Otherwise there is an extreme long list of imports.

/cc @sunag


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev34/examples/index.html

examples/webgl_materials_compile.html
```diff
@@ -33,205 +33,211 @@
...
+			import { TeapotBufferGeometry } from './jsm/geometries/TeapotBufferGeometry.js';
+			import { NodeFrame } from './jsm/nodes/core/NodeFrame.js';
+			import { MathNode } from './jsm/nodes/math/MathNode.js';
+			import { OperatorNode } from './jsm/nodes/math/OperatorNode.js';
+			import { ConstNode } from './jsm/nodes/core/ConstNode.js';
+			import { ColorNode } from './jsm/nodes/inputs/ColorNode.js';
+			import { ExpressionNode } from './jsm/nodes/core/ExpressionNode.js';
+			import { TimerNode } from './jsm/nodes/utils/TimerNode.js';
+			import { FloatNode } from './jsm/nodes/inputs/FloatNode.js';
+			import { PhongNodeMaterial } from './jsm/nodes/materials/PhongNodeMaterial.js';
```
donmccurdy: We don't import individual files from `src/`, but treat it as a single module. I think the same should apply to nodes, and we'd encourage users to write:

```js
import {
  NodeFrame,
  MathNode,
  OperatorNode,
  ...
} from './jsm/nodes/Nodes.js';

```

donmccurdy:
> `import * as Nodes from './jsm/nodes/Nodes.js';`

All of the places you used this look OK to me, but just for future reference, I would avoid using this shorthand (even for long lists of imports) within the `examples/jsm/nodes/*` subdirectory to ensure tree-shaking compatibility. The examples and NodeMaterialLoader are fine.

examples/webgl_materials_compile.html
donmccurdy: Perhaps `Nodes.js` should be renamed `index.js`?

Mugen87: Okay, I change the examples to this style:
```js
import * as Nodes from './jsm/nodes/Nodes.js';
```

Mugen87: > We don't import individual files from src/, but treat it as a single module.

Okay, done!

I'm not feeling strong about the name of `Nodes.js`. `index.js` seems to be fine, though.

gkjohnson:
Awesome! Is the intent still to transform the examples/jsm/Nodes folder into a single file in examples/js/Nodes? If so I can help modify the `build-examples` rollup script to help achieve that (and for other special cases like EffectComposer) once the notional switch has been flipped to start building the modules to require scripts.

donmccurdy:
> Awesome! Is the intent still to transform the examples/jsm/Nodes folder into a single file in examples/js/Nodes?

That would be my thought, and it probably applies to at least `postprocessing/`, `curves/`, and `lines/` too. Not sure when we want to "flip the switch" though. r107? 😇

Mugen87:
I would say yes. After finalizing `build-examples`, `modularize.js` should be removed since the JSM files are now the canonical versions.

gkjohnson:
I'll look into proposing some changes to `build-examples` to support this use soon.

examples/webgl_materials_compile.html
sunag: > Perhaps Nodes.js should be renamed index.js?

I vote to keep `Nodes.js` name, maintain the same standard of those on core like `Materials.js` files, that was inspired.
https://github.com/mrdoob/three.js/blob/dev/src/materials/Materials.js
https://github.com/mrdoob/three.js/blob/dev/src/Three.js

`src/nodes/Nodes.js`
`src/materials/Materials.js`
...

sunag:
@Mugen87 Thanks! :clap: :clap: :clap:

mrdoob:
Thanks!

PR #16921, opened 2019-06-27
Increase vector conversion function

FishOrBear:
The PR solves the following problem and writes the conversion code faster.

```
let v = new Vector2(1,2);
let v3 = v.asVector3();
```


Mugen87:
I don't think we need to enhance the API for this simple workflow.

PR #16931, opened 2019-06-27
Make STLExporter incorporate Vertex bone transformations on SkinnedMesh objects

cioddi:
…kinnedMesh objects

The previous exporter version would export all SkinnedMesh objects in resting pose even if it is animated. This will apply bone transformations to each vertex. Based on the raycasting boneTransform function. The boneTransform function can be removed from here once it is in master.

### DIFF
--- a/examples/js/exporters/STLExporter.js
+++ b/examples/js/exporters/STLExporter.js
@@ -3,6 +3,7 @@
  * @author mrdoob / http://mrdoob.com/
  * @author mudcube / http://mudcu.be/
  * @author Mugen87 / https://github.com/Mugen87
+ * @author cioddi / https://github.com/cioddi
  *
  * Usage:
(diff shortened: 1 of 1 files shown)


examples/js/exporters/STLExporter.js
```diff
@@ -97,6 +99,14 @@ THREE.STLExporter.prototype = {
 
 							vector.copy( vertices[ indices[ k ] ] ).applyMatrix4( matrixWorld );
 
+              if(
```
gkjohnson: Looks like indentation is off -- these should be tabs.

cioddi: Sorry, my editor setup messed that up. Created a new PR
https://github.com/mrdoob/three.js/pull/16942

cioddi:
New PR https://github.com/mrdoob/three.js/pull/16942

Issue #16922, opened 2019-06-27
Feature Suggestion: Support Global Uniforms


gkjohnson:
I've been using the Line example in a recent project and I've found the `resolution` uniform on `LineMaterial` a bit difficult to work with. It's per-renderer-specific and must be maintained and updated on every LineMaterial if it changes. I know a [comment mentions that the renderer could eventually set it](https://github.com/mrdoob/three.js/blob/dev/examples/js/lines/LineMaterial.js#L11) but it seems like there could be a general solution to this type of issue.

I'm proposing adding a list of global uniforms to the WebGLRenderer that override a given uniform on all materials being rendered:

```js
renderer.uniforms = {
    resolution: {
        value: new Vector2()
    }
};
```

Other uniforms such as environment maps, time incrementing variables, adn lighting variables could benefit from this pattern, as well. Unity implements something similar using the [Shader.SetGlobalFloat()](https://docs.unity3d.com/ScriptReference/Shader.SetGlobalFloat.html) etc functions.

I think my big open questions in the API design would be:
- How can an individual Material _not_ be affected by a global uniform?
- What should happen if the shape of the uniform values does does not match (Vector3 on global value, Vector2 on material)?

greggman:
just a suggestion but if you go down this path consider using a prefix, like all global properties are named `t3dg_` or something? Pretty much every project that does something like this and doesn't choose a way to separate magic stuff from user stuff ends up regretting it further down the line or at least that is my experience.

Usnul:
I'm not thrilled about this suggestion. It will only work for some cases the way you envision it, but as a result you will be submitting unnecessary uniforms in every "normal" material. Performance hit will not be huge for a couple of uniforms, but there is no justification for it in most of the standard use-cases.

gkjohnson:
I should clarify that this would be an entirely optional feature intended for the application developer to have complete control over. THREE should not automagically define global shader uniforms that override material values. This would simply be a tool of convenience to afford more control over how the scene and materials renders and materials would not require that global uniforms be set in order to function.

@greggman So in that case any prefixes would be defined as needed by the app developer and there shouldn't be any THREE magic going on.

@Usnul I don't see why unnecessary uniforms would have to be set on every material. A uniform should only be set if a material / shader requires it -- certainly the performance shouldn't be a whole lot different than now in that case?

greggman:
> @greggman So in that case any prefixes would be defined as needed by the app developer and there shouldn't be any THREE magic going on.

[...]

PR #16934, opened 2019-06-27
Object3D: Add `childadded` and `childremoved` events.

gkjohnson:
Reviving some of the ideas from #11459 this PR adds `childadded` and `childremoved` events that get dispatched from the parent when a child is added or removed.

I'd like to track when an object is added or removed from the scene anywhere in the hierarchy and there's currently no simple way to achieve that.

I can add the tests and docs if this looks merge-able.


Mugen87:
> I'd like to track when an object is added or removed from the scene anywhere in the hierarchy and there's currently no simple way to achieve that.

In this case, #8368 sounds more appropriate, right? I mean this PR only allows to track a removal or addition of direct child objects.

D504:
@Mugen87, do you think it has to work recursively?

gkjohnson:
> In this case, #8368 sounds more appropriate, right?

This PR is a lot simpler, introduces a useful new event notification and gives a lot of the same capability without the potential overhead of recursively firing events on every object in deep hierarchies.

>  I mean this PR only allows to track a removal or addition of direct child objects.

It does but it can be used to track every object addition and removal to and from the scene (or sub hierarchy) pretty easily. I've hacked this event into my application and am tracking object addition and removal like so (this is a simplified use case):
```js
    const allObjects = new Set();
    let addChildCb, removeChildCb;
    addChildCb = function(e) {
        const child = e.child;
        allObjects.add(child);
        child.traverse(c => {
            c.addEventListener('childadded', addChildCb);
            c.addEventListener('childremoved', removeChildCb);
            allObjects.add(c);
        });
    };

    removeChildCb = function(e) {
        const child = e.child;
        allObjects.delete(child);
        child.traverse(c => {
            c.removeEventListener('childadded', addChildCb);
            c.removeEventListener('childremoved', removeChildCb);
            allObjects.delete(c);
        });
    };

    addChildCb({ child: scene });
```

The only other way to really know this right now is to manually track the addition of every object or traverse the hierarchy every frame and look for differences. I think there's a lot of utility to easily and efficiently tracking which objects have been added to the world.

EliasHasle:
Do the extra events provide anything that cannot be achieved by accessing the parent of the target object of the added/removed events?

gkjohnson:
The example I showed in [the comment above](https://github.com/mrdoob/three.js/pull/16934#issuecomment-506793917) _cannot_ be achieved with just  the `add` and `remove` events. This allows a non-invasive and complete approach to tracking all of the additions and removals from a scene by notifying when children are changed. The `add` and `remove` event can't be used because you need to know about the child in order to add the event, which can't always be done because different libraries/ loaders or non-modifiable user code may add objects into the scene in an untrackable way.

My request for global uniforms (#16922) would be addressed by this PR because all Meshes with materials could be tracked and have their uniforms updated before render (traverse is slow and iterates over a lot of non-mesh objects). I feel in a lot of ways this allows a lot of flexibility and for features to be transparently added without the need for modifying core.

EliasHasle:
I am not sure I get it. Why can't you use the event `target` like this? :


    const allObjects = new Set();
    let addCb, removeCb;
    addCb = function(e) {
        const child = e.target;
        child.traverse(c => {
            c.addEventListener('added', addCb);
            c.addEventListener('removed', removeCb);
            allObjects.add(c);
        });
    };

    removeCb = function(e) {
        const child = e.target;
        child.traverse(c => {
            c.removeEventListener('added', addCb);
            c.removeEventListener('removed', removeCb);
            allObjects.delete(c);
        });
    };

    addCb({ target: scene });


For any cases where you need access to the parent, it could perhaps be an idea to have a `parent` property in the `'added'` and `'removed'` events? (In `'removed'` it cannot otherwise be retrieved, because the event is dispatched after the removal is complete.)

gkjohnson:
@EliasHasle 
Your solution does not allow for tracking additions of children to the scene, which is what I'm after. Please try it for yourself:

Using just the `add` and `remove` events the way you proposed:

```js
allObjects.add(scene);
addCb({ target: scene });
scene.add(new THREE.Object3D());

console.log(Array.from(allObjects.values()));
// logs [Scene] but we expect [Scene, Object3D]
```

And using the `childadded` and `childremoved` events I'm proposing:

```js
addChildCb({ child: scene });
scene.add(new THREE.Object3D());

console.log(Array.from(allObjects.values()));
// logs [Scene, Object3D]
```

EliasHasle:
Sorry. You are totally right. Maybe the `'childadded'` and `'childremoved'` events are generally more useful than the `'added'` and `'removed'` events? What are really the use cases for the existing ones, that cannot be covered by the proposed ones?

IRobot1:
I have some code that horizontally or vertically positions children within a group based on sort order defined in userData.  Without these events, I have to check if children.length has changed each frame to trigger if this positions need recalculating.
Please consider adding this change.  It would be really helpful.  
It doesn't need to be recursive.  Each group can notify any parent group with a childchanged event if needed.

IRobot1:
Without this change, every Object3D must listen for added and removed events...
```ts
class Thing extends Object3D {
  constructor() {
    super();
    this.addEventListener('added', () => {
      console.log('added', this.parent);
      this.parent?.dispatchEvent({ type: 'childadded', child: this });
    });

    this.addEventListener('removed', () => {
      console.log('removed', this.parent);
      this.parent?.dispatchEvent({ type: 'childremoved', child: this });
    });

  }
}

const group = new Group();
group.addEventListener('childadded', (event: any) => { console.log('child added', event) })
group.addEventListener('childremoved', (event: any) => { console.log('child removed', event) })

const thing = new Thing();
group.add(thing);
group.remove(thing);
```
Unfortunately, Object3D.remove sets parent to null before calling dispatchEvent, so the childremoved event never fires

```ts
object.parent = null;
this.children.splice( index, 1 );

object.dispatchEvent( _removedEvent );
```

artificial-jon:
What is the prognosis for this making it in?  It's a very safe and simple change, and I've had multiple instances where this functionality would be useful.

AdamEisfeld:
+1 to this, I didn't see this PR and had created this issue outlining the same solution [](https://github.com/mrdoob/three.js/issues/27563).

With the childadded / removed events, we can then intercept the addition /removal of objects to / from a Scene (via a subclass or a wrapper) and recursively set the childadded / childremoved events on every object to dispatch some event on the scene (eg "scenegraph_child_added" / "scenegraph_child_removed").

Then we can observe these scenegraph_child_added / scenegraph_child_removed events to perform some logic (eg, inspecting the class of the added object / any child objects and, if they pass some test, adding them to a list of objects to track / update each render loop).

AdamEisfeld:
An alternative approach would be to dispatch the old added / removed events globally via the document. If the new parent / old parent objects were added as parameters to this event, outside observers could listen for these events and then simply traverse the parent chain up to see if it ends on a Scene object in order to check if the scene graph has changed.

This would be a breaking change though, as users would need to start monitoring the document for the added / removed events instead of a specific object. The benefit being it would stick to just using the one event instead of adding a new one.

Mugen87 (approved):

src/core/Object3D.js
```diff
@@ -357,6 +357,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
 			this.children.push( object );
 
 			object.dispatchEvent( _addedEvent );
+			this.dispatchEvent( { type: 'childadded', child: object } );
```
Mugen87, line 360: From https://github.com/mrdoob/three.js/pull/11459#issuecomment-398116192.

I think it's okay creating an object per event however I also understand the concern for devs how perform many add and remove operations that would like to see a single object. 

Since the listeners in `dispatchEvent()` are processed in a sync fashion, I think it would be safe to have single event objects and just change the `child` reference. What do you think?

gkjohnson: I think it's good - we just have to set and then remove the objects from the cached event before and after dispatch. I'll merge this as is and then make a new PR with this.

hichemfantar:
I just discovered that this only works on direct children.

is there a way to listen for events in the entire hierarchy?

PR #16967, opened 2019-07-03

### LINKED ISSUE #16720: View FBX files exported from Autodesk Navisworks with threejs
Hi there

we are investigating the use of ThreeJS to display 3D models exported from Autodesk Navisworks.
We exported a test model from Navisworks and tried to view it using ThreeJS.
We experimented with the FBX export formats.
We’ve had success with FBX (after resaving it with Paint 3D).

We used FBXLoader. When trying to open the FBX file as exported form Navisworks, it fails to display. The error from the loader is: TypeError: Cannot read property ‘replace’ of undefined.

The console logs looks like this (I added logging the output of the loader):

THREE.WebGLRenderer 105
THREE.FBXLoader: FBX binary version: 7500
TypeError: Cannot read property ‘replace’ of undefined

When I however open the FBX file (as exported from Navisworks) with Paint 3D and save a copy of it as an FBX file, it displays correctly in ThreeJS .

console logs:
THREE.WebGLRenderer 105
THREE.FBXLoader: FBX binary version: 7400

Lewy had a look at it for us and confirmed that the FBXLoader were not able to parse the FBX file. 

Lewy requested me to open a new issue for this.

Originally submitted on discourse:
https://discourse.threejs.org/t/view-fbx-kmz-files-exported-from-autodesk-navisworks/7963/4

I attach the FBX file that was exported from Navisworks for your perusal.

Thank you
[stair.zip](https://github.com/mrdoob/three.js/files/3264510/stair.zip)

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

FishOrBear:
Solve this problem and this will not break the original code.

There are always some strange problems in FBX.

https://github.com/mrdoob/three.js/issues/16720

### DIFF
--- a/examples/jsm/loaders/FBXLoader.js
+++ b/examples/jsm/loaders/FBXLoader.js
@@ -962,7 +962,8 @@ var FBXLoader = ( function () {
 
 					}
 
-					model.name = PropertyBinding.sanitizeNodeName( node.attrName );
+					if ( node.attrName )
+						model.name = PropertyBinding.sanitizeNodeName( node.attrName );
 					model.ID = id;
 
 				}


WestLangley:
Do you need to fix that everywhere `sanitizeNodeName()` is called?

Would it be better to fix the method itself -- for example, return the string 'undefined' if the arg is undefined?

/ping @looeee

looeee:
This looks ok, but let me check it first before merging. I'll have some time over the weekend to do that.

looeee:
Thanks for waiting. This looks good 👍 

One thing - I don't think it's covered in the style guide, but we usually write if statements like this:

``` js
if ( node.attrName ) {

    model.name = PropertyBinding.sanitizeNodeName( node.attrName );

}
```

rather than 

``` js
if ( node.attrName )
    model.name = PropertyBinding.sanitizeNodeName( node.attrName );
```

@mrdoob are both styles OK?

WestLangley:
Maybe this is preferred:
``` js
if ( node.attrName !== undefined ) {

    model.name = PropertyBinding.sanitizeNodeName( node.attrName );

}
```

WestLangley:
@looeee Are you saying the answer to both of my questions is 'No'?

looeee:
@WestLangley

> Do you need to fix that everywhere sanitizeNodeName() is called?

No idea. 

However, my approach so far has only been to add checks like this when we find a file that needs them. Otherwise we'd have another hundred or so if statements littered through the loader. In this case, since `sanitizeNodeName` is only called 4 times it would be OK to add these checks preemptively if you prefer. 

> Would it be better to fix the method itself -- for example, return the string 'undefined' if the arg is undefined?

No preference on this one.

WestLangley:
>No idea.
>No preference

@looeee If I had known, I wouldn't have pinged you. Sorry.

/ping @takahirox

Mugen87:
To get things forward I suggest to add the check to each invocation of `sanitizeNodeName()` to make things more robust. I'm not a fan of enhancing `sanitizeNodeName()` so it can handle null or undefined values. The project does usually not perform this checks for performance reasons. If we start here, one could argue to enhance all math methods in the same way. So if parameter of a specific type (string, number, vector) is expected, the caller (in this context `FBXLoader`) have to ensure valid input.

@FishOrBear Are you willing to update your PR?

FishOrBear:
> To get things forward I suggest to add the check to each invocation of `sanitizeNodeName()` to make things more robust. I'm not a fan of enhancing `sanitizeNodeName()` so it can handle null or undefined values. The project does usually not perform this checks for performance reasons. If we start here, one could argue to enhance all math methods in the same way. So if parameter of a specific type (string, number, vector) is expected, the caller (in this context `FBXLoader`) have to ensure valid input.
> 
> @FishOrBear Are you willing to update your PR?


such?
```
function sanitizeNodeName(name)
{

    if (name === undefined)
        return "";

    return name.replace(/\s/g, '_').replace(reservedRe, '');

}
```

WestLangley:
@FishOrBear What @Mugen87 was trying to say is: Do not modify `sanitizeNodeName()`. Instead, check for a valid value of `name` before the function is called.

Mugen87:
Closing in favor of #17509

PR #16970, opened 2019-07-03
Examples: Clean up SelectionBox.

Mugen87:
see https://github.com/mrdoob/three.js/pull/16944#issuecomment-508142612


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev31/examples/misc_boxselection.html

WestLangley:
How about setting
```js
allSelected[ i ].material.emissive.set( 0xffffff );
```
At least it is visible.

Mugen87:
Good idea 👍

WestLangley:
Thanks!

mrdoob:
Thanks thanks!

Issue #16440, opened 2019-05-14
Promote Node-based Materials to Core and Polish


bhouston:
##### Description of the problem

@sunag did a great jobs on the node-based materials.  While he got some criticism for his PRs when he was submitted them, usually dealing with internal design of the code compiler, I've always thought his work was amazing.

We have adopted node-materials for our rendering recently.  We have found that it leads to more artistic creativity.  It also leads to less complex code base -- the current Uber materials in Three.Js are just insanely complex and hard to maintain.

Thus I would propose that we move the node-based materials into core and then we slowly move most of the examples to use the node-based materials instead of the complex do-everything uber materials.

It is a big change, but I think it moves Three.JS forward.

Some of the changes I think are needed in order to make node-based materials first class citizens of Three.JS:

- add ability to get specialized layers out of the node-based materials easily (normal, depth, velocity, etc.)  There is both an easy way and a hard way to do this.  We'll start with the easy way.
- add more advanced options to the BRDF/EDFs such as anisotropy, basic sub-surface scattering.
- add ability to read AO from a pre-pass in the BRDF/EDF so it can only affect indirect diffuse rather than being layered on top and add an example of this.
- rework the texture node a bit, it has some problematic dependencies that make the graph difficult to evaluate.  It was suggested by @jojobyte of NVIDIA to follow the model of UE4's texture node.
- have a separate clearcoat normal: https://github.com/mrdoob/three.js/issues/12867
- add trilinear mapping example using the node-materials: https://github.com/mrdoob/three.js/issues/15048
- add displacement map to the node-based BRDF/EDF (which would require a vertex shader change.)
- add bump to normal example using the node-materials.
- add transmission factor to the root BRDF/EDF and an example to three.js: https://github.com/mrdoob/three.js/issues/15941
- makes it easy to support arbitrary UV channels per map in a very simple way.  Lots of power and flexibility.  Thus negating the need for this: https://github.com/mrdoob/three.js/pull/8278
- allows for supporting world to UV mapping and other forms of procedural UV transforms.

This ties into my current proposal to glTF to adopt node-based materials:

https://docs.google.com/document/d/1Y6JFE2FV164IFDe7_cYhp2gzhSapB76fUNPgmsI6DDY/edit

I am not the first to suggest this but I can not find the previous discussion.  I think it is just a matter of time until this happens, so might as well do this now.  No other JavaScript 3D renderer has fully node-based materials. ;)

/ping @WestLangley @donmccurdy 

[...]

PR #16971, opened 2019-07-03
MeshPhysicalMaterial: Added sheen

DanielSturk:
Added sheen for cloth materials such as velvet. The BDRF used is the same as Google's Filament (NDF is Estevez's and Kulla's "Charlie" sheen + Neubelt's visibility term https://google.github.io/filament/Filament.md.html).

I also added support for sheen to the StandardNodeMaterial (@sunag)

The demo for testing is webgl_materials_sheen.html


DanielSturk:
By the way, this implementation only covers direct lighting for the time being. There's a bit of confusion/complication in regards to IBL and ambient lighting because sheen seems to be dependent on the direction of the light source

bhouston:
We should never support ambient lighting on this.

sunag:
Thank you for your contribution once again...
The parameter and example `sheen` does not seem to be working.
I believe that changes in the `core` and in the `NodeMaterial` should be sent separately but I think if possible unify this feature using nodes only without modify the core...

DanielSturk:
@sunag

> The parameter and example `sheen` does not seem to be working.

What do you mean by this? I asked my colleague tested it and it runs fine on his machine

sciecode:
@sunag 

If you are using something like "githack" to test, it won't work. 
`Three.module.js` needs to be updated, because of new `sheen` parameter. Running it locally works as intended.

bhouston:
> unify this feature using nodes only without modify the core...

The reason we are adding this is that it is a BRDF improvement and it is to align with the Enterprise PBR model:  https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel

bhouston:
This is being done in this later context of adopting PBR Next / Enterprise PBR https://github.com/mrdoob/three.js/issues/16977 and ensuring Node-graph materials are awesome https://github.com/mrdoob/three.js/issues/16440

sunag:
@DanielSturk @sciecode Yeah, I run locally and work thanks.

> The reason we are adding this is that it is a BRDF improvement and it is to align with the Enterprise PBR model: https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel

The roadmap is very good. My think is advance in BRDF using nodes only to create better support for it and facilitate the merger of PR. Once implemented, create other BRDF features will be easier and optimized since the code can be calculated modularly. `Sheen` could be one of these new BRDF nodes.

bhouston:
Usually the brdf parameters are part of the base brdf node.  Having brdf
parameters as separate nodes generally is not how brdfs are implemented.

sunag:
> Usually the brdf parameters are part of the base brdf node.  Having brdf
parameters as separate nodes generally is not how brdfs are implemented.

I was not thinking about separating the code from BRDF class, just implementing it as a process tree a bit more organized.
`brdf/SheenNode` for example...

So we can add relatives `sheen` functions if the parameter `.sheen` is really used.
Relative the performance, it is good too, even unused function adds an increased time to `program` creation.

sunag:
@mrdoob @WestLangley I do not want to block this PR, maybe I have caused this impression, it looks good to me. I can improve these details relative to the `NodeMaterial` later with other PR.

DanielSturk:
> @mrdoob @WestLangley I do not want to block this PR, maybe I have caused this impression, it looks good to me. I can improve these details relative to the `NodeMaterial` later with other PR.

examples/webgl_materials_sheen.html
```diff
@@ -0,0 +1,378 @@
+<html lang="en">
+	<head>
+		<title>Ammo.js softbody cloth demo</title>
```
Mugen87: Can you please add this example to `files.js`?  Otherwise it won't be visible in the side bar.

Mugen87:
https://raw.githack.com/DanielSturk/three.js/sheen/examples/webgl_materials_sheen.html

It seems it's not possible to control the arm with the keys like in the original [physics example](https://raw.githack.com/DanielSturk/three.js/sheen/examples/webgl_physics_cloth.html). Besides, the arm jitters right now in a strange way. At least on my system.

yomboprime:
> https://raw.githack.com/DanielSturk/three.js/sheen/examples/webgl_materials_sheen.html
> 
> It seems it's not possible to control the arm with the keys like in the original [physics example](https://raw.githack.com/DanielSturk/three.js/sheen/examples/webgl_physics_cloth.html). Besides, the arm jitters right now in a strange way. At least on my system.
> 
> If no interaction is intended, consider to remove the arm or make it static.

He has put a `Math.random()` in the arm control.

examples/webgl_materials_sheen.html
DanielSturk: Done

DanielSturk:
@Mugen87 @yomboprime I didn't expect the arm to be so controversial :p
My coworker made us this model to use instead
126e79a1512777e4966ebfeef3b0ebe7

Mugen87:
That's better^^! 

BTW: It seems that changing the `sheen` via `dat.gui` does not have a visual effect. Or do I miss something?

src/materials/MeshPhysicalMaterial.js
```diff
@@ -23,6 +23,8 @@ function MeshPhysicalMaterial( parameters ) {
 	this.clearCoat = 0.0;
 	this.clearCoatRoughness = 0.0;
 
+	this.sheen = 0.0;
```
Mugen87: It will be necessary to enhance the docs and TypeScript declaration files with this parameter. However, you can do this with a different PR after this one gets merged.

EliasHasle:
> BTW: It seems that changing `sheen` via `dat.gui` does not have a visual effect. Or do I miss something?

@DanielSturk That's right. The gui modifies properties of a separate params object and has no event handlers. The simplest solution will be something like:

    gui.add(material.uniforms.someFloatUniform, "value", 0, 1);

DanielSturk:
@EliasHasle
I'm almost certain the githack demo is broken because you're not supposed to submit the build files in PRs.

EliasHasle:
OK. Then I recommend that you do like me and create a personal github pages repository for these kinds of demos, where you can control which files are accessed. I would very much like to see this new velvet beast, or whatever it is. 😄

DanielSturk:
> OK. Then I recommend that you do like me and create a personal github pages repository for these kinds of demos

@EliasHasle  Great idea... instead I just created another personal branch for this demo :p

https://raw.githack.com/DanielSturk/three.js/sheen-demo/examples/webgl_materials_sheen.html

EliasHasle:
That is an impressive demo! Wow! :D

DanielSturk:
> That is an impressive demo! Wow!

Thank you!

Mugen87 (approved):

src/materials/MeshPhysicalMaterial.js
DanielSturk: Done

Mugen87:
@DanielSturk  Can you please resolve the merge conflicts?

Mugen87:
@mrdoob Looks ready to merge 👍

bhouston:
Is there a reason this isn't being merged?

Mugen87:
Sorry, I'm sure this PR was just overlooked. 

@DanielSturk Can you please resolve the merge conflicts one more time?

romainguy:
@DanielSturk What problems did you run into with IBL support for `sheen`? Folks on Babylon.js and I both got it to work, we can help if you'd like.

DanielSturk:
@Mugen87 Np. Merge conflicts resolved :)

DanielSturk:
> @DanielSturk What problems did you run into with IBL support for `sheen`? Folks on Babylon.js and I both got it to work, we can help if you'd like.

Thank you, I appreciate the support! I've got it working now (was a little confused by the DFG (DG) LUT but it's working now). As soon as we can get this merged, I'll create the PR for the IBL Sheen!

mrdoob:
Thanks!

mrdoob:
@DanielSturk @bhouston I have renamed the property to `sheen` to match `emissive`, `specular`, ... https://github.com/mrdoob/three.js/commit/3e9af1f44514f3bdf2d25d21757bd637498b83a1

mrdoob:
@sunag on one of my computers (MacBook Air, Early 2015, Intel HD Graphics 6000) the [sheen example](https://raw.githack.com/mrdoob/three.js/dev/examples/#webgl_materials_sheen) runs at 20fps when using the node material, and at 60fps when not using it.

In this screenshot, the fps graph shows when nodematerial was enabled and disabled:

Screenshot 2019-08-20 at 17 10 59

sunag:
> @sunag on one of my computers (MacBook Air, Early 2015, Intel HD Graphics 6000) the sheen example runs at 20fps when using the node material, and at 60fps when not using it.

This because the NodeMaterial is being compiled every frame: `nodeMaterial.needsCompile = true;` is added in render loop. I already send a PR to fix this.

mrdoob:
> This because the NodeMaterial is being compiled every frame: `nodeMaterial.needsCompile = true;` is added in render loop. I already send a PR to fix this.

Oh, what's the PR that fixes it?

sunag:
> Oh, what's the PR that fixes it?

Ops, I'm already sending, sry my english :flushed:

DanielSturk:
Thanks for the merge @mrdoob and everyone involved
Make sure to check out my follow up PR for IBL support @romainguy https://github.com/mrdoob/three.js/pull/17303

donmccurdy:
The glTF sheen extension ([KHR_materials_sheen](https://github.com/KhronosGroup/glTF/pull/1688)) is nearly complete, although still welcoming feedback for a bit longer. To implement in MeshPhysicalMaterial, it would require:

- sheen
- sheenMap
- sheenRoughness
- sheenRoughnessMap

The two maps use `.rgb` and `.a` channels respectively; they can be combined.

Issue #15866, opened 2019-03-01
FBXLoader loads just one morphtarget


nicolasdwolf:
##### Description of the problem

This fbx file has 45 morphtargets (checked with blender) but fbxloader loads just the last one (base_head)

provided fbx, texture and test html: 
[model.zip](https://github.com/mrdoob/three.js/files/2917933/model.zip)

![screenshot_20190301_010337](https://user-images.githubusercontent.com/48113737/53617146-b18dcd80-3bc4-11e9-85bb-3b335d3973da.png)

![screenshot_20190301_014344](https://user-images.githubusercontent.com/48113737/53617151-bbafcc00-3bc4-11e9-8dcc-f7a3c12c8a5c.png)

##### Three.js version

- [ ] Dev
- [x] r102
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

PR #16991, opened 2019-07-09
FBXLoader: load multiple morph targets

looeee:
Fix for #15866

It seems like morphtargets can be defined at two places for a geometry:

1. `deformers.morphTargets`

2. `deformers.morphTargets[ ID ].rawTargets`

Note that I'm now assuming that once `deformers.morphTargets[ ID ]` has been matched to a geometry ID, then all the `rawTargets` will match that geometry. All models I have for testing work fine with this assumption, but we might need to add an additional check here at some point.


mrdoob:
Thanks!

PR #16992, opened 2019-07-09
MipMap -> Mipmap

WestLangley:
Follow-up to #16981.

This PR updates the core library to use the new nomenclature. It also updates the docs.

I do not have the ability to test TS, but the examples appear to work, so both the old and the new nomenclature is supported.

I assume we want to continue to support both the old and the new constant names -- at least for awhile.

I'll update the examples to use the new nomenclature if this goes well.

### DIFF
--- a/docs/api/en/constants/Textures.html
+++ b/docs/api/en/constants/Textures.html
@@ -87,11 +87,11 @@
 		<h2>Minification Filters</h2>
 		<code>
 		THREE.NearestFilter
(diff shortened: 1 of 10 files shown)


WestLangley:
Let's give this a try.

@mrdoob Requires a rebuild

mrdoob:
Thanks! Done!

WestLangley:
>@mrdoob requires a rebuild

hehe. poor guy.

Issue #15941, opened 2019-03-10
Feature: Add transmissionFactor to Materials


gkjohnson:
When rendering a transparent object in THREE the specular highlight fades along with the diffuse surface of the material. I'd expect it retain the same brightness considering specular is supposed to be more of a mirror like reflection. Here's a picture of a sphere with the Standard material in THREE:

![image](https://user-images.githubusercontent.com/734200/54081450-06fb6600-42ba-11e9-9f9d-826012f1a861.png)

*metalness: 0, roughness: 0.3, opacity: 0.15*

And here's what a similar sphere looks like in Unity, which is more what I'd expect:

![image](https://user-images.githubusercontent.com/734200/54081475-43c75d00-42ba-11e9-9d1f-2cac293ef0c7.png)

*metallic: 0, smoothness: 0.8, opacity: 11 / 255*

As an aside I'm a bit surprised by the difference in darkness of the transparent objects between THREE and Unity. The THREE material has an opacity of 0.15 while the Unity material has an opacity of 0.043 (11 / 255) and yet they look the same. Switching Unity's color space from Linear to Gamma makes the results look more similar with the same with the same opacity. Is this a limitation of blending in the current WebGL APIs?

Thanks!

WestLangley:
See #5810, #8245, #8255.

gkjohnson:
Thanks! I figured this had been discussed previously but I couldn't find where -- it looks like https://github.com/mrdoob/three.js/pull/8276 is the last PR in the "premultiplied alpha" saga.

I may be missing something but is there a reason that it isn't enabled by default? From reading those issues it sounded like that's what was going to happen.

gkjohnson:
Even with `premultipliedAlpha = true` things don't look correct. Here's a another comparison with Unity (Unity on the left, THREE on the right):

![image](https://user-images.githubusercontent.com/734200/54089664-98a1bc80-4328-11e9-8f70-d3b71435f731.png)

**Unity** *color: red, opacity: 0, metalness: 0, smoothness: 0.8*
**THREE** *color: red, opacity: 0, metalness: 0, roughness: 0.2, premultipliedAlpha: true*

And what it looks like when objects are overlapping:

![image](https://user-images.githubusercontent.com/734200/54089755-9d1aa500-4329-11e9-8b4a-cbc776f93bdb.png)

**Unity** *opacity: 25 / 255*
**THREE** *opacity: 0.1*

All other settings are the same as the previous image.

[...]

PR #16996, opened 2019-07-09
Added transparency to StandardNodeMaterial

DanielSturk:
In an effort to align more with the Enterprise PBR model (https://github.com/mrdoob/three.js/issues/16977) I've implemented transparency (https://github.com/mrdoob/three.js/issues/15941).
3989eec928d4a7b990b50638e99f9fe6

Transparency differs from opacity in that transparency only hides the diffuse lighting (more physically accurate), whereas opacity completely fades the material (non-physical).
transparency vs opacity (taken from https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec.md.html)

Admittedly, this is somewhat hacked in, and mostly serves as a proof of concept for accelerating the changes towards Enterprise PBR. I'll spend the remainder of this post explaining what should be changed.

The blending is straight forward: at the end of the fragment shader, opacity still affects the final alpha, but an additional 'transparency' term is added which only affects the opacity of the diffuse lighting. For example, when transparency=1, the only visible light is coming from the specular (see the above Enterprise PBR chart). Unfortunately, there is no alpha value pertaining to the specular stored at this point, so I must infer one by **re**-calculating the fresnel term, and dividing it out of the specular light.

The obvious change that must be made is that the F term of the DFG should never be included in the `directSpecular` or `indirectSpecular` values, and it should only be calculated once at the end of the shader, where it is treated as the alpha term of the specular light.

Unfortunately I am struggling to seperate the fresnel term, mainly due to the opaque environment BRDF (https://github.com/mrdoob/three.js/issues/7489). F can easily be factored out of `BRDF_Specular_GGX()`, but not `BRDF_Specular_GGX_Environment()` or `BRDF_Specular_Multiscattering_Environment()`.

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -311,7 +311,7 @@ function WebGLRenderer( parameters ) {
 
 	// vr
 
-	var vr = ( typeof navigator !== 'undefined' && 'xr' in navigator && 'supportsSession' in navigator.xr ) ? new WebXRManager( _this ) : new WebVRManager( _this );
+	var vr = ( typeof navigator !== 'undefined' && 'xr' in navigator && 'supportsSession' in navigator.xr ) ? new WebXRManager( _this, _gl ) : new WebVRManager( _this );
 
 	this.vr = vr;
 
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -76,7 +76,6 @@ function WebGLRenderer( parameters ) {
 	// public properties
 
 	this.domElement = _canvas;
-	this.context = null;
 
 	// Debug configuration container
 	this.debug = {
@@ -299,7 +298,6 @@ function WebGLRenderer( parameters ) {
 
 		info.programs = programCache.programs;
 
-		_this.context = _gl;
 		_this.capabilities = capabilities;
 		_this.extensions = extensions;
 		_this.properties = properties;
@@ -2276,7 +2274,6 @@ function WebGLRenderer( parameters ) {
 
 		uniforms.clearCoat.value = material.clearCoat;
 		uniforms.clearCoatRoughness.value = material.clearCoatRoughness;
-		uniforms.transparency.value = material.transparency;
 
 	}
 
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -2276,6 +2276,7 @@ function WebGLRenderer( parameters ) {
 
 		uniforms.clearCoat.value = material.clearCoat;
 		uniforms.clearCoatRoughness.value = material.clearCoatRoughness;
+		uniforms.transparency.value = material.transparency;
 
 	}
 
--- a/src/renderers/shaders/ShaderChunk.d.ts
+++ b/src/renderers/shaders/ShaderChunk.d.ts
@@ -88,7 +88,7 @@ export let ShaderChunk: {
 	points_vert: string;
 	shadow_frag: string;
 	shadow_vert: string;
-	transparentcy_pars_fragment: string;
+	transparency_pars_fragment: string;
 
 	premultiplied_alpha_fragment: string;
 	project_vertex: string;
--- a/src/renderers/shaders/ShaderChunk.d.ts
+++ b/src/renderers/shaders/ShaderChunk.d.ts
@@ -88,6 +88,7 @@ export let ShaderChunk: {
 	points_vert: string;
 	shadow_frag: string;
 	shadow_vert: string;
+	transparentcy_pars_fragment: string;
 
 	premultiplied_alpha_fragment: string;
 	project_vertex: string;
--- a/src/renderers/shaders/ShaderChunk/transparency_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/transparency_pars_fragment.glsl.js
@@ -1,0 +1,23 @@
+export default /* glsl */`
+uniform float transparency;
+
+// makes surface transparent without hiding the specular term
+vec4 combineLight(const in vec3 diffuseLight, const in vec3 specularLight, const in GeometricContext geometry, const in PhysicalMaterial material) {
+
+	vec3 fresnel = BRDF_Specular_GGX_Environment(geometry, material.specularColor, material.specularRoughness);
+
+  // since we can't have per-channel opacity blending, we must settle for a single float (reflectance)
+  float reflectance = (fresnel.r + fresnel.g + fresnel.b) * (1. / 3.);
+
+  vec4 fragColor = vec4(specularLight, reflectance); // specularLight is already premultiplied by fresnel
+
+	float diffuseAlpha = (1. - reflectance) * (1. - transparency);
+	fragColor += vec4(diffuseLight * diffuseAlpha, diffuseAlpha);
+
+	// the above math is performed in premultiplied alpha
+  fragColor.rgb /= fragColor.a;
+
+  return fragColor;
+
+}
+`;
--- a/src/renderers/shaders/ShaderChunk/transparency_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/transparency_pars_fragment.glsl.js
@@ -7,7 +7,7 @@ uniform float transparency;
 	vec4 diffuseFinal = vec4( diffuseLight * ( 1. - transparency ), 1. - transparency );
 
 	vec3 fresnel = BRDF_Specular_GGX_Environment( geometry, material.specularColor, material.specularRoughness );
-	float fresnelApprox = ( fresnel.r + fresnel.g + fresnel.b ) * ( 1. / 3. ); // since we can't have per-channel opacity blending, we must approximate with a single blending factor
+	float fresnelApprox = linearToRelativeLuminance( fresnel ); // since we can't have per-channel opacity blending, we must approximate with a single blending factor
 
 	vec4 specularFinal = vec4( specularLight, fresnelApprox ); // specularLight is already premultiplied by fresnel (TODO: factor fresnel out of BSDF)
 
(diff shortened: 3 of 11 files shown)


donmccurdy:
Thank you for contributing this! ❤️

I’m traveling this week, but would be happy to review it next week.

src/renderers/shaders/ShaderChunk.d.ts
```diff
@@ -88,6 +88,7 @@ export let ShaderChunk: {
 	points_vert: string;
 	shadow_frag: string;
 	shadow_vert: string;
+	transparentcy_pars_fragment: string;
```
mrdoob: Typo, should be:

```
transparency_pars_fragment: string;
```

src/renderers/shaders/ShaderChunk/transparency_pars_fragment.glsl.js
```diff
@@ -0,0 +1,23 @@
+export default /* glsl */`
+uniform float transparency;
+
+// makes surface transparent without hiding the specular term
+vec4 combineLight(const in vec3 diffuseLight, const in vec3 specularLight, const in GeometricContext geometry, const in PhysicalMaterial material) {
+
+	vec3 fresnel = BRDF_Specular_GGX_Environment(geometry, material.specularColor, material.specularRoughness);
+
+  // since we can't have per-channel opacity blending, we must settle for a single float (reflectance)
```
mrdoob: This file seems like it needs some indentation clean up.

WestLangley:
>Admittedly, this is somewhat hacked in, and mostly serves as a proof of concept

Agreed. The rendering is pretty, but I am unable to comprehend the model on which your changes are based. Do you have a reference for your modeling approach?

WestLangley:
Babylon has another solution. Live demo: https://www.babylonjs-playground.com/#19JGPR#13

Instead of having both `material.opacity` and `material.transparency`, as in this PR, Babylon has some flags that can be set: `useRadianceOverAlpha` and/or `useSpecularOverAlpha`.

WestLangley:
`MeshPhysicalMaterial` was designed to model plastics and metals.

Perhaps a better approach would be to create `THREE.PBRGlassMaterial` and use a custom shader appropriate for modeling glass.

Related reading  [ADOBE_materials_thin_transparency](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/ADOBE_materials_thin_transparency).

mrdoob:
`MeshPhysicalThinMaterial`?

DanielSturk:
> Agreed. The rendering is pretty, but I am unable to comprehend the model on which your changes are based. Do you have a reference for your modeling approach?

I appreciate [the glTF paper](ADOBE_materials_thin_transparency) you linked. [This part](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/ADOBE_materials_thin_transparency#implementing-absorption) happens to describes what I'm actually trying to do, which is mix the diffuse/specular based on the fresnel. This may have not been obvious, so I went back and cleaned up my code
a6553b3abae64a4ba5d4c9d596627973
There's 2 major differences between with how we're doing it, however.
1. We should be factoring fresnel out of the BRDF, only incorporating it at the end for mixing. Right now I'm essentially re-calculating it at the end and factoring it out of the specular term (hacky)
2. Their fresnel term F is a vec3 (as is ours), and is used for mixing with the transmitted light. Since we're using WebGL blending, we can't do a component-wise blending with the background, so I must estimate with `fresnelApprox`. Real materials can have non-grayscale transparency, of course.

DanielSturk:
> Instead of having both `material.opacity` and `material.transparency`, as in this PR, Babylon has some flags that can be set: `useRadianceOverAlpha` and/or `useSpecularOverAlpha`.

Although I appreciate the elegance of Babylon's method, I support the [Enterprise PBR model](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec.md.html) (see https://github.com/mrdoob/three.js/issues/16977), which allows for separate transparency/opacity values.
Opacity and transparency are occasionally used in conjunction. Usually for fading out an object with transparent glass/whatever, or like this 
refractive_sphere_cutout (taken from [here](https://docs.arnoldrenderer.com/display/A5AFHUG/Transmission+and+Opacity))

WestLangley:
@DanielSturk Thank you for the link to the [Arnold doc](https://docs.arnoldrenderer.com/display/A5AFHUG/Transmission+and+Opacity). I think it explains quite well the justification for both transparency and opacity. The fact that such a detailed explanation is required leads me to believe the API was confusing to users, but I think the API is OK.

I also support adding the feature you are trying to implement.

I do not understand your physical modeling, however. Am I correct in guessing that this is something you developed on you own?

If that is the case, I suggest you use `linearToRelativeLuminance()` to compute an average fresnel value. (Doing so does not resolve my concerns, however. )

This is my attempt to reverse-engineer your code. Is it accurate?
```js
vec3 fresnel = BRDF_Specular_GGX_Environment( ... );

float F = luminance( fresnel );

vec4 fragColor = ( 1 - F ) * vec4( diffuseRGB * ( 1 - T ), 1 - T ) + F * vec4( specularRGB, F );

fragColor.rgb /= fragColor.a;

return fragColor;
```

DanielSturk:
@WestLangley Thanks for the help. I knew there was multiple ways of calculating luminescence but I didn't know we had a standard one in Three.JS, I'll add that in. As far as I can tell your code is correct (multiply diffuse by transparency, mix diffuse/specular using fresnel).

Also your are correct that I "developed" it myself, in relation to the luminescence thing. After some sleuthing I found that [Babylon does something similar](https://github.com/BabylonJS/Babylon.js/blob/650173395f9dffa614f300367a57ac3dea2d4d03/src/Shaders/default.fragment.fx#L399), but it seems less physical/accurate than our method, because it looks like they are calculating specular alpha from the specular light (wouldn't this mean brighter highlights are more opaque? seems wrong)

WestLangley:
@DanielSturk 

>As far as I can tell your code is correct

Just to clarify, it is not my code, it is a representation of your code. Unfortunately, I cannot understand any mathematical or physical justification for your equation. You are also dividing by `fragColor.a`, which can be zero.

Like I said, I support your efforts, but as of now, not the implementation. I do not have an alternate solution, however.

>wouldn't this mean brighter highlights are more opaque? seems wrong

Yes, that is what they claim. I think their claim is an accurate one. Their solution is to increase the opacity of the specular highlights using some heuristic.

donmccurdy:
The API surface related to alpha and transparency is becoming complex:

- `.opacity`
- `.transparency`
- `.transparent`
- `.alphaTest`
- `.alphaMap`
- `.refractionRatio`
- `.envMap` + `THREE.CubeRefractionMapping`

In particular, `.transparent` and `.transparency` have no relation to one another, which does seem confusing. The distinction between transparency and opacity is likely to cause confusion as well, although I agree with the reasons for that separation. For another reference, here is Blender's Principled BSDF node:

Screen Shot 2019-07-20 at 2 04 14 PM

Note the distinct "alpha" and "transmission" sockets.

My first, and simplest, suggestion would be that perhaps we should implement this only on NodeMaterial first, to buy some time to experiment with techniques and API. I'm nervous about the suggestion of a separate material type for physically-based transparency — Blender and glTF are both using a single ubermaterial.

We've also discussed having an equivalent of dither temporal AA or alpha hashing, which would throw another complication into the existing material API but is easier with NodeMaterial's `.clip` property.

src/renderers/WebGLRenderer.js
```diff
@@ -313,7 +311,7 @@ function WebGLRenderer( parameters ) {
 
 	// vr
 
-	var vr = ( typeof navigator !== 'undefined' && 'xr' in navigator && 'supportsSession' in navigator.xr ) ? new WebXRManager( _this ) : new WebVRManager( _this );
+	var vr = ( typeof navigator !== 'undefined' && 'xr' in navigator && 'supportsSession' in navigator.xr ) ? new WebXRManager( _this, _gl ) : new WebVRManager( _this );
```
DanielSturk: I didn't make these changes I just pulled the dev branch

src/renderers/webgl/WebGLProgram.js
```diff
@@ -238,7 +238,7 @@ function unrollLoops( string ) {
 
 function WebGLProgram( renderer, extensions, code, material, shader, parameters, capabilities ) {
 
-	var gl = renderer.context;
+	var gl = renderer.getContext();
```
DanielSturk: I didn't make these changes I just pulled the dev branch

DanielSturk:
@donmccurdy 

I'm totally satisfied with implementing this in the `StandardNodeMaterial` first (in fact I have no interest in non-node materials), so I undid the `MeshPhysicalMaterial` changes and applied the `StandardNodeMaterial` changes.

I don't have a solution for the names. It's already convoluted enough. That's one nice thing about node materials is having significantly less attributes (eg. alpha and transparency, but no .alphaMap, no .transparent (at the user level))

Also I don't see the point of adding yet another material, specifically for transparency. I think fewer materials is better

sunag:
@DanielSturk I found these changes great, this is exactly what I suggest here https://github.com/mrdoob/three.js/pull/16971#issuecomment-508847318, you can do this in `sheen` PR too?. 

To isolate this change completely of the `core` for now, you can transform `transparency_pars_fragment` in a `node` like this:

https://github.com/mrdoob/three.js/blob/dev/examples/jsm/nodes/bsdfs/RoughnessToBlinnExponentNode.js 

Removing this of `ShaderChuck`.

WestLangley:
The model makes no mathematical sense and produces bizarre results. Here is what happens when the `envMapIntensity` is reduced: the material turns black.

Screen Shot 2019-07-25 at 12 39 02 PM

It should be clear.

Please do not try to develop your own PBR shading models. Implement models you can reference elsewhere -- models that, at a minimum, are physically plausible.

DanielSturk:
@WestLangley I didn't think that was the desired behavior, since currently (including with opacity), `envMapIntensity` darkens the appearance of the material.
dd14a62b80717519ef857ccf12a3c395
Anyways, I see you've created your own PR. Ping me if you'd like me to close this one

WestLangley:
@DanielSturk Sorry, I do support your efforts, but I cannot support your model.

Perhaps you can have a look at #17114 and try to break it. That is the only way I think we will be able to arrive at a model that is reasonably acceptable.

Mugen87:
Enhancing the node material should based on #17114 now. It's probably better to create a new PR for this. Closing.

PR #17014, opened 2019-07-11
Water2: Added logarithmic depth buffer support

WestLangley:
As requested in https://github.com/mrdoob/three.js/issues/13963#issuecomment-510531549.


mrdoob:
Thanks!

WestLangley:
dang... I modified the jsm file. :/

PR #17015, opened 2019-07-11
Water2: Added logarithmic depth buffer support

WestLangley:
This time for the  js version, not the jsm version.


mrdoob:
Thanks!

PR #17024, opened 2019-07-13
Updated usage section with a cdn

praisedpk:
PageCDN creates a very tightly compressed `114KB` sized js file for three.min.js . Its a valuable addition to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,14 @@ The aim of the project is to create an easy to use, lightweight, 3D library with
 
 ### Usage ###
 
-Download the [minified library](http://threejs.org/build/three.min.js) and include it in your HTML, or install and import it as a [module](http://threejs.org/docs/#manual/introduction/Import-via-modules),
-Alternatively, see [how to build the library yourself](https://github.com/mrdoob/three.js/wiki/Build-instructions).
+Download the [minified library](http://threejs.org/build/three.min.js) and include it in your HTML:
 
 ```html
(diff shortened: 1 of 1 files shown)


mrdoob:
This looks like spam.

praisedpk:
Ooops... @mrdoob  No it is not spam. Its the best performing public CDN so far with very aggressive performance optimizations. You can test the CDN yourself, or I can provide more details if you want.

Thanks.

mrdoob:
Seems to be your business though?

praisedpk:
The three.js library is hosted as part of Public CDN that is totally free
and will always be free. There are hundreds of other libraries and themes
that are available for free.

There is a private CDN too that is paid one.

---
Thanks
Hamid Sarfraz

praisedpk:
The reason we created a Public CDN as part of our plan was because we
thought existing Public CDN solutions miss several optimizations that can
be achieved easily through better design. Lot more information about this
is available on the site.
---
Thanks
Hamid Sarfraz

praisedpk:
@mrdoob if you need more information, I can provide it here.

We also offer free private cdn to non-profit orgs or websites that publish
opensource software: https://pagecdn.com/about/non-profits

All this is to clarify that the pull request was not spam.

We are willing to work closely with you. And, your feedback is welcome.

Thanks.

mrdoob:
I don't think promoting your business by sending PRs to project's READMEs on GitHub is good etiquette...

PR #17043, opened 2019-07-17
Unify setting postprocessing pass size in non-module code

Oletus:
This completes some earlier work that did this change only in modularized code.


mrdoob:
Thanks!

PR #17057, opened 2019-07-19
GLTF file with KHR_MATERIALS_UNLIT Extension

wemamawe:
when i use GLTFLoader, i find a bug. the gltf file used the KHR_MATERIALS_UNLIT,but didn't mark in json.extensionsUsed. The EXTENSIONS.KHR_MATERIALS_UNLIT is a key of json.materials.extensions. so i add some code, and had a success test .

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -223,7 +223,48 @@ THREE.GLTFLoader = ( function () {
 				}
 
 			}
+			if ( json.materials&&json.materials.length>0 ) {
+			    for ( var i = 0; i < json.materials.length; ++ i ) {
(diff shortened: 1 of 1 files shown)


wemamawe:
选区_831
This is the json of my GLTF file

zeux:
This .gltf file violates the following part of the glTF standard:

> All extensions used in a glTF asset must be listed in the top-level extensionsUsed array object, e.g.,

It looks like the generator (cesiumlab?) needs to be fixed to specify KHR_materials_unlit in the extension list instead.

donmccurdy:
@zeux is correct – a glTF file cannot use an extension without _also_ listing it in `extensionsUsed`, and we should not maintain code that attempts to work around mistakes in invalid files here. I would recommend filing a bug against the Cesium converter.

PR #17058, opened 2019-07-19
TTFLoader: Fixed traverse logic

troy351:
- using `font.encoding.cmap.glyphIndexMap` instead of `font.glyphs.glyphs` for traverse all glyphs.

Note that `glyph.encoding` not always mapping to the corrent corresponding character especially for glyphs which has multiple encodes.

image
The glyph above has unicode `12092` which mapping to an unreadable character, but it's actually mapping to a Chinese character `心` which unicode is `24515`, you may see it at the second member of `unicodes`.

Actually `font` itself has the `charToGlyph` function which is using `font.encoding.cmap.glyphIndexMap` for mapping.

- Using `String.fromCodePoint` instead of `String.fromCharCode` for better unicode support.
image

- Fixed incorrectly getting fontFamily name error 
- Add an error log for `THREE.Font`


mrdoob:
Thanks!

PR #17062, opened 2019-07-20
OrbitControls: updated default dampingFactor

WestLangley:
As mentioned in https://github.com/mrdoob/three.js/pull/17060#issue-299580212, with the improvement in the damping logic, a new default value of 0.05 for `dampingFactor` is appropriate.

In fact, this new value makes much more sense.


WestLangley:
Ephemeral test links

old
https://threejs.org/examples/misc_controls_orbit.html

new
https://raw.githack.com/westlangley/three.js/dev_damping_factor/examples/misc_controls_orbit.html

Mugen87:
Much better!

Mugen87 (approved):

WestLangley:
Sweet. I'll leave the test links for awhile.

mrdoob:
Thanks!

PR #17063, opened 2019-07-21
BufferAttribute: Introduce dispose().

### LINKED ISSUE #15261: add dispose() to BufferAttribute

Mugen87:
This PR replaces #15308. Fixed #15261.

A first step to make `BufferAttribute` fixed sized. 

The new policy should be: _If you need to resize buffer data, you have to create a new instance of `BufferAttribute`._


yaoyao-cn:
@Mugen87 sorry for the delay,  is it necessary to resolve the conflicts caused by #15308 now ? if it is not necessary i will delete the relevant branch.

Mugen87:
Resolved the merge conflicts.

FMS-Cat:
I think I need this right now

Mugen87:
@mrdoob Any chances to merge this PR? If not, do you mind sharing your concerns about this change? Maybe we can find a different solution if this one is not satisfying?

Mugen87:
I'll make a PR on a current code base if this change is wanted. Resolving the merge conflicts seems more risky than starting from scratch^^.

OndrejSpanel:
Note: there is currently no method to properly dispose `instanceMatrix` and `instanceColor` in the `InstancedMesh`.

I think this is really a glaring hole in the resource management. A proper cleanup should always be possible. Someone may have a scene with a lot of dynamic instancing going on (I am now experimenting with exactly such scene) and the missing cleanup causes slowing growing allocation of WebGL resources in such case.

Mugen87:
After a deeper analysis I do not vote to introduce `BufferAttribute.dispose()` anymore since this easily breaks code in `WebGLBindingStates`. Geometry data are managed as a single entity. Calling `dispose()` on attribute level and thus deleting the WebGL buffer would not automatically update the respective VAO.

`InstancedMesh.dispose()` has to be implemented without `BufferAttribute.dispose()`.

PR #17067, opened 2019-07-22
WebVRManager: Fix VR camera before `VR ENTER`

hiroebe:
With my smartphone, VR does not work before pressing `VR ENTER` button made by `WEBVR.createButton`.
However when I tried an older release, it worked successfully even before `VR ENTER`.

This happens when `device !== null` and `device.isPresenting === false`, and it seems to be broken with the change in #15856.
So I reverted the change, and now it works successfully.

### DIFF
--- a/src/renderers/webvr/WebVRManager.js
+++ b/src/renderers/webvr/WebVRManager.js
@@ -256,7 +256,7 @@ function WebVRManager( renderer ) {
 
 		var userHeight = referenceSpaceType === 'local-floor' ? 1.6 : 0;
 
-		if ( isPresenting() === false ) {
+		if ( device === null ) {
 
 			camera.position.set( 0, userHeight, 0 );
 			camera.rotation.set( 0, 0, 0 );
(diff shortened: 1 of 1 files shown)


Mugen87:
> VR does not work before pressing VR ENTER button made by WEBVR.createButton.

What do you mean with "not work"? Are you referring to the missing "magic window" effect? (see #13634)

hiroebe:
Yes.
I'm sorry but I don't have enough knowledge to understand the PR you've shown.
Will the PR enable the magic window effect? And will it be merged?

Mugen87:
I don't think the PR is mergeable without rework it. Too many things have changed in `WebGLRenderer`. I just wanted to know if you are looking for the magic window effect.

hiroebe:
I'm looking for the magic window effect as you said.
I sent this PR because the effect is available in the older version and I considered it as the expected behavior.

Mugen87:
I don't think your PR is acceptable since essentially revert a fix for another issue. Restoring the magic window effect needs another approach.

I'm not sure why #13634 was not getting merged at that time. Maybe @mrdoob knows more about that topic.

hiroebe:
OK. I understand this PR is not good enough.
I will gather more information about the topic.
Thanks for giving me a lot of information!

PR #17073, opened 2019-07-23
Use onBeforeCompile to customize shader for instancing

Oletus:
Replacing shaders in THREE.ShaderLib is a bad practice, as it interferes with normal operation of built-in materials. When materials need to be customized, it's better to use onBeforeCompile.

### DIFF
--- a/examples/webgl_buffergeometry_instancing_lambert.html
+++ b/examples/webgl_buffergeometry_instancing_lambert.html
@@ -84,6 +84,7 @@
 			`
 			#define LAMBERT
 
+			// instanced
 			attribute vec3 instanceOffset;
 			attribute vec3 instanceColor;
 			attribute float instanceScale;
@@ -128,8 +129,8 @@
 				#include <defaultnormal_vertex>
 
 				#include <begin_vertex>
-				// position instanced
 
+				// position instanced
 				transformed *= instanceScale;
 				transformed = transformed + instanceOffset;
 
--- a/examples/webgl_buffergeometry_instancing_lambert.html
+++ b/examples/webgl_buffergeometry_instancing_lambert.html
@@ -34,65 +34,58 @@
 		import { OrbitControls } from './jsm/controls/OrbitControls.js';
 		import { Curves } from './jsm/curves/CurveExtras.js';
 
-		var customDepthToRGBAShader = { // this is a cut-and-paste of the depth shader -- modified to accommodate instancing for this app
+		// this is a cut-and-paste of the depth shader -- modified to accommodate instancing for this app
+		var customDepthVertexShader = 
+			`
+			// instanced
+			#ifdef INSTANCED
 
-			uniforms: THREE.ShaderLib.depth.uniforms,
+				attribute vec3 instanceOffset;
+				attribute float instanceScale;
 
-			vertexShader:
-				`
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

examples/webgl_buffergeometry_instancing_lambert.html
```diff
@@ -292,15 +285,16 @@
...
 
 
 			// custom depth material - required for instanced shadows
 
-			var shader = THREE.ShaderLib[ 'customDepthRGBA' ];
-
-			var uniforms = THREE.UniformsUtils.clone( shader.uniforms );
+			var uniforms = THREE.UniformsUtils.clone( customDepthToRGBAShader.uniforms );
 
 			var customDepthMaterial = new THREE.ShaderMaterial( {
```
sciecode: Custom shader can also be replaced with `THREE.MeshDepthMaterial` and explicitly modified with `onBeforeCompile` like the other suggestion.

sciecode (commented):
If we're no longer modifying `ShaderLib` we might as well hide the unnecessary shader code.

examples/webgl_buffergeometry_instancing_lambert.html
```diff
@@ -292,15 +285,16 @@
 
 			} );
 
+			material.onBeforeCompile = function( shader ) {
+				shader.vertexShader = customLambertVertexShader;
```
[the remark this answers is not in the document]
WestLangley: I created the example this way so users could clearly see the differences without having to assemble the shader chunks manually to understand the overall logic.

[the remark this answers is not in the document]
sciecode: Well, using `vertexShader.replace(<chunk>, ...)` you can still clearly see the differences, without needing to copy the entire shader. But I can see the reason for wanting to keep it this way.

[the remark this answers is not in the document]
WestLangley: I'll rephrase it to what I meant to say...

I created the example this way so users could easily read the shader without having to assemble the shader chunks manually.

examples/webgl_buffergeometry_instancing_lambert.html
Oletus: Done!

WestLangley: So, now, what is the point of `#ifdef INSTANCED`?

Oletus: True, it's not really needed anymore, removed it.

WestLangley: Do you think it is still easy to see the changes required to support instancing?

sciecode: I think so. I would just add the missing `//instanced` on `LambertVertexShader` and it should still be clear what are the changes necessary.

Oletus: Added the // instanced comment.

Mugen87:
@mrdoob This one looks good!

WestLangley:
Thanks!

mrdoob:
Thanks!

PR #17075, opened 2019-07-23
GridHelper: give different colors to X-centerLine and Z-centerLine

robyanne92:
With this modification I added a third parameter to the GridHelper construction, giving the possibility to set different colors to X-centerLine and Z-centerLine, like in Blender 2.80. This could be useful to mark the X-centerLine with the color of x-axis and the Z-centerLine with the color of z-axis.

image

### DIFF
--- a/docs/api/en/helpers/GridHelper.html
+++ b/docs/api/en/helpers/GridHelper.html
@@ -32,8 +32,9 @@
 		<p>
 		size -- The size of the grid. Default is 10. <br />
 		divisions -- The number of divisions across the grid. Default is 10. <br />
-		colorCenterLine -- The color of the centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444 <br />
-		colorGrid -- The color of the lines of the grid. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x888888
+		colorXCenterLine -- The color of the X-centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444 <br />
+		colorGrid -- The color of the lines of the grid. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x888888 <br />
+		colorZCenterLine -- The color of the Z-centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444
 		</p>
 		<p>
 		Creates a new [name] of size 'size' and divided into 'divisions' segments per side. Colors are optional.
--- a/src/helpers/GridHelper.js
+++ b/src/helpers/GridHelper.js
@@ -30,17 +30,20 @@ function GridHelper( size, divisions, color1, color2, color3 ) {
 
 		var color = i === center ? color1 : color2;
 
-		if(i === center){
+		if( i === center ) {
+			
 			color1.toArray( colors, j ); j += 3;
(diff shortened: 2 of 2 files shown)


docs/api/en/helpers/GridHelper.html
```diff
@@ -32,8 +32,9 @@ <h3>[name]( [param:number size], [param:Number divisions], [param:Color colorCen
 		<p>
 		size -- The size of the grid. Default is 10. <br />
 		divisions -- The number of divisions across the grid. Default is 10. <br />
-		colorCenterLine -- The color of the centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444 <br />
-		colorGrid -- The color of the lines of the grid. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x888888
+		colorXCenterLine -- The color of the X-centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444 <br />
+		colorGrid -- The color of the lines of the grid. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x888888 <br />
+		colorZCenterLine -- The color of the Z-centerline. This can be a [page:Color], a hexadecimal value and an CSS-Color name. Default is 0x444444
```
Mugen87: Unfortunately, the parameter order is now confusing. It should be at least: `colorXCenterLine`, `colorZCenterLine`, `colorGrid`. Or even `colorGrid`,  `colorXCenterLine`, `colorZCenterLine`. However, this will break backwards compatibility. 

In any event, I think we can't accept the change like that.

WestLangley:
I think this would be a useful new feature.

We can change the API in such a way the args can be checked and backwards-compatibility can be maintained.
```js
var options = { color: 0xffffff, colorXAxis: 0xff0000, colorZAxis: 0x0000ff );
new THREE.GridHelper( size, divisions, options );
```
Or, we can just add a method to reset the z-axis color:
```js
gridHelper.setZAxisColor( color );
```

Or, since it is just a helper, we can change order of the arguments and update the migration guide:
```js
new THREE.GridHelper( size, divisions, color, colorXAxis,  colorZAxis );
```

My preference is the last option.

@mrdoob will have to make this decision.

Mugen87:
> Or, since it is just a helper, we can change order of the arguments and update the migration guide:

Then we have to adjust many examples (approx. 25) since instances of `GridHelper` are often created with both color values.

WestLangley:
>Then we have to adjust many examples

That will take 10 minutes.

>This is probably also true for a lot of user level code.

Who knows, but it is not going to kill them.

I think for this Helper we can change the order of the arguments for clarity and support the suggested feature.

docs/api/en/helpers/GridHelper.html
robyanne92: > Unfortunately, the parameter order is now confusing. It should be at least: `colorXCenterLine`, `colorZCenterLine`, `colorGrid`. Or even better: `colorGrid`, `colorXCenterLine`, `colorZCenterLine`. However, this will break backwards compatibility.

I had noticed it too. Let me know if you need help with any code changes.

WestLangley:
/ping @mrdoob Can you please state your preference? Do you want the feature? Which API?

Mugen87:
Considering https://github.com/mrdoob/three.js/pull/21431#issue-586752972, this PR can be closed.

PR #17084, opened 2019-07-24
Improve some chinese docs translation

HypnosNova:
Improve some chinese docs translation


Mugen87:
It seems your PR does not contain any changes...

mrdoob:
Thanks!

PR #17091, opened 2019-07-25
 New hook function "onBeforeGLTFParser"

jiawenquan:
You can modify "GLTF json" by hook function before parsing "json"

```
	var loader = new THREE.GLTFLoader().setPath('models/gltf/DamagedHelmet/glTF/');

	loader.load('DamagedHelmet.gltf', function (gltf) {

		gltf.scene.traverse(function (child) {

			if (child.isMesh) {
				child.material.envMap = envMap;
			}

		});

		scene.add(gltf.scene);

	});

	loader.onBeforeGLTFParser = function (json) {

		// Modify GLTF json here
		json.images[0].uri = "Default_metalRoughness.jpg";

	}
```


donmccurdy:
Thank you for the PR – at the moment I would like for us to consider the plugin system @takahirox has suggested in https://github.com/mrdoob/three.js/issues/11682. I believe that would cover the use case this PR addresses.

Another option in this situation is to load the JSON with FileLoader, modify it, and then pass it into `loader.parse( ... )`.

jiawenquan:
Looking forward to "GLTFLoader: Extensibility."

Issue #17066, opened 2019-07-22
TextureNode colorspace/encoding issue


njarraud:
##### Description of the problem

I have a discrepancy between MeshStandardMaterial and StandardNodeMaterial used with color textures. Unless I missed something, the texture generated by TextureNode seems to have an issue with color space or encoding.

Renderer:
`renderer.gammaOutput = true;`

Texture encoding:
`texture.encoding = THREE.sRGBEncoding;`

Node texture:
`nodeTexture = new TextureNode(texture);`

Example with three.js standard material system:
MeshStandardMaterial 1 (solid color):
`mtl.color = new THREE.Color(_VALUE_).convertSRGBToLinear();`

MeshStandardMaterial 2 (color texture):
`mtl.map = texture;`

![MeshStandardMaterial](https://user-images.githubusercontent.com/45501180/61612327-5b741080-acb2-11e9-881b-f3cb040fbea5.jpg)

Example with three.js node system:
StandardNodeMaterial 1 (solid color):
```
mtl.color = new ColorNode(_VALUE_);
mtl.color.value.convertSRGBToLinear();
```

StandardNodeMaterial 2 (color texture):
`mtl.color = nodeTexture;`

![StandardNodeMaterial](https://user-images.githubusercontent.com/45501180/61612420-a5f58d00-acb2-11e9-83b0-9252c04e5560.jpg)

Both images shall be similar however we can notice that on the second one, the object with the texture is all washed out.

* [jsfiddle](https://jsfiddle.net/3onvw7kj/2/)

##### Three.js version

- [x] Dev

##### Browser

- [x] All of them

##### OS

- [x] All of them

n2k3s2it:
You might be using the wrong NodeMaterial. `StandardNodeMaterial` (as you write) is not the same as `MeshStandardNodeMaterial` unless that's a typo 😄

njarraud:
No it is not a typo. I want to show that texture encoding is properly handled using the standard material system of Three.js (MeshStandardMaterial) but that when switching to the node system with a similar PBR workflow (e.g StandardNodeMaterial), using TextureNode the result is completely different when it shall be similar.

njarraud:
I added a [jsfiddle](https://jsfiddle.net/3onvw7kj/2/) to explain a bit better the issue.
Both parts shall look identical however the part using the texture with StandardNodeMaterial (object1) is not as expected.

njarraud:
@sunag I find a fix for this issue

[...]

PR #17094, opened 2019-07-25
Fix color space conversion issue for TextureNode

njarraud:
The color space conversion for textures is currently inverted. Using fromDecoding instead of fromEncoding produces the expected result.

See this [post](https://github.com/mrdoob/three.js/issues/17066) for more information.


sunag:
Thanks!

sunag:
You can fix this to `CubeTextureNode.js` too?

Issue #17042, opened 2019-07-17
Normal issue with Bump Maps in StandardNodeMaterial


njarraud:
##### Description of the problem
Normals generated by BumpMapNode (and NormalMapNode) in StandardNodeMaterial seem to be incorrect.

I made 2 simple examples using only a HDR environment map to light the scene - One with StandardNodeMaterial and a second one with MeshStandardMaterial to show the difference. I also added a screenshot from Blender Eevee with the exact same set up which tends to prove that the result generated by MeshStandardMaterial is correct.

The scale factor also seem to be behaving differently.

![test](https://user-images.githubusercontent.com/45501180/61342495-799ad480-a89e-11e9-8a70-186c1f75e229.jpg)

* [jsfiddle](https://jsfiddle.net/dwkt8c5s/15/) StandardNodeMaterial
* [jsfiddle](https://jsfiddle.net/grfmnh6x/) MeshStandardMaterial

##### Three.js version

- [x] Dev
- [x] r106

##### Browser

- [x] All of them

##### OS

- [x] All of them

njarraud:
I keep experimenting with the node materials and I may have a suggestion regarding NormalMapNode.

If the geometry includes multiple uv sets, I would have thought that NormalMapNode would use the same uv set as the one specified in TextureNode, or at least that it would have been an argument of the constructor.

Currently, this is not very obvious that the uv index has to be set manually after:

`let normalTex = new THREE.TextureLoader().load('normal.png');`
`let normalNodeText = new Nodes.TextureNode(bevelTex, new Nodes.UVNode(1));`
`let normalMap = new Nodes.NormalMapNode(normalNodeText);`
`normalMap.uv = new Nodes.UVNode(1);`

sunag:
I am try revise the `BumpMapNode` in `jsfiddle` still without success . 
I am need to recreate the example locally and continue the tests...

sunag:
@njarraud try replaces the function `generate` of `NormalNode` for this:
It will fix `NormalNode.WORLD` problem.
Exist other bugs to fix I will launch a PR with theses fixes soon.
```javascript
NormalNode.prototype.generate = function ( builder, output ) {

	var result;

	switch ( this.scope ) {

		case NormalNode.LOCAL:

			// to use vObjectNormal as vertex normal
			//builder.requires.normal = true;

			result = 'normal';

			break;

		case NormalNode.WORLD:

			if ( builder.isShader( 'vertex' ) ) {

[...]

Issue #16975, opened 2019-07-04
NormalNode.scope='local' outputting incorrect values


fi4sk0:
##### Description of the problem

I think the NormalNode does not output the correct vector if the scope is NormalNode.LOCAL
This is visible in the nodes example [webgl_materials_nodes](https://threejs.org/examples/?q=nodes#webgl_materials_nodes), choosing "node / normal". If scope is "world" or "view" it looks correct to me. Selecting scope = "local" should look like "world", since the teapot is not moving, but the camera is orbiting, instead it is similar to "view".

Unfortunately I didn't understand all the code, but changing the following lines in https://github.com/mrdoob/three.js/blob/7f52269ff54590592bbf8a461a908b25e260a240/examples/jsm/nodes/accessors/NormalNode.js#L44 
```javascript
		case NormalNode.LOCAL:

			// to use vObjectNormal as vertex normal
			// builder.requires.normal = true;

			result = 'normal';

			break;
```
to 
```javascript
		case NormalNode.LOCAL:

			// to use vObjectNormal as vertex normal
			builder.requires.normal = true;

			result = 'vObjectNormal';

			break;
```
solves to problem, but I don't understand enough to foresee the side effects.

I hope this is helpful @sunag and thank you for your great work!!

##### Three.js version

- [ ] Dev
- [x] r106
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

##### Hardware Requirements (graphics card, VR Device, ...)

njarraud:
I just ran into the same issue with the broken LOCAL scope for NormalNode.
I didn't really studied the whole node code but your fix is working for me as well.

Here are screenshots with and without the fix - Triplanar texture projection with local normal and position nodes so that the projection remains identical when objects are moved around. Without the fix, the texture is being updated when the camera is moving.

![broken](https://user-images.githubusercontent.com/45501180/61344189-76a2e280-a8a4-11e9-9162-d2c5e220a8ed.jpg)

![fixed](https://user-images.githubusercontent.com/45501180/61344194-79053c80-a8a4-11e9-948b-8d99b1fcfcb9.jpg)

fi4sk0:
[...]

PR #17096, opened 2019-07-25
fix NormalNode

sunag:
Issues https://github.com/mrdoob/three.js/issues/16975, https://github.com/mrdoob/three.js/issues/17042


mrdoob:
Thanks!

PR #16893, opened 2019-06-24
Box3Helper: Fix ctor color parameter.

Mugen87:
Fixed #16891

PR #17103, opened 2019-07-25
Update Box3Helper.js

trusktr:
For some reason, when I run my app, there's a `Â` inside the Box3Helper code. I'm not sure if it is coming from this code base, or Webpack/Babel is injecting it for some reason.

Screen Shot 2019-07-25 at 4 57 22 PM

### DIFF
--- a/src/helpers/Box3Helper.js
+++ b/src/helpers/Box3Helper.js
@@ -15,7 +15,7 @@ function Box3Helper( box, color ) {
 
 	this.box = box;
 
-	color = color ||0xffff00;
+	color = color || 0xffff00;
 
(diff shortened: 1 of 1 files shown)


trusktr:
Re-opened, as I discovered that the Box3Helper.js file that is in `node_modules/three/src/helpers/Box3Helper.js`, present after running `npm install three`, indeed does contain this random `Â` character.

The Three.js source code does appear to contain the character. Seems like this character is only on npmjs.com.

trusktr:
Here's what I see when I run `vim node_modules/three/src/helpers/Box3Helper.js` with a syntax theme that makes the character visible (it makes a dot):

Screen Shot 2019-07-25 at 5 10 10 PM

trusktr:
Are you able to reproduce by installing three v0.106.0 with npm?

trusktr:
Ah, well! https://github.com/mrdoob/three.js/pull/16893

PR #17105, opened 2019-07-26
NodeMaterial Simplified Syntax

sunag:
This is a preview of ~~NodeMaterial Syntax 2~~ **NodeMaterial Simplified Syntax**, this syntax similar at glsl, this too is opening for a possible `threejs shader language`.

/feedbacks? @mrdoob @WestLangley @bhouston @donmccurdy @Mugen87 

This PR is a WIP

## Simplified Syntax
```javascript
import * as Nodes from './jsm/nodes/Nodes.js';
const { float, color, vec3, add, mul, dot, pow, normal } = Nodes.NodeLib.nodes;

var intensity = 1.3;
var power = float( 3 );
var clr = color( 0xFFFFFF );

var viewZ = dot( normal, vec3( 0, 0, - intensity ) );
var rim = add( viewZ, float( intensity ) );
var rimPower = pow( rim, power );
var rimColor = mul( rimPower, clr );

var mtl = new Nodes.PhongNodeMaterial();
mtl.color = color( 0x111111 );
mtl.emissive = rimColor;
```
## Default Syntax
```javascript
import * as Nodes from './jsm/nodes/Nodes.js';

var intensity = 1.3;
var power = new Nodes.FloatNode( 3 );
var color = new Nodes.ColorNode( 0xFFFFFF );

var viewZ = new Nodes.MathNode(
	new Nodes.NormalNode(),
	new Nodes.Vector3Node( 0, 0, - intensity ),
	Nodes.MathNode.DOT
);

var rim = new Nodes.OperatorNode(
	viewZ,
	new Nodes.FloatNode( intensity ),
	Nodes.OperatorNode.ADD
);

var rimPower = new Nodes.MathNode(
	rim,
	power,
	Nodes.MathNode.POW
);

var rimColor = new Nodes.OperatorNode(
	rimPower,
	color,
	Nodes.OperatorNode.MUL
);

var mtl = new Nodes.PhongNodeMaterial();
mtl.color = new Nodes.ColorNode( 0x111111 );
mtl.emissive = rimColor;
```

### DIFF
--- a/examples/jsm/nodes/inputs/TextureNode.js
+++ b/examples/jsm/nodes/inputs/TextureNode.js
@@ -6,6 +6,7 @@ import { InputNode } from '../core/InputNode.js';
 import { UVNode } from '../accessors/UVNode.js';
 import { ColorSpaceNode } from '../utils/ColorSpaceNode.js';
 import { ExpressionNode } from '../core/ExpressionNode.js';
+import { NodeLib } from '../core/NodeLib.js';
 
 function TextureNode( value, uv, bias, project ) {
 
@@ -113,5 +114,11 @@ TextureNode.prototype.toJSON = function ( meta ) {
 	return data;
 
 };
+
+NodeLib.addNode( 'texture', function () { 
+
+	return new TextureNode( ...arguments );
+
+} );
 
 export { TextureNode };
--- a/examples/jsm/nodes/accessors/NormalNode.js
+++ b/examples/jsm/nodes/accessors/NormalNode.js
@@ -86,19 +86,10 @@ NormalNode.prototype.toJSON = function ( meta ) {
 
 };
 
-NodeLib.addFunctionNode( 'normal', new NormalNode( NormalNode.LOCAL ) );
-NodeLib.addFunctionNode( 'worldNormal', new NormalNode( NormalNode.WORLD ) );
+NodeLib.addNode( 'normal', new NormalNode( NormalNode.LOCAL ) );
+NodeLib.addNode( 'worldNormal', new NormalNode( NormalNode.WORLD ) );
 
-NodeLib.addKeyword( 'normal', function () {
-
-	return new NormalNode();
-
-} );
-
-NodeLib.addKeyword( 'worldNormal', function () {
-
-	return new NormalNode( NormalNode.WORLD );
-
-} );
+NodeLib.addKeyword( 'normal', () => { return new NormalNode( NormalNode.LOCAL ); } );
+NodeLib.addKeyword( 'worldNormal', () => { return new NormalNode( NormalNode.WORLD ); } );
 
 export { NormalNode };
--- a/examples/jsm/nodes/accessors/NormalNode.js
+++ b/examples/jsm/nodes/accessors/NormalNode.js
@@ -86,6 +86,9 @@ NormalNode.prototype.toJSON = function ( meta ) {
 
 };
 
+NodeLib.addFunctionNode( 'normal', new NormalNode( NormalNode.LOCAL ) );
+NodeLib.addFunctionNode( 'worldNormal', new NormalNode( NormalNode.WORLD ) );
+
 NodeLib.addKeyword( 'normal', function () {
 
 	return new NormalNode();
--- a/examples/jsm/nodes/accessors/UVNode.js
+++ b/examples/jsm/nodes/accessors/UVNode.js
@@ -56,6 +56,12 @@ UVNode.prototype.toJSON = function ( meta ) {
 
 };
 
+NodeLib.addNode( 'uv', function () { 
+
+	return new UVNode( ...arguments );
+
+} );
+
 NodeLib.addKeyword( 'uv', function () {
 
 	return new UVNode();
--- a/examples/jsm/nodes/core/FunctionNode.js
+++ b/examples/jsm/nodes/core/FunctionNode.js
@@ -125,9 +125,9 @@ FunctionNode.prototype.generate = function ( builder, output ) {
 
 		}
 
-		if ( this.getIncludeByName( reference ) === undefined && NodeLib.contains( reference ) ) {
+		if ( this.getIncludeByName( reference ) === undefined && NodeLib.containsInclude( reference ) ) {
 
-			builder.include( NodeLib.get( reference ) );
+			builder.include( NodeLib.getInclude( reference ) );
 
 		}
 
--- a/examples/jsm/nodes/core/NodeBuilder.d.ts
+++ b/examples/jsm/nodes/core/NodeBuilder.d.ts
@@ -130,8 +130,8 @@ export class NodeBuilder {
   isTypeMatrix(format: string): boolean;
   getTypeLength(type: string): number;
   getTypeFromLength(len: number): string;
-  findNode(): Node;
-  resolve(): void;
+  findNode(...args: any[]): Node;
+  resolve(...args: any[]): void;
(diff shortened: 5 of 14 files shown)


sunag:
This is syntax is more readable and flexible to work ( http://www.mergely.com/TFTz578X/ ), mainly in big code, too is possible use namespace like `nodes.add( .. )` ( http://www.mergely.com/zJZE6x5A/ )

```javascript
import * as Nodes from './jsm/nodes/Nodes.js';
const nodes = Nodes.NodeLib.functions;

var intensity = 1.3;
var power = nodes.float( 3 );
var clr = nodes.color( 0xFFFFFF );

var viewZ = nodes.dot( nodes.normal, nodes.vec3( 0, 0, - intensity ) );
var rim = nodes.add( viewZ, nodes.float( intensity ) );
var rimPower = nodes.pow( rim, power );
var rimColor = nodes.mul( rimPower, clr );

var mtl = new Nodes.PhongNodeMaterial();
mtl.color = nodes.color( 0x111111 );
mtl.emissive = rimColor;
```

sunag:
The other name (maybe better) to this feature is `simplified syntax` or a full name `node material simplified syntax`, the idea is keeping both syntaxes, `simplified syntax` is generated almost automatically based on nodes created, e.g:

https://github.com/mrdoob/three.js/blob/d7119570abe2795c0c7d6a059bee5fc6dc77da33/examples/jsm/nodes/math/MathNode.js#L271-L275

Only this is necessary to declare all math functions in simplified syntax like `mix`, `sin`, `cos` ... 

```javascript
const { mix, sin, cos } = Nodes.NodeLib.nodes;
```

sunag:
@mrdoob I would like to know your opinion about this. With the concern of code style, what model you think is best.

Simplified 1
```javascript
var rim = add( viewZ, float( intensity ) );
```

Simplified 2
```javascript
var rim = Nodes.AddNode( viewZ, Nodes.FloatNode( intensity ) );
```

`Nodes.AddNode( a, b )` is a shortcut of `Nodes.NodesOperator( a, b, Nodes.NodesOperator.ADD )`

WestLangley:
I understand that @bhouston and @donmccurdy have a lot of interest in pushing Node forward. I think their feedback would be beneficial, too.

sunag:
> I understand that @bhouston and @donmccurdy have a lot of interest in pushing Node forward. I think their feedback would be beneficial, too.

Yes, no doubt.

bhouston:
I think this is okay.  The really short names can only be used when you know they will not conflict with other names.

The main issue with this syntax is that it wrecks tree shaking as you have made a single object that contains references to each possibly graph node.

I would prefer to just change things like Nodes.NodesOperator to NodeOp.

Simplified 3:
```
var rim = new NodeOp( '+', viewZ, new FloatNode( intensity ) );
```

I'd rather short class names overall, rather than two names, one long and one ultra short.  I guess I want a middle approach rather than a dual approach.  Dual approaches are harder to maintain.

sunag:
> I'd rather short class names overall, rather than two names, one long and one ultra short. I guess I want a middle approach rather than a dual approach. Dual approaches are harder to maintain.

Maybe the better alternative be this, a long and other ultra short name like GLSL. 
I am not sure if it really would be hard to maintain because I am using procedural code, this is practically automatic?

sunag:
> Maybe the better alternative be this, a long and other ultra short name like GLSL.

@bhouston Sorry I understand wrong your observation with relation the ultra-short... 

> I'd rather short class names overall,

Good, at least for me, unless we have a "big" change as:

Simplified 4:
```javascript
var rim = new AddNode( viewZ, new FloatNode( intensity ) );
```

to use nodes like `SinNode`, `MixNode` and others, it does not seem to make much necessary. use as short class sounds good to me.

bhouston:
Sunag,  I do like this last example you showed the best of all examples:
```
var rim = new AddNode( viewZ, new FloatNode( intensity ) );
```
And I agree with SinNode, MixNode, etc.  It makes sense.

examples/jsm/nodes/inputs/TextureNode.js
```diff
@@ -114,4 +115,10 @@ TextureNode.prototype.toJSON = function ( meta ) {
 
 };
 
+NodeLib.addNode( 'texture', function () { 
+
+	return new TextureNode( ...arguments );
+
+} );
```
donmccurdy: Is there a reason this can't be a one-liner like the FloatNode example above?

```js
NodeLib.addNode( 'texture', TextureNode );
```

donmccurdy:
**Factory function consistency**

> `var viewZ = dot( normal, vec3( 0, 0, - intensity ) );`

Is `normal` a function, or a Node? The factory functions should probably be consistent so that even functions taking no arguments still have to be invoked:

```js
var viewZ = dot( normal(), vec3( 0, 0, - intensity ) );
```

**Tree-shaking**

> The main issue with this syntax is that it wrecks tree shaking as you have made a single object that contains references to each possibly graph node.

Wouldn't flattening the exports also fix tree-shaking? For example:

```js
import { FloatNode, float, ... } from './jsm/nodes/Node.js';
```

Arguably, the `Node` constructors do not need to be exported at all, and we could avoid supporting two syntaxes.

**Syntax preference**

Comparing the "Simplified" syntax to "Simplified 4" ...

```js
// Simplified
import { dot, normal, vec3, add, float, pow } from './jsm/nodes/Nodes.js';
var viewZ = dot( normal(), vec3( 0, 0, -1.3 ) );
var rim = add( viewZ, float( 1.3 ) );
var rimPower = pow( rim, float( 3 ) );

// Simplified 4
import { DotNode, NormalNode, Vector3Node, AddNode, FloatNode, PowNode } from './jsm/nodes/Nodes.js';
var viewZ = new DotNode( new NormalNode(), new Vector3Node( 0, 0, -1.3 ) );
var rim = new AddNode( viewZ, new FloatNode( 1.3 ) );
var rimPower = new PowNode( rim, new FloatNode( 3 ) );
```

... I think the first is a bit easier to read, as it resembles GLSL more. But yes, keywords like "position" and "normal" could more easily conflict with user variable names. I'm OK with either of these.

**Node vs primitive arguments**

In both syntaxes, whether a function or constructor requires a node or a primitive type is not always obvious. Consider:

```js
// ERROR
var value = new Vector3Node( 0, 0, new FloatNode( -1.3 ) );

// OK
var value = new Vector3Node( 0, 0, -1.3 );

// OK
var value = new PowNode( rim, new FloatNode( 3 ) );

// ERROR
var value = new PowNode( rim, 3 );
```

If FloatNode and number primitives were consistently interchangeable, that would probably avoid some confusion.

sunag:
@donmccurdy Great review. Thanks! 

> Factory function consistency, Tree-shaking, Syntax preference
> ... I think the first is a bit easier to read, as it resembles GLSL more. But yes, keywords like "position" and "normal" could more easily conflict with user variable names. I'm OK with either of these.

I think that the @bhouston suggested, use `short class` is very good to solve all the issues you mentioned above like, still being a syntax accepted by the majority apparently.

- **Factory function consistency**
(_Keep classic names and threejs code style_)
- **Tree-shaking keeps classic JS `export` syntax.**
( _The approach of `short names` that I made is procedural, I try a declarative procedural using `export` but without success. This fixes this issue for use import `Nodes` instead of `NodeLib.nodes`_ ) 

> Node vs primitive arguments

Yeah, I have the same thought, to add this feature we could add a `NodeBuilder.resolveConst` in `Node.generate` of each node.

PR #17128, opened 2019-07-30
Updated Creating-a-scene.html to ECMAScript 6+

ghost:
Changed the vars to const


Mugen87:
I guess we are not yet at the point where we want to introduce `let`/`const` in the code base. Things might change when we have introduced classes in the core...

PR #17164, opened 2019-08-03
StereoCamera: Fix caching in update().

Mugen87:
`StereoCamera.update()` should only compute projection matrices for the right and left camera if necessary. However, because of wrong caching, the expensive computation is done per frame. You can verify this with the official example. Just set a breakpoint in the respective `update()` method: https://threejs.org/examples/webgl_effects_stereo

The evaluation `eyeSep !== this.eyeSep` is always `true` since `eyeSep` represents the half value of `this.eyeSep`. The PR fixes this so the computation is only done if parameters have actually changed.


mrdoob:
Nice catch!

mrdoob:
Thanks!

PR #17174, opened 2019-08-05
Create webgl_gpu_picking_point.html

1147079942:
it's an example of using depth value to calculate picking point

### DIFF
--- a/examples/webgl_gpu_picking_point.html
+++ b/examples/webgl_gpu_picking_point.html
@@ -84,7 +84,7 @@ function init() {
 
 	//
 
-	// let geometry = new THREE.SphereBufferGeometry(100, 20, 20);
+	// var geometry = new THREE.SphereBufferGeometry(100, 20, 20);
 	var geometry = new THREE.BoxBufferGeometry( 100, 100, 100 );
 
 	var material = new THREE.MeshNormalMaterial();
--- a/examples/webgl_gpu_picking_point.html
+++ b/examples/webgl_gpu_picking_point.html
@@ -57,6 +57,7 @@ var controls;
 var raycaster;
 
 var pickingCamera;
+var pickingTexture;	
 
 var GPUPickingPoint = new THREE.Vector3();
 var CPUPickingPoint = new THREE.Vector3();
@@ -71,8 +72,6 @@ function init() {
 	raycaster = new THREE.Raycaster();
 	renderCamera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 10000 );
 	pickingCamera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 10000 );
-	renderCamera.up.set( 0, 0, 1 );
-	pickingCamera.up.set( 0, 0, 1 );
 	renderCamera.position.x = 200;
 
 	scene = new THREE.Scene();
@@ -80,15 +79,15 @@ function init() {
 
 	//
 
-	let axesHelper = new THREE.AxesHelper( 100 );
+	var axesHelper = new THREE.AxesHelper( 100 );
 	scene.add( axesHelper );
 
 	//
 
 	// let geometry = new THREE.SphereBufferGeometry(100, 20, 20);
-	let geometry = new THREE.BoxBufferGeometry( 100, 100, 100 );
+	var geometry = new THREE.BoxBufferGeometry( 100, 100, 100 );
 
-	var material = new THREE.MeshBasicMaterial({color: new THREE.Color(0.5, 0.5, 0.5)});
+	var material = new THREE.MeshNormalMaterial();
 	var depthMaterial = new THREE.MeshDepthMaterial();
 	depthMaterial.depthPacking = THREE.RGBADepthPacking;
 
@@ -105,7 +104,11 @@ function init() {
 	document.body.appendChild( renderer.domElement );
 
 	//
+	
+	pickingTexture = new THREE.WebGLRenderTarget(renderer.domElement.width, renderer.domElement.height);
 
+	//
+	
 	stats = new Stats();
 	document.body.appendChild( stats.dom );
 
@@ -113,12 +116,12 @@ function init() {
 
 	gui = new GUI();
 	gui.add( renderer.capabilities, 'logarithmicDepthBuffer' );
-	let GPUPFolder = gui.addFolder( 'GPUPickingPoint' );
+	var GPUPFolder = gui.addFolder( 'GPUPickingPoint' );
 	GPUPFolder.open();
 	GPUPFolder.add( GPUPickingPoint, 'x' );
 	GPUPFolder.add( GPUPickingPoint, 'y' );
 	GPUPFolder.add( GPUPickingPoint, 'z' );
-	let CPUPFolder = gui.addFolder( 'CPUPickingPoint' );
+	var CPUPFolder = gui.addFolder( 'CPUPickingPoint' );
 	CPUPFolder.open();
 	CPUPFolder.add( CPUPickingPoint, 'x' );
 	CPUPFolder.add( CPUPickingPoint, 'y' );
@@ -126,15 +129,6 @@ function init() {
 	//
 
 	controls = new OrbitControls( renderCamera, renderer.domElement );
-	controls.rotateSpeed = 5.0;
-	controls.zoomSpeed = 5;
-	controls.panSpeed = 2;
-
-	controls.enableZoom = true;
-	controls.enablePan = true;
-
-	controls.enableDamping = false;
-	controls.dampingFactor = 0.3;
 
 	//
 
@@ -153,11 +147,10 @@ function pick( e ) {
 	var pX = 2 * x / renderer.domElement.width - 1;
 	var pY = 2 * y / renderer.domElement.height - 1;
 
-	let mouseVector = new THREE.Vector3();
+	var mouseVector = new THREE.Vector3();
 	mouseVector.set( pX, pY, 0.5 ); // set raycaster target
 	raycaster.setFromCamera( mouseVector, camera );
 
-	var pickingTexture = new THREE.WebGLRenderTarget( renderer.domElement.width, renderer.domElement.height );
 	//render the picking scene off-screen
 	// set the view offset to represent just a single pixel under the mouse
 	camera.setViewOffset( renderer.domElement.width, renderer.domElement.height, e.clientX * window.devicePixelRatio | 0, e.clientY * window.devicePixelRatio | 0, 1, 1 );
@@ -185,7 +178,7 @@ function pick( e ) {
 	// compute the depth far from the camera which between [0,far];
 	depth = camera.near * camera.far / ( camera.near * depth + camera.far * ( 1 - depth ) );
 
-	let direction = raycaster.ray.direction.normalize();
+	var direction = raycaster.ray.direction.normalize();
 	var target = new THREE.Vector3();
 	camera.getWorldDirection( target );
 	var dist = depth / target.dot( raycaster.ray.direction );
@@ -197,8 +190,8 @@ function pick( e ) {
 
 		camera = pickingCamera;
 
-		let pickingCameraPosition = new THREE.Vector3();
-		let pickingCameraLookAt = new THREE.Vector3();
+		var pickingCameraPosition = new THREE.Vector3();
+		var pickingCameraLookAt = new THREE.Vector3();
 		//why is 0.999 ,when depth is 10000 the point error is 1~10,if big than that value,the result could be wrong
 		pickingCameraPosition.addVectors( renderCamera.position, new THREE.Vector3().copy( direction ).multiplyScalar( ( dist - camera.near ) * 0.999 ) );
 		pickingCameraLookAt.addVectors( renderCamera.position, new THREE.Vector3().copy( direction ).multiplyScalar( dist ) );
@@ -207,7 +200,6 @@ function pick( e ) {
 		camera.lookAt( pickingCameraLookAt );
 		camera.updateProjectionMatrix();
 
-		var pickingTexture = new THREE.WebGLRenderTarget( renderer.domElement.width, renderer.domElement.height );
 
 		renderer.setRenderTarget( pickingTexture );
 		renderer.render( pickingScene, camera );
@@ -231,7 +223,7 @@ function pick( e ) {
 	}
 
 
-	let intersectObject = raycaster.intersectObject( mesh );
+	var intersectObject = raycaster.intersectObject( mesh );
 
 	if ( intersectObject.length ) {
 
--- a/examples/webgl_gpu_picking_point.html
+++ b/examples/webgl_gpu_picking_point.html
@@ -43,12 +43,11 @@
 	3.I can release the geometry data under the mesh, free more memory, and not affect the rendering.
 */
 import * as THREE from '../build/three.module.js';
-import Stats from './jsm/libs/stats.module.js';
 import { GUI } from './jsm/libs/dat.gui.module.js';
 import { OrbitControls } from './jsm/controls/OrbitControls.js';
 
 
-var renderCamera, scene, renderer, stats, pickingScene, gui;
+var renderCamera, scene, renderer, pickingScene, gui;
 
 var mesh;
 var mesh1;
@@ -108,14 +107,8 @@ function init() {
 	pickingTexture = new THREE.WebGLRenderTarget(renderer.domElement.width, renderer.domElement.height);
 
 	//
-	
-	stats = new Stats();
-	document.body.appendChild( stats.dom );
-
-	//
 
 	gui = new GUI();
-	gui.add( renderer.capabilities, 'logarithmicDepthBuffer' );
 	var GPUPFolder = gui.addFolder( 'GPUPickingPoint' );
 	GPUPFolder.open();
 	GPUPFolder.add( GPUPickingPoint, 'x' );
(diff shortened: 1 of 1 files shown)


Mugen87:
https://raw.githack.com/1147079942/three.js/patch-1/examples/webgl_gpu_picking_point.html

Mugen87:
It seems the example is broken.

> WEBGL is not defined

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,280 @@
...
+
+<div id="container"></div>
+<div id="info"><a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> webgl - buffergeometry - indexed
+</div>
+
+<script src="../build/three.js"></script>
+
+<script src="js/WebGL.js"></script>
+<script src="js/libs/stats.min.js"></script>
+<script src="./js/libs/dat.gui.min.js"></script>
```
Mugen87: Please avoid such imports and use ES6 modules. Like in all other examples...

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,280 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<title>three.js webgl - buffergeometry - indexed</title>
```
Mugen87: This title does not seems right.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,280 @@
...
+		a {
+			color: #0080ff;
+		}
+
+	</style>
+</head>
+<body>
+
+<div id="container"></div>
+<div id="info"><a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> webgl - buffergeometry - indexed
```
Mugen87: Wrong title...

1147079942:
I should use the newest release version,it works now.

Mugen87:
There are still some warnings in the browser console:

image

WestLangley:
>There are still some warnings in the browser console:

Oh man, I was just preparing to remove those 4-year-old warnings. I would have though 4 years was long enough...

Quite frankly, I still think it is.

mrdoob:
>> There are still some warnings in the browser console:
>
> Oh man, I was just preparing to remove those 4-year-old warnings. I would have though 4 years was long enough...
>
> Quite frankly, I still think it is.

I think it's too 👌

WestLangley:
>I think it's too

OK. I'll fix it in an upcoming PR.

1147079942:
> There are still some warnings in the browser console:
> 
> image

i have fixed the problem ,i will update this example later to reduce the point error

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,298 @@
...
+	// import {WEBGL} from './js/WebGL.js';
+	import Stats from './jsm/libs/stats.module.js';
+	import {GUI} from './jsm/libs/dat.gui.module.js';
+	import {OrbitControls} from './jsm/controls/OrbitControls.js';
+
+	// if (WEBGL.isWebGLAvailable() === false) {
+	//
+	// 	document.body.appendChild(WEBGL.getWebGLErrorMessage());
+	//
+	// }
```
Mugen87: Instead of commenting out, you can remove this code (and the import).

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,298 @@
...
+
+	var renderCamera, scene, renderer, stats, pickingScene, gui;
+
+	var mesh;
+	var mesh1;
+
+	var controls;
+	let raycaster;
```
Mugen87: The code base does not use `let` and `const` yet. Please stick to `var`.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,298 @@
...
+	var CPUPickingPoint = new THREE.Vector3();
+
+	init();
+	animate();
+
+	function init() {
+
+		//
+		raycaster = new THREE.Raycaster();
+		renderCamera = new THREE.PerspectiveCamera(60, window.innerWidth / window.innerHeight, 1, 10000);
```
Mugen87: The code of the example does not respect the code style yet: https://github.com/mrdoob/three.js/wiki/Mr.doob%27s-Code-Style%E2%84%A2

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,298 @@
...
+		// var ambientLight = new THREE.AmbientLight(0x222222);
+		// scene.add(ambientLight);
+		//
+		// var light1 = new THREE.DirectionalLight(0xffffff, 0.5);
+		// light1.position.set(1, 1, 1);
+		// scene.add(light1);
+		//
+		// var light2 = new THREE.DirectionalLight(0xffffff, 1);
+		// light2.position.set(0, -1, 0);
+		// scene.add(light2);
```
Mugen87: Remove code that is commented out.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,298 @@
...
+
+		//
+
+		// let geometry = new THREE.SphereBufferGeometry(100, 20, 20);
+		let geometry = new THREE.BoxBufferGeometry(100, 100, 100);
+
+		var depthMaterial = new THREE.MeshDepthMaterial();
+		depthMaterial.depthPacking = THREE.RGBADepthPacking;
+
+		mesh = new THREE.Mesh(geometry, depthMaterial);
```
Mugen87: Why using `MeshDepthMaterial`? It's intended for certain use cases but not as a typical material for rendering meshes. Use e.g.  `MeshBasicMaterial`, `MeshLamberMaterial` or `MeshPhongMaterial` instead.

Mugen87:
> it's an example of using depth value to calculate picking point

What is the advantage of using this approach compared to normal GPU picking?

https://threejs.org/examples/webgl_interactive_cubes_gpu

1147079942:
> > it's an example of using depth value to calculate picking point
> 
> What is the advantage of using this approach compared to the existing GPU picking code?
> 
> https://threejs.org/examples/webgl_interactive_cubes_gpu

1.Picking in cesium using gpu depth to calculate picking points，I think threejs can also have.
	2.It is more efficient to calculate the picking point using gpu depth when the number of faces of a single model is large.
	3.I can release the geometry data under the mesh, free more memory, and not affect the rendering.

examples/webgl_gpu_picking_point.html
1147079942: I can directly see the depth map effect used in the off-screen frame.

Mugen87: Um, I don't think this setup is good. It looks like a broken example.

examples/webgl_gpu_picking_point.html
```diff
@@ -88,10 +88,11 @@
 	// let geometry = new THREE.SphereBufferGeometry(100, 20, 20);
 	let geometry = new THREE.BoxBufferGeometry( 100, 100, 100 );
 
+	var material = new THREE.MeshBasicMaterial({color: new THREE.Color(0.5, 0.5, 0.5)});
```
WestLangley: Try `MeshNormalMaterial`. That will look better.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,291 @@
...
+
+function init() {
+
+	//
+
+	raycaster = new THREE.Raycaster();
+	renderCamera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 10000 );
+	pickingCamera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 10000 );
+	renderCamera.up.set( 0, 0, 1 );
+	pickingCamera.up.set( 0, 0, 1 );
```
WestLangley: Why z-up in this demo?

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,291 @@
...
+	CPUPFolder.open();
+	CPUPFolder.add( CPUPickingPoint, 'x' );
+	CPUPFolder.add( CPUPickingPoint, 'y' );
+	CPUPFolder.add( CPUPickingPoint, 'z' );
+	//
+
+	controls = new OrbitControls( renderCamera, renderer.domElement );
+	controls.rotateSpeed = 5.0;
+	controls.zoomSpeed = 5;
+	controls.panSpeed = 2;
```
WestLangley: Please remove these speed settings. They are too high and not needed.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,291 @@
...
+
+	controls.enableZoom = true;
+	controls.enablePan = true;
+
+	controls.enableDamping = false;
```
WestLangley: You should enable damping if you want to set a damping factor.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,291 @@
...
+	controls.dampingFactor = 0.3;
```
WestLangley: That is too high. The default of 0.05 should be fine.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,291 @@
...
+	var y = renderer.domElement.height - e.clientY;
+
+	var pX = 2 * x / renderer.domElement.width - 1;
+	var pY = 2 * y / renderer.domElement.height - 1;
+
+	let mouseVector = new THREE.Vector3();
+	mouseVector.set( pX, pY, 0.5 ); // set raycaster target
+	raycaster.setFromCamera( mouseVector, camera );
+
+	var pickingTexture = new THREE.WebGLRenderTarget( renderer.domElement.width, renderer.domElement.height );
```
WestLangley: Check the other gpu picking example. Can you create `WebGLRenderTarget` and reuse it?

1147079942:
@Mugen87 Thanks for help ,it looks better

examples/webgl_gpu_picking_point.html
1147079942: That's a habit caused by a model built by a colleague.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,283 @@
...
+	requestAnimationFrame( animate );
+
+	render();
+	stats.update();
+
+}
+
+function render() {
+
+	controls.update();
```
WestLangley: This is not needed if you remove damping. Nor is the animation loop since your scene is static.

You can enable damping again, or your can remove both the call to `update()` and the animation loop. Your choice.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,283 @@
...
+
+	//
+	
+	stats = new Stats();
+	document.body.appendChild( stats.dom );
+
+	//
+
+	gui = new GUI();
+	gui.add( renderer.capabilities, 'logarithmicDepthBuffer' );
```
WestLangley: What is the point of this menu item?

WestLangley: @Mugen87 @mrdoob  `renderer.capabilities` probably should not be exposed like this.

1147079942: I want know  that setting does useful for MeshDepthMaterial, maybe i should check the shader. if that's work, the way of compute the picking point will be different.

Mugen87: > renderer.capabilities probably should not be exposed like this.

I think it's okay like that. The `capabilities` object can be used to check for `WebGL2` (`renderer.capabilities.isWebGL2`) or get the maximum anisotropy (`renderer.capabilities.getMaxAnisotropy()`). I prefer this style than adding more methods/flags to `WebGLRenderer`.

WestLangley: @Mugen87 @mrdoob The OP is trying to _set_ the capability.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,273 @@
...
+var raycaster;
+
+var pickingCamera;
+var pickingTexture;	
+
+var GPUPickingPoint = new THREE.Vector3();
+var CPUPickingPoint = new THREE.Vector3();
+
+init();
+animate();
```
WestLangley: You do not need an animation loop. This is a static scene.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,273 @@
...
+
+	var x = e.clientX;
+	var y = renderer.domElement.height - e.clientY;
+
+	var pX = 2 * x / renderer.domElement.width - 1;
+	var pY = 2 * y / renderer.domElement.height - 1;
+
+	var mouseVector = new THREE.Vector3();
+	mouseVector.set( pX, pY, 0.5 ); // set raycaster target
+	raycaster.setFromCamera( mouseVector, camera );
```
WestLangley: `mouseVector` should be a `Vector2`.

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,273 @@
...
+
+		camera = pickingCamera;
+
+		var pickingCameraPosition = new THREE.Vector3();
+		var pickingCameraLookAt = new THREE.Vector3();
+		//why is 0.999 ,when depth is 10000 the point error is 1~10,if big than that value,the result could be wrong
+		pickingCameraPosition.addVectors( renderCamera.position, new THREE.Vector3().copy( direction ).multiplyScalar( ( dist - camera.near ) * 0.999 ) );
+		pickingCameraLookAt.addVectors( renderCamera.position, new THREE.Vector3().copy( direction ).multiplyScalar( dist ) );
+
+		camera.position.set( pickingCameraPosition.x, pickingCameraPosition.y, pickingCameraPosition.z );
```
WestLangley:
```js
camera.position.copy( pickingCameraPosition );
```

examples/webgl_gpu_picking_point.html
```diff
@@ -0,0 +1,273 @@
...
+		camera.lookAt( pickingCameraLookAt );
+		camera.updateProjectionMatrix();
```
WestLangley: If `updateProjectionMatrix()` is needed, it needs to go somewhere else.

examples/webgl_gpu_picking_point.html
Mugen87: @1147079942 Please perform no write operations on `renderer.capabilities`. It's best when you remove the respective line.

Mugen87: Well, a typical user error...

WestLangley: @Mugen87 I'll resolve my own questions.

1147079942: i event don't know you guys have understand what i express

examples/webgl_gpu_picking_point.html
1147079942: i don't get you point, that make me  confuse.

examples/webgl_gpu_picking_point.html
1147079942: let's try again, take this as a question.In my opinio, `logarithmicDepthBuffer` is work in shader, this use a different algorithm to calculate the value between near and far.I think this `logarithmicDepthBuffer` option design should work for `MeshDepthMaterial` too. If is the same in the threejs like what i say, or threejs change the `MeshDepthMaterial` shader, enable the `logarithmicDepthBuffer` in the `MeshDepthMaterial`, this example will be broken.

examples/webgl_gpu_picking_point.html
1147079942: @WestLangley  do you mean the object do not has an animation, i don't need continuous rendering.

WestLangley: Yes. See, for example, https://threejs.org/examples/webgl_materials_matcap.html.

WestLangley: If you want `OrbitControls` with damping enabled, then you need an animation loop. Your choice.

examples/webgl_gpu_picking_point.html
WestLangley: You can only enable or disable logarithmic depth buffer in the `WebGLRenderer` constructor.

See, for example, https://threejs.org/examples/webgl_camera_logarithmicdepthbuffer.html

WestLangley:
@Mugen87 I, too, am trying to understand this code... Doesn't this technique just provide the depth, and not the identity of the selected object? What is the use case for which that information would be sufficient?

1147079942:
Thanks for all suggests.Thanks for all patience.Thanks for all. This example will be useful to workers for BIM.

1147079942:
https://doc.babylonjs.com/how_to/using_logarithmic_depth_buffer

Mugen87:
I'm afraid the linked documentation does not answer @WestLangley's question...

1147079942:
sorry，that was not the ans of the question.That ans is given few days ago.The link is for the problem in meshDepthMaterial about the log depth buffer.

Mugen87:
So the answer is the approach is only usable for single meshes, right (since you can't differentiate between multiple 3D objects in your scene)?

1147079942:
The most important is calculate the point in the complex mesh. In our 3dtiles which conbined by gltf, we merge each tile's obj in one mesh. we use cpu pick tile box,use gpu to pick the obj id. But get point with CPU will cust hight price，we also hope use gpu to do
 it. The most important is with out use the cpu    
   intersectObject,we do not need keep the geometry under the mesh,this behavior will save large memory for us. I want threejs will do better than babylon or cesium which my colleague is using. That trick is useful to me. If threejs not need that,i just keep it for myself. I will close this request next week.
Thanks for your help,Thanks for your time,Thanks for all. Have a nice day.

Mugen87:
I understand your use case now, thanks for your explanation. However, I'm not feeling strong about adding the example to the lib. I leave it up to @WestLangley and @mrdoob.

PR #17182, opened 2019-08-06
Triangle: Improve closure variable check.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17159#discussion_r310823923


mrdoob:
Thanks!

PR #17199, opened 2019-08-08
WIP: Added Texture.contain, .cover, .fill

WestLangley:
When a flat texture is set as the scene.background, it stretches to fit by default.

I'm trying to find the right API to support something like the CSS object-fit "cover" and "contain" properties.

(This feature does not have to be supported by adding new methods to `Texture` as I have done. I could just show an example.)

Suggestions welcome...

### DIFF
--- a/src/renderers/shaders/ShaderLib/background_frag.glsl.js
+++ b/src/renderers/shaders/ShaderLib/background_frag.glsl.js
@@ -5,6 +5,9 @@ varying vec2 vUv;
 
 void main() {
 
+	if ( vUv.x < 0.0 || vUv.x > 1.0 ) discard; // works, but it messes up repeated textures in other use cases
+	if ( vUv.y < 0.0 || vUv.y > 1.0 ) discard;
+
 	vec4 texColor = texture2D( t2D, vUv );
 
 	gl_FragColor = mapTexelToLinear( texColor );
--- a/build/three.js
+++ b/build/three.js
@@ -2842,6 +2842,53 @@
 
 		},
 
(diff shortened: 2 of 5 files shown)


WestLangley:
It's alive !

https://raw.githack.com/westlangley/three.js/dev_object_fit/examples/webgl_background_cover.html

src/renderers/shaders/ShaderLib/background_frag.glsl.js
```diff
@@ -5,6 +5,9 @@ varying vec2 vUv;
 
 void main() {
 
+	if ( vUv.x < 0.0 || vUv.x > 1.0 ) discard; // works, but it messes up repeated textures in other use cases
+	if ( vUv.y < 0.0 || vUv.y > 1.0 ) discard;
```
WestLangley: This is required to support "contain". Is it worth supporting contain?

donmccurdy:
Perhaps `Scene` or `SceneUtils` might be a better place for this API? For example:

```js
import { Scene } from 'three';
import { SceneUtils } from 'three/examples/jsm/utils/SceneUtils.js';

const scene = new Scene();
scene.background = texture;
SceneUtils.setBackgroundFit( scene, 'contain', aspect );
```

WestLangley:
@donmccurdy That is probably better...

WestLangley:
@mrdoob Maybe we could have a `SceneBackground` object. You indicated something like that, if I remember correctly. I do not recall the motivation, though. One advantage would be having user-access to the background material.

PR #17209, opened 2019-08-08
Material.d.ts: Remove update().

### LINKED ISSUE #17200: Type definition: what is the `Material.update()` ?
https://github.com/mrdoob/three.js/blob/42d3129c0c8a9ead85a91f9378f5d1a0528c56d1/src/materials/Material.d.ts#L336-L339

I don't think we have `update()` for our materials...

@LuWangThreekit

Mugen87:
Fixed #17200


mrdoob:
Thanks!

PR #17211, opened 2019-08-08
Fix off-by-one error in computing shadow map texture coordinates

Oletus:
texture2DShadowLerp needs to calculate the texture coordinates of the four neighboring texels of the sample point.

Let's consider a texture with the size W * H. The centroids of the texels in this texture lie at texture coordinates [(X + 0.5) / W, (Y + 0.5) / H] where X and Y are integers.

To find the neighboring texel on the left of the sample point texture coordinate x, we want to find the maximum integer X for which:

(X + 0.5) / W < x <=>
X + 0.5 < x * W <=>
X < x * W - 0.5

We get this X value with floor( x * W - 0.5 ) and the corresponding texel centroid x coordinate is (floor( x * W - 0.5 ) + 0.5) / W.

For x = 1.3 and W = 2.0 this formula gives out 0.5 / 2.0. With the same values the previous formula would give out 1.0 / 2.0 which lies exactly at the texel boundary and so would end up sampling the wrong texel.

### DIFF
--- a/src/renderers/shaders/ShaderChunk/shadowmap_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/shadowmap_pars_fragment.glsl.js
@@ -41,7 +41,7 @@ export default /* glsl */`
 		const vec2 offset = vec2( 0.0, 1.0 );
 
 		vec2 texelSize = vec2( 1.0 ) / size;
-		vec2 centroidUV = floor( uv * size + 0.5 ) / size;
+		vec2 centroidUV = ( floor( uv * size - 0.5 ) + 0.5 ) * texelSize;
 
 		float lb = texture2DCompare( depths, centroidUV + texelSize * offset.xx, compare );
 		float lt = texture2DCompare( depths, centroidUV + texelSize * offset.xy, compare );


mrdoob:
I believe you 😊

mrdoob:
Thanks!

Oletus:
Thanks for the quick review!

guroseter:
@Oletus This change seems to have caused a lot of problems for my team, so I am trying to understand it a bit better. First of all, I am not sure what is meant by the four neighboring texels. I would expect it to be like this, see Figure 1. But from your equations that can’t be right, in your definition, the neighboring left texel will change based on where in the sample texel you are. 

Take as an example W=4.0 and sample coordinates x=0.3 and x=0.4. These two values belong to the same texel, texel 1, see Figure 2.

If I use your formula for the left neighboring texel: X < x * W - 0.5, then for 

x=0.3 => X < 0.7 => X = 0.

x=0.4 => X <0.11 => X = 1.

So the two points belonging to texel 1 have different left neighbors, is this how it is supposed to be? If so, can you explain what the neighboring texels of sample points lying in the different «quadrants» of a texel should be?

Figure1

Figure2

Oletus:
I think you may have misunderstood the sampling pattern implemented by the function in question - it's not a cross like in your illustration, but rather a 2x2 square where we're doing bilinear interpolation. Maybe my use of "neighboring" threw you off.

image

guroseter:
I see, thank you for taking the time to explain 🙂

PR #17222, opened 2019-08-09
Docs: Add live example to intro guide.

### LINKED ISSUE #17213: Docs: use jsfiddle or codepen?
On https://threejs.org/docs/index.html#manual/en/introduction/Creating-a-scene I read:

>Before you can use three.js, you need somewhere to display it. Save the following HTML to a file on your computer, along with a copy of three.js in the js/ directory, and open it in your browser.

It would be very nice if you could jsfiddle or codepen for the examples.

Same here: https://threejs.org/docs/index.html#manual/en/introduction/How-to-run-things-locally

Mugen87:
Fixed #17213


mrdoob:
Thanks!

Issue #17134, opened 2019-07-31
Add support for typing in EventDispatcher


amiorenstx:
Or take it a step further:
I would consider migrating the whole library to typescript and adding the rxjs library.
With rxjs you will be able to subscribe to different changes such as position changes etc
Also you will be able to know which type of messages are emitted from different events and be able to thorttle messages and buffer them

Mugen87:
As mentioned here #15545, `three.js` will not be converted to TS. The library provides type declaration files for the core and example JS code which should be sufficient for TS users.

> Also you will be able to know which type of messages are emitted from different events and be able to thorttle messages and buffer them

You can implement the same logic with JS, just differently (e.g. put the message type information in the event object). However, I don't think this stuff is necessary in the engine.

Usnul:
Ha, interesting discussion. At some point three.js started to create actual objects and send them as events, right now they are of form:
```js
{
type:"add"
}
```
but the reason why I know is, I re-populate my scene every frame, because of custom culling mechanism, and I found, via profiling, that three.js was generating a decent amount of 🗑 garbage, so I thought to myself 🤔 

"hey, three.js is usually super good at keeping GC to a minimum, what gives?"

turns out - new even objects. Now I bypass `scene.add` and manipulate `children` property directly.

My point is - I would prefer not to see more of an event framework in three.js, it's just about too heavy from my perspective as is. 🏋️‍♀️

gkjohnson:
> "hey, three.js is usually super good at keeping GC to a minimum, what gives?"
>
> turns out - new event objects.

This is interesting -- I hadn't considered that. Classes like [OrbitControls](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/OrbitControls.js#L281) create objects once and reuse them to dispatch events. It seems like that would be a reasonable improvement to the `add` and `remove` events in Object3D?

Mugen87:
> It seems like that would be a reasonable improvement to the add and remove events in Object3D?

I think so, too. I'm not sure if there was a specific reason to create new objects but reuse event objects for both methods seems to make more sense.

Mugen87:
Anyway, we've got sidetracked. The issue itself can be closed.

The idea of reusing event objects is important and should be discussed in another topic (it has nothing to do with TS).

PR #17224, opened 2019-08-09
Reuse "added" and "removed" event

gkjohnson:
As discussed [here](https://github.com/mrdoob/three.js/issues/17134#issuecomment-517128427).

Share the "added" and "removed" events across dispatches in Object3D to cut down on allocation when adding and removing objects a lot. The pattern works in this case because no fields are changed on the object between dispatches -- I wonder if there's a way to extend this to cases where event data must change?


mrdoob:
Thanks!

PR #17227, opened 2019-08-10
Feature/b3dm

lydonchandra:
support for b3dm file, which uses binaryGLTF as payload
(https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/1.0/specification/TileFormats/Batched3DModel)
seeking code feedback:
* should it inherit from GLTFLoader?
* should it be B3dmLoader, or should GLTFLoader be extended?

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -145,7 +145,23 @@ THREE.GLTFLoader = ( function () {
 
 				var magic = THREE.LoaderUtils.decodeText( new Uint8Array( data, 0, 4 ) );
 
-				if ( magic === BINARY_EXTENSION_HEADER_MAGIC ) {
+				if ( magic === "b3dm" ) {
+
+					try {
+						var b3dm = new b3dmExtension( data )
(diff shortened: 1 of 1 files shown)


gkjohnson:
Do you have any example data sets that can be tested?

/ping @donmccurdy

lydonchandra:
@gkjohnson  added sample data

donmccurdy:
I'd suggest having a separate B3DMLoader class for this. Whether that should extend GLTFLoader, or create a GLTFLoader instance used internally, is up to you.

Are you planning to just support the Batched 3D Model format, or more of the 3D Tiles specification? It might be worth building on https://loaders.gl/ and having a three.js integration of that, if the full specification is the goal.

lydonchandra:
@donmccurdy   was initially planning to create Chrome DevTools extension to render B3DM file as static image, to preview loaded B3DMs under network tab. Was not aiming for full 3d-tile spec implementation.

Mugen87:
Closing. Instead of modifying `GLTFLoader`, creating `B3DMLoader` which utilizes `GLTFLoader` is the better option. This approach was already used by [VRMLoader](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/VRMLoader.js).

PR #17228, opened 2019-08-10
Shadow cleanup and refactor

supereggbert:
Moved matrix calcs from webglshadows into shadow classes, removing shadow type conditions. Sorry this is quite a large one, but I couldn't see a way to break it down anymore and still have shadows functional. It should make extending shadows cleaner and easier. ie, adding CSM, LiSPSM etc
#17198

### DIFF
--- a/src/lights/PointLightShadow.d.ts
+++ b/src/lights/PointLightShadow.d.ts
@@ -1,0 +1,8 @@
+import { OrthographicCamera } from './../cameras/OrthographicCamera';
+import { LightShadow } from './LightShadow';
+
+export class PointLightShadow extends LightShadow {
+
+	camera: OrthographicCamera;
+
+}
--- a/src/lights/PointLightShadow.d.ts
+++ b/src/lights/PointLightShadow.d.ts
@@ -1,8 +1,8 @@
-import { OrthographicCamera } from './../cameras/OrthographicCamera';
+import { PerspectiveCamera } from './../cameras/PerspectiveCamera';
 import { LightShadow } from './LightShadow';
 
 export class PointLightShadow extends LightShadow {
 
-	camera: OrthographicCamera;
+	camera: PerspectiveCamera;
 
 }
--- a/docs/api/en/lights/shadows/LightShadow.html
+++ b/docs/api/en/lights/shadows/LightShadow.html
@@ -12,51 +12,9 @@
 		<h1>[name]</h1>
 
 		<p class="desc">
-			This is used internally by [page:PointLight PointLights] for calculating shadows, and also serves as
-			a base class for the other shadow classes.
+			Serves as a base class for the other shadow classes.
 		</p>
 
-
-		<h2>Example</h2>
-		<p>
-			<code>
-//Create a WebGLRenderer and turn on shadows in the renderer
-var renderer = new THREE.WebGLRenderer();
-renderer.shadowMap.enabled = true;
-renderer.shadowMap.type = THREE.PCFSoftShadowMap; // default THREE.PCFShadowMap
-
-//Create a PointLight and turn on shadows for the light
-var light = new THREE.PointLight( 0xffffff, 1, 100 );
-light.position.set( 0, 10, 0 );
-light.castShadow = true;            // default false
-scene.add( light );
-
-//Set up shadow properties for the light
-light.shadow.mapSize.width = 512;  // default
-light.shadow.mapSize.height = 512; // default
-light.shadow.camera.near = 0.5;       // default
-light.shadow.camera.far = 500      // default
-
-//Create a sphere that cast shadows (but does not receive them)
-var sphereGeometry = new THREE.SphereBufferGeometry( 5, 32, 32 );
-var sphereMaterial = new THREE.MeshStandardMaterial( { color: 0xff0000 } );
-var sphere = new THREE.Mesh( sphereGeometry, sphereMaterial );
-sphere.castShadow = true; //default is false
-sphere.receiveShadow = false; //default
-scene.add( sphere );
-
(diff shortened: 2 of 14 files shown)


src/lights/PointLightShadow.d.ts
```diff
@@ -0,0 +1,8 @@
+import { OrthographicCamera } from './../cameras/OrthographicCamera';
+import { LightShadow } from './LightShadow';
+
+export class PointLightShadow extends LightShadow {
+
+	camera: OrthographicCamera;
```
gkjohnson: This should be `PerspectiveCamera`, right?

supereggbert: Yep, nice spot.

mrdoob:
Which one should I merge first? This one or #17171?

supereggbert:
I'd do this one first, it will break #17171 but I already have a version with the conflicts resolved, so I can easily update it afterwards.

mrdoob:
Thanks!

mrdoob:
Ops, I guess this PR is going to break use code :/

The first example broke:

Screen Shot 2019-08-21 at 2 36 51 PM

supereggbert:
Doh!! This shouldn't break things like that, it's largely just refactoring. It looks like I missed something minor. Fixed in PR #17309

WestLangley:
According to bisect, this PR has caused breakage in commit 5c22b85a1e549443f38fd7149558fe485ce6762b.

Use cases that are now broken:
```js
// setting a custom shadow frustum
light.shadow = new THREE.LightShadow( new THREE.PerspectiveCamera( 22.5, 1, 200, 700 ) );

// setting a custom shadow frustum (note: SpotLightShadow, as it should be; not LightShadow)
light.shadow = new THREE.SpotLightShadow( new THREE.PerspectiveCamera( 22.5, 1, 200, 700 ) );

// setting a custom shadow frustum for DirectionalLight
light.shadow = new THREE.LightShadow( new THREE.OrthographicCamera( - 100, 100, 100, - 100, 200, 700 ) );
```

I have not investigated further than that.

supereggbert:
@WestLangley Thanks, I've fixed the LightShadow use case but I'm confused by the SpotLightShadow use case, if you can give me a bit more detail about it's purpose? Maybe it's a legacy use case? It looks like the SpotLightShadow didn't take a camera parameter even before this PR.

WestLangley:
@supereggbert Correct you did not break that. I was not able to get it working several years ago either. Do you know the reason why the `SpotLightShadow` pattern is not correct?

WestLangley:
Also, [these lines](https://github.com/mrdoob/three.js/blob/dev/examples/webgl_lights_spotlight.html#L65-L66) do not seem to have an effect either -- even if updating the projection matrix.

In my personal examples, similar lines are:
```js
light.shadow.camera.near = 200; // can change this
light.shadow.camera.far = 700; // can change this, too
light.shadow.camera.fov = THREE.Math.RAD2DEG * 2 * light.angle; // can't seem to change this, so maybe must set a new frustum...
light.shadow.camera.updateProjectionMatrix();
```

supereggbert:
@WestLangley looking at the code and without knowing the history, the SpotLightShadow has been automatically updating the shadow camera FOV to match the light angle. Setting a custom LightShadow with a perspective matrix should give the exact same result as SpotLightShadow but without wiping out any changes you make to the FOV. eg
`spotLight.shadow = new THREE.LightShadow(new THREE.PerspectiveCamera())
spotLight.shadow.mapSize.width = 1024;
spotLight.shadow.mapSize.height = 1024;
spotLight.shadow.camera.near = 10;
spotLight.shadow.camera.far = 200;
spotLight.shadow.camera.fov = THREE.Math.RAD2DEG * 2 * light.angle; // should now work
spotLight.shadow.camera.updateProjectionMatrix();`

WestLangley:
@supereggbert Thanks, getting there...

Does it work for you when you change the two lines in the Spotlight example I referenced.
```js
spotLight.shadow.camera.near = 10;
spotLight.shadow.camera.far = 200;
```

supereggbert:
@WestLangley yep, seems to be working:
http://www.paulbrunt.co.uk/shadowtest/examples/webgl_lights_spotlight.html
It's keeping whatever values I give it.

WestLangley:
Actually, that is not what I was asking. Please do not add this line
```js
 spotLight.shadow = new THREE.LightShadow(new THREE.PerspectiveCamera());
 ```
and try my question again.

supereggbert:
In which case only near is working the far is getting reset because distance has been set on the spotlight, removing the line `spotLight.distance = 200;` allow me to update both near and far.

See code for spot shadow projection updates:
https://github.com/mrdoob/three.js/blob/257f66c935ae4f0d0bbc09837c0e323b5e29a70c/src/lights/SpotLightShadow.js#L31

WestLangley:
Ugh...

I knew this was a mess. It is not your fault, of course.

Thanks for working on this... It is really confusing as to what patterns the user is supposed to follow.

WestLangley:
Users used to have the capability to focus the shadow frustum, say on an avatar, independently of the spotlight.

Now a certain pattern must be used to achieve that because of the magic that occurs behind-the-scenes.

The example, as I pointed out, is not correct in how it attempts to set the shadow camera.

supereggbert:
You're right that's confusing, hence the refactor as a start of a clean up. I wanted to add some more advance shadow mapping techniques, but doing it with the initially code base would have resulted in a mess. Hopefully, now the code is better segmented it'll be easier to sort out inconsistencies such as this and allow people to work on improving shadows.

WestLangley:
I'm very glad you are willing to work on improving this. Thanks!

I'm going to take a break from this now...   :-)

PR #17229, opened 2019-08-10
Docs: Add missing quotes in "Creating a scene" guide.

ghost:
When copy-pasting the HTML provided on the Creating a scene tutorial, glitch.com complains about it:

image

So I think this is a mistake and fixed it.


mrdoob:
Thanks!

PR #17232, opened 2019-08-11
Google Cardboard button

clintonman:
The Google Cardboard button only works with samples run from the official threejs server. These changes make it possible to run cardboard vr mode with the single button on other servers. The change also uses the camera transform to update the controller transform, because cardboard does not have pose data. The last change is that the button events for selectend and select have unique behaviors instead of being combined as one.


Mugen87:
WebVR support has been removed, see #18129.

PR #17233, opened 2019-08-12
PointerLockControls: Revert #16171.

### LINKED ISSUE #17216: Incorrect movement with pointerlock examples
I have a found a bug with the [controls/pointerlock](https://threejs.org/examples/#misc_controls_pointerlock) example. When looking straight down, the camera doesn't move at all (the player should be moving at a consistent speed along the x,z plane regardless of camera orientation). 

I have tried using `` 

```
var cameraDirection = controls.getDirection(new THREE.Vector3()).clone();
var angle = Math.atan2(cameraDirection.x, cameraDirection.z);

// and then update camera position (if forward key pressed with) with:
controls.getObject().position.z += (Math.cos(angle))
controls.getObject().position.x += (Math.sin(angle))
```

However, this results in the camera moving in another direction  (the `angle` variable snaps to the closest multiple of pi/2 radians) when the camera is facing straight down or straight up. 

What's the solution to this problem?

Mugen87:
The PR reverts a refactoring of `PointerLockControls` which did not took the primary usage of the control class into account: The control of a First-Person camera.

The distinction between a yaw and pitch object was done on purpose. It enables a clean implementation of FPS controls in the example or application layer. Why? Because the `yaw` object only represents the rotation around the y-axis whereas `pitch` represents the rotation along the x-axis. This approach easily ensures that the player always moves along the xz-plane which is unfortunately not true with #16171 anymore. However, this is the desired behavior for FPS controls.

Fixed #17216

### DIFF
--- a/examples/js/controls/PointerLockControls.js
+++ b/examples/js/controls/PointerLockControls.js
@@ -18,7 +18,14 @@ THREE.PointerLockControls = function ( camera, domElement ) {
 	var lockEvent = { type: 'lock' };
 	var unlockEvent = { type: 'unlock' };
 
-	var euler = new THREE.Euler( 0, 0, 0, 'YXZ' );
+	camera.rotation.set( 0, 0, 0 );
+
+	var pitchObject = new THREE.Object3D();
+	pitchObject.add( camera );
+
+	var yawObject = new THREE.Object3D();
+	yawObject.position.y = 10;
+	yawObject.add( pitchObject );
 
 	var PI_2 = Math.PI / 2;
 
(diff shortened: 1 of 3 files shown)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev32/examples/misc_controls_pointerlock.html

WestLangley:
>Because the yaw object only represents the rotation around the y-axis whereas pitch represents the rotation along the x-axis.

That is exactly what setting Euler rotation order to `YXZ` does.

There is nothing wrong with `PronterLockControls` as far as I know. The problem is with the example.

I think your concern can be addressed pretty easily. I'll try.

Mugen87:
> That is exactly what setting Euler rotation order to YXZ does.

I've meant something different when referring to `yaw` and `pitch`. The conceptual idea behind using two 3D objects is that it maps nicely to the body/head relationship of a virtual player. You can move the player's body through 3D space by just transforming the `yaw` object. Although the `yaw` object is rotated along the `y` axis, it never tilts or rolls. This makes the general handling easier, especially if you want to extend it.

I just wanna say that I've seen the hierarchy `yaw`-> `pitch` -> `camera` several times in game development and it does make sense.

Mugen87:
Closing for now, seehttps://github.com/mrdoob/three.js/pull/17241#issuecomment-524616981

PR #17239, opened 2019-08-12
Remove Vector4 deprecation warning

### LINKED ISSUE #17238: Vector4 incorrectly (?) deprecated as per typings
The [typings file](https://github.com/mrdoob/three.js/blob/dev/src/math/Vector4.d.ts) for `Vector4` has a deprecation for the entire class:

```ts
/**
 * @deprecated use {@link Vector3 THREE.Vector3} instead.
 */
```

I'm not sure if this was really intended because you cannot (generally) replace usages of `Vector4` with `Vector3`. For example, when transforming a direction vector with a `Matrix4` you need something like:

```ts
const result = new Vector4(dir.x, dir.y, dir.z, 0).applyMatrix4(matrix)
```

So I think this deprecation is incorrect, there's also no mention of deprecation in [`Vector4.js`](https://github.com/mrdoob/three.js/blob/dev/src/math/Vector4.js) or the deprecation list.

##### Three.js version

- [x] Dev
- [ ] r107
- [ ] ...

gkjohnson:
Fix #17238

### DIFF
--- a/src/math/Vector4.d.ts
+++ b/src/math/Vector4.d.ts
@@ -5,10 +5,6 @@ import { BufferAttribute } from './../core/BufferAttribute';
 import { Vector } from './Vector2';
 
 /**
- * @deprecated use {@link Vector3 THREE.Vector3} instead.
- */
-
-/**
  * 4D vector.
  *
  * ( class Vector4 implements Vector<Vector4> )


Mugen87 (approved):

mrdoob:
Thanks!

PR #17240, opened 2019-08-12
Node clearcoat normals

arobertson0:
Added clearcoat normals to nodes.

I had to rework part of the logic of ReflectNode, which assumed there was only one possible normal.

Now, when accessing the environment map, there is a contextual cubeNormal parameter that enables a parent node to switch which normal lights is being reflected of. I had to do the same for the normal map nodes.

Finally, I changed the way geometry.clearCoatNormal is being assigned. This way is equivalent, but it allows us to have `clearCoatNormal` as a node output if necessary.

I've added clear coat normal options to the physical material demo. It's also possible to just move them

/ping @sunag

### DIFF
--- a/examples/jsm/nodes/accessors/ReflectNode.js
+++ b/examples/jsm/nodes/accessors/ReflectNode.js
@@ -3,12 +3,14 @@
  */
 
 import { TempNode } from '../core/TempNode.js';
+import { NormalNode } from './NormalNode.js';
 
-function ReflectNode( scope ) {
+function ReflectNode( scope, normal ) {
 
-	TempNode.call( this, 'v3', { unique: true } );
+	TempNode.call( this, 'v3' );
 
 	this.scope = scope || ReflectNode.CUBE;
+	this.normal = normal || new NormalNode();
 
 }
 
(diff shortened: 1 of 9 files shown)


examples/jsm/nodes/accessors/ReflectNode.js
```diff
@@ -44,29 +46,25 @@ ReflectNode.prototype.generate = function ( builder, output ) {
...
-				result = 'reflectCubeVec';
+				result = 'vec3( -1.0 * ' + reflectVec + '.x, ' + reflectVec + '.yz )';
 
 				break;
 
 			case ReflectNode.SPHERE:
 
-				var reflectVec = new ReflectNode( ReflectNode.VECTOR ).build( builder, 'v3' );
-
-				builder.addNodeCode( 'vec2 reflectSphereVec = normalize( ( viewMatrix * vec4( ' + reflectVec + ', 0.0 ) ).xyz + vec3( 0.0, 0.0, 1.0 ) ).xy * 0.5 + 0.5;' );
```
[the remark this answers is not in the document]
sunag: This is used in together with `unique` to share the same code of the class and not only of node. For example:

```
material.color = new ReflectNode();
material.emissive = new ReflectNode();
// if is share he will create two codes for each slot
// if is unique he will share the same code for each slot
```

examples/jsm/nodes/inputs/CubeTextureNode.js
```diff
@@ -36,7 +36,19 @@ CubeTextureNode.prototype.generate = function ( builder, output ) {
 	}
 
 	var cubetex = this.getTexture( builder, output );
-	var uv = this.uv.build( builder, 'v3' );
+
+	
+	var uv = this.uv || builder.context.cubeUv || new ReflectNode();
+
+	if ( !uv ){
```
sunag: I think that here uv never will be undefined...

examples/jsm/nodes/misc/NormalMapNode.js
```diff
@@ -68,12 +68,17 @@ NormalMapNode.prototype.generate = function ( builder, output ) {
...
 		var perturbNormal2Arb = builder.include( NormalMapNode.Nodes.perturbNormal2Arb );
 
-		this.normal = this.normal || new NormalNode();
 		this.position = this.position || new PositionNode( PositionNode.VIEW );
 		this.uv = this.uv || new UVNode();
 
+		var normal = this.normal || builder.context.normal || new NormalNode();
+
+		if(this.normal){
+			console.log(this.normal);
```
sunag: debug code

sunag (changes requested):
@arobertson0 Thank you for your contribution, I reviewed what I could at the moment. I will have to analyze this more closely as soon as finish some other PR in line. 👍

examples/jsm/nodes/accessors/ReflectNode.js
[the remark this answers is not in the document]
arobertson0: The reason I changed it is because it is no longer unique. Some times we reflect normal, some times we reflect clearCoatNormals. I tough making a CleatCoatReflectNode was too much code duplication.

examples/jsm/nodes/inputs/CubeTextureNode.js
arobertson0: I removed the default value from the constructor. This allows uv to be contextually defined here.  If there is no contextual uv, the default value is still `new ReflectNode()`.

arobertson0:
> @arobertson0 Thank you for your contribution, I reviewed what I could at the moment. I will have to analyze this more closely as soon as finish some other PR in line.

Thanks for the quick review.

PR #17241, opened 2019-08-12
PointerLockControls: added .moveForward(), .moveRight() methods

WestLangley:
This is one way of addressing #17216.

To be honest, it is a bit confusing to me the way this example handles `velocity` and `direction`. I am not sure what behavior was intended... So maybe the changes made here are not what is preferred.

The two convenience methods should still be useful, however.

### DIFF
--- a/examples/js/controls/PointerLockControls.js
+++ b/examples/js/controls/PointerLockControls.js
@@ -22,6 +22,8 @@ THREE.PointerLockControls = function ( camera, domElement ) {
 
 	var PI_2 = Math.PI / 2;
 
+	var vec = new Vector3();
+
 	function onMouseMove( event ) {
 
 		if ( scope.isLocked === false ) return;
@@ -105,6 +107,27 @@ THREE.PointerLockControls = function ( camera, domElement ) {
 		};
 
 	}();
+
+	this.moveForward = function ( distance ) {
+
+		// move forward parallel to the xz-plane
+		// assumes camera.up is y-up
+
+		vec.setFromMatrixColumn( camera.matrix, 0 );
+
+		vec.crossVectors( camera.up, vec );
+
+		camera.position.addScaledVector( vec, distance );
+
+	};
+
+	this.moveRight = function ( distance ) {
+
+		vec.setFromMatrixColumn( camera.matrix, 0 );
+
+		camera.position.addScaledVector( vec, distance );
+
+	};
 
 	this.lock = function () {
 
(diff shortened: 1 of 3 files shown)


WestLangley:
https://raw.githack.com/WestLangley/three.js/dev_pointer_lock/examples/misc_controls_pointerlock.html

Mugen87:
The sensitivity of they `WASD` movement is extremely high. At least on my iMac. Feels like a "time lapse" mode.

WestLangley:
>direction represents the direction the player wants to head. It is directly derived from the keyboard input. velocity is the actual movement speed along a defined direction.

I understand that. It is the example code that I do not understand.

In any event, I made a PR change.

Mugen87:
> In any event, I made a PR change.

The movement is better now. I leave it up to @mrdoob which of both PRs (this one or #17233) is going to be merged.

WestLangley:
@Mugen87 I wish you would be helpful so we could make progress together, instead of focusing on reverting my PRs.

How about testing this and approving it, instead? That would actually be helpful.

Mugen87:
> instead of focusing on reverting my PRs.

That's sound a bit I would do this all the time^^. Come on, that's not fair.

I just think that the previous approach of `PointerLockControls` is better (see https://github.com/mrdoob/three.js/pull/17233#issuecomment-520753305).

supereggbert:
@Mugen87 @WestLangley Both methods have their merits, the old method is certainly clearer being more verbose and the new method is defiantly more elegant even if initially flawed. 

As a design preference, instead of separate functions for "moveForward" and "moveRight", it might be better with a single function that is just "move" which take a vector3. It can then accommodate all three axis and allow for the jump without treating it as an exception to the rule.

There isn't anything wrong with the example per say, but it is more confusing then it should be. Having four booleans, and casting them, instead of simply using two numbers. Plus, It's treating jump in a completely different way which does make it quite difficult to follow. The mechanics are clear when you realise that velocity was added to account for the acceleration of gravity; it's redundant in relation to lateral movement. The example overall does seem more improvised then planned. Jump looks like it was added as an extra bonus, not part of any design brief, which has probably led to much of the confusion.

WestLangley:
This improvement to `PointerLockControls` is both trivial (now that I figured how to do it) and correct. I am not sure what all the controversy is about.

If someone wants to change the example in a separate PR that is fine with me.

Mugen87:
I'm okay with merging this PR so the example is finally fixed. However, I would still like to hear @mrdoob feedback and act accordingly.

Mugen87:
Merging so the example is fixed for the next release. #17233 can be considered later.

Issue #17255, opened 2019-08-14
WebGL contextAttributes are ignored if you pass a context as a parameter


soadzoor:
##### Description of the problem

I've just spent like 5-6 hours trying to figure out why `canvas.toDataURL()` was constantly giving me a blank image, in other words: why `preserveDrawingBuffer: true` did not have any effect.

I was able to narrow down the problem to the three.js library, so I investigated the WebGLRenderer, and this is what I've found in the constructor:
`_gl = _context || _canvas.getContext( 'webgl', contextAttributes ) || _canvas.getContext( 'experimental-webgl', contextAttributes );`

This means, if you pass a context as a parameter, some of the context attributes will be ignored, including (but not limited to): `preserveDrawingBuffer`, `antialias`, `premultipliedAlpha`, and `alpha` just to mention a few important ones.

Now you might ask, why would one pass a WebGLRenderingContext to the WebGLRenderer's constructor. The answer is:
**That is the recommended way to create a WebGL2 renderer, based on the [docs](https://threejs.org/docs/#manual/en/introduction/How-to-use-WebGL2) !**

This is a pretty huge problem, I think we should implement (and recommend) a better way to get a WebGL2 Renderer,  or at least, put a huge **WARNING** to the related pages on the docs for now.

It's very misleading the way it is now, because I provide the necessary parameters, and they're simply ignored, not even a warning in the console appears, or whatsoever. Programmers can waste a lot of time figuring out why those parameters don't have any effect, just like I did.

Maybe we should put a `console.warn` above the `_gl = _context || _canvas.getContext( 'webgl', contextAttributes ) || _canvas.getContext( 'experimental-webgl', contextAttributes );` line: if the context was provided, warn the users that the given attributes (that were passed along with the context) will be ignored, and they should give those parameters to the context itself upon creation.
Modifying those parameters on the go according to the given ones would be even better, but it seems it's not possible: [link](https://stackoverflow.com/questions/17153833/how-do-i-set-the-attributes-of-a-webgl-context-after-it-has-been-created)

I can create a PR with a warning like that, but we should discuss it first.

##### Three.js version
- [x] r107

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
[...]

PR #17263, opened 2019-08-15
Added hint for context attributes to WebGL2 guide

soadzoor:
As discussed in [17255](https://github.com/mrdoob/three.js/issues/17255)

### DIFF
--- a/docs/manual/en/introduction/How-to-use-WebGL2.html
+++ b/docs/manual/en/introduction/How-to-use-WebGL2.html
@@ -48,9 +48,14 @@ if ( WEBGL.isWebGL2Available() === false ) {
 		automatically convert the built-in material's shader code to GLSL ES 3.00.
 	</p>
 
+	<p>
+		Since you are manually creating the WebGL 2 rendering context, you also have to pass in all necessary context attributes.
+		Note: It's not possible to modify these attributes after the context has been created, so passing them to the WebGLRenderer won't have any effect.
+	</p>
+
 	<code>
 var canvas = document.createElement( 'canvas' );
-var context = canvas.getContext( 'webgl2' );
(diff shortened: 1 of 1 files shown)


docs/manual/en/introduction/How-to-use-WebGL2.html
```diff
@@ -48,9 +48,14 @@ <h2>Workflow</h2>
...
 
+	<p>
+		Since you are manually creating the WebGL 2 rendering context, you also have to pass in all necessary context attributes.
+		Note: It's not possible to modify these attributes after the context has been created, so passing them to the WebGLRenderer won't have any effect.
+	</p>
+
 	<code>
 var canvas = document.createElement( 'canvas' );
-var context = canvas.getContext( 'webgl2' );
+var context = canvas.getContext( 'webgl2', { antialias: false } );
```
Mugen87: I would remove this change. It's a minimal code example and `{ antialias: false }` does not make sense here. It's set in `webgl2_multisampled_renderbuffers` because multisampled renderbuffers (offscreen MSAA) should be demonstrated.

soadzoor: Is there a contextAttribute that makes sense here? I'd just like see a demonstration on how one should do this the right way.

What if I change it to `{ alpha: false }`?

Mugen87: Sounds good.

soadzoor: Changed to `{ alpha: false }`

Mugen87 (approved):

mrdoob:
Thanks!

PR #17269, opened 2019-08-16
Audio: Added duration property

sttz:
This PR adds a `duration` property to the `Audio` class to precisely set how long the buffer should be played.

With this PR (and the fix in #17268), all parameters of `AudioBufferSourceNode.start()` will be usable through properties of the `Audio` class, which was the intention of my outdated PR #10254.

As mentioned in the previous PR, I need to set the duration in my implementation of audio sprites. Just calling `stop()` in JS events is unfortunately not precise enough.

### DIFF
--- a/src/audio/Audio.d.ts
+++ b/src/audio/Audio.d.ts
@@ -17,6 +17,7 @@ export class Audio extends Object3D {
 	loop: boolean;
 	startTime: number;
 	offset: number;
+	duration: number | undefined;
 	playbackRate: number;
 	isPlaying: boolean;
 	hasPlaybackControl: boolean;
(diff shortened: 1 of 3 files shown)


src/audio/Audio.d.ts
```diff
@@ -17,6 +17,7 @@ export class Audio extends Object3D {
 	loop: boolean;
 	startTime: number;
 	offset: number;
+	duration: number;
```
Mugen87: If the default value is `undefined`, we should do this:
```
duration: number | undefined;
```

sttz: Fixed, thanks for the catch!

Mugen87 (approved):

mrdoob:
Thanks!

Issue #11598, opened 2017-06-23
FBX loader problem with transparency


CarlBateman:
##### Description of the problem 

When importing FBX, some transparency information is lost.

Using the FBX loader example from Threejs.org site, just loading different model. Model at pearl_walk.fbx to http://webglworkshop.com/fbx/standing_idle.fbx

Looks like this
![image](https://user-images.githubusercontent.com/2813834/27481051-a9555454-5812-11e7-8dfd-615aa0d5da53.png)

Should look like this (from PlayCanvas) Note: PlayCanvas uses "Alpha to coverage"
![image](https://user-images.githubusercontent.com/2813834/27481301-da09ee7e-5813-11e7-9565-d5ea35c052aa.png)

or this (from SketchFab)
![image](https://user-images.githubusercontent.com/2813834/27481512-d4d4bbae-5814-11e7-97ba-888083923b82.png)

##### Three.js version

- [ ] Dev
- [x] r65

##### Browser

- [x] All of them

##### OS

- [ ] All of them
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

tauba:
I am having the very same problem here with another FBX file. I cannot use transparency on my model

And I am facing this output on the console

FBXLoader.js:542 
THREE.FBXLoader: Unknown texture application of type  "TransparentColor, skipping texture.

FBXLoader.js:542 
THREE.FBXLoader: Unknown texture application of type  "SpecularColor, skipping texture.

SntsDev:
Yes, FBXLoader.js doesn't manage some options, like transparency. 

I ended up adding `alphaTest` to `parseParameters` function of FBXLoader:
```javascript
function parseParameters( properties, textureMap, childrenRelationships ) {
    
      var parameters = {};
            
      if (properties.TransparentColor && properties.TransparentColor.value === 1) {
                
         parameters.alphaTest = 0.3;
                
      }
      ...
```
It works ok with use of PNG for maps.

TristanVALCKE:
I have a fix that work perfectly for make human export, maybe it will fix this too ! I will make a PR as soon as possible

looeee:
I added support for several more material parameters and map types in #12317

@CarlBateman can you test with the latest dev version of FBXLoader and see if this problem is fixed?

TristanVALCKE:
[...]

PR #17279, opened 2019-08-19
FBXLoader: Add TransparencyFactor

safu9:
I had warning message when loading fbx file,
`THREE.FBXLoader: TransparancyFactor map is not supported in three.js, skipping texture.`

Since TransparancyFactor seems to be another name of TransparentColor, I added it in FBXLoader.

Related: https://github.com/mrdoob/three.js/issues/11598


looeee:
> Since TransparancyFactor seems to be another name of TransparentColor

`TransparencyFactor` maps to `opacity`, while `TransparentColor` maps to `alphaMap` so I don't think this is correct.

There's also an FBX property `Opacity` which is what the loader currently uses, and `TransparencyFactor` is ignored. It's possible that's not correct though.

Are you testing this change with any models? If so, can you share them?

safu9:
Sorry for late reply.
I compared with the same model exported by Blender.

Before:
Screenshot from 2019-08-26 22-04-07
After this PR:
Screenshot from 2019-08-26 22-07-44

But I found some parts are not working correctly ... (z-buffer?)
Screenshot from 2019-08-26 22-08-40

safu9:
Model file is here. [Tree.zip](https://github.com/mrdoob/three.js/files/3541253/Tree.zip)

mrdoob:
@looeee what do you want to do with this one?

looeee:
Sorry it took me so long to check this. 

It seems like the easy solution here is to manually set `material.alphaTest = 0.5`:

2020-04-14 12_44_19-three js webgl - FBX loader

Treating TransparencyFactor as an alphaMap as in this PR is _not_ correct since that makes the leaves transparent. At least in this case, it's presumably meant to be an alpha mask map, which could be handled in the loader like this:

```
case 'TransparencyFactor':
        parameters.alphaMap = scope.getTexture( textureMap, child.ID );
	parameters.alphaTest = 0.5;
	break;
```

However, when I do this, the mask is inverted:

2020-04-14 12_57_59-three js webgl - FBX loader

I'm not sure why this is happening, seem to be that 

``` js
material.alphaMap = texture;
material.alphaTest = 0.5;
```

... and 

``` js
material.map = texture;
material.alphaTest = 0.5;
```

...give inverse results, for the same `texture`.

looeee:
I've found one more model that uses TransparencyFactor in this thread:

https://github.com/facebookincubator/FBX2glTF/issues/158

However, in that case it looks like a standard transparency map. They also note there that they've never encountered a model with this property before. Both of these files are created by a recent Blender version (2.79 and 2.80) so maybe they have started to use it. 

I've changed my mind on this - I think treating this the same way as `TransparentColor` is OK. Users will have to manually set `alphaTest` when required (as usual).

looeee (approved):

looeee:
@mrdoob this can be merged now.

mrdoob:
Thanks!

safu9:
Thank you!

looeee:
@safu9 thanks for your patience ^_^ 

And note that you probably still need to manually set `material.alphaTest = 0.5;` here.

PR #17281, opened 2019-08-19
fix X face inversion in LightProbeGenerator

nipmarsh:
The computation of the direction is inverted for the positive and negative X

### DIFF
--- a/examples/js/lights/LightProbeGenerator.js
+++ b/examples/js/lights/LightProbeGenerator.js
@@ -62,9 +62,9 @@ THREE.LightProbeGenerator = {
 
 				switch ( faceIndex ) {
 
-					case 0: coord.set( - 1, row, - col ); break;
+					case 0: coord.set( 1, row, col ); break;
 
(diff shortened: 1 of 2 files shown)


WestLangley:
The current code is correct when passing in a `CubeTexture`. See [this SO answer](https://stackoverflow.com/questions/45998274/why-are-the-faces-of-a-cubemap-labelled-px-nx-py-ny-pz-nz/45999554#45999554).

nipmarsh:
I see but this is related to another pull request (https://github.com/mrdoob/three.js/pull/17282)
If it was really like you say, i won't have to due it in the pull request https://github.com/mrdoob/three.js/pull/17282

nipmarsh:
Ok, i've checked and had an cubeCamera example in https://github.com/mrdoob/three.js/pull/17282
When generating the lightProbe from the cubeCamera, the X face should be inverted.
I have not double checked for a Cubemap, my bad
I close this pull request

PR #17282, opened 2019-08-19
Generate lightProbe from CubeCamera

nipmarsh:
Add a method to generate a lightProbe from a cubeCamera.
It assumes that the cubeCamera renderTarget is RGBA


WestLangley:
Is there a reason why you do not pass in a `WebGLRenderTarget.texture` instead of a `CubeCamera`?

See how `WebGLBackground` handles this:

https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLBackground.js#L105-L107

nipmarsh:
I could have only pass the renderTarget but the method is named fromCubeCamera so semanthically this is better to pass the CubeCamera

Mugen87:
Live demo: http://raw.githack.com/nipmarsh/three.js/lightprobegenerator_fromCubeCamera/examples/webgl_lightprobe_cubecamera.html

WestLangley:
>but the method is named fromCubeCamera

You can change the name, no? Can you reuse the existing method? See how `WebGLBackground` handles this.

examples/js/lights/LightProbeGenerator.js
```diff
@@ -116,6 +116,105 @@ THREE.LightProbeGenerator = {
...
+
+			var pixelSize = 2 / imageWidth;
+
+			for ( var i = 0, il = data.length; i < il; i += 4 ) { // RGBA assumed
+
+				// pixel color
+				color.setRGB( data[ i ] / 255, data[ i + 1 ] / 255, data[ i + 2 ] / 255 );
+				// pixel coordinate on unit cube
+
+				if (renderer.gammaOutput) {
```
WestLangley: All light probes must be in linear space. It has nothing to do with render settings.

WestLangley:
Your environment map is a Cornell box with colors of the form `[ 0, 1, 1 ]`, so converting to and from sRGB space does nothing. Hence it is difficult to test if your color-space is correct.

examples/js/lights/LightProbeGenerator.js
nipmarsh: Yes and this is why it depends on the renderer gammaOutput, keep in mind that the LightProbes is computed from a cubeCamera. During the rendering stage of the cubemap, gamma correction is applied if renderer.gammaOutput is true.

nipmarsh: The name is meaningful I think. I do not understand what is the matter with it
I have to read pixels back from the GPU to the CPU. The existing method is a little bit different and based on images. It seams a lot more easy for people who are not very confortable with CG to handle this seamlessly

nipmarsh: > Your environment map is a Cornell box with colors of the form `[ 0, 1, 1 ]`, so converting to and from sRGB space does nothing. Hence it is difficult to test if your color-space is correct.

True, I will check this tomorrow with another one

WestLangley: >During the rendering stage of the cubemap, gamma correction is applied if renderer.gammaOutput is true.

Right. And I have suggested repeatedly that you handle the cube camera at the application level and pass in `cubeCamera.renderTarget` to a properly-named method.

I do appreciate you doing this, however. :-)

In fact, it should be implemented on the GPU. Then we can auto-generate irradiance maps from any WebGLRenderTargetCube.

nipmarsh: A fromCubeRenderTarget will be ok for you ?
Yes, it should be on the GPU too, there is another pull request for this https://github.com/mrdoob/three.js/pull/16152

WestLangley: I feel pretty strongly that would be a better API. That is why I asked you if there was a reason why you didn't do it that way.

If you want, we can ask @mrdoob what he prefers. I don't want to waste your time.

nipmarsh: This is indeed a matter of point of view, maybe @mrdoob can give it conceptual advice ;-)

WestLangley:
You are getting there, but there is still something wrong with the way you handle colorspace.

These should produce the same light probe, and it should be in linear colorspace.
```js
lightProbe.copy( LightProbeGenerator.fromRenderTargetCube( renderer, cubeCamera.renderTarget ) );

lightProbe.copy( LightProbeGenerator.fromCubeTexture( cubeTexture ) );
```

WestLangley:
The texture files are not needed in this PR, right?

examples/js/lights/LightProbeGenerator.js
```diff
@@ -116,6 +116,105 @@ THREE.LightProbeGenerator = {
 
 		return new THREE.LightProbe( sh );
 
+	},
+
+	fromCubeRenderTarget: function ( renderer, renderTarget ) {
```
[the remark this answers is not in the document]
nipmarsh: Can you explain what is wrong for you?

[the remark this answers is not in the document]
nipmarsh: I think color correction in threejs is a little bit messy.
So i should simply assume that the renderTarget is linear. What do you think?
Color correction should be applied as a render pass if the user wants it.
It would be more simple to understand and all buffer would be in linear color space has it should be without wondering in which color space we are working on!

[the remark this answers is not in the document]
WestLangley: I think you can follow the same pattern in your method I used in #17385 and decode to linear based on the encoding of the `WebGLRenderTargetCube`.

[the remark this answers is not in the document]
WestLangley: I think the function signature should be
```js
fromRenderTargetCube: function ( renderer, renderTargetCube ) {
```

mrdoob:
I think this PR got messed up... 
https://github.com/mrdoob/three.js/pull/17282/files

nipmarsh:
> I think this PR got messed up...
> https://github.com/mrdoob/three.js/pull/17282/files

Ouch, i accidently merge instead of rebase dev branch, sorry...
I will redo a clean pull request and close this one

WestLangley:
@nipmarsh Would you be willing to redo a clean PR?

nipmarsh:
I already rebased the PR softly, is it still messed up?

PR #17290, opened 2019-08-20
CubeTexture mipmaps from CubeCamera render targets

jee7:
For #17072. If manual mipmaps are set from the render target of a CubeCamera, then the textures are on the GPU and need to be copied via framebuffer. This is different than just setting a manual mipmap from a loaded image on the CPU.


Mugen87:
It seems in the last years there was only one feature request for this special code path. If not more devs require the ability to use render targets as mipmaps for normal cube textures, it's better to close the PR and not add the additional logic to `WebGLTextures`.

PR #17294, opened 2019-08-20
TransformControls: Fix gizmo scale hover

enomado:
The bug was:

- https://threejs.org/examples/misc_controls_transform.html
- scale <kbd>R</kbd>
- hover mouse on white gizmo's cubes(any of it). All of them will change color only if hovered only left. Other is not.

Bug was caused by sharing a single color object and mutation of it by the last element.


Mugen87:
/ping @arodic Is this change okay like that?

mrdoob:
Looks good to me.

mrdoob:
Thanks!

mrdoob:
FYI @arodic

PR #17295, opened 2019-08-20
Separated the meanings of PHYSICAL

arobertson0:
For physical materials, the define `PHYSICAL` accidentally ended up having two meanings:
-  Rendering a material with reflectivity, clearcoat, etc.
-  Rendering a material that is a _MeshStandardMaterial_ (which includes _MeshPhysicalMaterial_)

And there is also the define `STANDARD`, used to mean:
 -  Rendering a material that is a _MeshStandardMaterial_ and **NOT** a _MeshPhysicalMaterial_

In this PR, I separate the two meanings thus:
- `PHYSICAL`: A _MeshStandardMaterial_ or subclass thereof.
- `ADVANCED_PHYSICAL`: A _MeshPhysicalMaterial_ or subclass thereof.

Additionally, each subfeature ( reflectivity, clearcoat, etc ) of `ADVANCED_PHYSICAL`  receives it's own define. This clearly delimits the contribution of each feature.


examples/jsm/nodes/materials/nodes/StandardNode.js
```diff
@@ -32,7 +32,15 @@ StandardNode.prototype.build = function ( builder ) {
 
 	var code;
 
-	builder.define( this.clearCoat || this.clearCoatRoughness || this.clearCoatNormal ? 'PHYSICAL' : 'STANDARD' );
+	builder.define('PHYSICAL');
+
+	var useClearCoat = this.clearCoat;
```
sunag: I loved your code, only here if you can preserve the previous logic, the code was writing thinking of using any of these inputs to active clearCoat.

```
this.clearCoat || this.clearCoatRoughness || this.clearCoatNormal
```

Mugen87:
@arobertson0 Can you also resolve the merge conflicts?

WestLangley:
Originally `STANDARD` corresponded to `MeshStandardMaterial` and `PHYSICAL` corresponded to `MeshPhysicalMaterial`. That is consistent with how the other materials are implemented.

Why do you prefer that `PHYSICAL` correspond to `MeshStandardMaterial`?

---

>For physical materials, the define PHYSICAL accidentally ended up having two meanings:
>Rendering a material that is a MeshStandardMaterial (which includes MeshPhysicalMaterial)

That should not have happened.

arobertson0:
> Why do you prefer that `PHYSICAL` correspond to `MeshStandardMaterial`?

The shader for both _MeshStandardMaterial_ and _MeshPhysicalMaterial_ always defined `PHYSICAL`, so we are in a strange situation where:
- The superclass _MeshStandardMaterial_ was defining `STANDARD` & `PHYSICAL`
- The subclass _MeshPhysicalMaterial_ was defining only `PHYSICAL`

In this situation, the pieces of shaders that are detecting if we are in a standard/physical material are actually checking if `PHYSICAL` is defined.

So, the reason I choose to have `PHYSICAL` correspond to _MeshStandardMaterial_ is to reduce the amount of changes.

What would you think of renaming the defines so that standard/physical materials define `STANDARD`, and the physical material defines an additional `PHYSICAL`? I think this would have the shader code match the js code, and it would correctly express the intent of the class hierarchy

WestLangley:
@arobertson0 Thank you for your reply.

Refreshing my memory a bit ( https://github.com/mrdoob/three.js/pull/8650 ), I'll support whatever nomenclature you and @Mugen87 think is best.

There has been a lot of changes to the shader code in the last few weeks.That may explain some of the confusion.

mrdoob:
Nice clean up!

Mugen87:
> What would you think of renaming the defines so that standard/physical materials define STANDARD, and the physical material defines an additional PHYSICAL? I think this would have the shader code match the js code, and it would correctly express the intent of the class hierarchy

Sounds good!

arobertson0:
Thanks @WestLangley and @Mugen87 for the review.

PR #17300, opened 2019-08-21
Loaders: Remove obsolete callbacks.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17293#issuecomment-523218182


mrdoob:
Thanks!

PR #17302, opened 2019-08-21
Loader: Introduce new API.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17293#issuecomment-523218182

`ObjectLoader` is now the first loader that is derived from `Loader`.


mrdoob:
Thanks!

mrdoob:
Ops, I guess this PR is going to break use code :/

The first example broke:

Screen Shot 2019-08-21 at 2 36 51 PM

PR #17305, opened 2019-08-21
Add functions to enable/disable all layers

Oletus:
This reduces the need to access Layers.mask directly when code needs to enable/disable all layers on an object.

### DIFF
--- a/docs/api/en/core/Layers.html
+++ b/docs/api/en/core/Layers.html
@@ -80,6 +80,16 @@
 			Toggle membership of *layer*.
 		</p>
 
+		<h3>[method:null enableAll]()</h3>
+		<p>
+			Add membership to all layers.
+		</p>
+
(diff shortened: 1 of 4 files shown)


Mugen87:
Maybe it's a good idea to enhance the respective [layers example](https://threejs.org/examples/webgl_layers) and add two new `dat.gui` buttons "enable all" and "disable all".

Oletus:
Agree completely, added the buttons to the example. I had to change the existing checkboxes to buttons as well so they wouldn't display outdated values once enableAll/disableAll is called.

Mugen87 (approved):
Tested your branch on my computer. The example works fine!

Oletus:
@mrdoob This should be ready to merge, if you think this feature is a good idea. We'd find it useful at least!

PR #17320, opened 2019-08-22
TS: add more fromArray and toArray function types

vlenhart:
This commit does the following:
- add arraylike variant of the fromArray function to Color, Matrix3, Matrix4, Quaternion, SphericalHarmonics3
- add arraylike variant of the toArray function to Matrix3, Matrix4, Quaternion, SphericalHarmonics3 (Color already had it)
- add the missing offset functionality to SphericalHarmonic3's fromArray and toArray function, also the toArray function can now insert into an existing array or array-like.

Also I added some comments and fixed existing ones.


mrdoob:
Thanks!

PR #17328, opened 2019-08-23
Add InterleavedBufferAttribute.prototype.toJSON

jsantell:
Serializing an object that has InterleavedBufferAttributes will throw -- not terribly familiar with interleaved buffers, not sure if any other values should be returned.

### DIFF
--- a/src/core/InterleavedBufferAttribute.js
+++ b/src/core/InterleavedBufferAttribute.js
@@ -1,3 +1,4 @@
+import { BufferAttribute } from './BufferAttribute.js';
 
 /**
  * @author benaadams / https://twitter.com/ben_a_adams
@@ -130,6 +131,16 @@ Object.assign( InterleavedBufferAttribute.prototype, {
 		this.data.array[ index + 3 ] = w;
 
 		return this;
+
+	},
+
+	toJSON: function ()	{
+
+		var data = BufferAttribute.prototype.toJSON.call( this );
+
+		data.isInterleavedBufferAttribute = true;
+
+		return data;
 
 	}
 


Mugen87:
There was a similar PR some time ago, see #16838

Fact is that interleaved attributes share a common interleaved buffer. That makes serialization and deserialization tricky since you can't serialize interleaved attributes in an isolated way. If you restored the attributes, they don't share a common buffer anymore. There are also no copy and clone methods for interleaved attributes because of their special architecture.

jsantell:
My main interest is making it possible to serialize a scene with InterleavedBufferAttributes -- much like WebGLRenderTargets (https://github.com/mrdoob/three.js/pull/16816), the existence of either of these currently makes the serialization *of the rest of the scene* impossible. I'd be happy with `toJSON()` existing and returning an empty object, but I think there is some extra information that can be provided (size, type, etc.).

As it is now, unserializable objects are not only un-deserializable via ObjectLoader (issue of serialization), they break `toJSON()` for the entire scene (issue of breaking serializing everything else in the scene)

jsantell:
Checking for `.toJSON` existence from BufferGeometry's toJSON (or any other object that may serialize an interleaved buffer attribute) would also solve the use case of serializing the rest of a scene if `toJSON()` should be reserved for completely serializable objects (e.g. can be hydrated via ObjectLoader) as another option

Mugen87:
> As it is now, unserializable objects are not only un-deserializable via ObjectLoader (issue of serialization), they break toJSON() for the entire scene

That's indeed a problem. I think I would prefer to ignore objects which are not serializable. Do you think this approach would also work in context of `WebGLRenderTarget`'s?

looeee:
>  I think I would prefer to ignore objects which are not serializable. 

Could we log a warning?

PR #17341, opened 2019-08-25
Relax class testing in TransformControls

jsantell:
Check only string properties (isPerspectiveCamera, isOrthographicCamera) rather than class membership in TransformControls to allow use different instances of camera classes. Most checks for three objects only test the `isFoo` properties, where `instanceof` is seldom used, outside of some internal classes.

Without this patch, a camera from a different THREE namespace than the TransformControls results in strangely broken controls.


Mugen87:
On latest `dev`, the `if` statement does not exists anymore, see bb42e1586fffffa6dce430bc0603bddb9231a11e

It seems this commit is missing on your branch.

jsantell:
Not sure how that happened!

PR #17342, opened 2019-08-25
Devtools: Expose revision

jsantell:
With the removal of the console.log messaging of the revision (https://github.com/mrdoob/three.js/commit/1013fbb9d8f25751c9f17a43f2ce654224e024bd#diff-c0e88b98497597a015ecf238e91ac3a0), there's no way to discover what version three is running when moduleified programmatically. One isn't really using three.js without using the renderer, and I think that'd be a good place to expose the revision, where it can be interpreted by devtools or folks debugging a page. Similarly removed the console.log messaging for `WebGL2Renderer`.

I don't feel strongly regarding `.REVISION` vs `.revision`, other than the capitalized form only makes sense if you're familiar with `THREE.REVISION`, but in the context of a renderer, `.revision` seems more appropriate.

### DIFF
--- a/src/Three.js
+++ b/src/Three.js
@@ -1,4 +1,5 @@
 import './polyfills.js';
+import { REVISION } from './constants.js';
 
 export { WebGLMultisampleRenderTarget } from './renderers/WebGLMultisampleRenderTarget.js';
 export { WebGLRenderTargetCube } from './renderers/WebGLRenderTargetCube.js';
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

donmccurdy:
What about `import { REVISION } from 'three';`? I haven't checked if that works with ES modules, but thought it did.

jsantell:
> What about `import { REVISION } from 'three';`? I haven't checked if that works with ES modules, but thought it did.

That'd work, but requires either source code already importing it like that snippet (unlikely), or using a namespace import (e.g. `import * as THREE from 'three'`, better chance than `REVISION` specifically, but also not necessary for the namespace to be available globally like `window.THREE`, otherwise there's no means of the information being available at runtime in a module scenario AFAICT

donmccurdy:
Oops, yes good point!

jsantell:
* Changed to an event fired for devtools in `src/Three.js` -- if this pattern works well, I can imagine a util (src/DevTools.js ?) that fires these events for consistency, and to be filtered out if needed.
  * This will fire before three is available. Could be delayed if needed, or add other objects for future use.
* Using event name `register`.
* There isn't any nice one-liner for the local usage + export of `src/constants.js` AFAIK :disappointed:

mrdoob:
Thanks!

PR #17349, opened 2019-08-26
Examples SkyShader: Fix parameter order

weiserhei:
I stumbled upon this one argument thats not in order with the others, this has no effect on the user.
Also there is no method `setLens` in THREE.Camera, so this comment should be removed.

### DIFF
--- a/examples/webgl_shaders_sky.html
+++ b/examples/webgl_shaders_sky.html
@@ -50,9 +50,9 @@
 				var effectController = {
 					turbidity: 10,
 					rayleigh: 2,
+					luminance: 1,
 					mieCoefficient: 0.005,
 					mieDirectionalG: 0.8,
-					luminance: 1,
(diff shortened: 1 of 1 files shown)


Mugen87 (dismissed):

examples/webgl_shaders_sky.html
```diff
@@ -50,9 +50,9 @@
 				var effectController = {
 					turbidity: 10,
 					rayleigh: 2,
+					luminance: 1,
```
Mugen87: > I stumbled upon this one argument thats not in order with the others

On what order are you referring to? Now the parameter order is different compared to the UI (see `guiChanged()`).

weiserhei: I refer to the order of the uniforms. I have updated the PR now to how it shoul have been, just moving the uniforms line.

PR #17359, opened 2019-08-27
Shaders: decouple light probe irradiance from other indirect light sources

### LINKED ISSUE #17253: Ambient light brightes metal materials with high roughness
In the current dev version materials with a metalness of 1 get brighter through the ambient light. The rougher they are the stronger the efffect is.

### Testscene
Sphere left side: `metalness: 1, roughness: 0`, 
Sphere right side: `metalness: 1, roughness: 1`
AmbientLight Intensity: `2`

### Expected (r107)
Spheres r107
The ambient light intensity does not change the brightness of metal materials  
[jsfiddle master](https://jsfiddle.net/JohannesDeml/536szath/)

### Actual (dev)
Spheres dev
The ambient light intensity changes the brightness of the smooth metal a bit and the brightness of the rough metal a lot.  
[jsfiddle dev](https://jsfiddle.net/JohannesDeml/n3q46fku/)

##### Three.js version

- [x] Dev
- [ ] r107

##### Browser

- [x] All of them

##### OS

- [x] Windows

WestLangley:
Fixes #17253.

As suggested in https://github.com/mrdoob/three.js/issues/17102#issuecomment-518053612 and https://github.com/mrdoob/three.js/issues/17253#issuecomment-522624543.

This also fixes an unreported bug. The multi-scattering, energy-preservation model introduced in #15644 should only have been applied to irradiance from IBLs, but it was applied to all indirect diffuse light sources.

/ping @jsantell 
/ping @elalish  
/ping @sunag 
/ping @bhouston

### DIFF
--- a/src/renderers/shaders/ShaderChunk/lights_fragment_end.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/lights_fragment_end.glsl.js
@@ -7,7 +7,7 @@ export default /* glsl */`
 
 #if defined( RE_IndirectSpecular )
 
-	RE_IndirectSpecular( radiance, irradiance, clearcoatRadiance, geometry, material, reflectedLight );
+	RE_IndirectSpecular( radiance, iblIrradiance, clearcoatRadiance, geometry, material, reflectedLight );
 
 #endif
 `;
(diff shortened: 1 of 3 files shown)


sunag:
@WestLangley Is not necessary create an `iblRadiance` property too?

WestLangley:
You could, but this was the minimal change. `radiance` is only used in one place.

The entire shader probably needs to be rewritten at some point.

PR #17369, opened 2019-08-28
Add support for raycasting meshes with triangle strip drawMode

InfiniteLee:


Mugen87:
> If we want to fully support all draw modes, I'm thinking the draw mode should be abstracted away somehow with getters.

Indeed, that would be nice. `Mesh.raycast()` is already so complex...

Mugen87:
`Mesh.drawMode` has been removed. Please check the migration guide for `R112` to see the recommended migration task.

https://github.com/mrdoob/three.js/wiki/Migration-Guide#r111--r112

Issue #17299, opened 2019-08-21
clone issue: SpotLight.target which is placed in scene graph is cloned in two roles


EliasHasle:
##### Description of the problem

Case: Make many similar lamps with geometry and attached SpotLight.

````
let scene = new THREE.Scene();
let lampMesh = new THREE.Mesh(fooPars);
let spotLight = new THREE.SpotLight(barPars);
spotLight.target.position.set(1,2,3);
lampMesh.add(spotLight, spotLight.target);

for (let i = 0; i < numLights; i++) {
    let lightClone = lampMesh.clone();
    lightClone.position.set(0,0,i);
    scene.add(lightClone);
}
````

The cloning will copy the target independently in the roles of `lampMesh.children[1]` and of `lampMesh.target`. This will cause the lamps to interpret the target position as a global position, rather than a local one, and moving the targets in the scene will have no effect on the lights, because their actual `target` properties are independent clones with no `parent`. And the error is/was really hard to locate.

Proposed solution: I am not sure. The only one I can think of is adding an optional parameter to _all_ clone methods, that contains a map (literal object) from original ID/UUID to already cloned objects, such that whenever an already cloned object is encountered, the existing clone will be reused in the new role. The map is passed down the recursion, and must also be passed to the `copy` methods. This seems like a drastic solution which will require a substantial amount of work, but it is also a general solution which will solve similar problems that _may_ exist elsewhere. I am not sure it is without unintended consequences, though.

User solution with current implementation: Don't use `clone` without knowing how it will behave. For the case above, consider either adding targets to scene after cloning, or building the copies without using `clone`, e.g. by using instances of a class which has static geometry and material.

Please also include a live example if possible (**maybe later**). You can start from these templates:

* [jsfiddle](https://jsfiddle.net/3foLr7sn/) (latest release branch)
* [jsfiddle](https://jsfiddle.net/qgu17w5o/) (dev branch)
* [codepen](https://codepen.io/anon/pen/aEBKxR) (latest release branch)
* [codepen](https://codepen.io/anon/pen/BJWzaN) (dev branch)

Mugen87:
Many issues with `clone()`/`copy()` methods were exposed in the past like #16076 or #16224. This is obviously another one.

[...]

PR #17370, opened 2019-08-28
Clone Proposal: Add ID cache to Object3D, DirectionalLight copy and clone functions

gkjohnson:
Related to https://github.com/mrdoob/three.js/issues/17299#issuecomment-525441729

This adds an id cache to the Object3D `copy` and `clone` functions to fix the clone issues described in the above issue and updates `Light.copy` and `DirectionalLight.copy` to take advantage of it.

The short of the problem is that if objects have separate references to objects that live in the scene (such as `DirectionalLight.target`) then clone a subtree with the objects results in losing the relationship between those objects after the clone / copy (as happens with the light and target). Here's some code demonstrating the problem:

```js
const obj = new THREE.Object3D();
const directionalLight = new THREE.DirectionalLight();
obj.add( directionalLight, directionalLight.target );

const clone = obj.clone();

const objEqual = obj.children[ 0 ].target === obj.children[ 1 ];
const cloneEqual = clone.children[ 0 ].target === clone.children[ 1 ];

console.log( objEqual, cloneEqual );

// in dev: "true false" is logged
// this PR: "true true" is logged
```

The proposed pattern enables the following behavior with DirectionalLight, which should extend to other objects:
- If `target` is not parented to anything then the directional light is cloned with a new `target` object also unparented.
- If `target` is parented within the subtree that is being copied / cloned then `target` is cloned and is parented at the same position in the hierarchy in the new subtree.
- If `target` is parented _outside_ the subtree that is being copied / cloned then `target` is cloned but is left unparented.

### DIFF
--- a/src/lights/DirectionalLight.js
+++ b/src/lights/DirectionalLight.js
@@ -36,7 +36,7 @@ DirectionalLight.prototype = Object.assign( Object.create( Light.prototype ), {
 
 		if ( ( source.target.uuid in cache ) === false ) {
 
-			cache[ source.target.uuid ] = source.target.clone();
+			cache[ source.target.uuid ] = source.target.clone( recursive, cache );
 
 		}
 		this.target = cache[ source.target.uuid ];
--- a/src/lights/DirectionalLight.js
+++ b/src/lights/DirectionalLight.js
@@ -28,11 +28,18 @@ DirectionalLight.prototype = Object.assign( Object.create( Light.prototype ), {
 
 	isDirectionalLight: true,
 
-	copy: function ( source ) {
+	copy: function ( source, recursive, cache ) {
 
-		Light.prototype.copy.call( this, source );
+		if ( cache === undefined ) cache = {};
 
-		this.target = source.target.clone();
+		Light.prototype.copy.call( this, source, recursive, cache );
+
+		if ( ( source.target.uuid in cache ) === false ) {
+
+			cache[ source.target.uuid ] = source.target.clone();
+
+		}
+		this.target = cache[ source.target.uuid ];
 
 		this.shadow = source.shadow.clone();
 
--- a/src/lights/DirectionalLight.js
+++ b/src/lights/DirectionalLight.js
@@ -36,7 +36,7 @@ DirectionalLight.prototype = Object.assign( Object.create( Light.prototype ), {
 
 		if ( ( source.target.uuid in cache ) === false ) {
 
-			cache[ source.target.uuid ] = source.target.clone( recursive, cache );
+			source.target.clone( recursive, cache );
 
 		}
 		this.target = cache[ source.target.uuid ];
--- a/src/lights/Light.js
+++ b/src/lights/Light.js
@@ -27,8 +27,6 @@ Light.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 	copy: function ( source, recursive, cache ) {
 
-		if ( cache === undefined ) cache = {};
-
 		Object3D.prototype.copy.call( this, source, recursive, cache );
 
 		this.color.copy( source.color );
--- a/src/lights/Light.js
+++ b/src/lights/Light.js
@@ -25,9 +25,11 @@ Light.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 	isLight: true,
 
-	copy: function ( source ) {
+	copy: function ( source, recursive, cache ) {
 
-		Object3D.prototype.copy.call( this, source );
+		if ( cache === undefined ) cache = {};
+
+		Object3D.prototype.copy.call( this, source, recursive, cache );
 
 		this.color.copy( source.color );
 		this.intensity = source.intensity;
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -802,9 +802,9 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	},
 
-	clone: function ( recursive ) {
+	clone: function ( recursive, cache ) {
 
-		return new this.constructor().copy( this, recursive );
+		return new this.constructor().copy( this, recursive, cache );
 
 	},
 
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -808,9 +808,13 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	},
 
-	copy: function ( source, recursive ) {
+	copy: function ( source, recursive, cache ) {
 
 		if ( recursive === undefined ) recursive = true;
+
+		if ( cache === undefined ) cache = {};
+
+		cache[ source.uuid ] = this;
 
 		this.name = source.name;
 
@@ -842,7 +846,7 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 			for ( var i = 0; i < source.children.length; i ++ ) {
 
 				var child = source.children[ i ];
-				this.add( child.clone() );
+				this.add( cache[ child.uuid ] || child.clone( recursive, cache ) );
 
 			}
 


EliasHasle:
I think SpotLight should be fixed in the same PR. And then we have a similar situation with CameraHelper.camera: https://github.com/mrdoob/three.js/issues/17299#issuecomment-523404518

src/lights/DirectionalLight.js
```diff
@@ -28,11 +28,18 @@ DirectionalLight.prototype = Object.assign( Object.create( Light.prototype ), {
...
 
-		Light.prototype.copy.call( this, source );
+		if ( cache === undefined ) cache = {};
 
-		this.target = source.target.clone();
+		Light.prototype.copy.call( this, source, recursive, cache );
+
+		if ( ( source.target.uuid in cache ) === false ) {
+
+			cache[ source.target.uuid ] = source.target.clone( recursive, cache );
```
EliasHasle: Here, the cache will only be updated with the target _after_ cloning the target subtree. This means other references (e.g. other lights pointing at the same target) in the target subtree will not use the cached version. If I understand right.

src/lights/Light.js
```diff
@@ -25,9 +25,11 @@ Light.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 	isLight: true,
 
-	copy: function ( source ) {
+	copy: function ( source, recursive, cache ) {
 
-		Object3D.prototype.copy.call( this, source );
+		if ( cache === undefined ) cache = {};
```
EliasHasle: Why is this line needed? `cache` will be defined in `Object3D.copy`.

EliasHasle:
> Ha hopefully it doesn't feel like I stole your thunder!

Likewise. 😉 No, I am honored. 😃

Looks like my review comments got sorted out well. Have you run any tests with normal and more exotic cases?

gkjohnson:
> Have you run any tests with normal and more exotic cases?

Only the one I mentioned above. Here are the cyclic cases the @donmccurdy mentioned in https://github.com/mrdoob/three.js/issues/17299#issuecomment-525539687, which might count as exotic example cases:

**two lights being eachothers' targets**
```js
const dl1 = new THREE.DirectionalLight();
const dl2 = new THREE.DirectionalLight();

dl1.target = dl2;
dl2.target = dl1;

const clone = dl1.clone();
console.log(clone.target.target === clone);
// in dev: throws stackoverflow exception on "clone"
// this pr: "true" is logged
```

**lights target is its parent**
```js
const dl = new THREE.DirectionalLight();
dl.target.add(dl);

const clone = dl.clone();
console.log(clone.target.children[0] === clone);
// in dev: throws stackoverflow exception on "clone"
// this pr: "true" is logged
```

Both of these cases actually fail entirely in dev because the cyclic nature of the relationships causes an infinite recursion but it works properly in this branch. Regarding "normal" use cases -- calling `new DirectionalLight().clone()` works just fine unless you had something else in mind? I can test any others, too.

EliasHasle:
How about: One target attached to an object, and many lights with different relation to it pointing at it.

gkjohnson:
> One target attached to an object, and many lights with different relation to it pointing at it.

Yeah that will work, too:

```js
const dl1 = new THREE.DirectionalLight();
const dl2 = new THREE.DirectionalLight();
dl2.target = dl1.target;
scene.add(dl1, dl2, dl1.target);

const clone = scene.clone();
console.log(
    clone.children[0].target === clone.children[2],
    clone.children[1].target === clone.children[2],
);

// in dev: "false false" is logged
// this pr: "true true" is logged
```

EliasHasle:
> with different relation to it

I meant like one light is a descendant, one is an ancestor.? Not that I expect it to fail or anything.

But what if the target is an ancestor to the light, but the clone three does not include the target? Will there be an extra copy of the entire target tree, detached from everything else? And will this copy use cache and "steal" objects from the real scene? If so, this is becoming problematic, and one may need to keep track of the subtree set from the start, as I suggested in https://github.com/mrdoob/three.js/issues/17299#issuecomment-524789765, and have another default behavior than having a detached clone. I think intuitively that it is better to let the cloned light refer to the original target tree when the target is outside the clone tree.

gkjohnson:
Can you sketch out a hierarchy that you expect to be a problem along with which node should be cloned to illustrate the issue? I'm not sure if I'm following the description.

EliasHasle:
````
//Hierarchy:
scene: {
	children: [
		obj: {
			children: [
				dirLight: {
					target: scene
				}
			]
		}
	]
}

obj.clone();
````

**Comment by gkjohnson** [comment]
With the PR as it is now cloning `obj` will result in a clone of the complete tree as you've sketched it including the scene which will be the clone obj's parent and the cloned directional light's target.

I don't have any strong opinions on what should happen here mostly because I don't really feel there's a one-size-fits-all solution so I'm okay with whatever solution the project owners feel is best.

The behavior you're describing where a reference is maintained to objects outside the subtree would require something like the following changes to this PR:

```js
Object3D.copy = function(...) {

  if ( cache === undefined ) {

    // prep by saving all the source uuids  in the subtree 
    // we know we'll want to clone ahead of time
    cache = {};
    source.traverse(c => cache[ c.uuid ] = null );    

  }

  // ...

}

DirectionalLight.copy = function(...) {

  // ...

  // If the id is within the cache then it should be cloned otherwise
  // just use the existing object thta's outside the subtree.
  if ( source.target.uuid in cache ) {

    if ( cache[ source.target.uuid ] === null ) {

      source.target.clone( recursive, cache );

    }
    this.target = cache[ source.target.uuid ];

  } else {
    
    this.target = source.target;

  }

  // ...    

}

**Comment by EliasHasle** [comment]
> With the PR as it is now cloning `obj` will result in a clone of the complete tree as you've sketched it including the scene which will be the clone obj's parent and the cloned directional light's target.

Ah, I see. Because the clone root will be initially orphan and added to cache, and when it is encountered during recursive cloning of the target/scene it will be taken care of from the cache by a wonderful cloned parent.

> The behavior you're describing where a reference is maintained to objects outside the subtree would require something like the following changes to this PR:

So elegantly written.

I am thinking in practical applications here. How often does an average user want the whole scene to be cloned when he clones a subtree which contains a light that targets the scene, compared to how often he wants clones with lights that target the _same_ scene? I don't know all practical use cases of cloning, but I have only ever used it to duplicate objects for use _within_ a scene.

Another case:

````
//Hierarchy:
scene: {
	children: [
                movingObj
		dirLight: {
			children: [
				lightMesh
			],
			target: movingObj
		}
	]
}

dirLight.clone();
````

The user's intention is to have multiple lights with attached geometry, following the moving object. But since `movingObj` is outside the clone tree, it will be cloned, and the clone will be an orphan. Unless you use the second version. Right?

**Comment by trusktr** [comment]
The default behavior is very reasonable. It can be described as a general concept for Object3D.clone, that applies to _any object_ that may reference _any other sort of object_ located somewhere else in the tree outside of the tree being cloned.

**Comment by trusktr** [comment]
As far as the implementation, I imagine doing it a little differently, encapsulating it all in Object3D, and not requiring a separate `cache` arg to the public methods.

**Comment by EliasHasle** [comment]
@trusktr Are you referring to the behavior in the original post of this PR, or to the one described in https://github.com/mrdoob/three.js/pull/17370#issuecomment-526226077?

Regarding implementation, do you have a suggestion on how to encapsulate the behavior in `Object3D`, a solution that does not involve passing a cache?

@gkjohnson I will hide away a lot of my noise from the thread to make it more readable.

**Comment by gkjohnson** [comment]
It turns out this is causing issues with GLTF files when you clone them, as well, since the target object is no longer added as a child of the directional light as it [originally when loading a GLTF file](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js#L550). You can see in this fiddle that if you clone the gltf model before adding it to a scene and it has a directional light embedded the displayed result is different.

https://jsfiddle.net/8795teLf/

| | w/o clone() | clone() |
|---|---|---|
| screenshots | image | image |
| computed light direction | 0, 0, 1 | 0, 0, 0 |

This also effects model-viewer, as well, and can cause the lighting of the model to change depending on the model position (cc @elalish):

https://jsfiddle.net/keLcuptz/

When you clone an object it should clone with as many internal references intact, as possible instead of losing all the references. This would be really nice to have fixed in some way. This was a massive pain to track down and understand why there were lighting differences between MV and my project.

**Comment by trusktr** [comment]
> Regarding implementation, do you have a suggestion on how to encapsulate the behavior in `Object3D`, a solution that does not involve passing a cache?

My reply is a little late, so I don't remember exactly what I thought back then, but I think that this code:

```js
		if ( ( source.target.uuid in cache ) === false ) {

			source.target.clone( recursive, cache );

		}
		this.target = cache[ source.target.uuid ];
```

could just be this:

```js
		this.target = source.target.clone( recursive, cache );
```

where `.clone()` can just return a cached object if any, then this type of conditional check does not need to be repeated in subclasses of `Object3D` because it is encapsulated in `clone()`. Is there a downside to that?

PR #17400, opened 2019-08-31
TS: Add missing setFromObject method

### LINKED ISSUE #17374: BoxHelper's TypeScript definition file is missing setFromObject

I'm trying to use the BoxHelper in my TypeScript project and I noticed that the typescript definition file found here: 

`https://github.com/mrdoob/three.js/blob/dev/src/helpers/BoxHelper.d.ts`

is missing the definition for the setFromObject function defined in:

`https://github.com/mrdoob/three.js/blob/master/src/helpers/BoxHelper.js`

##### Three.js version

- [ ] Dev
- [X] r107
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

chkvskyi:
Fixed https://github.com/mrdoob/three.js/issues/17374

### DIFF
--- a/src/helpers/BoxHelper.d.ts
+++ b/src/helpers/BoxHelper.d.ts
@@ -8,4 +8,6 @@ export class BoxHelper extends LineSegments {
 
 	update( object?: Object3D ): void;
 
+	setFromObject( object: Object3D ): this;
+
 }


mrdoob:
Thanks!

PR #17402, opened 2019-09-01
Docs: Fix link to example in TGALoader page.

Mugen87:
see https://github.com/mrdoob/three.js/issues/17399#issuecomment-526868655


mrdoob:
Thanks!

PR #17410, opened 2019-09-02
Examples: Fix TS files.

Mugen87:
see https://github.com/mrdoob/three.js/issues/17409#issuecomment-527213768


mrdoob:
Thanks!

sunag:
@Mugen87 Thanks!

PR #17413, opened 2019-09-03
Fix toggling receiveShadow without changing material

Oletus:
Fix the case where there are two objects with the same material in the scene and one has receiveShadow=true and the other has receiveShadow=false.

This also prevents possible extra shader compiles when the same material is reused with receiveShadow enabled or disabled in the same scene.

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1792,6 +1792,13 @@ function WebGLRenderer( parameters ) {
 
 		}
 
+		if ( refreshMaterial || materialProperties.receiveShadow !== object.receiveShadow ) {
+
+			materialProperties.receiveShadow = object.receiveShadow;
+			p_uniforms.setValue( _gl, 'receiveShadow', object.receiveShadow );
+
+		}
+
 		if ( refreshMaterial ) {
 
 			p_uniforms.setValue( _gl, 'toneMappingExposure', _this.toneMappingExposure );
--- a/examples/webgl_shadowmap_receiveshadow.html
+++ b/examples/webgl_shadowmap_receiveshadow.html
@@ -1,0 +1,88 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - shadow map - per-object receive shadow</title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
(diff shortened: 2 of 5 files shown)


Mugen87:
I can confirm that this PR fixes #11400. The related [fiddle](https://jsfiddle.net/s1nh95y1/) now works with the mentioned workarounds.

I'm not sure we need a new example though. It feels strange to demonstrate something that should have been worked right from the beginning^^.

mrdoob:
Beautiful!

> I'm not sure we need a new example though. It feels strange to demonstrate something that should have been worked right from the beginning^^.

Agreed. I'll delete the example.

mrdoob:
Thanks!

src/renderers/WebGLRenderer.js
```diff
@@ -1792,6 +1792,13 @@ function WebGLRenderer( parameters ) {
 
 		}
 
+		if ( refreshMaterial || materialProperties.receiveShadow !== object.receiveShadow ) {
+
+			materialProperties.receiveShadow = object.receiveShadow;
```
mrdoob: This line acts as a cache, right?

mrdoob:
Hmm... I only see this locally, but seems to me that the cloth is self-shadowing? 🤔

Screen Shot 2019-09-03 at 5 56 51 PM

Compared to `r108`:

Screen Shot 2019-09-03 at 5 57 26 PM

src/renderers/WebGLRenderer.js
Oletus: Yes, precisely.

Oletus:
Ah yes, still needed to fix the shadowmask code that's used by MeshLambertMaterial. Here's a PR with the remaining fix: https://github.com/mrdoob/three.js/pull/17423

ThorstenBux:
Hi everyone, I'm fairly new to three.js. I'm creating an object which should receive a shadow but not cast any. While searching for a solution I found this issue https://github.com/mrdoob/three.js/issues/11400 and then this PR which has been merged to dev and from what I understand should be available inside build directory on the dev branch. However, the issue still persists that an object keeps casting a shadow even if castShadow is set to false and it only stops casting a shadow if receiveShadow is set to false as well. 

The fiddle is here with a link to three.js from the latest build of dev branch
https://jsfiddle.net/Thor_Bux/g1Loqb08/13/

Forum question is here:
https://discourse.threejs.org/t/create-object-that-receives-shadow-but-doesnt-cast-any/9598

(I would have opened a new issue but I thought it might be related to this one)

PR #17428, opened 2019-09-04
WebXR: add "mode" and "sessionInit" options to createButton

klausw:
This should help make WebXR support more flexible and future-proof by providing a way to specify the mode and sessionInit required/optional features directly. The default behavior is unchanged.


klausw:
The "Handle the requested referenceSpaceType as a feature" commit is an optional convenience feature. The default "local-floor" reference space is no longer available automatically, it must be requested as a required (or optional) feature in requestSession. That change automatically adds the referenceSpaceType as a required feature.

klausw:
Closing as per discussion, see simpler version in https://github.com/mrdoob/three.js/pull/17429.

PR #17429, opened 2019-09-04
WebXR: request common reference spaces as features

klausw:
WebXR's requestReferenceSpace only works if the corresponding feature
was requested at session creation time. For simplicity, just ask for
the interesting ones as optional features, but be aware that the
requestReferenceSpace call will fail if it turns out to be unavailable.
('local' is always available for immersive sessions and doesn't need to
be requested separately.)


mrdoob:
Thanks!

PR #17432, opened 2019-09-05
Update Loading-3D-models.html

fabriziocacicia:
To add another troubleshooting step that could be useful. The problem I had in order to make the model visible was that it was hidden in the dark. Once I added and positioned a light source, the model appeared. Maybe someone else could have the same problem and could solve it more quickly than me.


mrdoob:
Thanks!

PR #17443, opened 2019-09-07
Physical Material Sheen parameter serialization and deserialization and exposed in editor

sdinesh86:
Also update the setAnimationLoop typescript def to include null


Mugen87:
Can you please remove the build files from your PR?

src/loaders/MaterialLoader.js
```diff
@@ -61,6 +61,7 @@ MaterialLoader.prototype = Object.assign( Object.create( Loader.prototype ), {
 		if ( json.color !== undefined ) material.color.setHex( json.color );
 		if ( json.roughness !== undefined ) material.roughness = json.roughness;
 		if ( json.metalness !== undefined ) material.metalness = json.metalness;
+		if ( json.sheen !== undefined ) material.sheen = (new Color()).setHex( json.sheen );
```
Mugen87: Minot style issue:
```js
if ( json.sheen !== undefined ) material.sheen = new Color().setHex( json.sheen );
```

sdinesh86: I saw this style beign used at some place so i added it

Mugen87 (approved):

mrdoob:
Thanks!

sdinesh86:
no problem.

I feel the editor to be not getting the attention it needs and I would love to open PRs for it, but i see
there are many PRs related to it not accepted and still pending so i am confused on the roadmap for it.

would love to hear from you guys on the roadmap and also love to contribute if any.

mrdoob:
Porting the code to modules is definitely a priority now.

The pending PRs... I have not been able to go through them yet 😕

sdinesh86:
I have few tasks listed out for myself for now,

1. A rough implementation of AssetBrowser ( only Materials and probably textures ).
2. Updating the UI for the Outliner to be either resizeable or the treeview to be collapsible.

Atleast if i can get an idea on how the implementations are expected i could make them ready for and then these can be reviewed and accepted later.

mrdoob:
> 1. A rough implementation of AssetBrowser ( only Materials and probably textures ).

Yes, I wanted to do that too. In the `PROJECT` section, I was thinking of adding a `UI.Select` with all the geometries in `editor.geometries` and another `UI.Select` with all the materials in `editor.materials`. We could then have a button under each to apply the selected geometry/material to the selected mesh in the scene.

How does that sound?

> 2. Updating the UI for the Outliner to be either resizeable or the treeview to be collapsible.

That would be great. I think there were a couple of PRs trying to do that.

The problem of some of those PRs is that the implementation was a bit complicated, so I didn't have time to give feedback or to think how I would implement it myself.

sdinesh86:
> Yes, I wanted to do that too. In the PROJECT section, I was thinking of adding a UI.Select

I was thinking of a panel at the bottom of the viewport, with explorer style view ( more typical asset browser that you find in every game engine. ),
I think adding it to the project section will 
1. Hide it.
2. Might make it unintutive due to the small space avaliable.

> The problem of some of those PRs is that the implementation was a bit complicated,

I will try to find best of those and simplify it

I will make PR for each one of these and we can take these dicussions their

mrdoob:
The reason I suggest the PROJECT section is because I was seeing it more as resources in the project.

Also, keep in mind the editor runs on mobile too. Adding these in the PROJECT section won't require extra ui design.

Screenshot 2019-09-09 at 23 45 39

sdinesh86:
Also, keep in mind the editor runs on mobile too. Adding these in the PROJECT section won't require extra ui design.

agreed

PR #17457, opened 2019-09-10
WebGLGeometries: Announce stop of THREE.Geometry support.

Mugen87:
Instead of directly stop rendering `THREE.Geometry` (see #15514), it might be more acceptable to announce this change so users can easier prepare for it. 

The warning is logged once for each automatic geometry conversion of `WebGLRenderer`.

### DIFF
--- a/src/renderers/webgl/WebGLGeometries.js
+++ b/src/renderers/webgl/WebGLGeometries.js
@@ -63,6 +63,8 @@ function WebGLGeometries( gl, attributes, info ) {
 
 			if ( geometry._bufferGeometry === undefined ) {
 
+				console.warn( 'THREE.WebGLGeometries: three.js will stop rendering geometries of type "THREE.Geometry" in R111. Please use "THREE.BufferGeometry" instead.' );
+
 				geometry._bufferGeometry = new BufferGeometry().setFromObject( object );
 
 			}


looeee:
I'd say we should give at least 6 months warning. 

Once we get treeshaking working, most of the unused code like this should be taken out of user's bundles, right?

Mugen87:
> Once we get treeshaking working, most of the unused code like this should be taken out of user's bundles, right?

Not sure what you mean. Are you referring to the console.log?

looeee:
No, I mean the actual `Geometry` classes. 

For example `BoxGeometry.js` has two exports:

``` js
export { BoxGeometry, BoxBufferGeometry };
```

When a user does this:

``` js
import { BoxBufferGeometry } from 'three';
```

...they currently get both exports included in their bundle (well actually, they get _all_ the exports from `three.module.js`) . With tree shaking, the `BoxGeometry` class should get identified as dead code and removed along with all the other unused classes.

Mugen87:
Ah, now I see what you mean. Yes, this is the intended outcome!

Mugen87:
Closing for now. I should have created this PR as a "Draft PR" since I primarily wanted some feedback about this approach. As soon as @mrdoob feels comfortable with deprecating `Geometry`, we can add such a message until it the code paths are finally removed.

PR #17458, opened 2019-09-10
ArrowHelper: Make .setLength() more robust.

Mugen87:
see https://discourse.threejs.org/t/cant-get-the-animation-to-work-three-matrix3-getinverse-cant-invert-matrix-determinant-is-0/5249/3?u=mugen87

This change will avoid the following warning if `headLength` is bigger than `length`.

> THREE.Matrix3: .getInverse() can't invert matrix, determinant is 0

### DIFF
--- a/src/helpers/ArrowHelper.js
+++ b/src/helpers/ArrowHelper.js
@@ -97,7 +97,7 @@ ArrowHelper.prototype.setLength = function ( length, headLength, headWidth ) {
 	if ( headLength === undefined ) headLength = 0.2 * length;
 	if ( headWidth === undefined ) headWidth = 0.2 * headLength;
 
-	this.line.scale.set( 1, Math.max( 0, length - headLength ), 1 );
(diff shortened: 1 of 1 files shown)


WestLangley:
```js
this.line.scale.set( 1, Math.max( Number.EPSILON, length - headLength ), 1 );
```

Setting the scale to be `Number.EPSILON` has consequences elsewhere. For example, the normal matrix in this case is:

Screen Shot 2019-09-10 at 1 56 56 PM

It is true that the normal matrix is not used in this case because the material is of the basic type.

However, since humans can only see 3-significant digits -- or so -- I would suggest doing something like this, instead:

```js
this.line.scale.set( 1, Math.max( 0.0001, length - headLength ), 1 ); // ref this post
```

That should not have bad numerical behavior, and the line would not be visible because it is so short.

Mugen87:
I'm fine with that 👍

mrdoob:
Thanks!

PR #17461, opened 2019-09-10
Examples: Rename DragControls example.

Mugen87:
Renaming `webgl_interactive_draggablecubes` to `misc_controls_drag` so it's easier to find the example for `DragControls`.

The example is slightly updated, too. It does not use `TrackballControls` anymore. If you drag a mesh, it is now highlighted.


Mugen87:
http://raw.githack.com/Mugen87/three.js/dev19/examples/misc_controls_drag.html

mrdoob:
Thanks!

Issue #12678, opened 2017-11-16
Outline pass can't handle transparent background


kevinkleine:
##### Description of the problem 

After commit b3cf1c73c279d7d9788d4a9735d17dcb7893b50c the outlinePass renders transparent backgrounds in black.

##### Three.js version

- [x] Dev
- [x] r88

##### Browser

- [x] Chrome latest
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [x] Windows 10
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
This commit was reverted, see #12281

Maybe a subsequent commit introduced the problem. Does it work with `R87`?

yuluyi:
Same problem in r89

DIeHarke:
I can also confirm that this issue still exists in r89.

Mugen87:
@DIeHarke Can you provide a live demo for debugging? 😇 

Or maybe you can tell how to change the example [webgl_postprocessing_outline.html](https://threejs.org/examples/webgl_postprocessing_outline.html) in order to reproduce the issue.

DIeHarke:
@Mugen87 I create a pastebin with an updated example. You can find it here:

[https://pastebin.com/u7reMDFg](url)

Things I changed:
- Created a css color gradient background (line: 9)
- Activated alpha mode of WebGlRenderer (including setting of clearColor, gammaInput and gammaOutput) (line: 150 - 155)
- (Re-)Enabled resetting to empty array, when no object intersection was found (line: 357)

Now when you hover over the scene and you are not intersecting any objects you can see the gradient background. Otherwise if an intersection was found the outline gets drawn and the background changes to black.

Mugen87:
Okay, i've created a fiddle so we can easily test the issue: https://jsfiddle.net/9a3c8o5a/3/

WestLangley:
Are you sure this is a bug? Try the original version of `OutlinePass` and see if transparent background was ever supported. If not, this is a feature request.

Mugen87:
You are right. I've tested with an [older version](https://github.com/mrdoob/three.js/blob/eafc8ad66ded672dcc9fc6457f43ece04d7aff7c/examples/js/postprocessing/OutlinePass.js) and there is the same behavior:

https://jsfiddle.net/9a3c8o5a/4/

DIeHarke:
[...]

PR #17462, opened 2019-09-10
OutlinePass: Preserve alpha in blurring shader

nirmal:


Mugen87:
This PR actually fixed #12678! Great!

PR #17495, opened 2019-09-13
Editor: Initial Material Browser implementation

sdinesh86:
This is a very basic initial implementation of the Material Browser

https://raw.githack.com/sdinesh86/three.js/editor-materialbrowser/editor/index.html

As discussed Material Browser can now be found under the project panel
image

This PR for now only implements creating and applying of Materials.
In order to apply the materials, i chose the drag and drop approach, which i am not sure if its too complicated or intiutive.

If this implementation is ok then i will go ahead and complete the implementation.

Feedbacks will be most appreciated.

### DIFF
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -46,6 +46,38 @@ select {
 		background-color: #444;
 	}
 
+
+/* Panel */
+
+.Panel {
+	color: #888;
+	border-top: 1px solid #222;
+}
+
+/* TabbedPanel */
+
+.TabbedPanel .Tabs {
+	background-color: #1b1b1b;
+	border-top: 1px solid #222;
+}
+
+	.TabbedPanel .Tab {
+		color: #555;
+		border-right: 1px solid #222;
+	}
+
+	.TabbedPanel .Tab.selected {
+		color: #888;
+		background-color: #111;
+	}
+
+/* Listbox */
+
+.Listbox {
+	color: #888;
+	background: #222;
+}
+
 /* UI */
 
 #viewport {
@@ -251,30 +283,6 @@ select {
 		background-color: rgba(21,60,94,1);
 	}
 
-.TabbedPanel .Tabs {
-	background-color: #1b1b1b;
-	border-top: 1px solid #222;
-}
-
-	.TabbedPanel .Tab {
-		color: #555;
-		border-right: 1px solid #222;
-	}
-
-	.TabbedPanel .Tab.selected {
-		color: #888;
-		background-color: #111;
-	}
-
-.Listbox {
-	color: #888;
-	background: #222;
-}
-
-.Panel {
-	color: #888;
-	border-top: 1px solid #222;
-}
 /* */
 
 @media all and ( max-width: 600px ) {
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -56,9 +56,12 @@ select {
 
 /* TabbedPanel */
 
+.TabbedPanel {
+	border-top: 1px solid #222;
+}
+
 .TabbedPanel .Tabs {
 	background-color: #1b1b1b;
-	border-top: 1px solid #222;
 }
 
 	.TabbedPanel .Tab {
@@ -186,7 +189,7 @@ select {
 #sidebar {
 	position: absolute;
 	right: 0;
-	top: 32px;
+	top: 33px;
 	bottom: 0;
 	width: 300px;
 	background-color: #111;
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -200,7 +200,6 @@ select {
 	#sidebar .Panel {
 		color: #888;
 		padding: 10px;
-		border-top: 1px solid #222;
 	}
 
 	#sidebar .Panel.collapsed {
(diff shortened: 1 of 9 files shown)


mrdoob:
> In order to apply the materials, i chose the drag and drop approach, which i am not sure if its too complicated or intiutive.

I would just have an `APPLY` button for now (which applies to the selected object).

mrdoob:
Some CSS added an extra line here:
64891257-0f7c9100-d68f-11e9-96ad-24b646d504e6

mrdoob:
I think the additional line was added by mistake in the previous PR:

sdinesh86:
I had added it on purpose for Panel, 
I was thinking of having some common styling for the components according to its purpose.

So accordingly Panel will hold other UI elements, will have border on top and possibly a padding of some sorts.

For puposes like the above ones ( TabbedPanel ) we can reuse the UI.Div element with `container` class. which essentially is just simple container without any styling.

So i will replace the Panel wrapper in TabbedPanel with simple UI.Div

sdinesh86:
> > In order to apply the materials, i chose the drag and drop approach, which i am not sure if its too complicated or intiutive.
> 
> I would just have an `APPLY` button for now (which applies to the selected object).

I will add the apply button and also keep the drag and drop functionality if we need it just in case,

also a general question, is it ok to have commented code.
for example : Just incase if we dont want to have the drag and drop for now can i just comment it instead of removing the code ?

mrdoob:
> I had added it on purpose for Panel,
I was thinking of having some common styling for the components according to its purpose.

Be aware that pretty much everything is a Panel. So now there is a border at the top of the Editor because the Menubar is a Panel.

mrdoob:
> I will add the apply button and also keep the drag and drop functionality if we need it just in case,

Ok!

> also a general question, is it ok to have commented code.

Sure!

mrdoob:
> So i will replace the Panel wrapper in TabbedPanel with simple UI.Div

Make sure styles remain the same though. Try opening a new tab with https://threejs.org/editor/ and switch tabs with your changes to make sure still looks the same.

So far I've noticed the extra borders and TabbedPanel sections are no longer uppercase.

sdinesh86:
> So far I've noticed the extra borders and TabbedPanel sections are no longer uppercase.

Thanks for pointing it out, I completely missed it.

I am making the changes we will try to make these changes incrimentally and by the end everything should be consistent.

editor/js/Editor.js
```diff
@@ -8,6 +8,10 @@ var Editor = function () {
 	this.DEFAULT_CAMERA.name = 'Camera';
 	this.DEFAULT_CAMERA.position.set( 0, 5, 10 );
 	this.DEFAULT_CAMERA.lookAt( new THREE.Vector3() );
+	this.DEFAULT_MATERIAL = new THREE.MeshBasicMaterial( {
+		color: 0xff00ff
+	} );
+	this.DEFAULT_MATERIAL.uuid = 'AFE5AF61-A88E-493B-90D1-6AB205395C0B';
```
sdinesh86: I have added option to delete material from the material browser and fallback to default editor material.
@mrdoob can you please review the change for the default material and let me know your thoughts on it.

mrdoob:
> I have added option to delete material from the material browser and fallback to default editor material.

That is logic that the library itself does not have; the library doesn't reuse a single material, it creates a random colored material if the user doesn't pass one. 

I'm trying to make the editor reflect the library's logic so it serves to teach users how it works.

So, for now, I vote for not adding the `DEFAULT_MATERIAL` logic.

mrdoob:
By the way, there is still a border at the top of the editor.

Before:
Screenshot 2019-09-15 at 12 00 26

After:
Screenshot 2019-09-15 at 12 00 32

editor/css/dark.css
```diff
@@ -154,7 +189,7 @@ select {
 #sidebar {
 	position: absolute;
 	right: 0;
-	top: 32px;
```
mrdoob: `32px` was fine. The problem is the new border that has been added to Panel, which is affecting `menubar`.

mrdoob:
When doing PRs, it's better to keep them focused to the a feature it's adding.
If you change other things around it, it may be that one of these changes is not desired and make the PR not ready to merge.

mrdoob:
Looking good overall though!

sdinesh86:
> That is logic that the library itself does not have; the library doesn't reuse a single material, it creates a random colored material if the user doesn't pass one.
> 
> I'm trying to make the editor reflect the library's logic so it serves to teach users how it works.
> 
> So, for now, I vote for not adding the `DEFAULT_MATERIAL` logic.

But then dont you think its waste of resources if new materials are created everytime a mesh is added, I myself got irritated a few times while implementing the feature when its created

sdinesh86:
Sorry mistakenly closed the pull request

sdinesh86:
> When doing PRs, it's better to keep them focused to the a feature it's adding.
> 
> If you change other things around it, it may be that one of these changes is not desired and make > the PR not ready to merge.

I will close this PR for now and open a new one with Material Browser specific changes and then make another PR for the styling related changes

PR #17479, opened 2019-09-12
Docs: Add EditorControls page.

Mugen87:
see #17382

PR #17498, opened 2019-09-14
Examples: Move EditorControls to editor/js.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17479#issuecomment-531443156


mrdoob:
Thanks!

PR #17509, opened 2019-09-16

Mugen87:
see #16967

This PR ensures a name property is always present (with a proper initial value).


mrdoob:
Thanks!

PR #17528, opened 2019-09-18
SpotLight: rectangular spot light shape

nipmarsh:
This PR add the possibility to change the spot light shape from Circle to Rectangular


Mugen87:
Um, I don't think it's a good idea to support different shapes in a hard coded way. #13057 seems to be more promising.

nipmarsh:
> Um, I don't think it's a good idea to support different shapes in a hard coded way. #13057 seems to be more promising.

I saw it but it is completly inactive for a while now... Maybe i can create a new type of light call RectangularSpotLight?

Mugen87:
What if then someone wants a TriangularSpotLight^^? It seems this approach does not scale very well...

nipmarsh:
> What if then someone wants a TriangularSpotLight^^? It seems this approach does not scale very well...

I've just implemented what was suggested by @mrdoob in this comment https://github.com/mrdoob/three.js/pull/13057#issuecomment-359005867

Regarding this comment, this PR seems to be a good idea

Mugen87:
I think your approach is too limited. Using a texture in the sense of [Cookies](https://docs.unity3d.com/Manual/Cookies.html) is much more flexibel, IMO. Anyway, this needs to be discussed in #13057.

WestLangley:
I think that cookies are the best way to model the "barn doors" on studio lights. I would prefer to go the cookie route.

Mugen87:
@nipmarsh Thanks for your work on this but it's better to focus on #14621 instead.

PR #17532, opened 2019-09-18
Docs: Clarify LoadingManager page.

Mugen87:
see https://github.com/mrdoob/three.js/issues/5141#issuecomment-532834837


mrdoob:
Thanks!

PR #17533, opened 2019-09-19
Linting: Lint Example Shaders PR 1

gkjohnson:
Related to #17225 

This PR fixes string indenting in a subset of shaders in the `examples/*/shaders` folder (see https://github.com/mrdoob/three.js/pull/17225#issuecomment-532184616). I'm doing them bit by bit because its not just a straight forward find and replace. Once the string indentation is fixed the remaining lint errors should be simple to address with eslint --fix.

I'll submit follow up PRs to fix the rest of the /shaders files. My plan is to fix the linting errors in the example files, [mark the examples/js/ files as deprecated with a warning](https://github.com/mrdoob/three.js/pull/16920#issuecomment-532746836), and then lastly [enable linting on all example files](https://github.com/mrdoob/three.js/pull/17225).


Mugen87 (approved):

mrdoob:
Thanks!

PR #17546, opened 2019-09-20
BasisTextureLoader: ASTC and PVRTC alpha support update

shrekshao:
Update Basis Texture Loader with 
* ASTC support
* PVRTC RGBA support
* The refactored wasm export need no workaround any more.

**Question**: 
* Are the jsm files hand crafted, or is there any kind of build script?
* Is Three.js using Tabs?

#16524


examples/js/loaders/BasisTextureLoader.js
```diff
@@ -388,28 +406,22 @@ THREE.BasisTextureLoader.BasisWorker = function () {
 
 	function init( wasmBinary ) {
 
+		var m;
```
donmccurdy: Could this be renamed to something more descriptive, like `BasisModule`?

examples/js/loaders/BasisTextureLoader.js
```diff
@@ -22,6 +20,7 @@ THREE.BasisTextureLoader = function ( manager ) {
 
 	THREE.Loader.call( this, manager );
 
+	this.useAlpha = true;
```
donmccurdy: Do we need this option? We can detect that the file has alpha and choose the right format that way, right?

If it's necessary, the option would need to be part of the task configuration rather than the loader configuration. A single loader should be able to transcode images with and without alpha in parallel, without downloading another copy of the transcoder. For example:

```js
const map = loader.loadRGB( 'foo.basis', { ...options } );
const alphaMap = loader.loadAlpha( 'foo.basis', { ...options } );
```

^That's likely to be a hassle, as the texture can't be transferred to two workers at once without copying, so I'm hoping we can avoid it for now.

donmccurdy:
Awesome, thanks for the quick update and simplifying the WASM export!

>  Are the jsm files hand crafted, or is there any kind of build script?

There's a script, yes: `node utils/modularize.js`

> Is Three.js using Tabs?

Tabs, correct. This looks OK to me except line 167.

donmccurdy:
Do we know why the WASM transcoder has doubled in size? I'm not sure whether that's expected with the additional supported formats or not.

examples/js/loaders/BasisTextureLoader.js
shrekshao: My understanding is the use case should be able to choose use alpha channel or not regardless of existence of alpha channel in basis texture. As some compressed format (PVRTC) with alpha has fewer bits for RGB channels. It's possible to use a compressed format without alpha channel when transparency is not needed. What's your thought?

shrekshao:
> Do we know why the WASM transcoder has doubled in size? I'm not sure whether that's expected with the additional supported formats or not.

According to Rich's commit message: This went from ~200kb to 900kb due to a compiler macro issue. Recompiling on OSX reduce size to ~400kb.

Seems @richgel999 is aware of this change. I assume it should be good? It'd be better if we could have Rich's confirmation.

examples/js/loaders/BasisTextureLoader.js
shrekshao: Maybe I will use auto-detected alpha for now. If there's any issue brought up by devs we could then try the refactor

richgel999:
The transcoder supports more texture formats, so we added new lookup tables. (The lookup tables are the bulk of the problem, not code.) The older transcoders weren't changed, so it must be the new stuff.

With everything enabled the transcoder compiles to around 900K of WebAssembly. We did have a version checked in that enabled it all including BC7 (by accident), but I fixed this by disabling BC7 completely. 

I recommend when compiling the transcoder for Web use that you set these macros to 0:

BASISD_SUPPORT_BC7=0
BASISD_SUPPORT_ATC=0 
BASISD_SUPPORT_ASTC_HIGHER_OPAQUE_QUALITY=0

The version that's checked in (under the "webgl" directory) only disables BC7. I'll rebuild the transcoder here and see how large it is with these three macros.

richgel999:
I have just checked in a new CMakeLists.txt file that disables ATC and higher quality ASTC (in addition to the previously disabled BC7). Also, I rebuilt the transcoder ans it's now 358K. If this is too large then I can work on optimizing its size further.

https://github.com/BinomialLLC/basis_universal/commit/b57a362f6520b40ef7c36d372c9a912d3c02e88a

richgel999:
The new ASTC tables are less than 64KB in size total (or 128K with BASISD_SUPPORT_ASTC_HIGHER_OPAQUE_QUALITY=1). So I'm guessing it's a combination of the new code for ASTC and these new ASTC tables. I'll have to dig in and see what's going on.

richgel999:
It's possible to compress all the lookup tables so they are 30-40% smaller without losing any quality, which should noticeably reduce the overall size of the compiled transcoder. How significant would such savings me to your project?

donmccurdy:
> How significant would such savings be to your project?

What percent of the transcoder size do the lookup tables represent?

For comparison, the transcoder in this PR gzips to 172 kb, whereas three.js itself gzips to 142 kb. So I'm sure we'll have a few users complain the transcoder is larger than the library, and other users very happy that Basis saves them on net filesize when textures are factored in.

I'm not familiar enough with WASM toolchains to know whether this is a reasonable idea, but I could imagine users wanting to pick and choose lookup tables, and sideload them as separate binary files. For example:

```js
Promise.all( [

  fetch( 'basis_transcoder.wasm' ).then( r => r.arrayBuffer() ),
  fetch( 'lookup_table_1.bin' ).then( r => r.arrayBuffer() ),
  fetch( 'lookup_table_2.bin' ).then( r => r.arrayBuffer() )

] ).then( ( [ transcoder, lt1, lt2 ] ) => {

  transcoder.setLookupTables( [ lt1, lt2 ] );
  // ...

} );
```

^That might also be a totally unreasonable thing to support, and we should just encourage users to compile the transcoder themselves for production use. 😇

donmccurdy:
But for now, I'm happy to approve this PR with whatever settings you think are appropriate for most use cases on the web. We can figure out how to help users optimize their applications later on.

richgel999:
> What percent of the transcoder size do the lookup tables represent?

Around half of the total size are the lookup tables. DXT1's/BC1's tables are the largest (because it uses both 5 and 6 bit endpoints). 

I'm going to remove std::vector usage, which should help. We can also bitpack the tables more tightly, which I'll be looking into next week.

mrdoob:
@donmccurdy Should this be merged in the meantime?

examples/jsm/loaders/BasisTextureLoader.d.ts
```diff
@@ -8,12 +8,15 @@ import {
 export class BasisTextureLoader extends Loader {
 
 	constructor( manager?: LoadingManager );
+	useAlpha: boolean;
```
donmccurdy: ^This property has been removed.

examples/jsm/loaders/BasisTextureLoader.d.ts
```diff
@@ -8,12 +8,15 @@ import {
...
 	transcoderBinary: ArrayBuffer | null;
 	transcoderPath: string;
 	transcoderPending: Promise<void> | null;
 
 	workerConfig: {
 		format: number;
+		useAlpha: boolean,
```
donmccurdy: ^This property has been removed.

examples/js/loaders/BasisTextureLoader.js
```diff
@@ -427,6 +433,16 @@ THREE.BasisTextureLoader.BasisWorker = function () {
...
 
+		if ( !hasAlpha ) {
+			switch(config.format) {
+				case 9: // Hardcoded: THREE.BasisTextureLoader.BASIS_FORMAT.cTFPVRTC1_4_RGBA
+					config.format = 8; // Hardcoded: THREE.BasisTextureLoader.BASIS_FORMAT.cTFPVRTC1_4_RGB;
+					break;
+				default:
+					break;
+			}
+		}
```
donmccurdy: Could you try running:

```
npm install --global eslint @typescript-eslint/eslint-plugin @typescript-eslint/parser eslint-config-mdcs eslint-plugin-html
eslint --fix examples/js/loaders/BasisTextureLoader.js
```

That should fix the styleguide issues on the file. If that doesn't work, see https://github.com/mrdoob/three.js/wiki/Mr.doob%27s-Code-Style%E2%84%A2. We typically have empty lines after opening brackets and before closing brackets, and spaces after opening parens and before closing parens.

```js
if ( ! hasAlpha ) {

  switch ( config.format ) {

    // ...
```

donmccurdy (commented):
Couple small comments to resolve, but I don't think we need to wait for a smaller transcoder before merging this change.

shrekshao:
Style and .d.ts fixes. Thanks for the review!

donmccurdy (approved):

mrdoob:
Thanks!

nh2:
@shrekshao ^

shrekshao:
Can you try with a local change to see if that fixes the issue? (I don't have a handy ios device to test) @nh2  (Sorry for the late reply)

examples/js/loaders/BasisTextureLoader.js
```diff
@@ -441,12 +454,6 @@ THREE.BasisTextureLoader.BasisWorker = function () {
 
 		}
 
-		if ( basisFile.getHasAlpha() ) {
-
-			console.warn( 'THREE.BasisTextureLoader: Alpha not yet implemented.' );
-
-		}
```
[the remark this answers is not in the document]
Mugen87: Isn't it there?

https://github.com/mrdoob/three.js/pull/17546/files#diff-d40f7b109a16ad9c476442a1e99bc31bL455

nh2:
@shrekshao For clarity, it doesn't fix the issue I reported with Basis->PVRTC warping -- see https://github.com/BinomialLLC/basis_universal/issues/78#issuecomment-536163508:

> I notice the problem on iOS no matter whether I set `pvrtcWrapAddressing` to `0` or `1`.

What I reported above is an unrelated issue I discovered when searching for a solution to my issue. I suspect that the likely-incorrect change to the parameter passing I pointed out above will result in alpha not working when expected, but I'm not using any textures with alpha currently so I don't observe it.

I suspect the change

```diff
      var status = basisFile.transcodeImage(
        dst,
        0,
        mip,
        config.format,
-       config.etcSupported ? 0 : ( config.dxtSupported ? 1 : 0 ),
+       config.useAlpha,
        0
      );
```

should have been this instead:

```diff
      var status = basisFile.transcodeImage(
        dst,
        0,
        mip,
        config.format,
        config.etcSupported ? 0 : ( config.dxtSupported ? 1 : 0 ),
-       0
+       config.useAlpha
      );
```

> I don't have a handy ios device to test

I didn't manage to make anything transparent with a quick playing-around with the values on iOS, but according to

```javascript
    } else if ( config.dxtSupported ) {

			config.format = this.useAlpha ? BasisTextureLoader.BASIS_FORMAT.cTFBC3 : BasisTextureLoader.BASIS_FORMAT.cTFBC1;
```

in the file it should also be possible to try this with a DXT supporting device, is that right?

examples/js/loaders/BasisTextureLoader.js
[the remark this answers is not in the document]
nh2: Ah, you're right. It's in a different commit in the PR, that confused me.

shrekshao:
Oh I see what you mean. It's a different issue from the PVRTC edge bleeding one. (Thanks for the clarification)
Based on the transcoder wrapper code:
https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp#L261
```
self.transcodeImage(dst, imageIndex, levelIndex, format, unused, getAlphaForOpaqueFormats);
```

I think it would be:
```
var status = basisFile.transcodeImage(
				dst,
				0,
				mip,
				config.format,
				0,
				hasAlpha
			);
```

I will make a patch for this later.

bleighb:
This (or at least the jump to r109) appears to have fixed not being able to load a mipmapped basis texture using PVRTC due to this error:

`WebGL: INVALID_VALUE: compressedTexImage2D: length of ArrayBufferView is not correct for dimensions`

🙏🙏🙏  THANK YOU 🙏🙏🙏

jespertheend:
For some reason all my textures with alpha are rendering only the alpha on iOS.
This is what my textures look like on desktop:
image
And on iOS:
image

Do I need to do something special to make alpha textures work on iOS or should it work out of the box?
I'm loading basis textures like this
```js
let basisLoader = new BasisTextureLoader();
basisLoader.setTranscoderPath("libs/basis/");
basisLoader.load(url, compressedTexture => {
    myObject.material.map = compressedTexture;
});
```

I believe I need to get both an rgb as well as an alpha texture separately. But I'm only getting the alpha here. How do I tell the BasisTextureLoader that the image has an alpha channel and how do I get the rgb and alpha separately?

arpu:
what you can try is -seperate_rg_to_color_alpha with the basisu commandline arg

jespertheend:
Hmm well that certainly did something. It seems like it is showing one of the rgb channels correctly, but there is still no alpha. I think I'll just separate the rgb and alpha into two separate images in our asset pipeline. We also have fallback jpg support in case a browser doesn't support basis textures. Though I haven't seen any browser that doesn't. The upside of this approach is that the jpg images will support alpha as well now.

richgel999:
It's a long thread, and I can't tell - what texture format are you using on iOS?

donmccurdy:
@jespertheend could you file an issue?

@richgel999 this should be `cTFPVRTC1_4_RGBA`, although if @jespertheend has a demo I wouldn't mind double-checking that.

> We also have fallback jpg support in case a browser doesn't support basis textures. Though I haven't seen any browser that doesn't.

The browser doesn't have to — The Basis library transcodes to whatever GPU format the browser understands. Worst case, it can transcode to uncompressed RGBA data that all browsers support. In theory you shouldn't need the JPG fallback, but there may be a device-specific (or target format-specific) bug that you're running into here, which we'd need to investigate.

jespertheend:
Alright I'll try to get a demo running.

jespertheend:
Hmm I think I might be messing up somewhere because if I replace the texture of [the basis example](https://threejs.org/examples/?q=basis#webgl_loader_texture_basis) with one of my own it works just fine on iOS. I'll let you know once I figure out what's going on.

donmccurdy:
@jespertheend you may also need to call `basisLoader.detectSupport( renderer )` before loading, so the transcoder knows which formats it can use.

jespertheend:
@donmccurdy that's what I'm doing. I'm trying to figure out the difference between the example and my implementation, but they seem pretty similar.

jespertheend:
Alright I have some more info, it seems like it's only happening when requesting a lot of basis textures at the same time. Sometimes the alpha seems to load first, and then the rgb loads shortly afterwards. But when you request a lot of basis textures at once, it sometimes stays stuck at the alpha texture.

I'm trying to replicate the issue in the example, but when I call `BasisTextureLoader.load()` more than once I get an error `Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Worker': ArrayBuffer at index 0 is already detached.` Are you not supposed to reuse the same loader multiple times.

Anyway, I fixed my issue by separating all rgb and alpha images for now. This works nicely for us so I'll leave that as it is. But if anyone needs this to be fixed, I'd be glad to help out.

donmccurdy:
> Failed to execute 'postMessage' on 'Worker': ArrayBuffer at index 0 is already detached.

This could occur if the same `.basis` texture is loaded more than once – the fix used for DRACOLoader (in https://github.com/mrdoob/three.js/pull/18136) is also needed in BasisTextureLoader, but not yet implemented. Basically the texture data is transferred to a worker thread, invalidating it on the main thread, and if you try to transcode that texture again it will fail.

If you ensure the same `.basis` texture is never requested twice that _should_ fix it, otherwise you'd need to wait until that bug is fixed.

donmccurdy:
> Are you not supposed to reuse the same loader multiple times.

I would recommend _always_ reusing the same BasisTextureLoader, to efficiently share web worker resources and caches.

jespertheend:
> This could occur if the same .basis texture is loaded more than once

Ah well that explains the error on desktop. Though on iOS I was also having this issue when requesting two different basis files, but only with transparent textures. And the `ArrayBuffer at index 0 is already detached` error was not visible in the console.

> I would recommend always reusing the same BasisTextureLoader

Alright nice, that's what I expected.

donmccurdy:
> Though on iOS I was also having this issue when requesting two different basis files, but only with transparent textures. 

Hm ok, I don't know what's going on there. If you could file a (new) issue with an example please do!

ameliemaia:
@jespertheend I recently ran into a similar issue with alpha textures on IOS. I'm loading 60+ alpha textures along with other rgb ones. I noticed if one of the alpha textures fails to load all the rest of the alpha textures only show the alpha channel. This issue would only happen on iOS 13 (safari and chrome) though. IOS 14 seems to work fine even with some of the alpha textures failing to load.

jespertheend:
@ameliemaia hmm I think safari might have added support for alpha and rgb in the same compressed texture? Not sure. Either way, I’m glad to hear it has fixed itself.
I’m not sure if the issue you are describing is similar to mine though. From what I can remember I didn’t actually see any of my textures failing to load in the network panel.

PR #17547, opened 2019-09-20
orbit controls expected methods

sergsar:
typings has this but no public methods in sources


Mugen87:
The type declaration files are actually wrong. It's not supposed that these methods are public.

sergsar:
> The type declaration files are actually wrong. It's not supposed that these methods are public.

Why? So we have no way for changing this camera distance. It's ok to support this methods.

Mugen87:
We had this discussion already here: #9005

PR #17556, opened 2019-09-22
Editor : Basic Material browser implementation

sdinesh86:
(...) #17495 

Cleaned up the code and now making specific PR for material browser

### DIFF
--- a/editor/js/Editor.js
+++ b/editor/js/Editor.js
@@ -57,7 +57,9 @@ var Editor = function () {
 		helperAdded: new Signal(),
 		helperRemoved: new Signal(),
 
+		materialAdded: new Signal(),
 		materialChanged: new Signal(),
+		materialRemoved: new Signal(),
 
 		scriptAdded: new Signal(),
 		scriptChanged: new Signal(),
@@ -234,6 +236,32 @@ Editor.prototype = {
 	addMaterial: function ( material ) {
 
 		this.materials[ material.uuid ] = material;
+		this.signals.materialAdded.dispatch();
+
+	},
+
+	removeMaterial: function ( material ) {
+
+		delete this.materials[ material.uuid ];
+		this.signals.materialRemoved.dispatch();
+
+	},
+
+	getMaterialById ( id ) {
+
+		var material;
+		var materials = Object.values( this.materials );
+
+		for( var i = 0; i < materials.length; i ++ ) {
+			
+			if ( materials[i].id === id ) {
+				material = materials[i];
+				break;
+			}
+
+		}
+
+		return material;
 
 	},
 
--- a/editor/css/dark.css
+++ b/editor/css/dark.css
@@ -273,7 +273,6 @@ select {
 
 .Panel {
 	color: #888;
-	border-top: 1px solid #222;
 }
 /* */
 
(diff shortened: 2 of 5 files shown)


editor/js/Editor.js
```diff
@@ -234,6 +236,32 @@ Editor.prototype = {
...
+	},
+
+	getMaterialById ( id ) {
+
+		var material;
+		var materials = Object.values( this.materials );
+
+		for( var i = 0; i < materials.length; i ++ ) {
+			
+			if ( materials[i].id === id ) {
```
Mugen87: There are some minor style issues in your code. I suggest you setup your editor in a way so it automatically respect the project's style guide (defined in [mdcs](https://github.com/mrdoob/three.js/blob/2a8aa96d69c6a2e7daf4689db95f136f9164adb8/package.json#L79)).

You can then automatically fix such issues and produce outputs like:
```js
if ( materials[ i ].id === id ) {

    material = materials[ i ];
    break;

}

mrdoob:
I think this is good enough. I'll clean things after merging 👌

mrdoob:
Thanks!

mrdoob:
@sdinesh86 FYI https://github.com/mrdoob/three.js/commit/86335608ca8b9b88a82ecb4d5c0dea69f862847b https://github.com/mrdoob/three.js/commit/c26e154419c09b2cf532a464f73ff8875c25b1e7 https://github.com/mrdoob/three.js/commit/5adaa9d840abd57f32aa38e03a3c26e4b1e408e0 https://github.com/mrdoob/three.js/commit/a9d50e7f63c1510e5e3e2fe07c04ba9e7ae1b3a8

mrdoob:
@sdinesh86 Also https://github.com/mrdoob/three.js/commit/3b9016bf16a5dc7b7b37229e3b695c6e987e35c6 https://github.com/mrdoob/three.js/commit/294246bf8c63b126374b84c91a1932cfa2874425 https://github.com/mrdoob/three.js/commit/b8903a43a87f0e0d844614bc34b212792a3bf34b

sdinesh86:
@mrdoob THANKS for the cleanup and update !!!

I will keep a note of these things in next PR

editor/js/Editor.js
sdinesh86: Will do it,

Will need some getting used to, to these linters :)

Mugen87:
While testing the ES6 module code, I've encountered something I'm not sure about:

- Go to https://threejs.org/editor/
- Add a Box
- Go to "PROJECT"  tab. You see the instance of `MeshStandardMaterial` in the material section.
- Go back to "SCENE" and change the box's material to `MeshBasicMaterial`
- Go to "PROJECT"  tab. The material browser shows `MeshStandardMaterial` but not the newly added `MeshBasicMaterial`

Is this a bug or am I doing something wrong.

Besides, when using the dark mode CSS, it's really hard to recognize the active list element:

image

The bottom list element is actually active^^.

mrdoob:
> Is this a bug or am I doing something wrong?

It is definitely a bug. We're probably not dispatching a signal when changing the material.

Mugen87:
I've tried to fix it via e66bcfe0030aefb520956f960dac04cf90c897ae in #18201.

PR #17591, opened 2019-09-27
fix double WebGLUniforms upload for ShaderMaterial if uniformsNeedUpdate

amakaseev:
I use mostly ShaderMaterials and have encountered such a problem on mobile devices very where a limited number of texture units, in the sources found that if the shader material is set to uniformsNeedUpdate = true, there is a double update of uniformity which is generally not critical, but since the first upgrade took a certain number of texture units to second update is already missing them.

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1982,9 +1982,7 @@ function WebGLRenderer( parameters ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 
-		}
-
-		if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
+		} else if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 			material.uniformsNeedUpdate = false;
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1982,7 +1982,15 @@ function WebGLRenderer( parameters ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 
-		} else if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
+			if ( material.isShaderMaterial ) {
+
+				material.uniformsNeedUpdate = false;
+
+			}
+
+		}
+
+		if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 			material.uniformsNeedUpdate = false;


Mugen87:
For reference: #15581

I hope we get this merged this time...

Mugen87 (dismissed):

src/renderers/WebGLRenderer.js
```diff
@@ -1982,9 +1982,7 @@ function WebGLRenderer( parameters ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 
-		}
-
-		if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
+		} else if ( material.isShaderMaterial && material.uniformsNeedUpdate === true ) {
```
gkjohnson: Should `uniformsNeedUpdate` be set to false in the above `refreshMaterial` condition block, as well? If `refreshMaterial` _and_ `uniformsNeedUpdate` are true on the same frame then the uniforms will be updated but `uniformsNeedUpdate` will not be set to false meaning the uniforms will be updated unnecessarily the next frame, right?

Mugen87: Um, but it's an `else if` statement...

gkjohnson: Right but consider this situation (or something similar):

**Frame 1**

- ShaderMaterial is newly created, uniforms set, and `uniformsNeedUpdate` set to true.
- `setProgram` is called for the ShaderMaterial.
- `refreshMaterial` is true because it's the first time rendering it, meaning the first block in this condition runs.
- The uniforms are updated but `uniformsNeedUpdate` remains true because the second block is not run even though the uniforms have already been updated.

**Frame 2**

- `setProgram` is called again for the ShaderMaterial.
- `refreshMaterial` is false because nothing else has changed.
- `uniformsNeedUpdate` is true because it was never set to false in frame 1 after updating the uniforms.
- The second `material.isShaderMaterial && material.uniformsNeedUpdate` block is run and needlessly updates the uniforms.

It doesn't update the uniforms twice in one from but it will needlessly update the uniforms in the next frame unless I'm missing something.

Mugen87: Okay, I see now. We can basically do the following (it's actually the code from #15581. Unfortunately, the branch was deleted so I remembered incorrectly the code).
```
if ( material.isShaderMaterial ) material.uniformsNeedUpdate = false; // #15581
```
@amakaseev Can you please undo your change and put the following line at the end of the previous `if` block.

Mugen87 (approved):

src/renderers/WebGLRenderer.js
```diff
@@ -1982,6 +1982,12 @@ function WebGLRenderer( parameters ) {
 
 			WebGLUniforms.upload( _gl, materialProperties.uniformsList, m_uniforms, textures );
 
+			if ( material.isShaderMaterial ) {
+
+				material.uniformsNeedUpdate = false;
```
WestLangley: @mrdoob @Mugen87 https://github.com/mrdoob/three.js/pull/15581#issuecomment-454833117 I think this, at a minimum, needs an explanatory comment.

Mugen87: You mean something like `// #15581`? Or a textual description?

WestLangley: I think `// #15581` is OK.

PR #17355, opened 2019-08-26
Broken PR: RangeFog and DensityFog

EliasHasle:
**Update: #17592 replaces this PR.**

Step 2 of the plan in #17264, adapted to comply with https://github.com/mrdoob/three.js/issues/17420#issuecomment-534592079.

<details><summary>Outdated</summary>Also correct typescript docstring and misspelled name in test file for FogExp2, as well as docs for standard Fog.</details>

Includes the change to distance-based depth, as well as corrections to better support mediump and avoid unnecessary squaring of a square root. Makes #17316 and #17324 obsolete.

Requires new build, update to the library build used in the editor, doc build, examples build.

I added an example named `webgl_geometry_terrain_fog_types`, which is basically `webgl_geometry_terrain_fog` without flight and with a GUI to try different fog setups. Everything seems to work. <details><summary>Outdated link... Will be updated soon</summary>Running version here: https://eliashasle.github.io/testing/fog_types/ (I think `FogExp` looks much more natural than the others.)</details>

I have added additional versions of the new example, that test legacy signatures, `mediump` and `lowp`, respectively. They should probably be removed before merging.

I am currently struggling with git. Hoping to get everything in order soon.

<details><summary>Resolved</summary>Some examples give an apparently unrelated warning message about using `pow` with negative input. I would really appreciate pointers into what goes wrong, as I am not using `pow` in my code or, as far as I can see, outputting anything that can be negative.</details>

PR #17592, opened 2019-09-27
RangeFog and DensityFog

EliasHasle:
Replaces #17355.

`RangeFog(color, near, far)` and `DensityFog(color, density, squared)`, as requested in https://github.com/mrdoob/three.js/issues/17420#issuecomment-534592079

Includes the change to distance-based depth, as well as corrections to better support `mediump` and `lowp`. Makes #17316 and #17324 obsolete.

<s>`DensityFog`: To make sure a change to `squared` causes the necessary material update, a `needsUpdate` is used internally and handled by the renderer. (Does `needsUpdate` also have to be declared in `DensityFog.d.ts`?)

The fog types example is [live on my website](https://eliashasle.github.io/testing/fog_types/webgl_geometry_terrain_fog_types.html).
[Test of legacy signatures](https://eliashasle.github.io/testing/fog_types/webgl_geometry_terrain_fog_types_legacy_test.html)
[Test of mediump](https://eliashasle.github.io/testing/fog_types/webgl_geometry_terrain_fog_types_mediump_test.html)
[Test of lowp](https://eliashasle.github.io/testing/fog_types/webgl_geometry_terrain_fog_types_lowp_test.html)
It looks suspiciously good with `mediump` and `lowp` on my desktop, probably because the precision specifiers are _minimal_ requirements. These ones must be tested on actual `mediump`/`lowp` devices.</s>

The editor works too. It is not perfectly user-friendly to only have an unlabeled checkbox to represent `squared`, though. (Unlabeled controls seems to be a global problem in the editor).


Mugen87:
> But I think it will be ready for r109 after all.

No rush. I don't think merging this in that short amount if time is good. Better to target a later release.

EliasHasle:
@Mugen87 Now that you are apparently here, do you think I should go with ES6 classes right away? I think I can do that in no time. :-)

Mugen87:
I think it's better to stick to the old style for now. I'm not sure how fast we move on with this topic.

src/renderers/WebGLRenderer.d.ts
```diff
@@ -315,7 +315,7 @@ export class WebGLRenderer implements Renderer {
 
 	renderBufferDirect(
 		camera: Camera,
-		fog: Fog,
+		fog: IFog,
```
EliasHasle: I am uncertain about this, as I don't really know TypeScript. It didn't look correct before anyway, since FogExp2 was not included.

src/renderers/WebGLRenderer.d.ts
```diff
@@ -315,7 +315,7 @@ export class WebGLRenderer implements Renderer {
 
 	renderBufferDirect(
 		camera: Camera,
-		fog: Fog,
+		fog: IFog | null,
```
EliasHasle: This type declaration is my best guess after reading a bit about interfaces and nullable types. I don't really know TypeScript. On the other hand, it seemed incomplete before.

The change of putting IFog (the fog interface) in a separate d.ts file felt natural, but I don't know if it is correct.

EliasHasle:
In light of #17662, should I change `fog_vertex.glsl.js` to:

```
export default /* glsl */`
#ifdef USE_FOG

	vFogPosition = mvPosition.xyz;
	// Consider only depth for orthographic camera:
	float p23 = projectionMatrix[2][3];
	vFogPosition += (-1.-p23)*vec(1.,1.,0);

#endif
`;
```

?

I feel a bit bad about using a separate varying for fog, when `vViewPosition` is conditionally defined in so many cases. The branch-free modification above can also be moved into the fragment shader, so that it still applies only to fog.

EliasHasle:
In light of #17767, should I change `fog_vertex` to:

```
#ifdef USE_FOG

	vFogPosition = vec3( ( 1. - isOrthographic ) * mvPosition.xy, mvPosition.z );

#endif
```

to make orthographic camera fall back on depth-based fog, rather than eye-distance-based fog?

Top-view maps are a typical use case for orthographical cameras. I have some trouble imagining fog is desired in that case, but if it were, I would think it most natural to have the same fog depth over the whole map.

Mugen87:
I normally don't do that if files in the core change. In this case, it's best to download your branch and test locally.

EliasHasle:
For some reason, the "Resolve conflicts" button is greyed-out.

Mugen87:
Read the tooltip 😊 

image

src/renderers/WebGLRenderer.d.ts
EliasHasle: OK. @Mugen87 The (too complex) conflicts are with the ts files. I think I am able to fix them, but I feel I am still shooting in the dark with TS (TypeScript, not Tourettes Syndrome, although that would be a funny interpretation). Is there a TS expert around here, or are we all shooting in the dark for the time being? ;-)

Mugen87: I would say try your best and then let the commit review by others.

EliasHasle:
PS (@mrdoob): I referred to the PR as "up-to-date", but that was meant in a relative sense. 😁 (It clearly has a few merge conflicts now, after months without care. Poor little thing.)

Feirell:
Well this could fix things for me, I really hope this gets some traction again and not vanish in the deep dark.

LeviPesin:
@EliasHasle Can you please update this PR?

src/renderers/WebGLRenderer.js
```diff
@@ -1582,7 +1583,7 @@ class WebGLRenderer {
 
 					needsProgramChange = true;
 
-				} else if ( material.fog === true && materialProperties.fog !== fog ) {
+				} else if ( material.fog === true && ( materialProperties.fog !== fog || ( fog && ( materialProperties.fogSquared !== fog.squared ) ) ) ) {
```
EliasHasle, line 1586: This and line 1397 seems to work, but I am not 100% confident that all conceivable edge cases are perfectly covered.

examples/screenshots/webgl_geometry_terrain_fog_types.jpg
EliasHasle, line 1: I made this manually, matching the file type and image dimensions, while upscaling the controls a bit to emphasize them.

I notice that the automatic tests that fail seem to take diffs between screenshots. Are the thumbnail screenshots used for those too?

LeviPesin: Yes. Please run `npm run make-screenshot webgl_geometry_terrain_fog_types`.

(Also please run `npm run lint-fix` to fix linting of the example)

src/loaders/ObjectLoader.js
```diff
@@ -772,13 +772,14 @@ class ObjectLoader extends Loader {
...
+					if ( data.fog.type === 'RangeFog' || data.fog.type === 'Fog' ) {
 
-						object.fog = new Fog( data.fog.color, data.fog.near, data.fog.far );
+						object.fog = new RangeFog( data.fog.color, data.fog.near, data.fog.far );
 
-					} else if ( data.fog.type === 'FogExp2' ) {
+					} else if ( data.fog.type === 'DensityFog' || data.fog.type === 'FogExp2' ) {
 
-						object.fog = new FogExp2( data.fog.color, data.fog.density );
+						const squared = ( data.fog.squared == undefined ) || !! data.fog.squared;
```
LeviPesin, line 781: Did you mean `===` here?

src/Three.Legacy.js
```diff
@@ -227,3 +230,25 @@ export class TubeBufferGeometry extends TubeGeometry { // @deprecated, r144
...
+
+	}
+
+}
+
+export class FogExp2 extends DensityFog {
+
+	constructor( color, near ) {
+
+		console.warn( 'THREE.FogExp2 has been deprecated. Use THREE.DensityFog( color, near, squared ) with squared == true instead.' );
```
LeviPesin, line 249:
```suggestion
		console.warn( 'THREE.FogExp2 has been deprecated. Use THREE.DensityFog( color, near, true ) instead.' );
```

examples/screenshots/webgl_geometry_terrain_fog_types.jpg
LeviPesin: And from failing screenshots it seems like this PR breaks quite a few examples.

src/renderers/WebGLRenderer.js
EliasHasle: Alternatives to checking `squared` every time:
1. Make it truly or practically immutable after construction. I don't know what pattern would be preferred here.
2. Have separate fog classes for FOG_EXP and FOG_EXP2 (`DensityFogPhysical` and `DensityFogSquared` or whatever, still kind of clunky). Independent default densities would be a bonus.

One could also have an abstract parent class. Sadly, the name "Fog" will still be in use, but maybe `SceneFog`? With inheritable `color`, `isSceneFog`, and default shader chunks, and extensible with custom shader chunks, instead of juggling with the `#define`s in the renderer and separate shader chunk files. The proposed three standard fogs could be implemented using this, and it would also be straightforward to, e.g., make a fog with angle-dependent density, combine density with range, sneak in a fog texture, etc.. (If the default chunks match RangeFog, then the name `Fog` can even be reused, and the "abstract" class will be concrete and backwards compatible. In that case, I would suggest using the more "traditional" names `FogExp` and `FogExp2` on the subclasses.)

examples/screenshots/webgl_geometry_terrain_fog_types.jpg
LeviPesin: > Regarding "broken" examples, I suppose some differences in the 0.1% range (and sometimes more) are to be expected when changing to distance-based fog, as this PR does.

Please then also update screenshots for broken examples (with the same command, i.e. `npm run make-screenshot example1 example2 example3`).

> It is also possible to postpone that change and only change the API now.

@mrdoob What do you think?

src/renderers/WebGLRenderer.js
LeviPesin: > Make it truly or practically immutable after construction. I don't know what pattern would be preferred here.

Is there a use case in which it would be updated after construction? We can just say that it shouldn't be changed.

EliasHasle: I update it after construction in the example. That is how I encountered the updating issue. It seems to work, so I guess there is no urgent need to make it _less_ user friendly. The `color` and `density` are uniforms that can be changed after construction, and as a user, I would expect the same flexibility for `squared`.

examples/screenshots/webgl_geometry_terrain_fog_types.jpg
EliasHasle: My understanding is that `npm run make-screenshot example1` will make a new _reference_ screenshot for example1. If that is right, I should first make sure that the output is as expected. 0.1% difference sounds OK anyway, but there were a few in the 2-3 % range that I may have to inspect more closely.

src/loaders/ObjectLoader.js
EliasHasle: The issue here is that I don't know if `undefined` is really stored as `undefined` in JSON, and that both `undefined` and `null` are "falsey" values, whereas DensityFog defaults to `squared=true`, as requested by someone at some time (sorry). I would prefer to have `squared=false` as default, due to realism, and that would certainly facilitate less error-prone saving, loading, and dirtiness checks. I mean, as long as the to-be-deprecated `FogExp2` still results in construction with `squared=true`, are there any compelling reasons to keep that as the default?

examples/screenshots/webgl_geometry_terrain_fog_types.jpg
EliasHasle: > Makes some sense, while considering whether(/when/how) `vViewPosition` should be made near-universal instead of bloating the advanced materials with an equivalent `vFogPosition`.

One weak reason to keep a separate `vFogPosition` could be that one wanted to define it differently for orthographic cameras. Something like:

```
#ifdef USE_FOG

	vFogPosition = mvPosition.xyz;
	// Consider only depth for orthographic camera:
	float p23 = projectionMatrix[2][3];
	vFogPosition += ( - 1. - p23 ) * vec3( 1., 1., 0 );

#endif
```

EliasHasle:
The rationale for using only depth rather than Euclidean distance from a single point for fog with orthographic projection is that every pixel on the screen represents a parallel "eye beam", whereas with perspective projection, the camera position represents the focal point, ideally located at the position of the observer in front of the screen, where the screen is the projection plane (or "window" to the virtual world). (Nevermind that most demos and games inflate the FOV way beyond the typical eye-screen angle, for practical gameplay/immersion purposes.)

EliasHasle:
> Well this could fix things for me, I really hope this gets some traction again and not vanish in the deep dark.

@Feirell Do you mind testing the newest version?

This PR should be ready for review now. @mrdoob @Mugen87 @LeviPesin

Given the parameterisation with an optional `squared`, I would prefer to have it `false` by default, because `FOG_EXP` retains at least a rudimentary physical realism that `FOG_EXP2` sacrifices. (It could also simplify some logic for JSON etc.) But for now, I have left it `true` by default, as requested.

Feirell:
@EliasHasle sorry to disappoint but the usecase I had has long moved on. I switched to a different approach entirely. Depending on if you believe that another tester is useful, I could cobble together a quick scene and test if it works as I would expect it to.

Non the less, thank you for your work and effort!

Mugen87:
Closing see https://github.com/mrdoob/three.js/issues/17420#issuecomment-3462364972. With the new node material, we have a much more flexible approach for implementing fog factors. We already support range, density and exponential height fogs this way. Custom fog factors can be implemented without modifying the core.

https://github.com/mrdoob/three.js/blob/477da6140bffbd2803a9e7e363ca2831f6ea9667/src/nodes/fog/Fog.js#L32-L82

Issue #12938, opened 2017-12-21
Was THREE.ShapeUtils.triangulate removed without warning?


loldrup:
from v88 to v89

A warning could have been raised, and the warning could even have pointed to an appropriate alternative.

By the way, https://threejs.org/docs still shows nothing when one searches for 'triangulation'.

Mugen87:
I actually thought this method was only used interally by the library, sry. In any event, the new polygon triangulation algorithm ist noted in the migration guide.

TBH, i don't think we need an extra page for the triangulation algorithm in the docs. The documentation to ShapeUtils should be sufficient.

loldrup:
The ShapeUtils documentation looks just fine, but how does one know to look for a 'ShapeUtils' class when one wants to do triangulation? Could an invisible tag 'triangulation' be added to that page?
https://threejs.org/docs/#api/extras/ShapeUtils

Thanks for the tip regarding there being a migration guide :). Though it doesn't say that ShapeUtils has been moved to the 'extras' folder.
https://github.com/mrdoob/three.js/wiki/Migration-Guide

⁣

Den 21. dec. 2017 13.54, fra 13.54, Michael Herzog <notifications@github.com> skrev:

looeee:
The ShapeUtils page is at least the first result when you google for "three.js triangulation".

loldrup:
Ah, I use duck Duck Duck Go, and ShapeUtils is not even on the first page of search results (though the page has infinite scroll, so I could look on forever..).


⁣

Den 21. dec. 2017 14.10, fra 14.10, Lewy Blue <notifications@github.com> skrev:

loldrup:
Ahh, so what happened was that 'triangulate' was removed and triangulateShape has been added (or was always there). So all that's missing is:
* The migration guide could mention that 'triangulate' has been removed, deferring to triangulateShape.
* Duck Duck Go should start pointing to ShapeUtils when searching for three.js triangulation (could three.js help the poor search engine with a #triangulation tag in the ShapeUtils doc?)

⁣

Den 21. dec. 2017 14.19, fra 14.19, Michael Herzog <notifications@github.com> skrev:

mrdoob:
> * The migration guide could mention that 'triangulate' has been removed, deferring to triangulateShape.

The migration page is a wiki. Feel free to complete the gaps 😊

Mugen87:
@loldrup I've added a note that `ShapeUtils.triangulate()` has been removed. Thanks for that hint!

[...]

PR #17605, opened 2019-09-28
Fix ShapeUtils typing

flostellbrink:
ShapeUtils don't have a `triangulate` method anymore.
It's not part of the [Documentation](https://threejs.org/docs/#api/en/extras/ShapeUtils) and has been [removed from the codebase](https://github.com/mrdoob/three.js/issues/12938). It also does not exist in [`ShapeUtils.js`](https://github.com/mrdoob/three.js/blob/dev/src/extras/ShapeUtils.js)

The holes parameter in the `triangulateShape` method is essentially an array of contours.
That means it is a `Vec2[][]` and not a `Vec2[]`:
https://github.com/mrdoob/three.js/blob/5feb4ba4955376ed05572a90545015a126df9a3a/src/extras/ShapeUtils.js#L32

This pull request solves both problems.

### DIFF
--- a/src/extras/ShapeUtils.d.ts
+++ b/src/extras/ShapeUtils.d.ts
@@ -5,7 +5,6 @@ interface Vec2 {
 
 export namespace ShapeUtils {
 	export function area( contour: Vec2[] ): number;
-	export function triangulate( contour: Vec2[], indices: boolean ): number[];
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

PR #17620, opened 2019-09-30
Global constant faceDirection instead of gl_FrontFacing

sunag:
Still I did not make any deep-test... but I can finish. Apparently work for normal/bump maps.

Considering that Adreno GPUs not work property `gl_FrontFacing`, otherwise, it can be more simple.

https://github.com/mrdoob/three.js/pull/17598

**This is a WIP**

/ping @WestLangley @mrdoob

### DIFF
--- a/examples/jsm/nodes/misc/BumpMapNode.js
+++ b/examples/jsm/nodes/misc/BumpMapNode.js
@@ -119,7 +119,7 @@ BumpMapNode.prototype.generate = function ( builder, output ) {
 
 			var derivativeHeightCode = derivativeHeight + '( ' + this.value.build( builder, 'sampler2D' ) + ', ' +
 				this.value.uv.build( builder, 'v2' ) + ', ' +
-				this.scale.build( builder, 'f' ) + ' )';
+				this.scale.build( builder, 'f' ) + ' * faceDirection )';
 
 			return builder.format( perturbNormalArb + '( -' + this.position.build( builder, 'v3' ) + ', ' +
 				this.normal.build( builder, 'v3' ) + ', ' +
--- a/examples/jsm/nodes/misc/NormalMapNode.js
+++ b/examples/jsm/nodes/misc/NormalMapNode.js
@@ -48,18 +48,6 @@ NormalMapNode.Nodes = ( function () {
 
 	mapN.xy *= normalScale;
 
-	#ifdef DOUBLE_SIDED
-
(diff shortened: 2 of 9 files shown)


sunag:
> Considering that Adreno GPUs not work property gl_FrontFacing, otherwise, it can be more simple.

This is affirmation is true? 
If yes, I think that this will prevent other bugs related, maybe not published in github.
If no, I think that the code can be more simple.

/ping @WestLangley @mrdoob 

https://github.com/mrdoob/three.js/blob/26df5e2fe581898ea2ca72a8e7e9f1e016f4344b/src/renderers/shaders/ShaderChunk/normalmap_pars_fragment.glsl.js#L41

https://github.com/mrdoob/three.js/issues/15850#issuecomment-517216599

makc:
fwiw, I have seen like 2 other drivers having problems with gl_FrontFacing, and heard of [3rd](https://github.com/Kitware/VTK/commit/221ba52da6313c7944b35b79a7f6077fbe41be7f) one. what I did was
```
#extension GL_OES_standard_derivatives : enable
vec3 fdx = dFdx(v_position.xyz);
vec3 fdy = dFdy(v_position.xyz);
vec3 faceNormal = normalize(cross(fdx,fdy));
vec3 normal = v_normal;
if (dot (normal, faceNormal) < 0.0) {
	normal *= -1.0;
}
```
instead of
```
vec3 normal = v_normal;
if (!gl_FrontFacing) {
    normal *= -1.0;
}
```

WestLangley:
>Still I did not make any deep-test... but I can finish.

Thanks, @sunag. I think this is definitely worth further study.

makc:
hmm, funny thing here. my workaround above should be the same as in https://github.com/mrdoob/three.js/issues/17804#issuecomment-546132110 but that one is causing problems for @WestLangley 🤔

Mugen87:
Notice that #17958 has removed the Adreno workaround.

FWIW, while study the shader code of Sketchfab, I've noticed that they use `gl_FrontFacing` in the shader, too. AFAIK they don't use hacks like `dot( cross( S, T ), N )`, so introducing a global constant seems not necessary. Hence, I vote to close the PR.

sunag:
> ...so introducing a global constant seems not necessary. Hence, I vote to close the PR.

@Mugen87 I think so, this PR is too outdated... Maybe in future simplify the operation `a *= ( float( gl_FrontFacing ) * 2.0 - 1.0 )` for `a *= faceDir` can be an option for optimisation. closing...

PR #17624, opened 2019-10-01
Typedefinition add EquirectangularToCubeGenerator variables

JohannesDeml:
Adding class variables to type definition, especially renderTarget is interesting to access.

### DIFF
--- a/examples/jsm/loaders/EquirectangularToCubeGenerator.d.ts
+++ b/examples/jsm/loaders/EquirectangularToCubeGenerator.d.ts
@@ -23,6 +23,7 @@ export interface EquirectangularToCubeGeneratorOptions {
 export class CubemapGenerator {
 
 	constructor( renderer: WebGLRenderer );
+	renderer: WebGLRenderer;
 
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

kevor:
EquirectangularToCubeGenerator disappeared from examples/jsm/loaders in r110 @JohannesDeml @Mugen87 

According to https://threejs.org/examples/?q=gltf#webgl_loader_gltf line 56 the workflow is different, ok to me, but not clear from the r110 changelog!

Maybe I missed something, if that's the case just ignore this commet :P

Mugen87:
> but not clear from the r110 changelog!

There is actually a note in the change log:

image

The [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide) also describes what do:

> EquirectangularToCubeGenerator has been removed. Please use WebGLRenderTargetCube.fromEquirectangularTexture() instead.

PR #17625, opened 2019-10-01
WebVRManager: Account for both left and right eye parameters when computing render buffer size

jbaicoianu:
Currently, when WebVRManager determines the size of the underlying render buffer, it only looks at the left eye's parameters and assumes the right eye's parameters are the same.  This patch accounts for both eyes when computing buffer size.

This fixes rendering issues encountered when using newer versions of Three.js with Exokit.


Mugen87:
WebVR support has been removed, see #18129

PR #17626, opened 2019-10-01
WebGLPrograms: Add displacementMap to list of texture attributes which cause vUv to be defined.

jbaicoianu:
This pull request fixes one of the concerns @WestLangley raised in #17487 about vUv being undefined if other qualifying textures aren't defined.  This is fixed by checking for the presence of material.displacementMap when deciding whether to ```#define USE_UV```

There's still some concern that this is a change from previous behavior.  Would it help to have a define which lets us toggle between using transformed vs untransformed UVs for the displacementMap?

### DIFF
--- a/src/renderers/webgl/WebGLPrograms.js
+++ b/src/renderers/webgl/WebGLPrograms.js
@@ -176,7 +176,7 @@ function WebGLPrograms( renderer, extensions, capabilities ) {
 
 			vertexTangents: ( material.normalMap && material.vertexTangents ),
 			vertexColors: material.vertexColors,
-			vertexUvs: !! material.map || !! material.bumpMap || !! material.normalMap || !! material.specularMap || !! material.alphaMap || !! material.emissiveMap || !! material.roughnessMap || !! material.metalnessMap || !! material.clearcoatNormalMap,
+			vertexUvs: !! material.map || !! material.bumpMap || !! material.normalMap || !! material.specularMap || !! material.alphaMap || !! material.emissiveMap || !! material.roughnessMap || !! material.metalnessMap || !! material.clearcoatNormalMap || !! material.displacementMap,
 
 			fog: !! fog,
 			useFog: material.fog,


Mugen87 (approved):

WestLangley:
`MeshDepthMaterial` warning with a custom depth material and this change:

THREE.WebGLProgram: gl.getProgramInfoLog() – "WARNING: Output of vertex shader 'vUv' not read by fragment shader"

jbaicoianu:
@WestLangley hmm, ok. If you can provide a link to an example that shows off this problem I'll take a look, it seems your case is different than the one I was running into.

That case might be a bit more tricky, we have to know ahead of time if the fragment shader plans to sample from any of our textures, then either use a varying if they do, or just a local variable in the vertex shader otherwise..  Probably not too difficult to handle with some ifdef gymnastics, so an example would be a great help.

WestLangley:
>If you can provide a link to an example that shows off this problem I'll take a look

In 'webgl_animation_cloth.html', make this change:
```js
object.customDepthMaterial = new THREE.MeshDepthMaterial( {
	depthPacking: THREE.RGBADepthPacking,
	displacementMap: clothTexture,
} );
```

Mugen87:
Merging for now in order to solve the breakage in `webgl_materials_channels`. The warning:

> three.module.js:18065 THREE.WebGLProgram: gl.getProgramInfoLog() WARNING: Output of vertex shader 'vUv' not read by fragment shader

can be fixed with a different PR.

WestLangley:
>The warning can be fixed with a different PR.

@Mugen87 Please fix it.

If it is not fixed, I will have to revert this PR and #17487 before the next release.

Mugen87:
> If it is not fixed, I will have to revert this PR and #17487 before the next release.

@mrdoob What do you think about this? I mean the warning is only logged in a rare case.

WestLangley:
> I mean the warning is only logged in a rare case.

Rare? Based on what?

Mugen87:
Using a custom depth material with an displacement map seems not something usual to me.

Mugen87:
Anyway, #17784 should remove the warning.

Mugen87:
Oh wait, the warning is logged with all materials if you just use a single displacement map. Anyway, the new PR should solve the issue.

jbaicoianu:
Hey guys, sorry for the delay, I haven't had the time I'd like to spend on this.  @Mugen87 thanks for taking on the work of fixing the USE_UVS define, your solution is a lot more comprehensive than what I had in mind.  I actually ended up running into the situation @WestLangley described too, so yeah, while it's not the most common case, I certainly don't want to push breakage on anyone.

I do see one potential oversight in #17784 - I'll comment with more details there

PR #17651, opened 2019-10-03
WebGLPrograms: Add missing ParameterNames

gero3:
This is continuation of #17552. This adds all missing parameters to the parameterNames Array.


Mugen87:
> This is continuation of #17552.

So it's a replacement for #17552, right?

gero3:
> > This is continuation of #17552.
> 
> So it's a replacement for #17552, correct?

I would prefer #17552 over this, but this indeed fixes the same problem.

> In essence, the entries in `parameterNames` should match the object properties of `parameters`, right (except `shaderID`)?
> 
> https://github.com/mrdoob/three.js/blob/a5c9edd6864bf7b0b0be4644c5388271f39f0d5d/src/renderers/webgl/WebGLPrograms.js#L143-L235

I believe so because parameters are used to create different shaders and different shaders are detected by parameterNames

Mugen87:
Um, why not changing:
```js
for ( var i = 0; i < parameterNames.length; i ++ ) {

   array.push( parameters[ parameterNames[ i ] ] );

}
```
to
```js
Array.prototype.push.apply( array, Object.values( parameters ) );
```
In this way, `parameterNames` is not needed. Would it be a problem if `shaderID` is in the array?

gero3:
It doesn't matter if shaderID is added. ShaderID will just be added twice in that array then. ([Line 245](https://github.com/mrdoob/three.js/blob/a5c9edd6864bf7b0b0be4644c5388271f39f0d5d/src/renderers/webgl/WebGLPrograms.js#L245))

It can be in a critical path, so we should check performance. Also [compability](https://github.com/mrdoob/three.js/blob/a5c9edd6864bf7b0b0be4644c5388271f39f0d5d/src/renderers/webgl/WebGLPrograms.js#L245) could be a problem if we expect internet explorer to still work.

Mugen87:
> It doesn't matter if shaderID is added. ShaderID will just be added twice in that array then.

We could refactor the mentioned lines to the following in order to solve this problem:
```js
if ( parameters.shaderID === undefined )

    array.push( material.fragmentShader );
    array.push( material.vertexShader );

}
```
> It can be in a critical path, so we should check performance. Also compability could be a problem if we expect internet explorer to still work.

True. Unfortunately, IE does not support `Object.values()` 😞

Mugen87:
With ES6 we could even write^^:
```js
array.push(  ...Object.values( parameters ) );
```

Mugen87:
Anyway, I don't care which PR is going to be merged. Both are fine. My suggested refactoring can be done later, too.

Mugen87:
@gero3 Can you please resolve the merge conflicts? I'd like to approve this change.

gero3:
@Mugen87 The conflicts are resolved.

Mugen87 (approved):

Mugen87:
Merging this one in favor of #17552.

Mugen87:
Updated builds: fcb84f52316f29649a86e236127b42e607e3399e

PR #17658, opened 2019-10-03
WebGLProgram: Vertex and fragment shaders have the same defines.

gero3:
This reduces the clutter around standard defines. 

It adds the same defines to both vertex and fragement shader, which allows users to make use of the defines even in the other shader. It also makes sure these defines stay in sync between 2 shaders. 

It also makes it clearer which defines are set for Three.js.

### DIFF
--- a/src/renderers/webgl/WebGLProgram.js
+++ b/src/renderers/webgl/WebGLProgram.js
@@ -377,6 +377,93 @@ function generateEnvMapBlendingDefine( parameters ) {
 
 }
 
+function generateStandardDefines( renderer, parameters, extensions, material ) {
+
+	var shadowMapTypeDefine = generateShadowMapTypeDefine( parameters );
+	var envMapTypeDefine = generateEnvMapTypeDefine( parameters );
+	var envMapModeDefine = generateEnvMapModeDefine( parameters );
+	var envMapBlendingDefine = generateEnvMapBlendingDefine( parameters );
+	var gammaFactorDefine = ( renderer.gammaFactor > 0 ) ? renderer.gammaFactor : 1.0;
+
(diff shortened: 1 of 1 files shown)


src/renderers/webgl/WebGLProgram.js
```diff
@@ -385,17 +472,11 @@ function WebGLProgram( renderer, extensions, code, material, shader, parameters
...
-	var envMapTypeDefine = generateEnvMapTypeDefine( parameters );
-	var envMapModeDefine = generateEnvMapModeDefine( parameters );
-	var envMapBlendingDefine = generateEnvMapBlendingDefine( parameters );
-
-
-	var gammaFactorDefine = ( renderer.gammaFactor > 0 ) ? renderer.gammaFactor : 1.0;
 
 	var customExtensions = parameters.isWebGL2 ? '' : generateExtensions( material.extensions, parameters, extensions );
 
 	var customDefines = generateDefines( defines );
```
Mugen87: In this case, I would rename `generateDefines()` to `generateCustomDefines()`.

src/renderers/webgl/WebGLProgram.js
```diff
@@ -377,6 +377,93 @@ function generateEnvMapBlendingDefine( parameters ) {
 
 }
 
+function generateStandardDefines( renderer, parameters, extensions, material ) {
```
Mugen87: I guess I would rename the method to `generateDefines()`. The term `Standard` is already related to `MeshStandardMaterial`.

Mugen87:
Feel free to reopen the issue if the given feedback is implemented (and the merge conflicts resolved).

PR #17672, opened 2019-10-06
Docs: Add note about onBeforeCompile().

### LINKED ISSUE #14009: [Enhancement] onBeforeCompile may break material's .clone() method

Mugen87:
Replaces #15464, fixes #14009

### DIFF
--- a/docs/api/en/materials/Material.html
+++ b/docs/api/en/materials/Material.html
@@ -342,6 +342,9 @@
 		An optional callback that is executed immediately before the shader program is compiled.
 		This function is called with the shader source code as a parameter. Useful for the modification of built-in materials.
 		</p>
+		<p>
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

PR #17675, opened 2019-10-07
SpriteMaterial: add alphaMap support

WestLangley:
This support for alpha maps is consistent with the implementation of the other built-in materials.

A use case is a sprite in which the map is one of many images, and the sprite shape is controlled by the alpha map. This new feature prevents having to bake the alpha map into every image.

UV `offset/repeat` of the map and alpha map continues to be coupled. This is a current limitation of the library.


Mugen87 (approved):

WestLangley:
OK. Thanks. Merging...

/ping @sunag ... FYI

mrdoob:
Thanks!

PR #17684, opened 2019-10-08
Fixing "context must be XR compatible" issue

Artyom17:
See the issue #17683 for details.

### DIFF
--- a/src/renderers/webvr/WebXRManager.js
+++ b/src/renderers/webvr/WebXRManager.js
@@ -135,6 +135,12 @@ function WebXRManager( renderer, gl ) {
 			session.addEventListener( 'selectend', onSessionEvent );
 			session.addEventListener( 'end', onSessionEnd );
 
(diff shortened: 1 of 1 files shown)


src/renderers/webvr/WebXRManager.js
```diff
@@ -135,6 +135,12 @@ function WebXRManager( renderer, gl ) {
 			session.addEventListener( 'selectend', onSessionEvent );
 			session.addEventListener( 'end', onSessionEnd );
 
+			if ( gl.makeXRCompatible !== undefined ) {
```
Artyom17: Actually, I might be a little bit too paranoid to check for presence of makeXRCompatible. If a Browser supports WebXR in the latest spec, then this method must be there.

mrdoob:
https://github.com/mrdoob/three.js/issues/17683#issuecomment-540095805

Mugen87:
I think this can be closed.

PR #17697, opened 2019-10-09
WebGLRenderer: Added .initTexture().

### LINKED ISSUE #16338: Feature Request: Re-integrate setTexture2D
I've noticed setTexture2D has been removed in recent versions of ThreeJS. In my experience this has been a really important function in production applications to give the developer more control about when this texture is first uploaded to the GPU. For example, uploading textures during a preloader rather than waiting until first render (which leads to a huge 'jank' in common WebGL apps). Currently, the functionality still seems to exist in the WebGLTextures interface, but not exposed to the user. The only option is to render a scene with the texture visible (such as on a full screen quad in a plain scene with ortho camera), and then immediately afterwards clear the screen in order to ensure there is no 'flash' of content displayed to the user.

Is there any particular reason this functionality has been removed?

Thanks!

Mugen87:
Fixed #16338

Also added `WebGLState.unbindTexture()` in order to unbind a texture after certain operations. This is actually not required but the more clear approach (IMO).

### DIFF
--- a/docs/api/en/renderers/WebGLRenderer.html
+++ b/docs/api/en/renderers/WebGLRenderer.html
@@ -394,6 +394,9 @@
 		Returns the viewport.
 		</p>
 
+		<h3>[method:null initTexture]( [param:Texture texture] )</h3>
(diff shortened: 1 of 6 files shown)


Mugen87:
Screenshot from performance analysis:

image

mrdoob:
Thanks!

PR #17703, opened 2019-10-10
PointsMaterial: add alphaMap support

WestLangley:
This support for alpha maps is consistent with the implementation of the other built-in materials.

Alpha maps are particularly appropriate for controlling the shape of `Points`. Single-channel (`Luminance`) alpha maps are supported -- as they are elsewhere in the library.

UV `offset/repeat` of the map and alpha map continues to be coupled. This is a current limitation of the library.

### DIFF
--- a/docs/api/en/materials/PointsMaterial.html
+++ b/docs/api/en/materials/PointsMaterial.html
@@ -65,6 +65,17 @@ scene.add( starField );
 		<h2>Properties</h2>
 		<p>See the base [page:Material] class for common properties.</p>
 
+		<h3>[property:Texture alphaMap]</h3>
+		<p>The alpha map is a grayscale texture that controls the opacity across the surface
+			(black: fully transparent; white: fully opaque). Default is null.<br /><br />
+
+			Only the color of the texture is used, ignoring the alpha channel if one exists.
+			For RGB and RGBA textures, the [page:WebGLRenderer WebGL] renderer will use the
+			green channel when sampling this texture due to the extra bit of precision provided
+			for green in DXT-compressed and uncompressed RGB 565 formats. Luminance-only and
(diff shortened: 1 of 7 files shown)


WestLangley:
Since the uv and map Shader Chunks were already combined into one file for the `PointsMaterial` shader, I continued that pattern and included the alpha map Chunk in the same file. I did not see any reason to separate them.

I hope that is not a problem for Node...

/ping @sunag ...

sunag (approved):

Mugen87 (approved):

WestLangley:
Thanks. Merging...

mrdoob:
Thanks!

EliasHasle:
In my experience, alpha does _not_ seem to be a convenient way of controlling the shape of points, as there so easily arise conflicts between points that are drawn in the same draw call. Having the option is good, though. With a lot of care, one may be able to get good results, such as by sorting points by distance before drawing. (This thread on the forum is a good example: https://discourse.threejs.org/t/shader-to-create-an-offset-inward-growing-stroke/6060)

WestLangley:
I think that is an unfair statement. For controlling shape, alpha is either zero or one. The alpha map is a _cutout_, transparency is not required for rendering, and the draw order does not matter.

EliasHasle:
Hm, I have had no success with that. I have not tried explicitly setting only zeros and ones, but I have used functions that generate mostly pure zeros and ones, and this has caused the problems referred to. (I guess graphics drivers may convert shaders that are guaranteed to set alpha to only zeros and ones into `discard` branches, avoiding the problems. But a map, unless read with nearest filtering, will be interpolated.)

If you have an example for shape control using alpha, which works with many particles and no sorting, I would be really interested in seeing it. :)

PR #17707, opened 2019-10-10
Docs: Clarify Mesh.drawMode.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17694#issuecomment-540072331


mrdoob:
Thanks!

Issue #16778, opened 2019-06-13
RGB floating point textures broken on latest Firefox


silvainSayduck:
The EXRLoader recently broke on Firefox, tested on MacOS and Windows.

See https://threejs.org/examples/webgl_materials_envmaps_exr.html

![](https://i.imgur.com/TEOBIEH.png)

It works on Chrome (setting aside visual artefacts as reported on https://github.com/mrdoob/three.js/issues/16409)

silvainSayduck:
I'm almost positive that the example (and my own code using EXR) worked on Firefox just a few weeks ago, so I guess something has changed on Firefox end.

In any case, this workaround for Android seems to also work in this case: https://github.com/mrdoob/three.js/issues/15343#issuecomment-457528822

Mugen87:
I wonder if it's not better if `EXRLoader` would always return RGBA textures. As mentioned by @gman at stackoverflow, it's not guaranteed that a WebGL (1) implementation needs to support RGB floating point textures.

WestLangley:
Similarly, [the HDR example](https://threejs.org/examples/webgl_materials_envmaps_hdr.html) , which uses `HDRCubeTextureLoader` is also broken on Firefox  _when the output data type is changed_ to `THREE.FloatType` or `THREE.HalfFloatType`.

silvainSayduck:
Still broken on Firefox 68.0.1 (MacOS)

Mugen87:
We can fix the issue if we switch to `RGBA`. However, I'm not sure everybody is okay with this change..

silvainSayduck:
Hi @Mugen87 ,

In fact, a lot of devices are just breaking with the EXRLoader, I'm guessing related to the lack of support for RGB Floating point textures.
Right now, we need to wait for users to report that it doesn't work, and add a manual fix per reported device, which is not ideal since a lot of devices out there could still be breaking. Would it be possible to add a check of some sorts to the EXRLoader that would make sure it works (ie. at least check for RGB Floating point texture support)?
I realise this is half a support question instead of an issue, but it also still seems to me that this would be a good approach to make the EXRLoader production-ready.
I'm willing to add the needed checks, but I'm not sure how to figure out the exhaustive webGL extension/feature list that is required by the EXRLoader. Is there a way to figure that one out?

silvainSayduck:
[...]

PR #17708, opened 2019-10-10
Examples: Remove EquirectangularToCubeGenerator.

Mugen87:
...and use `WebGLRenderTargetCube.fromEquirectangularTexture()` instead.


Mugen87:
http://raw.githack.com/Mugen87/three.js/dev39/examples/webgl_loader_gltf.html
http://raw.githack.com/Mugen87/three.js/dev39/examples/webgl_loader_gltf_extensions.html
http://raw.githack.com/Mugen87/three.js/dev39/examples/webgl_materials_envmaps_exr.html

examples/webgl_materials_envmaps_exr.html
```diff
@@ -79,14 +78,15 @@
...
 
 						texture.minFilter = THREE.NearestFilter;
-						// texture.magFilter = THREE.NearestFilter;
 						texture.encoding = THREE.LinearEncoding;
 
-						var cubemapGenerator = new EquirectangularToCubeGenerator( texture, { resolution: 512, type: THREE.HalfFloatType } );
-						exrBackground = cubemapGenerator.renderTarget;
-						var cubeMapTexture = cubemapGenerator.update( renderer );
+						var options = {
+							minFilter: THREE.NearestFilter
```
WestLangley: I'm not sure that makes sense. Here, or in the lines above. Probably should be `Linear`.

You will have to step through the code to see if it matters.

Also, you will almost surely have to change the loader output data type to `HalfFloat` to duplicate the previous results.
```js
.setDataType( THREE.HalfFloatType )
```

examples/webgl_materials_envmaps_exr.html
```diff
@@ -79,14 +78,15 @@
...
+						};
 
-						var pmremGenerator = new PMREMGenerator( cubeMapTexture );
+						exrBackground = new THREE.WebGLRenderTargetCube( 512, 512, options ).fromEquirectangularTexture( renderer, texture );
+
+						var pmremGenerator = new PMREMGenerator( exrBackground.texture );
```
WestLangley: _PMREM changes the parameters of the input texture_. This is of course not desirable. So now you don't know what the params of the background texture are. PMREM is going to be replaced, however.

examples/webgl_materials_envmaps_exr.html
Mugen87: > I'm not sure that makes sense. Here, or in the lines above. Probably should be Linear.

Why do the previous texture filtering settings are wrong. It does actually look good.

> Also, you will almost surely have to change the loader output data type to HalfFloat to duplicate the previous results.

That's true. Let me fix this!

Mugen87: Links for direct comparison:

https://threejs.org/examples/webgl_materials_envmaps_exr
https://raw.githack.com/Mugen87/three.js/dev39/examples/webgl_materials_envmaps_exr.html

examples/webgl_materials_envmaps_exr.html
Mugen87: I don't know a good solution here. Cloning the render target does not work so the only option would to be call `WebGLRenderTargetCube.fromEquirectangularTexture()` twice.

Mugen87: It looks like this was already an issue with `EquirectangularToCubeGenerator.update()`. The method returned a reference to the internal rendertarget's texture, see

https://github.com/mrdoob/three.js/pull/17708/files#diff-db685e7af40e5c3f125f34b618daf01bL80

WestLangley:
@Mugen87

1.  I forked your branch and tested `webgl_loader_gltf_extensions.html`, and it appears to be OK.

2. `webgl_materials_envmaps_exr.html` has the issues I mentioned above.

3. I did not study `webgl_loader_gltf.html`. You can do that one.  ;-)

---

As I did, you need to make sure `encoding`, `generateMipmaps`, `magFilter`, `minFilter`, and `type` are unchanged from the previous examples -- for both the background and the environment map. If the settings are unchanged, but do not make sense, then this is an opportunity to fix them.

As I said above, PMREM changes the input texture, so you need to check the background texture after PMREM is finished.

---

Personally, I think setting a background cube map with mipmapping disabled is probably not what we want to be recommending, but I guess that can be a separate issue.

examples/webgl_materials_envmaps_exr.html
WestLangley: Your example is throwing errors on my machine.

I have always used Linear interpolation when the encoding supports it. Some encodings like RGBE are not compatible with linear, so nearest must be used. I have never seen Linear and Nearest mixed in the same texture. If there is a use case for that, I am not aware of it.

Mugen87:
I've checked all examples again and made sure the properties of the background's texture are the same as before. Let me list those values in the following table:

Background:

| Example | encoding | generateMipmaps  | minFiler | magFilter | type
| ------------- |-------------| -----| -----| -----| -----|
| gltf    | RGBEEncoding | false | NearestFilter | NearestFilter | UnsignedByteType
| gltf_ext    | RGBEEncoding | false | NearestFilter | NearestFilter | UnsignedByteType
| exr (exr)   | LinearEncoding | false | NearestFilter | LinearFilter | HalfFloatType
| exr (png)   | sRGBEncoding | true | LinearMipmapLinearFilter | LinearFilter | UnsignedByteType

EnvMap:

| Example | encoding | generateMipmaps  | minFiler | magFilter | type
| ------------- |-------------| -----| -----| -----| -----|
| gltf    | RGBM16Encoding | false | LinearFilter | LinearFilter | UnsignedByteType
| gltf_ext    | RGBM16Encoding | false | LinearFilter | LinearFilter | UnsignedByteType
| exr (exr)   | LinearEncoding | false | LinearFilter | LinearFilter | HalfFloatType
| exr (png)   | sRGBEncoding | true | LinearFilter | LinearFilter | UnsignedByteType

examples/webgl_materials_envmaps_exr.html
Mugen87: > Your example is throwing errors on my machine.

Um, can't reproduce this on my iMac with Chrome. What errors are logged?

Mugen87:
> If the settings are unchanged, but do not make sense, then this is an opportunity to fix them.

The PNG env map in the EXR example has set `generateMipmaps` to `true` although only `LinearFilter` is used. Should we change this setting?

examples/webgl_materials_envmaps_exr.html
WestLangley: Failure is on macOS Safari.
>WebGL: INVALID_OPERATION: texImage2D: type HALF_FLOAT_OES but ArrayBufferView is not NULL

WestLangley:
Right. Fix it if you can. Otherwise, wait until the PMREM rewrite is finalized.

examples/webgl_materials_envmaps_exr.html
Mugen87: Okay, I can reproduce. It seems `WebGLRenderTargetCube.fromEquirectangularTexture()` throws this error on Safari when the input texture is of `THREE.HalfFloatType`. Using `THREE.FloatType` with `EXRLoader` as before solves the issue. However, since `WebGLRenderTargetCube.fromEquirectangularTexture()` strictly derives the `type` of the source texture (unlike `EquirectangularToCubeGenerator` which does `type: options.type || this.sourceTexture.type`), I can't fix this issue without changing previous values (see https://github.com/mrdoob/three.js/pull/17708#issuecomment-541005429).

Mugen87:
Merging for now. The settings in `webgl_materials_envmaps_exr` can be optimized with a different PR. 

Sidenote: Let's not forget `webgl_materials_envmaps_exr` does not work in FF (see #16778).

examples/webgl_materials_envmaps_exr.html
Mugen87: see #17727

PR #17718, opened 2019-10-12
Layers for WebVR

AmadeusJohansson:
Before the VR camera was hardcoded to only show layer 0 and 1 in the left eye, and layer 0 and 2 in the right eye. Because of this, layers were unusable since the layers set on the original perspective camera was not carried over to the new VR camera.

This is very confusing if you do not know about this since it simply seems like layers are broken in VR mode. Most people using this functionality probably want to use layers normally and not render different views for different eyes. If they do, they will now have to use the new leftLayers and rightLayers fields.


Mugen87:
`WebVRManager` has been removed #18129. So this change needs to be done for `WebXRManager` instead.

Mugen87:
Better to create a new PR targeting `WebXRManager`.

Issue #17662, opened 2019-10-04
Shading models using incorrect view direction for orthographic projections.


sciecode:
#17379 exposed a problem with the current orthographic behavior of light. However it was quickly dismissed as a help question, but I don't think this is the case.

It was thought to be a problem with physical and standard models, but I believe the real issue at hand is the incorrect calculation of the view direction for orthographic cameras.

Currently the view direction is being calculated as follows `vViewPosition = - mvPosition.xyz;`
However this is not a correct model for an orthographic view. This is supposed to be fixed as the camera direction, disregarding the object position. 

This is causing multiple unexpected behaviours. Given a fixed orthographic camera and a scene being lit by a directional light, the specular light component shouldn't move if the object moves. This is not what's happening on Dev - [dev example](https://rawcdn.githack.com/sciecode/three.js/307fbf53e2b24ca5c5908ad8ea77a58b2bc0fc62/examples/webgl_interactive_ortho_problem.html)

![image](https://user-images.githubusercontent.com/3927951/66229311-f5473900-e6b7-11e9-840a-3e6b5e4174b2.png)

From the image, you can see the fresnel component is being incorrectly placed on the object. As previously noted, the specular component is also incorrectly moving when it shouldn't be.

____

In order to fix this behavior, I just replaced the `vViewPosition` calculation for every shading model with the following:
`vViewPosition = isPerspectiveMatrix( projectionMatrix ) ? - mvPosition.xyz : vec3( 0, 0, 1 );`

This is the example of what should be actually happening - [fix example](https://rawcdn.githack.com/sciecode/three.js/c82bde8e3481de03b276af1f665012b76e772815/examples/webgl_interactive_ortho_problem.html)

![image](https://user-images.githubusercontent.com/3927951/66229533-6c7ccd00-e6b8-11e9-88e6-084438a4fec8.png)

Would appreciate if other contributors, that are more familiar with orthographic projection, could confirm this is indeed the correct way of handling the issue at hand.

WestLangley:
While you are thinking about this, see if you can also study reflections from an environment map.

It is not clear to me how we should be handling lighting with an orthographic camera.

It might be of interest to compare with other engines.

alelepd:
Thanks for this @sciecode! Glad to know there was a fix.

I compared this fix with blender and they book look the same now.

Recorded a video from blender with similar settings:
https://jumpshare.com/v/kS3GlDUihE64q8AsmMdq

sciecode:
[...]

Issue #17623, opened 2019-10-01
Logarithmic Depth Artifacts


vanruesc:
#### Description of the problem

Logarithmic depth appears to be broken in `three@0.109.0`. Possibly related: #17442.

On my system, I see lots of depth artifacts in setups that use `logarithmicDepthBuffer`. I don't know if this problem is hardware-related; would be cool if someone could verify!

#### Screenshots

[logarithmicdepthbuffer example](https://threejs.org/examples/?q=log#webgl_camera_logarithmicdepthbuffer)

![logdepthbuf](https://user-images.githubusercontent.com/9214917/65923543-a1073500-e3e9-11e9-8cbb-b39c76a54c5e.png)

I get the same artifacts in one of my [projects](https://vanruesc.github.io/postprocessing/public/demo/#bokeh) where I use logarithmic depth for testing purposes:

| r108 (worked fine) | r109 (broken) |
|------|------|
| ![logdepthbuf1](https://user-images.githubusercontent.com/9214917/65923515-8b920b00-e3e9-11e9-8d71-9991482f6fec.png) | ![logdepthbuf2](https://user-images.githubusercontent.com/9214917/65923534-9a78bd80-e3e9-11e9-8afb-02e7be93956e.png) |

#### Three.js version

- [x] Dev
- [x] r109
- [ ] r108

#### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer
- [ ] Edge

#### OS

- [ ] All of them
- [x] Windows 10 Pro, 64-bit
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

#### Hardware Requirements (graphics card, VR Device, ...)

GPU: NVIDIA GTX 1060 6GB

#### Drivers

Graphics Driver: 436.30 (Standard Type)

gkjohnson:
Hmm that PR shouldn't have affected any of the logarithmic depth buffer logic when using a perspective camera. Because of how noisy the output is my best guess is that it's something to do with the varying `vIsPerspective` value comparison that is causing the issue?

[...]

PR #17742, opened 2019-10-15
WebGLProgram: camera-specific shader define.

sciecode:
As per suggested on https://github.com/mrdoob/three.js/issues/17662#issuecomment-541164744

This PR introduces `#define ORTHO` on programShaders. In my opinion this is long overdue.

As noted by @EliasHasle, this would facilitate a lot the handling of orthographic projection specific behavior inside `ShaderChunks`. It would also solve the issue of `varying vIsPerspective` hardware dependant calculation, dismiss the problems involved in colliding varyings and avoid several conditional branching currently happening inside the shaders to deal with orthographic cameras.

I'm also waiting for this change to push a fix for https://github.com/mrdoob/three.js/issues/17379 and https://github.com/mrdoob/three.js/issues/17662

**sidenote**: went for `ORTHO` instead of `ORTHOGRAPHIC` for shorter notation, but open to suggestions.

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1059,13 +1059,13 @@ function WebGLRenderer( parameters ) {
 
 					for ( var i = 0; i < object.material.length; i ++ ) {
 
-						initMaterial( object.material[ i ], scene.fog, object );
+						initMaterial( object.material[ i ], scene.fog, object, camera );
 
 					}
 
 				} else {
 
-					initMaterial( object.material, scene.fog, object );
+					initMaterial( object.material, scene.fog, object, camera );
 
 				}
 
@@ -1485,7 +1485,7 @@ function WebGLRenderer( parameters ) {
 
 	}
 
-	function initMaterial( material, fog, object ) {
+	function initMaterial( material, fog, object, camera ) {
 
 		var materialProperties = properties.get( material );
 
@@ -1495,7 +1495,7 @@ function WebGLRenderer( parameters ) {
 		var lightsStateVersion = lights.state.version;
 
 		var parameters = programCache.getParameters(
-			material, lights.state, shadowsArray, fog, _clipping.numPlanes, _clipping.numIntersection, object );
+			material, lights.state, shadowsArray, fog, _clipping.numPlanes, _clipping.numIntersection, object, camera );
 
 		var code = programCache.getProgramCode( material, parameters );
 
@@ -1701,7 +1701,7 @@ function WebGLRenderer( parameters ) {
 
 		if ( material.needsUpdate ) {
 
-			initMaterial( material, fog, object );
+			initMaterial( material, fog, object, camera );
 			material.needsUpdate = false;
 
 		}
--- a/src/renderers/webgl/WebGLProgram.js
+++ b/src/renderers/webgl/WebGLProgram.js
@@ -449,6 +449,8 @@ function WebGLProgram( renderer, extensions, code, material, shader, parameters 
 			( parameters.useFog && parameters.fog ) ? '#define USE_FOG' : '',
 			( parameters.useFog && parameters.fogExp2 ) ? '#define FOG_EXP2' : '',
 
+			parameters.orthographic ? '#define ORTHO ' : '',
+
 			parameters.map ? '#define USE_MAP' : '',
 			parameters.envMap ? '#define USE_ENVMAP' : '',
 			parameters.envMap ? '#define ' + envMapModeDefine : '',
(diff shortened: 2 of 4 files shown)


EliasHasle:
Without complete overview, I think this is a step in the right direction. The implementation looks clean and complete too.

The obivious downside is that it will cause more compiles (and more _programs_) for cases where both perspective and orthographic cameras are to be used on the same objects.

Regarding my (arguably minor) point in the same comment of opening for more exotic cameras, it would be an advantage to also define `MATRIX_CAMERA` (or something like that) for both orthographic and perspective, and let the default `project_vertex` `#ifdef`-depend on that.

sciecode:
> The obivious downside is that it will cause more compiles (and more programs) for cases where both perspective and orthographic cameras are to be used on the same objects.

Yes, this has the side effect of doubling the number of programs for materials being used by two different cameras. 

However, considering all the benefits exposed for single camera ( overwhelming majority of use-cases), I feel the pros far out-weight the cons in this case.

Mugen87:
> However, considering all the benefits exposed for single camera ( overwhelming majority of use-cases), I feel the pros far out-weight the cons in this case.

I'm not sure about that. Having more programs will lead to more WebGL state changes which can noticeable downgrade rendering performance (depending on the scenario). 

Why is the current shader-based solution `isPerspectiveMatrix()` not sufficient?

sciecode:
@Mugen87 I know you edited the answer, but I feel like the question "why `isPerspectiveMatrix` isn't sufficient" is relevant and I'd like to address it.

The current problem with that approach is that the `projectionMatrix` isn't available on the fragmentShader, thus requiring us to set a varying on the vertex to pass this information onwards. This `varying vIsPerspective` is how Log Depth chunks are currently dealing orthographic vs perspective handling.

I'd say this is probably the best option if there's a low number of conditional branching happening, but this won't be the case after I start handling problems discussed on https://github.com/mrdoob/three.js/issues/17662. Let me illustrate the problem:

On the PR in question, I proposed a fix that involved directly changing `vViewPosition`, however this is not the correct way of handling it, as it causes problems with clipping planes and flat shading. 

Thus I'd need to change `geometry.viewDir` declaration conditionally, which is on itself always executed. `viewDir`is calculated on the fragment for most material models, but not on lambert model. Which already introduces the problem of when and where it should be the declared and defined, additionally we when using a `envMap` or `logdepth` this varying will be used on the every material model. 

In short, `isPerspectiveMatrix` with `vIsPerspective` are sufficient to handle the problem if done correctly, which will likely require the introduction of few new shaderchunks <ortho_pars>, <ortho_vertex>, <ortho_fragment> and so on.

>I'm not sure about that. Having more programs can potentially lead to more WebGL state changes which can downgrade rendering performance (depending on the scenario).

The current model makes the code way harder to maintain and it affects performance of every material even when not using orthogonal projection. There's, also, the fact that we're checking and branching if the camera is orthographic on every frame and every fragment when we could simply be calculating it once at compile time. Which is making every use-case slower, instead of making it possibly slower on a very specific use-case.

sciecode:
I prefer @gkjohnson proposal over this PR actually.

sciecode:
This would solve the problems regarding varying and make it a lot easier to mantain, but it would still require the multiple conditional branching. It’s an improvement over our current approach nonetheless

gkjohnson:
The modern impacts of branching in shaders are also not so clear to me. I know it used to be a big deal but if you have any documents on the current thinking on this topic that would be great. I've heard that 

1. branching is always bad and incurs a performance hit
2. it's only really bad if multiple fragments in a batch take different branches
3. that it's really not a big deal on modern hardware

I'm sure it really depends on the graphics hardware in question but my strategy is typically to assume the worst case and use `mix` in order to avoid a condition:

```hlsl
float3 viewDir = mix( perspViewDir, orthoViewDir, isOrthographic );
```

where `isOrthographic` is `1.0` or `0.0`.  But I don't know if this is _actually_ better because both "branches" here still have to be calculated in order to call `mix`. I'd definitely be interested in whatever the de-facto best practices on this are if they exist.

An aside but if `isOrthographic` is passed as a uniform I think that would remove the rounding precision issue that caused a logarithmic depth rendering regression we ran into due to varying field interpolation (see https://github.com/mrdoob/three.js/issues/17623#issuecomment-537125504).

sciecode:
> 1. branching is always bad and incurs a performance hit
>2. it's only really bad if multiple fragments in a batch take different branches
>3. that it's really not a big deal on modern hardware

I'm also not completely clear on the impacts of conditional branching and I have read similar things. I know it is highly dependant on specific hardware, therefore I've always used the same `mix` approach as you presented.

But, I also remember reading that the built-in ternary operator is supposed to be better for modern hardware when compared with `mix`. I think if we were to use the `isOrthographic` uniform, we could probably stick the ternary operator.

```glsl
vec3 viewDir = ( isOrthographic == 0.0 ) ? pespViewDir : orthoViewDir;
```

My only worry is that, by using multiple of these conditionals, we have negative impact on the performance for older hardware.

EliasHasle:
> I'm also not completely clear on the impacts of conditional branching

Here is a relevant question on StackExchange which has attracted some seemingly high-quality answers (that I personally feel I learned from reading): https://gamedev.stackexchange.com/questions/158173/does-cause-branching-in-glsl

PR #17743, opened 2019-10-16
Update issue template with correct arguments to OrbitControls

EliasHasle:
...in jsfiddles, that is.

https://discourse.threejs.org/t/jsfiddle-templates-should-supply-renderer-domelement-to-orbitcontrols/9267/11

#17612

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -11,8 +11,8 @@ Always include a code snippet, screenshots, and any relevant models or textures 
 
 Please also include a live example if possible. You can start from these templates:
 
-* [jsfiddle](https://jsfiddle.net/yn8azeg1/) (latest release branch)
-* [jsfiddle](https://jsfiddle.net/kLwhx875/) (dev branch)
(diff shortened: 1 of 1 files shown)


looeee:
We should probably update them to use buffer geometry as well.

EliasHasle:
Ah, I didn't know the issue template fiddles were different. I will have to adapt the current ones instead, then.

EliasHasle:
I was going to update the new fiddles from the old, but now the old fails with syntax error on `*` (presumably `import *`). 😕

EliasHasle:
Anyway, it may be inconvenient/uncomfortable that I own the official fiddles. An alternative surely is that the owner of the official fiddles updates them to pass `renderer.domElement` after `camera`.

looeee:
> Ah, I didn't know the issue template fiddles were different. 

This isn't intentional. The issue templates were updated sometime in the last month and the forum templates got left behind.

Mugen87:
Solved via https://github.com/mrdoob/three.js/commit/35eaaa18bd5fae9033bdac30a857a281122c9536

PR #17753, opened 2019-10-17
WebGLProgram: Add isOrthographic uniform.

sciecode:
Alternative to #17742 

Based on https://github.com/mrdoob/three.js/pull/17742#issuecomment-542796736, <del> makes  `varying isPerspective` no longer needed and </del> sets up for future orthographic behavior handling.

@gkjohnson @Mugen87 Does it look ok to you guys?

### DIFF
--- a/src/renderers/WebGLRenderer.js
+++ b/src/renderers/WebGLRenderer.js
@@ -1753,6 +1753,8 @@ function WebGLRenderer( parameters ) {
 
 				_currentCamera = camera;
 
+				p_uniforms.setValue( _gl, 'isOrthographic', camera.isOrthographicCamera === true );
+
 				// lighting uniforms depend on the camera so enforce an update
 				// now, in case this material supports lights - or later, when
 				// the next material that does gets activated:
--- a/src/renderers/webgl/WebGLProgram.js
+++ b/src/renderers/webgl/WebGLProgram.js
@@ -495,6 +495,7 @@ function WebGLProgram( renderer, extensions, code, material, shader, parameters 
 			'uniform mat4 viewMatrix;',
 			'uniform mat3 normalMatrix;',
 			'uniform vec3 cameraPosition;',
+			'uniform bool isOrthographic;',
 
 			'#ifdef USE_INSTANCING',
 
@@ -617,6 +618,7 @@ function WebGLProgram( renderer, extensions, code, material, shader, parameters 
 
 			'uniform mat4 viewMatrix;',
 			'uniform vec3 cameraPosition;',
+			'uniform bool isOrthographic;',
 
 			( parameters.toneMapping !== NoToneMapping ) ? '#define TONE_MAPPING' : '',
 			( parameters.toneMapping !== NoToneMapping ) ? ShaderChunk[ 'tonemapping_pars_fragment' ] : '', // this code is required here because it is used by the toneMapping() function defined below


src/renderers/shaders/ShaderLib/sprite_vert.glsl.js
```diff
@@ -20,9 +20,7 @@ void main() {
 
 	#ifndef USE_SIZEATTENUATION
 
-		bool isPerspective = isPerspectiveMatrix( projectionMatrix );
-
-		if ( isPerspective ) scale *= - mvPosition.z;
+		// if ( ! isOrthographic ) scale *= - mvPosition.z;
```
Mugen87: Why is this line commented out?

sciecode: Just pushed a fix, noticed just before you commented.

Mugen87:
I definitely prefer this solution over #17742

gkjohnson:
I prefer this approach, too -- looks good!

Mugen87:
@sciecode How about closing #17742 and focus on this PR instead? I think it should not be hard for the other devs to review this small PR 👍

sciecode:
> @sciecode How about closing #17742 and focus on this PR instead? 

Sure thing 👍

EliasHasle:
This is from the older PR discussion, but I would like to comment it here, as it is still relevant:
 > The current problem with that approach is that the projectionMatrix isn't available on the fragmentShader, thus requiring us to set a varying on the vertex to pass this information onwards.

How is it that `projectionMatrix` is not available in the fragment shader? Uniforms are available both in vertex and fragment shaders. Whether it is declared in the fragment shader is another question. As is how referencing cells of it in the fragment shader (the `isPerSpectiveMatrix(projectionMatrix)` solution) will impact performance, compared to using a separate `float` or `bool` uniform, or a varying.

WestLangley:
FYI - "is-perspective" is not the same as "not-orthographic". Cameras do not have to be one or the other.

WestLangley:
FYI - [Unity code](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl)

```glsl
// Returns 'true' if the current view performs a perspective projection.
bool IsPerspectiveProjection()
{
#if defined(SHADERPASS) && (SHADERPASS != SHADERPASS_SHADOWS)
    return (unity_OrthoParams.w == 0);
#else
    // This is a generic solution.
    // However, using 'unity_OrthoParams' is better for cache locality.
    // TODO: set 'unity_OrthoParams' during the shadow pass.
    return UNITY_MATRIX_P[3][3] == 0;
#endif
}
```

sciecode:
> FYI - "is-perspective" is not the same as "not-orthographic". Cameras do not have to be one or the other.

I'm aware of that, but we aren't currently dealing with other types of cameras that need to be distinctly treated. I understand it isn't ideal if we look to expand this in the future, do you have any suggestions?

WestLangley:
>we aren't currently dealing with other types of cameras

The render only requires an instance of `THREE.Camera`.  See `GPUComputationRenderer`, for example.

sciecode:
I see, default camera wouldn't execute size attenuation and logarithmic depth buffer previously, but the would on this PR. Need to address this, thanks for noting.

sciecode:
This actually creates quite a problem, because `isOrthographic` becomes insufficient to handle all the use cases. So we still need to use `vIsPerspective`, and thus making this PR a lot less meaningful.

sciecode:
Limited this PR to adding `isOrthographic` uniform.

WestLangley:
This entire thread make no sense in light of the completely rewritten commit.

How about closing this.

You can incorporate the three (3) lines as part of another PR, where you will be expected to provide the context to justify these changes.

Mugen87:
Just for my understanding: I see the code snipper from Unity but it's not clear to me if its part of the vertex or fragment shader. If it's the latter one, they provide the projection matrix as a uniform. Couldn't we do the same so the varying `vIsPerspective` is not necessary anymore?

PR #17765, opened 2019-10-18
Added  parameter to raycaster.intersectObject/s

fernandojsg:
Many times I have been facing this issue and I just hacked the raycaster code. Basically you have a scene with one mesh as a collider, but you have it invisible but you still want to do raycasting against it.
I added an `ignoreVisibility` parameter, if `true` it will just skip the `object.visible === false` check, otherwise it will act as till now.

Maybe we could also transform the extra/optional parameters as an object:
```javascript
intersectObject( object, raycaster, params ) {}

intersectObject( object, raycaster, {
  intersects: intersects,
  recursive: false,
  ignoreVisibility: true
});
```

### DIFF
--- a/src/core/Raycaster.js
+++ b/src/core/Raycaster.js
@@ -42,9 +42,9 @@ function ascSort( a, b ) {
 
 }
 
-function intersectObject( object, raycaster, intersects, recursive ) {
+function intersectObject( object, raycaster, intersects, recursive, ignoreVisibility ) {
 
-	if ( object.visible === false ) return;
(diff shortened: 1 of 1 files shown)


Mugen87:
Related #14700

I guess a proposal like https://github.com/mrdoob/three.js/issues/14700#issuecomment-475235430 should also solve this issue.

fernandojsg:
Ops I didn't noticed that issue. I like your layers proposal, so closing it in favor of that one

PR #17784, opened 2019-10-21
WebGLPrograms: Enable uvs only in vertex shader.

Mugen87:
Related #17626

### DIFF
--- a/src/renderers/shaders/ShaderChunk/uv_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/uv_pars_fragment.glsl.js
@@ -1,5 +1,5 @@
 export default /* glsl */`
-#ifdef USE_UV
+#if ( defined( USE_UV ) && ! defined( UVS_VERTEX_ONLY ) )
 
 	varying vec2 vUv;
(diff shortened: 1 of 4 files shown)


jbaicoianu:
This seems like it fixes the warning by declaring the variable as local to the vertex shader in the case where the fragment shader doesn't use it, but I think in that case it's initializing the vector to the default (0,0).  Correct me if I'm missing something, but it seems we'll also need to modify the vertex shader to assign the values if they're not defined:

https://github.com/mrdoob/three.js/blob/970f47a2addd1b965d2c43d63ca47170649eba73/src/renderers/shaders/ShaderChunk/uv_vertex.glsl.js#L2-L6

Mugen87:
I'm afraid I don't understand what you mean^^. The idea of the PR is to avoid the usage of a varying. Essentially `vUv` is just declared differently. 

> Correct me if I'm missing something, but it seems we'll also need to modify the vertex shader to assign the values if they're not defined:

How would this modification look like?

Mugen87:
Merging. We can finally verify the change on `dev`.

Issue #17793, opened 2019-10-22
Geometry.fromBufferGeometry has an issue when reading InterleavedBuffer


Julianouyang:
##### Geometry.fromBufferGeometry produces wrong data when working with InterleavedBuffer

Assume we have a bufferGeometry with interleavedBufferAttribute as its attribute type. 

There is an issue in [geometry.js](https://github.com/mrdoob/three.js/blob/f7727f85502cbe34600b65b727c8e6204e9bf505/src/core/Geometry.js#L199) if we call `Geometry.fromBufferGeometry(bufferGeometry)`. 
The current code assumes that the type of attribute is `BufferAttribute`, but it also can be `InterleavedBufferAttribute`, and they cannot be read in the same way.

For `BufferAttribute`, we read attribute like position, normal individually. But for `InterleavedBufferAttribute`, it returns a typed array that bundles all attributes together. So we need to read the data based on its offset and itemsize.

It might be better to support reading attribute individually in `InterleavedBufferAttribute `class, but at least we need a quick fix in the geometry method. 

If you all think this is a bug and nobody is fixing it right now, I can open a PR. I can also create a jsfiddle if you think it's necessary to demonstrate the wrong data output.

##### Three.js version

- [x] Dev
- [x] r109
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
`Geometry` can be considered as deprecated so its unlikely that this will ever be supported. What is your use case for still using `Geometry`?

Julianouyang:
@Mugen87 I tried to create an `EdgesGeometry` and it calls [Geometry.fromBufferGeometry](https://github.com/mrdoob/three.js/blob/e0a31ea77c0242ae29ef6e951fa588216e1d247f/src/geometries/EdgesGeometry.js#L40) internally.

Mugen87:
Oh dear, this is indeed an unfortunate dependency to `Geometry`. Thanks for pointing this out!

@mrdoob TBH, I was actually not aware of this detail. I guess we have to migrate `EdgesGeometry` to `BufferGeometry` in the near future.

WestLangley:
`Geometry.fromBufferGeometry` is used internally in three.js in multiple places.

SubdivisionModifier

[...]

PR #17794, opened 2019-10-22
Geometry.fromBufferGeometry: support interleaved buffer

Julianouyang:
As discussed in issue #17793, `geometry.fromBufferGeometry` doesn't support `InterleavedBuffer`. With this change it should read data correctly.


Mugen87:
Closing in favor of #18783.

PR #17796, opened 2019-10-23
CSS3DRenderer: Ensure child DOM elements are correctly removed.

viniciuslrangel:


examples/jsm/renderers/CSS3DRenderer.js
```diff
@@ -10,21 +10,33 @@ import {
 	Vector3
 } from "../../../build/three.module.js";
 
+const _removedEvent = { type: 'cssRemoved' };
```
Mugen87: Please use `var` here.

examples/jsm/renderers/CSS3DRenderer.js
```diff
@@ -10,21 +10,33 @@ import {
...
-	this.addEventListener( 'removed', function () {
+	this.addEventListener( 'cssRemoved', function () {
 
 		if ( this.element.parentNode !== null ) {
 
 			this.element.parentNode.removeChild( this.element );
 
 		}
 
+		this.children.forEach( function (e) {
```
Mugen87: Minor code style issues. The code should look like so.
```js
this.children.forEach( function ( e ) {

    e.dispatchEvent( _removedEvent );

} );

**Comment by Mugen87** [comment]
Can you please explain a bit what problem the PR is going to solve?

File: examples/jsm/renderers/CSS3DRenderer.js [hunk 3]
```diff
@@ -10,21 +10,33 @@ import {
...
+			e.dispatchEvent( _removedEvent );
+		} )
+
+	} );
+
+	this.addEventListener( 'removed', function () {
+
+		this.dispatchEvent(_removedEvent)
```
Mugen87:
```js
this.dispatchEvent( _removedEvent );
```

viniciuslrangel:
The CSS3DObject uses the remove event to remove itself from DOM. But if the removed element has children, they don't receive remove too and aren't removed from DOM.

yomotsu:
I think this could be done in userland because children are not `THREE.CSS3DObject`.

viniciuslrangel:
Actually, this is removing the node itself. remove Child is being called from the parent to this

yomotsu:
that's right, however, I think `removeChild` remove all descendants too.


https://jsfiddle.net/nksdLmru/

viniciuslrangel:
CSS3DRenderer doesn't organize elements in a DOM tree this way. Every object goes in the same root element.

yomotsu:
I got your point. Here is the problem to solve.
https://jsfiddle.net/cefv0r5q/

yomotsu:
You can do that with `traverse`.
https://jsfiddle.net/3bvsfdp6/

However, I'm not sure removing child-objects is good or bad because `Objecr3D` doesn't `traverse`-removing.

WestLangley:
If the remaining issues are resolved, please edit the` js/` file, not the `jsm/` file. The `jsm` flles are auto-generated.

viniciuslrangel:
Done

viniciuslrangel:
> You can do that with `traverse`.
> https://jsfiddle.net/26mhruqc/
> 
> ~However, I'm not sure removing child-objects is good or bad because `THREE.Object3D` doesn't `traverse`-removing.~ <- seems this [works well](https://jsfiddle.net/cL29x5dn/). Could you replace with `traverse` please?

Actually, this method would require checking if the object is a CSS3DObject and contains element property. Maybe using a custom event would be better. Don't you think?

examples/js/renderers/CSS3DRenderer.js
```diff
@@ -13,11 +13,15 @@ THREE.CSS3DObject = function ( element ) {
 
 	this.addEventListener( 'removed', function () {
 
-		if ( this.element.parentNode !== null ) {
+		this.traverse( function( object ) {
 
-			this.element.parentNode.removeChild( this.element );
+			if ( object.element !== null && object.element.parentNode !== null ) {
```
yomotsu: `object.element` could be `undefined` or other value. How about this?
```suggestion
			if ( object.element instanceof Element && object.element.parentNode !== null ) {
```

yomotsu: `element` won't be null and should be `instanceof Element`, because of https://github.com/mrdoob/three.js/blob/5bf9029a0416ddf739582f998c507c3bfc7602a4/examples/js/renderers/CSS3DRenderer.js#L12
Otherwise, it shows an error at initializing.

I think `object.element !== null` can be omitted

yomotsu: sorry, forget about above, `CSS3DObject` may contain `Object3D`.

maybe, 
```suggestion
			if ( object.element instanceof Element  && object.element.parentNode !== null ) {
```

viniciuslrangel: But the child can be just an Object3D e.g. a Group

yomotsu:
Thanks! LGTM👍

Mugen87 (approved):

Mugen87:
Updated JSM modules: a4e573604cf1694db8f0059798bc97653fd538a6

mrdoob:
Thanks!

PR #17797, opened 2019-10-23
EXRLoader: added load() method

WestLangley:
Improves `EXRLoader` by setting more reasonable texture parameters.

I have modified the relevant examples, too. We will just have to see how how well this works --  or if there are device issues.


WestLangley:
Related: https://github.com/mrdoob/three.js/pull/15026#issuecomment-501104760.

mrdoob:
Thanks!

PR #17811, opened 2019-10-25
Minimap demo

LeonYuanYao:
Added examples/jsm/misc/Minimap.js & examples/webgl_camera_minimap.html for demo purpose.

### DIFF
--- a/examples/js/misc/Minimap.js
+++ b/examples/js/misc/Minimap.js
@@ -1,0 +1,160 @@
+/**
+ * @author Yuan Yao https://github.com/LeonYuanYao
+ * 
+ * A Minimap viewing tool for three.js
+ *
+ * @param {THREE.WebGLRenderer} renderer
+ * @param {THREE.Scene} scene
+ * @param {THREE.Camera} mainCamera
+ * @param {Object} params
+ */
+
+THREE.Minimap = function(renderer, scene, mainCamera, params) {
+
+    this.renderer = renderer;
+    this.scene = scene;
+    this.object = mainCamera;
+    this.viewRange = params.viewRange !== undefined ? params.viewRange : 1000;
+    this.mapSize = params.mapSize !== undefined ? params.mapSize : 300;
+    this.heightOffset = params.heightOffset !== undefined ? params.heightOffset : 10;
+
+    var scope = this;
+    var clientWidth, clientHeight;
+    var updatePos, idcPos;
+
+    var viewport = new THREE.Vector4();
+    renderer.getCurrentViewport(viewport);
+
+    var up = mainCamera.up.clone(), lookat = new THREE.Vector3(), lastLookat = new THREE.Vector3();
+    getLookatVecProjected(mainCamera, lastLookat);
+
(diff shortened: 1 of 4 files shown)


Mugen87:
Can you please remove the changes to `three.module.js` and `WebGLRenderer.js`?

Mugen87:
For testing: http://raw.githack.com/LeonYuanYao/three.js/minimap-demo/examples/webgl_camera_minimap.html

WestLangley:
Nice demo. However, I think it is more appropriate as a forum showcase than as a three.js example.

The three.js examples demonstrate -- and serve as test cases of -- three.js features.

This is a three.js application.

LeonYuanYao:
> Can you please remove the changes to `three.module.js` and `WebGLRenderer.js`?
> 
> Normally you would create `js/misc/Minimap.js` instead and generate the module version via `node utils/modularize.js`. However, I remember that @mrdoob is okay when new files are only added as modules, see [#16920 (comment)](https://github.com/mrdoob/three.js/pull/16920#issuecomment-531446949).

Hi, Mugen87
I have restored the changes in the WebGLRenderer.js and re-build the three.js and three.module.js. It seems some '__proto__: null,' codes were added into the built files. Is that acceptable?

LeonYuanYao:
> Can you please remove the changes to `three.module.js` and `WebGLRenderer.js`?
> 
> Normally you would create `js/misc/Minimap.js` instead and generate the module version via `node utils/modularize.js`. However, I remember that @mrdoob is okay when new files are only added as modules, see [#16920 (comment)](https://github.com/mrdoob/three.js/pull/16920#issuecomment-531446949).

Hi, Mugen87
I have restored the changes in the WebGLRenderer.js and re-build the three.js and three.module.js. It seems some '__proto__: null,' codes were added into the built files. Is that acceptable?

LeonYuanYao:
> Nice demo. However, I think it is more appropriate as a forum showcase than as a three.js example.
> 
> The three.js examples demonstrate -- and serve as test cases of -- three.js features.
> 
> This is a three.js application.

Hi, WestLangley 
Thank you for the advice! Yes, maybe I should consider submitting this as a showcase. But I don't know how to do that yet. Can you please provide some instructions or maybe accept this as a showcase? Just want to give a demo and module codes for people who need 'minimap' functionality. =)

Mugen87:
> I have restored the changes in the WebGLRenderer.js and re-build the three.js and three.module.js. It seems some 'proto: null,' codes were added into the built files. Is that acceptable?

Not really. You should actually never commit build files.

> Can you please provide some instructions or maybe accept this as a showcase?

Showcases are shared at the official [forum](https://discourse.threejs.org/). You can login with your github account and then create a new topic in the category `Showcase`.

LeonYuanYao:
It's a clean commit now with Minimap.js, the moduled version and the demo html files only!

Mugen87:
I agree with @WestLangley. In the meanwhile, it's recommended to use the forum for such work since we try to avoid application specific code in the library as much as possible.

PR #17816, opened 2019-10-25
Allow TrackballControls to work in a static scene like OrbitControls does

mkhorton:
Adds `_this.update()` to `mousewheel()`, `mousemove()` and `touchmove()` events in `TrackballControls`.

From the documentation, TrackballControls and OrbitControls look like they're drop-in replacements for one another, they both have the same constructor, etc. However, unlike OrbitControls, TrackballControls does not work in a static scene with no animation loop present.

See this [Stack Overflow thread](https://stackoverflow.com/questions/28746354/trackballcontrols-change-events) for a discussion.

PR would allow TrackballControls to work in static scenes.

### DIFF
--- a/docs/examples/en/controls/TrackballControls.html
+++ b/docs/examples/en/controls/TrackballControls.html
@@ -16,6 +16,8 @@
 			<p>
 				[name] is similar to [page:OrbitControls]. However, it does not maintain a constant camera [page:Object3D.up up] vector.
 				That means if the camera orbits over the “north” and “south” poles, it does not flip to stay "right side up".
+				
+				If using [name] in a static scene, make sure to disable the damping effect by setting [page:.staticMoving staticMoving] to *true*.
 			</p>
 		</p>
 
--- a/examples/js/controls/TrackballControls.js
+++ b/examples/js/controls/TrackballControls.js
@@ -515,6 +515,8 @@ THREE.TrackballControls = function ( object, domElement ) {
 			_panEnd.copy( getMouseOnScreen( event.pageX, event.pageY ) );
 
 		}
+		
+		_this.update();
 
 	}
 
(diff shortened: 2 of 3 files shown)


Mugen87:
Current version: https://jsfiddle.net/6dagmj0c/
With your change: https://jsfiddle.net/dp1ry53v/2/

> PR would allow TrackballControls to work in static scenes.

Yes, it is necessary to set `TrackballControls.staticMoving` to `true`. Otherwise the interaction feels weird.

Can you please also update the JSM version by running `node utils/modularize.js`? In this way, it's easier to verify the existing examples via: http://raw.githack.com/mkhorton/three.js/patch-1/examples/index.html

mkhorton:
Thanks @Mugen87, I've run `modularize` and added a note to the TrackballControls docs about setting `TrackballControls.staticMoving` to `true` in static scenes.

Mugen87:
The issue with this change is that it makes the call of `TrackballControls.update()` obsolete when no damping is used. The official example `misc_controls_trackball` now works even if the update call is removed from the animation loop.

Hence, the documentation of `update()` should be changed to something similar like [OrbitControls.update()](https://threejs.org/docs/index.html#examples/en/controls/OrbitControls.update). Besides, it's necessary to check the examples and remove `controls.update()` when `TrackballControls. staticMoving` is set to `true`. Otherwise unnecessary updates are performed which is not good.

Mugen87:
I think it would be also good to use the new `change` event in `misc_controls_trackball` and then remove the animation loop.

Mugen87:
Let's revisit this change if it's clear what happens to `TrackballControls`, see #17842, #17857, #17858.

mkhorton:
Thanks for the update!

PR #17823, opened 2019-10-28
Add DeviceOrientationEvent.requestPermission for IOS 13

HypnosNova:
When using "deviceorientation" or "orientationchange" in iOS 13+ devices, we should call DeviceOrientationEvent.requestPermission first to accept the page using gyro.

See https://medium.com/flawless-app-stories/how-to-request-device-motion-and-orientation-permission-in-ios-13-74fc9d6cd140


Mugen87:
Can you please edit the [js](https://github.com/mrdoob/three.js/blob/dev/examples/js/controls/DeviceOrientationControls.js) version instead? `jsm` files are auto-generated via `node utils/modularize.js`.

I can't test this PR since I don't have an iPhone at hand. However, the changes look good.

Mugen87:
For testing: http://raw.githack.com/HypnosNova/three.js/ios/examples/misc_controls_deviceorientation.html

examples/js/controls/DeviceOrientationControls.js
```diff
@@ -61,8 +61,27 @@ THREE.DeviceOrientationControls = function ( object ) {
 
 		onScreenOrientationChangeEvent(); // run once on load
 
-		window.addEventListener( 'orientationchange', onScreenOrientationChangeEvent, false );
-		window.addEventListener( 'deviceorientation', onDeviceOrientationChangeEvent, false );
+		if ( typeof window.DeviceMotionEvent.requestPermission === 'function' ) {
```
Mugen87: In order to make the code more robust, write this statement like so:
```js
if ( window.DeviceMotionEvent !=== undefined && typeof window.DeviceMotionEvent.requestPermission === 'function' ) {
```
When using the example without `https`, your code introduces a runtime error otherwise.

Mugen87 (changes requested):

sirxemic:
Was this actually tested? By looking at the code it shouldn't work, since `DeviceMotionEvent.requestPermission` is not triggered in a user action (e.g. a button click and such). Just for completion's sake I also tested it on my phone, and it does not work.

HypnosNova:
> Was this actually tested? By looking at the code it shouldn't work, since `DeviceMotionEvent.requestPermission` is not triggered in a user action (e.g. a button click and such). Just for completion's sake I also tested it on my phone, and it does not work.

Actully the 'requestPermission' could be called automatically but I tried several iphones I found that if the page is in http or not be called manually, the response is 'denied' so I change the demo with a button when iOS 13+ device open the page. And we should test it in https.

Thanks for finding the problem.

HypnosNova:
I tested the link and the demo works fine now.
But one more think, we found that we should write the logic both in the demo and controls. 

Is it a good way to let the controls having a dom element with a click event to call 'requestPermission', so users don't need to do anything about that case.
or just romove the whole logic from controls and tell users to add logic in their project by themselves and in this way we just change the demo page instead of changing the controls.

examples/misc_controls_deviceorientation.html
```diff
@@ -13,13 +13,30 @@
 			photo by <a href="http://www.flickr.com/photos/jonragnarsson/2294472375/" target="_blank" rel="noopener">Jón Ragnarsson</a>.
 		</div>
 
+		<button id="deviceBtn">Open device orientation(Only in IOS 13+)</button>
```
Mugen87: Actually, it's no good approach to show optional UI elements by default and hide them via JavaScript. When you refresh the page, you can see right at the beginning how the button is displayed and then disappears. The location at the top left is not optimal, too. 

I suggest you develop the code similar to [webgl_materials_video](https://threejs.org/examples/webgl_materials_video) and always request a user interaction at the beginning (with the text `Using device orientation might require user interaction.`).

examples/misc_controls_deviceorientation.html
```diff
@@ -13,13 +13,30 @@
...
+			if ( window.DeviceOrientationEvent === undefined || typeof window.DeviceOrientationEvent.requestPermission !== 'function' ) {
+
+				document.body.removeChild( deviceBtn );
+
+			} else {
+
+				deviceBtn.addEventListener( 'click', () => {
+
+					document.body.removeChild( deviceBtn );
+					controls.connect();
```
Mugen87: `.connect()` is also called in the ctor of `DeviceOrientationControls`. TBH, it's a bit hacky calling it called twice.

Developing the example according to https://github.com/mrdoob/three.js/pull/17823/files#r341047206 should make this code obsolete.

Mugen87:
@sirxemic Can you please verify if this updated example works on iOS 13?

https://raw.githack.com/Mugen87/three.js/dev33/examples/misc_controls_deviceorientation.html

sciecode:
Works perfectly @Mugen87, both Safari and Chrome.

Mugen87:
Closing in favor of #17884

PR #17842, opened 2019-10-31
TrackballControls: Refactor API.

Mugen87:
This PR makes the API more similar to `OrbitControls`.


WestLangley:
The only significant difference between `OrbitControls` and `TrackballControls` should be the `update()` method.

Rather than changing the API at this point, I suggest freezing that file, and extending `OrbitControls` to accommodate `TrackballControls` -- just as we did with `MapControls`.

Mugen87:
In any event, the API of `TrackballControls` will change whatever we are going to do. Merging this PR or merge it with `OrbitControls`. 

Hence, I see no problem to merge this now since there will be a migration effort in any case.

PR #17847, opened 2019-11-01
3MFLoader: Add support for vertex colors.

Mugen87:
The loader can now handle `color groups` :tada:


WestLangley:
You are adding a new file `vertexcolors.3mf` file that does not appear to be used, but it does render when tested separately.

I suggest adding flat normals or `flatShading`, however.
 
Screen Shot 2019-11-01 at 11 46 24 AM

Mugen87:
Yes, I'm aware of that. We have several models which are not necessarily used in the examples but present for validation.

Mugen87:
> I suggest adding flat normals or flatShading, however.

Good point! Totally forgot setting the flag...

mrdoob:
Thanks!

PR #17849, opened 2019-11-01
3MFLoader: Added missing flatShading.

Mugen87:
see https://github.com/mrdoob/three.js/pull/17847#issuecomment-548840245


mrdoob:
Thanks!

PR #17857, opened 2019-11-03
TrackballControls: extends OrbitControls - API refactor

sciecode:
Based on @WestLangley's https://github.com/mrdoob/three.js/pull/17842#issuecomment-548659998

Similar to what was done to `MapControls` on https://github.com/mrdoob/three.js/pull/16961

____
**Notes:**

- Introduces `fixedUp` which controls whether or not to maintain `camera.up` direction on rotation. 

- Noticeable `rotationSpeed` changes, due to the different mouse coordinates used in `OrbitControls`.

- Implements @Mugen87's backwards compatibility changes on #17842

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -59,6 +59,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 	// Set to false to disable rotating
 	this.enableRotate = true;
 	this.rotateSpeed = 1.0;
+	this.fixedUp = true; // maintains camera.up constant
 
 	// Set to false to disable panning
 	this.enablePan = true;
@@ -131,11 +132,14 @@ THREE.OrbitControls = function ( object, domElement ) {
 	this.update = function () {
 
 		var offset = new THREE.Vector3();
+		var yUp =  new THREE.Vector3( 0, 1, 0 );
 
-		// so camera.up is the orbit axis
-		var quat = new THREE.Quaternion().setFromUnitVectors( object.up, new THREE.Vector3( 0, 1, 0 ) );
-		var quatInverse = quat.clone().inverse();
+		var quat = new THREE.Quaternion();
+		var quatInverse = new THREE.Quaternion();
+		var axisQuat = new THREE.Quaternion();
 
+		var lastUp = new THREE.Vector3();
+		var lastDir = new THREE.Vector3();
 		var lastPosition = new THREE.Vector3();
 		var lastQuaternion = new THREE.Quaternion();
 
@@ -144,6 +148,21 @@ THREE.OrbitControls = function ( object, domElement ) {
 			var position = scope.object.position;
 
 			offset.copy( position ).sub( scope.target );
+
+			if ( ! scope.fixedUp ) {
+
+				lastDir.copy( offset ).normalize();
+
+			}
+
+			if ( ! scope.object.up.equals( lastUp ) ) {
+
+				quat.setFromUnitVectors( scope.object.up, yUp );
+				quatInverse.copy( quat ).inverse();
+
+				lastUp.copy( scope.object.up );
+
+			}
 
 			// rotate offset to "y-axis-is-up" space
 			offset.applyQuaternion( quat );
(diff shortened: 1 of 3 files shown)


sciecode:
GitHack is acting up, will post example link as soon as it gets back up. 👍

WestLangley:
I posted #17858 just to show how I implemented it. It may, or may not, be a better implementation.

I'd only change `OrbitControls` js,  jsm, and one example file at this point, however. Don't remove the original `TrackballControls`. Why change 32 files when you can change 3 for now?

sciecode:
> I'd only change OrbitControls js, jsm, and one example file at this point

I will force-push a simpler PR, no problem.

> Don't remove the original `TrackballControls`.

We did a similar thing when extending `OrbitControls` for `MapControls`, how's this case different? 
Is it because `TrackballControls` is more broadly used? Given that we implement the backwards compatibilities changes proposed by @Mugen87 on his PR (which I did here), I don't see this being an issue.

sciecode:
As far as implementations go, I feel like both PRs are handling the `TrackballControls` behavior adequately. However, I can see two major differences:

This PR:
- Modifies `camera.up`.
- Works when user changes `camera.up` after controls instantiation (dev also doesn't support this, it's just an added bonus)

West PR:
- Doesn't modify `camera.up`.
- Fails to handle `autoRotate = true` option.
- Fails to handle `screenSpacePanning = false` option (correctly).

I don't know how impactful is keeping `camera.up` constant in the overall scheme, nor do I know if it's easy to adapt his PR to work with `autoRotate = true` and `screenSpacePanning = false`.

WestLangley:
>We did a similar thing when extending OrbitControls for MapControls, how's this case different?

The sensitivity and keyboard control are different from the original `TrackballControls`. I think we should keep the old code around for awhile since the new `TrackballControls` is different.

>Fails to handle autoRotate = true option.

IMO, auto rotate does not make sense with `TrackballControls` because there is no natural 'up direction'.

>Fails to handle screenSpacePanning = false option (correctly).

Since there is no natural 'up direction', only screen-space panning makes sense for `TrackballControls`, IMO.

sciecode:
> The sensitivity and keyboard control are different from the original TrackballControls. I think we should keep the old code around for awhile since the new TrackballControls is different.

Makes sense 👍 

[GitHack - Trackball Example](https://rawcdn.githack.com/sciecode/three.js/3a9c25798c56ef75d02cabddb7038dd858fd54b0/examples/misc_controls_trackball.html) ( currently down )

examples/js/controls/OrbitControls.js
```diff
@@ -1178,3 +1207,117 @@ THREE.MapControls = function ( object, domElement ) {
...
 THREE.MapControls.prototype.constructor = THREE.MapControls;
+
+// This set of controls performs orbiting, dollying (zooming), and panning.
+// Unlike OrbitControls, it rotates the "up" direction camera.up.
+//
+//    Orbit - left mouse / touch: one-finger move
+//    Zoom - middle mouse, or mousewheel / touch: two-finger spread or squish
+//    Pan - right mouse, or left mouse + ctrl/meta/shiftKey, or arrow keys / touch: two-finger move
+
+THREE.TrackballControls = function ( object, domElement ) {
```
Mugen87: I really think this code and the definition of `MapControls` should go in its own module file. The following code looks weird, IMO:
```js
import { TrackballControls } from './jsm/controls/OrbitControls.js';
```

Yes, I'm aware that the JS versions would have a dependency to `OrbitControls` but I think this is still better than having strange imports.

sciecode: Hmm, not sure actually. This is more of a preference / library consistency kind of deal, so I think that's up to you guys. I'll be happy to change if we decide on that course of action.

Mugen87: Sorry, I really don't understand your argumentation. We normally remove legacy/obsolete classes with very few exceptions. If we would keep all files, the repository would not be manageable anymore.

If users can't invest time in migration, it's okay to use one of the previous release. Hence, I don't get your goal with retaining the API of `TrackballControls`. It's good to align it without keeping the old stuff around.

WestLangley: >If we would keep all files, the repository would not be manageable anymore.

Come on, dude.

WestLangley: > I don't get your goal with retaining the API of TrackballControls. 

https://github.com/mrdoob/three.js/pull/17858#issuecomment-550055124

Mugen87:
@sciecode I think this can be closed now. See the discussion in #18483 for more information.

PR #12849, opened 2017-12-10
Bump JSZip to the latest version

takahirox:
Let's update `examples/js/libs/jszip.min.js` to the latest version.

Its API has changed, mainly from sync methods to async ones.
So I also needed to update `3MF/AMF/KMZLoader` and `Editor` where they use `JSZip`.

https://stuk.github.io/jszip/documentation/upgrade_guide.html

Advantage
- With async methods, zip (un)compression doesn't block

Disadvantage
- `jszip.min.js` size increases a bit, from 73KB to 100KB
- new `JSZip` requires `Promises`. (Update) With `JSZip.external.Promise`, non-Promise-support browser doesn't need to load extra `Promise` polyfill
- those loaders `parse` methods will be async from sync

I'm trying `GLTFLoader/Exporter` zip archive support. I wanna work with the latest `JSZip`.

PR #17864, opened 2019-11-04
Update 3MFLoader.js to use JSZip 3 (via jszip-sync)

silvainSayduck:


Mugen87:
Just for my understanding: What are the benefits of this change?

silvainSayduck:
The latest version of JSZip (3.x) crashes with the current implementation in 3MFLoader. The syntax has changed for almost everything, including the constructor, and the synchronous versions of the methods do not exist anymore.

Mugen87:
Unfortunately, this change breaks the official 3MF examples. They crash with the error:

> Uncaught TypeError: zip.sync is not a function

So it's necessary to upgrade [jszip.min.js](https://github.com/mrdoob/three.js/blob/dev/examples/js/libs/jszip.min.js), too. And all other code that depends on this lib (at least `AMFLoader`, `KMZLoader` and the example `webgl2_materials_texture2darray`).

silvainSayduck:
Sorry, I didn't see that other files where depending on JSZip. But yes, this PR is about updating JSZip. If it is not interesting or required, then it can be closed.

I just thought it might make sense to work with the latest version of JSZip which I am using in my project.

Mugen87:
> I just thought it might make sense to work with the latest version of JSZip which I am using in my project.

Definitely. The upgrade was tried once (see #12849) but it seems `zip.sync()` was not available at that time. Would be nice to have a consistent solution in the repo.

silvainSayduck:
Do you mean that this PR should update all other impacted files too? Do we have an exhaustive list of those files somewhere? Not sure when I'll have the time for this, but I might give it a try if I can.

Mugen87:
> Do you mean that this PR should update all other impacted files too?

I think we won't get this PR merged otherwise.

>  Do we have an exhaustive list of those files somewhere?

Nope, but you search in the repo where `jszip.min.js` is included. AFAIK, it should be `3MFLoader`, `AMFLoader`, `KMZLoader` and `webgl2_materials_texture2darray`. Besides, the editor has a dependency to `jszip.min.js`.

> Not sure when I'll have the time for this, but I might give it a try if I can.

I think there is no rush. We should also try to put `getZippedFileContentSync()` in a helper method so we don't copy the code over and over again.

BTW: I did not find any references to `jszip-sync` at the JZIP repo? Where does it come from?

silvainSayduck:
It's a fork, but seems maintained in parallel with the original JSZip. I put the link to the package in the warning message, but here it is again: https://www.npmjs.com/package/jszip-sync

Mugen87:
Um, too bad. I thought it's part of real `JSZip` project. Are there any other libs outside there that could replace `JSZip`?

silvainSayduck:
There probably is, but I haven't used any.

Mugen87:
I'm actually not happy with the amount of code that is necessary to work with JSZip async code path. And we also want to keep this pattern:
```js
const loader = new THREE.3MFLoader();
const model = loader.parse( data );
```
As long as there is no more elegant way to handle decompression, I think it's better to stay at the current version of JSZip.

silvainSayduck:
Hi,

Yes, I agree. Using the async approach would require a lot of code update and paradigm change, which is why I went with the JSZip sync fork. But if this is not good enough for THREE, then feel free to close the PR :)

PR #17901, opened 2019-11-09
VRMLLoader: Fix background color computation.

Mugen87:
The computation of the [background node](http://www.x-3-x.net/vrml/archive/annotatedVRML2/CH3.HTM#Background)'s color values is actually wrong since the respective interpolation in `paintFaces()` is broken.

This PR introduces a new implementation according to the VRML standard.


mrdoob:
Thanks!

PR #12321, opened 2017-10-03
AxisHelper renamed to AxesHelper

WestLangley:
...as proposed in https://github.com/mrdoob/three.js/pull/12316#issuecomment-333673832.

PR #17903, opened 2019-11-10
changes AxesHelper to AxisHelper

7imo:
Somehow AxesHelper does not work in practice, but AxisHelper does.


Mugen87:
> Somehow AxesHelper does not work in practice

https://jsfiddle.net/xug70jy3/

Maybe you are not using the latest version of `three.js`? `AxisHelper` was once renamed to `AxesHelper`, see #12321.

PR #17922, opened 2019-11-14
Examples: Added panoramic 6DOF example 🌐

juniorxsound:
**What?**
Add an example that shows how to use 360 panoramas with depth maps to apply vertex displacement and create 6 degrees of freedom 360 panoramic environment.

Desktop example 
6DoF

VR example
quest


mrdoob:
Very nice!

mrdoob:
Thanks!

takahirox:
Looking nice! BTW some code indentation seems broken. Do you fix in your side @mrdoob ?

mrdoob:
I'll have time to clean it up next week, unless someone does before me.

I'm also curious to test if the `ShaderMaterial` is needed, or if the effect can be also achieved with a `MeshBasicMaterial` and the `displacementMap` property.

juniorxsound:
@mrdoob @takahirox Thanks 🎉 I can spend some time cleaning up the example and investigating whether it could be done without the `ShaderMaterial`

mrdoob:
That'd be great!

juniorxsound:
After a quick test I realized this could totally be done with a `MeshStandardMaterial` and a negative displacement scale - will commit the simplified version! Thanks for the suggestion @mrdoob 🙏🏽

meshstandardmaterial

kanybekasanbekov:
Can anyone explain me how can view 360 panorama like this?

Mugen87:
@kanybekasanbekov Please use the forum for [help requests](https://discourse.threejs.org/).

mrdoob:
Actually, this probably should be done with `MeshBasicMaterial` too. But it doesn't support `displacementMap` 🤔

juniorxsound:
@mrdoob I can definitely take a look at adding `displacementMap`, `displacementScale` and `displacementBias` to `MeshBasicMaterial`, do you think it makes sense to add it?

mrdoob:
I was trying to think why it was only in `MeshStandardMaterial`, but I don't see why it couldn't be in `MeshBasicMaterial`, `MeshLambertMaterial` and `MeshPhongMaterial` too yeah.

PR #17929, opened 2019-11-14
TS: Clean up.

linbingquan:


Mugen87:
Already fixed here: #17928

Thanks anyway!

PR #17932, opened 2019-11-14
Add Object3D.onBeforeShadow

NataliaDSmirnova:
This PR is a result of a conversation about #14921.

I suggest closing the small gap in shadowmap building: some objects that use `onBeforeRender` to update their material uniforms in cases when it cannot be done in every other place suffers from an impossibility of update the uniforms for shadowmap building step.
So I've added `Object3D.onBeforeShadow` function and used it during shadowmap building. It works.

We use procedural sprites for rendering cylinders and spheres and I want them to cast shadows.
I've prepared an online example with three modes (you can change the mode at the script beginning): 
GEO - geometry cylinders
SPRITE - common sprites with textures
and Z_SPRITE representing our procedural sprites. 

Here you can find online example,  which uses your threejs lib https://jsfiddle.net/NataliaDSmirnova/pqeyft5o/
And here is the same example, which uses threejs library with suggested changes
https://jsfiddle.net/NataliaDSmirnova/e4awmk62/
As you can see the shadow appears!

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -128,6 +128,9 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 	onBeforeRender: function () {},
 	onAfterRender: function () {},
 
+	onBeforeShadow: function () {},
+	onAfterShadow: function () {},
+
 	applyMatrix: function ( matrix ) {
 
 		if ( this.matrixAutoUpdate ) this.updateMatrix();
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -129,7 +129,6 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 	onAfterRender: function () {},
 
 	onBeforeShadow: function () {},
-	onAfterShadow: function () {},
 
 	applyMatrix: function ( matrix ) {
 
--- a/src/renderers/webgl/WebGLShadowMap.js
+++ b/src/renderers/webgl/WebGLShadowMap.js
@@ -427,7 +427,7 @@ function WebGLShadowMap( _renderer, _objects, maxTextureSize ) {
 						if ( groupMaterial && groupMaterial.visible ) {
 
 							var depthMaterial = getDepthMaterial( object, groupMaterial, light, shadowCamera.near, shadowCamera.far, type );
-
+							object.onBeforeShadow( shadowCamera, geometry, depthMaterial );
 							_renderer.renderBufferDirect( shadowCamera, null, geometry, depthMaterial, object, group );
 
 						}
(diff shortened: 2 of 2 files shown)


Mugen87:
I'm fine with this change. But we should not forget to update the TS definition file and the docs. However, you can make a separate PR after this one is merged.

WestLangley:
@NataliaDSmirnova 

Please provide the addition info I requested in https://github.com/mrdoob/three.js/issues/14921#issuecomment-553709572.

Also, please explain why you think this is the best signature:

```js
object.onBeforeShadow( shadowCamera, geometry, depthMaterial );
```

gkjohnson:
I would expect that if `onBeforeShadow` is going to be added that `onAfterShadow` would be added, as well, and that both should follow the same signatures as their `onBeforeRender` and `onAfterRender` counterparts? Is there a reason to pass different arguments?

NataliaDSmirnova:
> I'm fine with this change. But we should not forget to update the TS definition file and the docs. However, you can make a separate PR after this one is merged.

@Mugen87  You're right. I'm not attentive enough. I'll file an additional PR with pleasure. 

> Please provide the addition info I requested in [#14921 (comment)](https://github.com/mrdoob/three.js/issues/14921#issuecomment-553709572).

Done. @WestLangley, please, see [my comment](https://github.com/mrdoob/three.js/issues/14921#issuecomment-554291130) 

> I would expect that if `onBeforeShadow` is going to be added that `onAfterShadow` would be added, as well, and that both should follow the same signatures as their `onBeforeRender` and `onAfterRender` counterparts? Is there a reason to pass different arguments?

@gkjohnson  I'm happy, that you ask about `onAfterShadow` addition. It hasn't been added because I don't use it in the examples. I just try to minimize the changes in threejs. If you read my first commit, you'll find it there 😃 

> Also, please explain why you think this is the best signature:
> 
> ```js
> object.onBeforeShadow( shadowCamera, geometry, depthMaterial );
> ```
@WestLangley, @gkjohnson I agree with your opinion about onBeforeShadow arguments. It'll be nicer if the functions have the same signature. It was a mistake to make the signature fitting only my needs. The only question is what should I set as a `scene` to `onBeforeShadow` function, there is just an `object`. `WebGLRenderer.renderObject` which calls `onBeforeRender` and `WebGLShadowMap.renderObject` which calls `onBeforeShadow` have slightly different context.

I'm ready to realize all your requests in the hope it helps to add the feature 😉

WestLangley:
@NataliaDSmirnova Thank you. I do not have a recommendation as to what the signature should be. I guess I'd have to see more use cases to determine that.

I'm OK with whatever @mrdoob wants to do regarding this PR.

NataliaDSmirnova:
@mrdoob It'll be great if you could comment on the PR. Have I got any chance to help it being accepted? 😃

Mugen87:
Closing in favor of #25933.

Alexander-103:
Уже чуть врубился!

PR #17935, opened 2019-11-15
packing.glsl: fix HalfToRGBA artifacts

munrocket:
This PR is fixing artifacts that appears on some iPhone's, Androids, iPads.
Related bug: https://github.com/mapbox/webgl-wind/issues/12

Here live examples where you can see how it works:
before fix: https://codepen.io/munrocket/pen/yLLQBxB
after fix: https://codepen.io/munrocket/pen/zYYMOeK

I am sure this solution can solve also this #9092 but I don't have robust solution for packDepthToRGBA right now.

### DIFF
--- a/src/renderers/shaders/ShaderChunk/packing.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/packing.glsl.js
@@ -39,6 +39,7 @@ vec4 encodeHalfRGBA ( vec2 v ) {
 }
 
 vec2 decodeHalfRGBA( vec4 v ) {
+	v = floor( v * 255.0 + 0.5 ) / 255.0;
 	return vec2( v.x + ( v.y / 255.0 ), v.z + ( v.w / 255.0 ) );
 }
 
--- a/src/renderers/shaders/ShaderChunk/packing.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/packing.glsl.js
@@ -25,20 +25,11 @@ float unpackRGBAToDepth( const in vec4 v ) {
 	return dot( v, UnpackFactors );
 }
 
-vec4 encodeHalfRGBA ( vec2 v ) {
-	vec4 encoded = vec4( 0.0 );
-	const vec2 offset = vec2( 1.0 / 255.0, 0.0 );
-
-	encoded.xy = vec2( v.x, fract( v.x * 255.0 ) );
-	encoded.xy = encoded.xy - ( encoded.yy * offset );
-
-	encoded.zw = vec2( v.y, fract( v.y * 255.0 ) );
-	encoded.zw = encoded.zw - ( encoded.ww * offset );
-
-	return encoded;
+vec4 packHalfToRGBA( vec2 v ) {
+	vec4 r = vec4( v.x, fract( v.x * 255.0 ), v.y, fract( v.y * 255.0 ));
+	return vec4( r.x - r.y / 255.0, r.y, r.z - r.w / 255.0, r.w);
 }
-
-vec2 decodeHalfRGBA( vec4 v ) {
+vec2 unpackHalfToRGBA( vec4 v ) {
 	v = floor( v * 255.0 + 0.5 ) / 255.0;
 	return vec2( v.x + ( v.y / 255.0 ), v.z + ( v.w / 255.0 ) );
 }
--- a/src/renderers/shaders/ShaderChunk/shadowmap_pars_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/shadowmap_pars_fragment.glsl.js
@@ -38,7 +38,7 @@ export default /* glsl */`
 
 	vec2 texture2DDistribution( sampler2D shadow, vec2 uv ) {
 
-		return decodeHalfRGBA( texture2D( shadow, uv ) );
+		return unpackHalfToRGBA( texture2D( shadow, uv ) );
 
 	}
 
--- a/src/renderers/shaders/ShaderLib/vsm_frag.glsl.js
+++ b/src/renderers/shaders/ShaderLib/vsm_frag.glsl.js
@@ -9,7 +9,7 @@ void main() {
 
   float mean = 0.0;
   float squared_mean = 0.0;
-  
+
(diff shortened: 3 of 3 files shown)


mrdoob:
@supereggbert I think you added that method? What do you think?

munrocket:
Here examples where this artifact more noticeable on a devices from target list:
before: https://codepen.io/munrocket/pen/yLLGaaX
after: https://codepen.io/munrocket/pen/eYYbdJG

Also naming of this functions doesn't make any sense in overall context. Since we already have functions with names `packNormalToRGB` and `packDepthToRGBA` in file `packing.glsl` I am suggest to rename `encodeHalfRGBA` to `packHalfToRGBA`. Are everybody ok with this renaming?

munrocket:
We also can create a version with “tidy” overflow, GPGPU examples will look the same but rendered pictures is differs.

WestLangley:
>Are everybody ok with this renaming?

I think the renaming is fine. (Sorry, I can't comment on the implementation.)

mrdoob:
Looks good to me!

mrdoob:
Thanks!

mrdoob:
Also, I'm seeing lines in the `webgl_shadowmap_vsm` examples (MacOS, Intel Iris Graphics 550)...

Screen Shot 2019-11-18 at 1 00 06 AM

munrocket:
>Also, I'm seeing lines in the webgl_shadowmap_vsm examples (MacOS, Intel Iris Graphics 550)...

It was before fix, or we solve problem on one devices and get another on Intel Iris?

supereggbert:
New function names make much more sense. The packing function looks to be the same but more concise. However, I'm also seeing the banding issue on vsm demo, it looks like precision issue cause by `v = floor( v * 255.0 + 0.5 ) / 255.0;` I think the initial issue was being caused by my silly mistake, using 255 to shift when it should have been 256. Everything looks good to me using:
```c
vec4 packHalfToRGBA( vec2 v ) {
	vec4 r = vec4( v.x, fract( v.x * 256.0 ), v.y, fract( v.y * 256.0 ));
	return vec4( r.x - r.y / 256.0, r.y, r.z - r.w / 256.0, r.w);
}
vec2 unpackHalfToRGBA( vec4 v ) {
	return vec2( v.x + ( v.y / 256.0 ), v.z + ( v.w / 256.0 ) );
}
```
I've not got any apple devices to test on though, @munrocket does that work for you?

supereggbert:
I've found the root cause of the issue, it's not precision related. The reason it's not working with VSM is because VSM is using linear interpolation; producing a result that fast and accurate "enough" to nicely filter the shadows and produce a much smoother result. However, it means the new unpack function cannot work as it assumes 8 bit float as input. There are a few options:
- We simply flip vsm shadows to nearest, but that does mean we loose much of the smoothing
- We flips to nearest and do "proper" interpolation in fragment shader but that will mean more samples and will be slower.
- We have two sets of packing functions one for GPGPU and one specifically for VSM.

munrocket:
Not understand how it can be possible, from artistic perspective it's ok but with math not stable. We don't have another interpolation without new samples? What if exist better fix, than simple 0.5 rounding? Anyway I am ok with old version of packing, after all GPGPU is not part of a library. Just wanted to share solution.

supereggbert:
@munrocket your solution is 100% right and the best way to do it. It's the VSM implementation that's off mathematically and technically wrong, but it isn't noticeable in the final result and it's cheaper then the alternatives.

munrocket:
Ok, I find solution, we needed to round before packing and use "tidy" version.
https://munrocket.github.io/three.js/examples/#webgl_shadowmap_vsm
https://codepen.io/munrocket/pen/GRRPzeO
This examples works on iPhone's and PC with VSM and GPGPU without any modifications.

mrdoob:
So what's the solution? 🤔

mrdoob:
@munrocket

> after: https://munrocket.github.io/three.js/examples/#webgl_shadowmap_vsm

This one looks good on my system 🤔

munrocket:
@mrdoob current status of a problem: we have one solution for VSM and another for GPGPU. In example earlier gpgpu become [broken](https://codepen.io/munrocket/pen/GRRPzeO). IMHO we need to revert back old VSM version, because it is part of a library. But I am not believe that we can’t find common solution that will fit to all examples and trying to check packing algorithms from papers like [this](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf).

Mugen87:
> IMHO we need to revert back old VSM version, because it is part of a library.

Can you please perform this revert? Unfortunately, we can't just revert a single PR since this one and 62a012d934351d0a796cbbba904a4e2d8d44aa83

munrocket:
So, I tried 3 different packing and for all of them iOS fix worked, but only for GPGPU and not for VSM. Here probably the simplest packing ever:
```c
vec4 pack2HalfToRGBA( vec2 v ) {
	v *= 255.;
	return vec4( floor( v.x ) / 255., fract( v.x ), floor( v.y ) / 255., fract( v.y ) );
}
vec2 unpack2HalfToRGBA( vec4 v ) {
	v.xz = floor( v.xz * 255. + 0.5 ) / 255.;    // fix for iOS that not fit to VSM example
	return vec2( v.x + v.y / 255., v.z + v.w / 255. );
}
```
"simple" packing demo: https://codepen.io/munrocket/pen/YzzBWMj

But VSM is so fragile and it become broken or partially fixed only by adding 5e-5 in one place.
```
//vsm_frag.glsl.js: 22
squared_mean += distribution.y * distribution.y + distribution.x * distribution.x + 0.00005;
```
partially fixed vsm: https://munrocket.github.io/three.js/examples/#webgl_shadowmap_vsm


Conclusion: I don't know what actually do this line `v.xz = floor( v.xz * 255. + 0.5 ) / 255.;` with high bits (rounding?), but It fixing GPGPU examples for some of devices like iOS and breaks VSM for all of them.

Mugen87:
We really should focus on the official example for now. The live link (https://munrocket.github.io/three.js/examples/#webgl_shadowmap_vsm) still shows unacceptable artifacts on my iMac.

So I think it's better to revert the changes in `dev` for now and then give it another go.

PR #17949, opened 2019-11-17
WebGLTextures: Avoid unnecessary texture uploads.

### LINKED ISSUE #5821: Texture.clone() reuse of images
I use the Texture.clone() function a lot in my code, because I need to set the UV coordinates differently on every model, and this slows down the initiation of the page considerably. If I don't clone the texture, then the scene renders instantly. So something is not optimal in the rendering pipeline.

It seems that the Texture.clone() function indeed creates a shallow clone that reuses the image, but then when data is sent to the graphics card these two textures are seen as having two separate images, sending the same image twice. Is this correct? Perhaps this is where the bottleneck is.

Coming from XNA and DirectX, where textures are not much more than containers for bitmap data, I wonder why Three.js also includes UV data in them, instead of specifying UV once the texture is applied to a mesh?

### LINKED ISSUE #17766: Introducing THREE.Image
Moved the discussion from #5728 to a separate issue.

The idea is to add `THREE.Image` as an additional class which is used by `THREE.Texture`. An instance of `THREE.Image` is directly mapped to an instance of `WebGLTexture` (the raw WebGL texture object). That means multiple textures can refer to the same instance of `THREE.Image` and not causing redundant texture uploads anymore (related issues #5728, #5821, #7223)

Ideally, `THREE.Image` will make the handling of image objects transparent (e.g. the difference in `HTMLImageElement`, `ImageBitmap` or the image stub in `DataTexture` is encapsulated in this new class).

Mugen87:
Fixed #17766
Fixed #5821

This is a draft PR that should demonstrate what kind of changes are necessary to avoid unnecessary texture uploads in `three.js`.

##### Background 

`three.js` is currently not able to detect shared images in textures. For example when you clone a texture multiple times, `three.js` is going to create an internal `WebGLTexture` object for each texture. And that means unnecessary processing and memory allocation.

##### The Current Solution

I've tried to approach this problem in different ways. This PR represents a change with the least impact on the code base. However, it requires a hack in the `THREE.Texture`. This hack clearly shows what kind of new logic is required to avoid multiple uploads.

The idea is to add a `version` property to `Texture.image`. That makes it possible to determine whether a texture upload is required or not. Besides, `WebGLTextures` only creates raw WebGL texture objects if necessary now. This is done by introducing a new weak map that maps WebGL texture objects to images. The problem is that a single image can have multiple raw WebGL textures because of different texture properties. Some of these properties are mostly static and closely coupled to the actual texture data (like `format`, `type` or `encoding`). Other parameters like `wrapS`/`wrapT` might be different from texture to texture. To accommodate this issue, `WebGLTextures` now computes a key via `getTextureCacheKey()` based on the texture properties and uses it to identify raw WebGL texture objects. It's similar to how `three.js` uses material parameters to reuse existing shader programs.

In any event, the enhancement of `Texture` is a hack and thus not a clean solution. However, introducing `THREE.Image` will be a breaking change because the semantics of `Texture.image` is going to change.

##### The Better Solution

I think I would rename `THREE.Image` to something like `THREE.TextureImage` in order to avoid name conflicts with the native `Image` ctor. The type of `Texture.image` would be changed to `THREE.TextureImage` _or_ a new property introduced (so it's easier to ensure backwards compatibility). Both cases will require refactoring in loaders and serialization/deserialization logic however hacks like this won't be necessary anymore:

https://github.com/mrdoob/three.js/blob/7680fb5f51ae4315ff789c788fef98f68d91413f/src/textures/Texture.js#L186

Note that the same type of hack (pollute objects with custom properties) is used to assign uuids to images. Introducing `THREE.TextureImage` would avoid this by wrapping the image into a separate object. Something like:

```js
function TextureImage( image ) {

    this.image = image;
    this.version = 0;

}

Object.assign( TextureImage.prototype, {

    toJSON( meta ) {

        // ...

    }

} );
```
I'm trying to experiment with this approach in the next time to better understand the impact on the code base and if it's possible to introduce `TextureImage` without breaking code. Fortunately, I can reuse code from this PR, especially the changes in `WebGLTextures` might be useful.


Mugen87:
Updated the PR by introducing `TextureImage` (c934e4de64d4cc31e16fa6b5f296bf4704be133c). The code does not pollute raw image objects anymore^^.

donmccurdy:
@Mugen87 Thank you for working on this!

You claim that your previous solution was a hack — do you see anything wrong with it other than pollution of the Image objects? I'm not opposed to adding a new TextureImage class, but if modifying Image properties was the only problem with the older solution (I don't see any other issues) I think there may be an option that could be hidden from the public API:

Instead of using `WeakMap<Image, WebglTexture>`, what about something like `WeakMap<Image, UploadedTextureCache>`, similar to:

```js
uploadedTextureCache = _images.get( image );

uploadedTextureCache[ textureCacheKey ] = {
  uuid: ...,
  version: ...,
  glTexture: ...
};

_images.set( image, uploadedTextureCache );
```

All else being equal, it might be nice to hide this implementation detail from the public API.

Mugen87:
> do you see anything wrong with it other than pollution of the Image objects?

A new class be the ideal place for the respective `toJSON()` logic so it would be definitely a more proper solution. The current implementation in `Texture` is a bit of a hack.

donmccurdy:
Could you express what makes it seem like a hack? BufferAttribute does not require a separate subclass to represent and serialize its `.array` property, for example. The additional class is a breaking change, and an increase in user-facing complexity, so I think it would be helpful to justify that more clearly.

Mugen87:
> The additional class is a breaking change,

Can you please explain why it's a breaking change? Since the `image` property is still available, user level code should not notice the changes of this PR.

donmccurdy:
Oops I see. Agreed, it's not a breaking change.

It's still a bit complex; the meanings of `Texture`, `TextureImage`, and `Image` classes are not self-explanatory given the names. Alternative names for `TextureImage` might improve that:

- GPUImage
- GLImage
- UploadedImage
- InternalImage

^Arguably the word "Image" could be replaced with "Texture" for any of those.

Or the `toJSON` code could be put in a utility function without requiring a wrapper class. ¯\_(ツ)_/¯

Mugen87:
That sounds everything very reasonable 😊. Let's see what @mrdoob and the others prefer from these options.

DavidPeicho:
It think the name `PixelBuffer` / `PixelData` makes a lot of sense. Textures use a `PixelBuffer` / `PixelData` that is uploaded on the GPU. The only bad thing is that it makes a lot of sense for the `DataTexture.data` field, but sounds weird for a HTML image element.

stevesan:
+1 to this! I encountered a similar problem when using GLTFLoader with KHR_texture_transform. Although there are other issues to be fixed there, having this PR would be crucial for a *proper* solution.

zeux:
Just noting that this is important to fix for KHR_mesh_quantization extension as well since it relies on KHR_texture_transform for providing dequantization data.

munrocket:
What is the status of this problem? I find that I can't easily optimize loader.

munrocket:
Thank you for working on this.

mrdoob:
Lets try to add this in the next cycle! 🤞

Mugen87:
Not sure why the CI still fails. E.g. it complains about `css3d_periodictable` although it looks okay on my local machine (besides it's unaffected by this change anyway).

munrocket:
Thank you for this PR again. We have another screenshots in dev repo and migrated to .jpg. Doesn’t know how the PR can be affected by this, but rebasing commit by hand on new dev branch probably will solve this.

munrocket:
On my PC all screenshot is good with 0 diff, also I am tried to recreate it. But CI [fails](https://github.com/munrocket/three.js/actions/runs/111358692).
Probably I need to download result from CI, because I have no idea what's going on.

src/textures/Texture.js
```diff
@@ -97,7 +112,7 @@ Texture.prototype = Object.assign( Object.create( EventDispatcher.prototype ), {
 
 		this.name = source.name;
 
-		this.image = source.image;
+		this.textureImage = source.textureImage;
 		this.mipmaps = source.mipmaps.slice( 0 );
```
munrocket, line 116: Rebuild a repo please, this PR seems wrong.

munrocket: Lack in deterministic Math.Random().

Mugen87: The color of the point light is random. Could this be a problem?

https://github.com/mrdoob/three.js/blob/2c73923c3ee412e048f544603dce04c73ebec44b/examples/webgl_loader_ttf.html#L74

Mugen87:
It seems all examples that fail during the CI use `Math.random()` in some extent.

munrocket:
This happens a second time after major changes in the core #19028.

I am suggesting to switch random in [here](https://github.com/mrdoob/three.js/blob/2c73923c3ee412e048f544603dce04c73ebec44b/src/math/MathUtils.js#L25) to deterministic one. We need to rethink how we use random functions, such PRs should not change lot of screenshots anymore.

WestLangley:
See #16655.

munrocket:
@WestLangley funny, you know about this :)

We need to update 92 screenshot with every big update in core without it.
```
npm run make-screenshot css3d_periodictable css3d_sandbox css3d_sprites misc_boxselection misc_controls_drag misc_controls_map misc_controls_orbit misc_controls_pointerlock misc_controls_trackball misc_lookat physics_ammo_rope physics_cannon_instancing svg_sandbox webgl2_buffergeometry_attributes_integer webgl2_multisampled_renderbuffers webgl2_sandbox webgl_buffergeometry webgl_buffergeometry_custom_attributes_particles webgl_buffergeometry_instancing webgl_buffergeometry_instancing_billboards webgl_buffergeometry_lines webgl_buffergeometry_points webgl_buffergeometry_points_interleaved webgl_buffergeometry_rawshader webgl_buffergeometry_uint webgl_camera_cinematic webgl_camera_logarithmicdepthbuffer webgl_clipping_intersection webgl_custom_attributes_points webgl_custom_attributes_points3 webgl_depth_texture webgl_effects_anaglyph webgl_effects_parallaxbarrier webgl_effects_peppersghost webgl_effects_stereo webgl_geometry_extrude_shapes webgl_geometry_hierarchy webgl_geometry_minecraft webgl_geometry_minecraft_ao webgl_geometry_terrain webgl_geometry_terrain_raycast webgl_geometry_text webgl_gpgpu_birds_gltf webgl_gpgpu_water webgl_instancing_modified webgl_interactive_buffergeometry webgl_interactive_cubes webgl_interactive_cubes_gpu webgl_interactive_cubes_ortho webgl_layers webgl_lensflares webgl_lights_pointlights2 webgl_loader_imagebitmap webgl_loader_ttf webgl_lod webgl_materials_cubemap_balls_reflection webgl_materials_cubemap_balls_refraction webgl_materials_shaders_fresnel webgl_math_obb webgl_modifier_tessellation webgl_multiple_elements webgl_nearestneighbour webgl_performance webgl_performance_doublesided webgl_performance_nodes webgl_performance_static webgl_points_billboards webgl_points_sprites webgl_postprocessing webgl_postprocessing_crossfade webgl_postprocessing_fxaa webgl_postprocessing_glitch webgl_postprocessing_nodes webgl_postprocessing_nodes_pass webgl_postprocessing_outline webgl_postprocessing_pixel webgl_postprocessing_rgb_halftone webgl_postprocessing_sao webgl_postprocessing_ssaa_unbiased webgl_postprocessing_ssao webgl_postprocessing_unreal_bloom_selective webgl_sandbox webgl_shaders_ocean2 webgl_shadowmap webgl_shadowmap_pcss webgl_shadowmap_performance webgl_simple_gi webgl_sprites webgl_test_memory webxr_vr_ballshooter webxr_vr_cubes webxr_vr_dragging
```

I maid a new PR that will fix this for all future commits #19411

sciecode:
@Mugen87 I was having a closer look at this PR and I noticed something that could potentially be improved in the future. It appears that the cache test for creating new `WebGLTextures` is taking some unnecessary parameters into account and forcing us to create new internal textures when we might not want to.

As far as I understand and what I've read on WebGL Insights, we want to avoid updating things like `format`, `type`, `width` and `height`, because it is more costly than creating a new texture. However, we might not want to create new textures for other parameters like `wrapS/wrapT`, `anisotropy`, etc..

If bind the same `WebGLTexture` on different slots, we might be able to adjust the just necessary "sampling" parameters, which might be something we want. I'm not sure if there's any hardware performance for re-uploading the same texture on different slots, we would need a vendor guy to confirm this for us, but either way we could potentially save significant memory usage by trying to optimize this.

Sounds like a reasonable thing to consider, what your thoughts?

makc:
@Mugen87 if you dont want to add prng into uuid, nor you want to remove random from the exampls, consider using [getRandomValues](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) in uuid instead. this is pretty good random generator that is built into the browser just like Math.random(), but you will need to test its prformance.

Mugen87:
In this particular case, I think I just update the screenshot. I don't think it's appropriate to change code just so the CI works...

makc:
but then the repo will get bigger and bigger every time thanks to the old pngs.

Mugen87:
TBH, I think it was no good idea to put the screenshots under version control.

DefinitelyMaybe:
yup

sciecode:
So, by now, I've reviewed this PR a couple of times, but I'm still struggling to understand what exactly that we are trying to solve with this.

> The problem is that a single image can have multiple raw WebGL textures because of different texture properties. Some of these properties are mostly static and closely coupled to the actual texture data (like format, type or encoding). Other parameters like wrapS/wrapT might be different from texture to texture.

As explained by @Mugen87, the original idea of mapping a single `THREE.Image` to a single `WebGLTexture` object is not applicable, as initially intended. 

The only real case where we would avoid texture uploads is when using `Texture.clone()` and not modifying any of the aforementioned parameters. We clone textures internally ( e.g `UniformsUtils.clone()`) sort of as a fail-safe to prevent users from modifying the original texture without realizing, but in most of these cases we really would want to just keep a reference to the original texture instead, since they are, by all means, the same thing.

@donmccurdy exposed the same opinion in the original issue https://github.com/mrdoob/three.js/issues/17766#issuecomment-543903167

I wonder if, instead of adding this extra complexity, it wouldn't be better to address the design decisions regarding the internal policy of cloning textures instead of just referencing them.

With that being said, if the actual idea is to be able to use different `uvTransform` for the same `WebGLTexture` object. I believe we should look for a different solution, that doesn't involve adding this much complexity to core.

zeux:
If we're merely interested in improving the behavior of GLTFLoader given texture transform extensions, one alternative option is to introduce explicit control over UV transform for both channels into Material. `.repeat`/`.scale` are a bit of a lie anyway because if their values are inconsistent between diffuse & normal map, the diffuse map transform is used to transform the normal map data.

WebGLRenderer could use the transform specified in the material, if present, and fall back to the current logic of using the .repeat/.scale values from the first matching texture.

GLTFLoader could then start using these values instead of the .repeat/.scale, thereby avoiding the texture cloning altogether (unless the same texture is used with different addressing modes, but that requires cloning of WebGL texture anyway).

donmccurdy:
>  If the actual idea is to be able to use different uvTransform for the same WebGLTexture object. I believe we should look for a different solution, that doesn't involve adding this much complexity to core.

Ability to use two UV transforms without uploading a texture twice is the important goal, yes. Other texture properties seem secondary, in my opinion.

> ...one alternative option is to introduce explicit control over UV transform for both channels into Material.

This is more like the proposal of https://github.com/mrdoob/three.js/pull/8278. That would be at least as useful, and add per-map UV slots too. But also seems more complex than this PR.

***

The simpler and more flexible option would still be NodeMaterial, but it doesn't seem like we've had resources to fully develop that recently. It would be nice to have the duplicate upload issue resolved sooner.

sciecode:
Just to be clear, I'm not against the PR, it's very well written and it works as intended.

I just felt like it would be good to discuss if this is still the solution we want to go for, after realizing that the initial idea of a 1-1 mapping between image/texture is not actually possible in WebGL1 ( it is in WebGL2 with sampler objects )

Mugen87:
Since #8278 was not merged after long considerations and node material usage in the core is still far away, this approach represents some sort of low-hanging fruit that we can have now. I've actually see similar approaches in other engines, too (like e.g. pixi.js). So the idea to decouple the engine's texture class from the actual native (WeBGL) texture objects and use _something_ in between. In this case, it's `THREE.ImageTexture`.

zeux:
@donmccurdy 
> This is more like the proposal of #8278. That would be at least as useful, and add per-map UV slots too. But also seems more complex than this PR.

That one is more complex because it adds full support for per-texture UV transforms as far as I can see. Exposing uvScale, uvOffset, uv2Scale and uv2Offset on Material should be a very small change.

This is also more consistent with how GPU actually works - UV transforms are *NOT* properties of textures! They are properties of materials, similar to any other uniforms.

I'm not opposed to this PR but I'd like to see the problem resolved. If this PR is going to get merged - great! :D I'm happy. If it's not, the material route is an option to consider.

mrdoob:
@Mugen87 Could you resolve the conflicts here when you have a chance?

Mugen87:
Yes. I'll put this on my list for tomorrow 👍

Mugen87:
Okay, I've cleaned up the PR and did a few more tests offline. Do you think we can give this a try at the beginning of `r121dev`?

manucorporat:
So happy to see this PR moving forward!

sciecode:
I ran some tests against this PR and it's looking good. Have you considered how we could extend on this logic for WebGL2 [WebGLSampler](https://developer.mozilla.org/en-US/docs/Web/API/WebGLSampler)?

I ask because, now that we are advancing development onto WebGL2 features. I find sampler objects, specifically, is one of the nicest features that we could benefit from. As mentioned previously, it would allow a 1-1 mapping we were looking for. However it would be better if we considered this when designing the WebGL1 fallback as well, so that we can easily extend later.

Again, I don't mean to block the PR. Just take the opportunity to consider how to move forward afterwards.

Mugen87:
I suggest we start discussing this as soon as this PR lands. I've planned to give this a try when this PR and some refactoring in `WebGLTexture`s is done.

Right now, there are a few important PRs in the pipeline waiting to be merged. Unfortunately, we can only focus on a limited amount of work and we also have to ensure that things stay stable. Especially the later point is already a big challenge and it's going to be even more complicated the more logic is added.

sciecode:
Sounds good to me 👍 

I had given a bit of thought on how we could incorporate both WebGL1 and WebGL2 APIs in a single and fail-safe implementation. I must say it is far from a trivial problem, so I just thought it would be good to ask.

Noobulater:
This seems to be getting pushed back, would really love to see this get implemented and willing to help get this one moving. What still needs to be done on this?

mrdoob:
More ideas on this subject...

Instead of adding a new class under `THREE.Texture`, we could add a new class on top of it.
We could introduce `THREE.Sampler` and move the relevant properties from `THREE.Texture`.

```js
const texture = new THREE.TextureLoader().load( 'foo.png' );
const material = new THREE.MeshBasicMaterial();
material.map = new THREE.Sampler( texture );
material.map.offset.x = 0.5;
```

One could still do `material.map = texture` but they would lose the transforms so things wouldn't completely break.

Another name option would be `THREE.TextureSampler`.

donmccurdy:
Adding a class on top might also fit better with the node material pattern (like how `UVTransformNode` works now).

Mugen87:
Closing. The PR is full of merge conflicts and the `THREE.Sampler` looks so promising that it's worth giving it a try. If it turned out the approach is problematic for some reasons, we can still introduce `THREE.TextureImage` to fix the issue.

mrdoob:
@Mugen87 Would you like to give it a go at implementing `THREE.Sampler`?

Mugen87:
Yes 😊 . Already thinking about it^^. It's actually more complex to introduce `Sampler` above `Texture` than `TextureImage` (which only holds the image property) below `Texture`.

mrdoob:
Seeing that the `THREE.Sampler` idea is way more complicated than the approach here...

How about renaming `THREE.TextureImage` to `THREE.ImageSource` (which could be followed by `THREE.DataSource`, `THREE.ImageArraySource`, ...)?

We can then introduce a new `source` property to `THREE.Texture` to transition from the legacy approach.

mrdoob:
Or, to keep things more open it could just be `THREE.Source` with `uuid` and `data` as properties:

```js
const image = new THREE.ImageLoader().load( 'foo.png' );
const source = new THREE.Source( image ); // source.data = image

const material = new THREE.MeshBasicMaterial();
material.map = new THREE.Texture( source );
material.map.offset.x = 0.5;
```

Sorry for the back and forth 😅

Mugen87:
Let me create a new PR with this approach.

PR #17953, opened 2019-11-18
fix: use outline pass to select line

AmyFoxFN:


examples/js/postprocessing/OutlinePass.js
```diff
@@ -158,7 +158,7 @@ THREE.OutlinePass.prototype = Object.assign( Object.create( THREE.Pass.prototype
 
 		function gatherSelectedMeshesCallBack( object ) {
 
-			if ( object.isMesh ) {
```
Mugen87: I think it's better to inline the check. A separate function is not necessary.

Besides, id does not make sense to push lines to an array called `selectedMeshes`. Also the function names `gatherSelectedMeshesCallBack()` does not work anymore since it also handles lines.

BTW: Can you please show how the outline pass looks like with lines? Maybe with a live example (or at least a screenshot)?

Mugen87:
Closing for now. The PR can be reopen when the requested information are provided.

PR #17957, opened 2019-11-18
Alternative examples page comprised of thumbnails

callumprentice:
As per issues discussion here https://github.com/mrdoob/three.js/issues/17692, I am submitting a PR for an alternative version of the examples index page that displays an autogenerated thumbnail for each.

The thumbnails are included in the PR but it is expected that a new set will be created for each release using the Google Puppeteer script provided in the `utils/examples folder`. There is a README file in the same folder with some context, usage info and suggestions for the future.

Cheers!


looeee:
[Live index_thumbs.html for testing](https://raw.githack.com/callumprentice/three.js/dev/examples/index_thumbs.html)

looeee:
@callumprentice were all of the thumbs autogenerated using the script?

Also, `thumbs/webgl_loader_assimp2json.jpg` seems to be missing.

callumprentice:
@looeee apologies for the delay replying - we were traveling for the holidays

> @callumprentice were all of the thumbs autogenerated using the script?

Yep they were - I wonder if it's worth considering taking a set of ones that look "right" and then autogenerating just new ones as they're added - without a signal from the app itself, it's hard to know exactly when to take the snapshot and get a perfect shot.

> Also, `thumbs/webgl_loader_assimp2json.jpg` seems to be missing.

I noticed that too - plus one other IIRC.. I think it's because the version of files.js I am using is a bit out of date - I'll make it work with the one checked into the repo vs local.

looeee:
> Yep they were - I wonder if it's worth considering taking a set of ones that look "right" and then autogenerating just new ones as they're added 

Well, I think we should do whatever is least effort and looks good enough. Ideally this would be a script that can be run in a few minutes before each release without any further intervention. 

That way we don't need to track which examples have been changed in a given release.

Actually I asked this because I was surprised how good the thumbnails look. Admittedly I didn't examine each one in detail but overall they look great.

callumprentice:
> Well, I think we should do whatever is least effort and looks good enough. Ideally this would be a script that can be run in a few minutes before each release without any further intervention.
> 
> That way we don't need to track which examples have been changed in a given release.

Understood - I hadn't consider that some examples might just change vs be new.
 
I'll make sure the `files.js` being used correctly.

callumprentice:
I pushed out some changes I made over the holidays - latest version is here: https://github.com/callumprentice/three.js - do I need to make a new PR or will GitHub take care of it?

I got the latest version of files.js and use that now to capture the thumbnails.  I also changed the script (`./utils/examples/generate_thumbnails.js`) that generates thumbnails to be more efficient and easier to work out.  I used it to create a new set of thumbnails in the `./examples./thumbs/` folder. Finally, I pushed out the new set of thumbnails to my test site here https://3ve.surge.sh/examples/index_thumbs.html

The only issue I know about is that the script loads each example using the URL to the main three.js example site but some of the newer examples are only in the dev branch - "instancing / modified" and "instancing / performance" for example - so they 404.  Not sure how to fix that without a 2 pass approach (deploy dev to main then recapture the thumbnails) - or perhaps there is a dev version of the examples HTML page? 

I am going to be heading off to the UK at some point - my mum passed away unexpectedly a couple of days ago and I need to go out there and make the usual arrangements.  Not sure when I'll be going or for how long but I might be less responsive here for a bit.

munrocket:
@callumprentice this PR is awesome! We now have official screenshot pack in threejs repo here  `/examples/screenshots`. How to make new main index.html with thumbnails?

@Mugen87 @looeee Should we leave a scrollbar? Should we upload thumbnails somewhere in order to fast page loading and/or increase their sizes? Can you update index.html page generator according to exist screenshot pack?

callumprentice:
@munrocket It was a trivial change to make the index_thumbs.html page reference the `screenshots` folder vs my thumbs `folder`.

The change is reflected in https://github.com/mrdoob/three.js/pull/18714

How are the screenshots generated?  The script that is also part of the PR generates a folder of thumbnails automatically - looks like maybe the official screenshots one is done by hand?

munrocket:
No, they generated with puppeteer too, but with much complex one, because I cannot simplify it without loosing coverage or robustness in e2e tests. You can review it here #18574. Resolution is 800x600 and in jpg, probably too much and with wrong format for web purposes.

Anyway this PR is great and thumbnails is much better than blank page. We need it, because we already have ~360 examples. Hard to navigate through it right now.

looeee:
> Should we leave a scrollbar?

Yes, I'd say so. 

> Should we upload thumbnails somewhere in order to fast page loading and/or increase their sizes?

Do you mean like a CDN? I'd say there's no need, the images are already served from github's servers. Unless we end up with thousands of screenshot stored in the repo. How many screenshots per example does your regression test add? 

> I am going to be heading off to the UK at some point - my mum passed away unexpectedly a couple of days ago and I need to go out there and make the usual arrangements. Not sure when I'll be going or for how long but I might be less responsive here for a bit.

Sorry to hear that @callumprentice. Hope you're doing OK.

callumprentice:
> Sorry to hear that @callumprentice. Hope you're doing OK.

@looeee - thank you and yes, all good.

munrocket:
>How many screenshots per example does your regression test add?

All of them minus 4, they in exception list. Probably having 2 puppeteer scripts is even better, you will make thumbnails with proper sizes and formats (jpg is much better for web and png for per pixel matching, no need to reuse it).

>Do you mean like a CDN? I'd say there's no need, the images are already served from github's servers.

Right now thumbnails little bit laggy, they not on github. By the way main page in threejs is super fast.

>should we leave a scrollbar?

Asked a wrong question, I mean should we leave sidebar? Or should we leave some navigation when we click on example? I usually click on all of them one by one, right now I need to return back on thumbnails page in current design. Or maybe some another modifications, that you will like to see in order to make it as a default page.

looeee:
> main page in threejs is super fast.

It's a [pages.github.com](https://pages.github.com/) site so it's all served through github's servers. They are very fast. Same with the docs and examples page. However the test on sites like https://3ve.surge.sh or raw.githack.com aren't, so it's probably not fair to compare speed yet.  

[Here's a githack live version](https://raw.githack.com/mrdoob/three.js/9fa5ecbc596a5fce8d98a5b53c2e1a28f432fe33/examples/index_thumbs.html), seems like it's a bit faster than surge.sh.

munrocket:
So cool! It's just one big `media` tag. On 1920x1080 thumbnails triggering to mobile version.

munrocket:
Also we can add this in script and VR examples will be not blank and datgui and stats.js will be invisible.
```
await page.evaluate( async () => {
  let button = document.getElementById( 'startButton' );
  if ( button ) {

    button.click();

  }

  let style = document.createElement( 'style' );
  style.type = 'text/css';
  style.innerHTML = `#info, button, input, body > div.dg.ac, body > div.lbl { display: none !important; }`;
  let head = document.getElementsByTagName( 'head' );
  if ( head.length > 0 ) {

    head[ 0 ].appendChild( style );

  }

  let canvas = document.getElementsByTagName( 'canvas' );
  for ( let i = 0; i < canvas.length; ++ i ) {

    if ( canvas[i].height === 48 ) {

      canvas[i].style.display = 'none';

    }

  }
} );
```

Mugen87:
Closing in favor of #18714.

PR #17969, opened 2019-11-20
added support for legacyJSONLoader in the Editor

atul-mourya:
It was a missing loader in the index.html
It would be nice to have the backward compatibility issue resolved in this regards.
Thanks

### DIFF
--- a/editor/index.html
+++ b/editor/index.html
@@ -38,6 +38,7 @@
 		<script src="../examples/js/loaders/TDSLoader.js"></script>
 		<script src="../examples/js/loaders/VRMLLoader.js"></script>
 		<script src="../examples/js/loaders/VTKLoader.js"></script>
+		<script src="../examples/js/loaders/LegacyJSONLoader.js"></script>
 
 		<script src="../examples/js/exporters/ColladaExporter.js"></script>
 		<script src="../examples/js/exporters/GLTFExporter.js"></script>


editor/index.html
```diff
@@ -38,6 +38,7 @@
 		<script src="../examples/js/loaders/TDSLoader.js"></script>
 		<script src="../examples/js/loaders/VRMLLoader.js"></script>
 		<script src="../examples/js/loaders/VTKLoader.js"></script>
+		<script src="../examples/js/loaders/LegacyJSONLoader.js"></script>
```
Mugen87: This is not the correct path. It should be:
```js
<script src="../examples/js/loaders/deprecated/LegacyJSONLoader.js"></script>
```

Mugen87:
In any event, we should consider to completely remove `LegacyGLTFLoader` and `LegacyJSONLoader` from the repo. If user still need them, they can checkout a previous release at any time.

editor/index.html
atul-mourya: Oh yes!! My bad. Thank you :)

atul-mourya:
I agree. But I'm afraid, it still might be too early for the people who have adopted this.
There isn't any alternative way as of now to test out 3d models imported from old JSON format.

donmccurdy:
> we should consider to completely remove LegacyGLTFLoader...

I think we could do this any time.

> There isn't any alternative way as of now to test out 3d models imported from old JSON format exported from blender.

Note that there are never going to be any more ways to test the old JSON format. That Blender exporter has been deleted, and we encourage using other formats with better interoperability. There's an open PR (https://github.com/mrdoob/three.js/pull/15552) providing a converter from legacy JSON to glTF, if that's helpful.

I'm not sure adding support for the legacy JSON format in new places is helpful to users in the long run.

PR #17975, opened 2019-11-21
Update .d.ts for BufferAttributes

0b5vr:

- [BufferAttribute](https://github.com/FMS-Cat/three.js/blob/dev/src/core/BufferAttribute.js)
- [InterleavedBufferAttribute](https://github.com/FMS-Cat/three.js/blob/dev/src/core/InterleavedBufferAttribute.js)

I have no knowledge about the InterleavedBufferAttribute though...


Mugen87 (approved):

mrdoob:
Thanks!

Issue #17974, opened 2019-11-21
toJSON doesn't work on CompressedTextures


jespertheend:
##### Description of the problem

When calling `toJson()` on an object that contains a CompressedTexture, an error gets thrown:
```
Uncaught TypeError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The provided value is not of type '(CSSImageValue or HTMLImageElement or SVGImageElement or HTMLVideoElement or HTMLCanvasElement or ImageBitmap or OffscreenCanvas)'
    at Object.getDataURL (js.js?v=000:4184)
    at CompressedTexture.toJSON (js.js?v=000:4398)
    at MeshStandardMaterial.toJSON (js.js?v=000:10171)
    at serialize (js.js?v=000:7295)
    at Mesh.toJSON (js.js?v=000:7349)
    at Object3D.toJSON (js.js?v=000:7363)
    at Scene.toJSON (js.js?v=000:7363)
    at Scene.toJSON (js.js?v=000:7507)
```
My specific use case is that I load textures in the basis universal format, and I call `toJSON()` to export my entire scene to a file in order to debug it in the threejs editor.

I don't think it's easy to convert a compressed texture to a data url. But I'll try to make a pull request that at leasts gives a warning and exports it without the texture. Instead of throwing an error.

##### Three.js version

r110

##### Browser

Chrome

##### OS

MacOS

Mugen87:
I guess #17745 should already implement a better error handling. It would be great if you can give it  a try 👍

jespertheend:
Hmm it seems like the issue is a bit more complex than I thought, but I think I can make this work. Do you know by any chance if there are situations where it is actually possible to serialize CompressedTextures? Or would it be ok to just generate a warning whenever you try to serialize any CompressedTexture? I don't want to break existing functionality.

Mugen87:
> Do you know by any chance if there are situations where it is actually possible to serialize CompressedTextures?

I'm not aware of anything like that. That would mean you have to decode the compressed texture on JS side and then transform it into something that can be serialize via `getDataURL()`. Better to log a warning.

donmccurdy:
> Do you know by any chance if there are situations where it is actually possible to serialize CompressedTextures?

[...]

PR #17978, opened 2019-11-21
log warning when trying to serialize compressed texture

jespertheend:


Mugen87:
Merging #17745 should also solve #17974

jespertheend:
@Mugen87 oh hmm I thought #17745 would only fix it for DataTextures

Mugen87:
The following branch should also be executed for compressed textures:

https://github.com/mrdoob/three.js/pull/17745/files#diff-955123916e8856f9a39f789ef5bb406dR363-R368

jespertheend:
Ah I see that makes sense now.

Mugen87:
Closing in favor of #17745. It's the more holistic approach.

PR #17980, opened 2019-11-22
Update typedef for WebGLGeometries and others

0b5vr:

- [WebGLGeometries](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLGeometries.js)
- [WebGLAttributes](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLAttributes.js)
- [WebGLInfo](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webgl/WebGLInfo.js)

This is just my learning process of internal architecture of Three.js...


mrdoob:
Thanks!

Issue #17970, opened 2019-11-20
DRACOLoader: exposing _initDecoder as public method


TyLindberg:
Previously one could call `THREE.DRACOLoader.getDecoderModule()` to preload the DRACO libraries, but there is no respective public instance method on the current DRACOLoader module. However, the private method `_initDecoder` seems to accomplish preloading. Would it be possible to make this method public or make a public wrapper for it?

https://github.com/mrdoob/three.js/blob/219e3c1bc24163e2c31ab6f5cf4bb66f81b689bc/examples/jsm/loaders/DRACOLoader.js#L230

donmccurdy:
I think that would be a good addition, yes. The method can just be renamed. 

I'm having a hard time deciding what to name it. Using "init" or "initialize" sounds like it should only be called once, when it's actually called each time `load` is called. Maybe just `.preload()`?

TyLindberg:
Yeah, `.preload()` seems like a good name to me. Want me to tackle changing that and opening up a PR or are you planning to make that change?

donmccurdy:
Please do! :)

TyLindberg:
Implemented in #17983.

donmccurdy:
Thanks @TyLindberg!

PR #17983, opened 2019-11-22
DRACOLoader: Add preload method

TyLindberg:
Addresses #17970.

This PR adds the public method `preload` to DRACOLoader (previously a private method `_initDecoder`). This allows the DRACO libraries to be loaded before loading any assets, similar to the behavior achieved in past releases using `DRACOLoader.getDecoderModule`.

I took the documentation from the old documentation for `getDecoderModule`, but there may be a better description we could use.

### DIFF
--- a/examples/js/loaders/DRACOLoader.js
+++ b/examples/js/loaders/DRACOLoader.js
@@ -220,7 +220,7 @@ THREE.DRACOLoader.prototype = Object.assign( Object.create( THREE.Loader.prototy
 
 	},
 
-	_initDecoder: function () {
+	preload: function () {
 
 		if ( this.decoderPending ) return this.decoderPending;
 
@@ -269,7 +269,7 @@ THREE.DRACOLoader.prototype = Object.assign( Object.create( THREE.Loader.prototy
 
 	_getWorker: function ( taskID, taskCost ) {
 
-		return this._initDecoder().then( () => {
+		return this.preload().then( () => {
 
 			if ( this.workerPool.length < this.workerLimit ) {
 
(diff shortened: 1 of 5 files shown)


examples/js/loaders/DRACOLoader.js
```diff
@@ -220,7 +220,7 @@ THREE.DRACOLoader.prototype = Object.assign( Object.create( THREE.Loader.prototy
 
 	},
 
-	_initDecoder: function () {
+	preload: function () {
```
donmccurdy: Maybe it would be better to have a second method: public `preload() : this` and private `_initDecoder() : Promise`. No other public APIs in three.js return Promises, so this would be a bit unusual otherwise. I don't think will users actually need to know when the preloading is complete.

TyLindberg: Yeah that seems pretty reasonable to me. Previously `getDecoderModule` returned a Promise so that's why I carried it over here, but keeping consistent with the rest of three.js is a smart move.

As for whether or not users would want to know when preloading has completed, I imagine the majority of use cases wouldn't need it. If we wanted to preserve that functionality we could add an optional onLoad callback parameter, but it may be better to wait until developers request it before adding such a feature.

TyLindberg: Just updated to include those changes

donmccurdy (approved):

examples/js/loaders/DRACOLoader.js
donmccurdy: Thanks!

mrdoob:
Thanks!

PR #17987, opened 2019-11-24
OutlineEffect: Refactor shaders.

Mugen87:
See https://github.com/mrdoob/three.js/issues/15468#issuecomment-557850638

This is an alternative implementation of `OutlineEffect`. Instead of injecting code in the vertex shader, it uses an own custom vertex shader.

In this way, the vertex shader only contains code which is actually relevant for the respective fragment shader. The mentioned warnings of the above issue are gone. I personally think this is the more robust approach which avoids unexpected side effects.

### DIFF
--- a/examples/js/effects/OutlineEffect.js
+++ b/examples/js/effects/OutlineEffect.js
@@ -54,9 +54,6 @@
  * 	visible: true,
  * 	keepAlive: true
  * };
- *
- * TODO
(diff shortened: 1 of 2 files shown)


mrdoob:
Thaaaaanks!

WestLangley:
@Mugen87 This is great. Thanks!

Issue #18009, opened 2019-11-26
THREE.Line2 and THREE.LineSegments2 Cannot be cropped by clippingPlanes;


389652405:
THREE.Line2 and THREE.LineSegments2 Cannot be cropped by clippingPlanes;
http://drm.gitee.io/clippaln/webgl_clipping.html

Mugen87:
Try setting `material.clipping` of the line material to `true`.

WestLangley:
Fat lines support clipping -- both local and global. Set `material.clipping = true;`

>It should be okay if LineMaterial sets this property always to true.

`ShaderMaterial` requires that `.clipping` be set to `true` to use user-defined clipping planes. `LineMaterial` extends `ShaderMaterial`.

If Fat lines are promoted into the library as a built-in material, then we can change the behavior.

Mugen87:
I just wanna say that there is no problem if the default value would be `true`, even if no clipping planes are used. In this case, @389652405 would never run into this issue.

WestLangley:
I prefer to adhere to the convention for `ShaderMaterial`, but we can make this change if you want. 

It just means users need to set `.clipping`  to `false` if they don't want the lines clipped by the clipping planes.

389652405:
Thank you very much.

PR #18011, opened 2019-11-26
FatLines: set .clipping to true

WestLangley:
As proposed in https://github.com/mrdoob/three.js/issues/18009#issuecomment-558667349.


mrdoob:
Thanks!

PR #18013, opened 2019-11-27
Update GLTFLoader.d.ts

0b5vr:

- [GLTFLoader](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/loaders/GLTFLoader.js)

Tried not to be exhaustive.

### DIFF
--- a/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/examples/jsm/loaders/GLTFLoader.d.ts
@@ -42,12 +42,6 @@ export class GLTFLoader extends Loader {
 export class GLTFParser {
 
 	json: any;
-	extensions: any;
-	options: {
-		path: string,
-		corssOrigin: string,
-		manager: LoadingManager
-	};
 
 	getDependency: (type: string, index: number) => Promise<any>;
 	getDependencies: (type: string) => Promise<any[]>;
--- a/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/examples/jsm/loaders/GLTFLoader.d.ts
@@ -14,7 +14,14 @@ export interface GLTF {
 	scene: Scene;
 	scenes: Scene[];
 	cameras: Camera[];
-	asset: any;
+	asset: {
+		copyright?: string;
+		generator?: string;
+		version?: string;
+		minVersion?: string;
+		extensions?: any;
+		extras?: any;
+	};
 	parser: GLTFParser;
 	userData: any;
 }
--- a/examples/jsm/loaders/GLTFLoader.d.ts
+++ b/examples/jsm/loaders/GLTFLoader.d.ts
@@ -14,7 +14,9 @@ export interface GLTF {
 	scene: Scene;
 	scenes: Scene[];
 	cameras: Camera[];
-	asset: object;
+	asset: any;
+	parser: GLTFParser;
+	userData: any;
 }
 
 export class GLTFLoader extends Loader {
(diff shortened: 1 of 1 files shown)


examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -14,7 +14,9 @@ export interface GLTF {
 	scene: Scene;
 	scenes: Scene[];
 	cameras: Camera[];
-	asset: object;
+	asset: any;
```
FMS-Cat: We might want to make it:

```suggestion
	asset: {
		copyright?: string;
		generator?: string;
		version?: string;
		minVersion?: string;
		extensions?: any;
		extras?: any;
	};
```

Note that there is no description about GLTF schema in anywhere in this repository though

examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -14,7 +14,9 @@ export interface GLTF {
...
+	parser: GLTFParser;
```
FMS-Cat: Is the `parser` intended to be exposed?

examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -29,3 +31,19 @@ export class GLTFLoader extends Loader {
...
+	extensions: any;
+	options: {
+		path: string,
+		corssOrigin: string,
+		manager: LoadingManager
+	};
+
+	parse: ( onLoad: ( result: GLTF ) => void, onError: ( reason: any ) => void ) => void;
+	getDependency: (type: string, index: number) => Promise<any>;
+	getDependencies: (type: string) => Promise<any[]>;
```
FMS-Cat: I picked members that I assumed that is intended to be exposed. might be more or less

examples/jsm/loaders/GLTFLoader.d.ts
donmccurdy: Looks good to me!

examples/jsm/loaders/GLTFLoader.d.ts
donmccurdy: Yes – it's useful enough to include here, I think.

examples/jsm/loaders/GLTFLoader.d.ts
```diff
@@ -29,3 +31,19 @@ export class GLTFLoader extends Loader {
...
+
+export class GLTFParser {
+
+	json: any;
+	extensions: any;
+	options: {
+		path: string,
+		corssOrigin: string,
+		manager: LoadingManager
+	};
```
donmccurdy: I think I would omit the `extensions` and `options` properties; those aren't really meant to be used externally.

examples/jsm/loaders/GLTFLoader.d.ts
donmccurdy: I don't think users should need to call `parse()` directly; at least in the current API parsing will already be done by the time they receive the GLTFParser. The other two methods are useful, though.

FMS-Cat: sure, I'm going to drop the `parse`

examples/jsm/loaders/GLTFLoader.d.ts
FMS-Cat: sure!

FMS-Cat:
Addressed to all the recommendations by @donmccurdy -san so far 🙌

donmccurdy (approved):

mrdoob:
Thanks!

PR #18016, opened 2019-11-27
WebGLAttributes: Add uploadCallback when attribute gets updated

Yxwww:
Ensure `.onUploadCallback` gets called when attribute gets updated. 
Context: https://discourse.threejs.org/t/how-to-dispose-bufferattribute-array-on-updates/11044/2

### DIFF
--- a/src/renderers/webgl/WebGLAttributes.js
+++ b/src/renderers/webgl/WebGLAttributes.js
@@ -70,6 +70,8 @@ function WebGLAttributes( gl ) {
 
 		gl.bindBuffer( bufferType, buffer );
 
+		attribute.onUploadCallback();
+
 		if ( updateRange.count === - 1 ) {
 
 			// Not using update ranges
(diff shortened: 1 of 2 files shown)


Mugen87 (approved):

Mugen87:
@Yxwww Sorry, but the PR did not get enough support in the last months. Closing for now.

Yxwww:
Thanks for the update 👍 

Does it mean this approach is not ideal ? Is there any alternative approach I should take ?

Mugen87:
> Is there any alternative approach I should take ?

You can remove the array data in the next frame manually since the renderer will perform the update until then.

BTW: It's actually no good approach to create a new `BufferAttribute` like mentioned in the forum. This will only create a memory leak since it's not possible to free single buffer attributes (meaning you can call `dispose()` only on the geometry object, not on attributes).

Yxwww:
Ah after reading some dispose code. It makes more sense to me. Thanks for explaining !

PR #18022, opened 2019-11-28
Update .d.ts for audio

f-a24:
Updated some d.ts of audio.


mrdoob:
Thanks!

f-a24:
AudioBuffer has its interface defined in typescript (lib.dom.d.ts).
Removed because it is not a three.js type.

PR #18025, opened 2019-11-28
WebGLRenderer: VR -> XR.

### LINKED ISSUE #18017: renderer.vr should be renamed to renderer.xr

From https://twitter.com/superhoge/status/1199804367638949888

I'd like to suggest to rename `renderer.vr` to `renderer.xr` because WebXR is used not only for VR but also AR. The property name "vr" may be confusing for AR devs.

##### Three.js version

- [x] Dev
- [ ] r111
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
Fixed #18017

### DIFF
--- a/docs/api/en/renderers/WebGLRenderer.html
+++ b/docs/api/en/renderers/WebGLRenderer.html
@@ -283,6 +283,16 @@
 		Tone mapping white point. Default is *1*.
 		</p>
 
+		<h3>[property:Object xr]</h3>
+		<p>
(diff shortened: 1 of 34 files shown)


mrdoob:
Thanks!

PR #18028, opened 2019-11-29
Path Class Documentation, clockwise properties types for arcs incorrectly indicated as float.

bartlomiej-przymus:
Path Class has methods for drawing arcs and these have clockwise property listed as Float. I changed it to Boolean.


Mugen87:
Good catch! Can you please also correct the Chinese version of the doc page?

https://github.com/mrdoob/three.js/blob/dev/docs/api/zh/extras/core/Path.html

bartlomiej-przymus:
Will do, as soon as i get home

mrdoob:
Thanks!

PR #18030, opened 2019-11-29
Make `extractUrlBase()` return empty string for data URIs

exelotl:
This fixes an issue where the resource path for GLTFLoader (and others) would be incorrect when loading a model from a data URI, which would cause trouble if the GLTF data contained relative paths to textures and buffers.

e.g. my GLTF contains the following:
```
    "buffers" : [
        {
            "byteLength" : 140,
            "uri" : "back-wall.bin"
        }
    ]
```

If I convert the GLTF to base64 and use a LoadingManager to inspect the URLs that get passed

```js
var manager = new THREE.LoadingManager();
manager.setURLModifier(function (url) {
  console.log(url);
  return url;
});
var loader = new THREE.GLTFLoader(manager);
loader.load('data:application/octet-stream;base64,ewogICAg...........', function () {});
```

It tries to load the URL `data:application/back-wall.bin` where it should have used `back-wall.bin`

image

### DIFF
--- a/src/loaders/LoaderUtils.js
+++ b/src/loaders/LoaderUtils.js
@@ -40,6 +40,12 @@ var LoaderUtils = {
 
 	extractUrlBase: function ( url ) {
 
+		if ( /^data:.*,.*$/i.test( url ) ) {
+
+			return '';
+
+		}
+
 		var index = url.lastIndexOf( '/' );
 
 		if ( index === - 1 ) return './';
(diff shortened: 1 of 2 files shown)


Mugen87:
Why don't you directly call `GLTFLoader.parse()`? In this way, you can pass `""` as the second parameter. It seems all what you have to do is to convert the base64 string into an array buffer.

PR #18042, opened 2019-12-02
Geometric anti-aliasing

elalish:
This is a small follow-on to my PMREM revamp, also upstreaming existing work from `<model-viewer>`. Since the new PMREM uses custom filtering and no standard mipmaps, it needs this anti-aliasing to be added manually. Here I am simply increasing the roughness (which equates to looking up a deeper mip level) according to the screen space derivatives of the geometric normal, just like standard mipmapping would. I am approximating the relationship between normal variation and roughness as linear, when it should probably be square, but I didn't think the extra sqrt would be worth the difference. 

It might seem natural to use the actual normal variation (including the normal map) here instead of just the vertex normals, however I found that the structural errors from texture interpolation simply manifested as their own source of aliasing rather than fixing the problem. I'll have another PR coming shortly to address the roughness component from normal map variation.

### DIFF
--- a/src/renderers/shaders/ShaderChunk/lights_physical_fragment.glsl.js
+++ b/src/renderers/shaders/ShaderChunk/lights_physical_fragment.glsl.js
@@ -1,7 +1,13 @@
 export default /* glsl */`
 PhysicalMaterial material;
 material.diffuseColor = diffuseColor.rgb * ( 1.0 - metalnessFactor );
-material.specularRoughness = clamp( roughnessFactor, 0.04, 1.0 );
+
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

mrdoob:
I think this PR broke clearcoat?

This is how it used to look:

https://threejs.org/examples/webgl_materials_physical_clearcoat.html

Screen Shot 2019-12-03 at 10 56 12 AM

And this is how it looks now:

https://raw.githack.com/mrdoob/three.js/dev/examples/webgl_materials_physical_clearcoat.html

Screen Shot 2019-12-03 at 10 55 37 AM

PR #18077, opened 2019-12-05
TS: Undeprecate Skeleton

### LINKED ISSUE #18056: Skeleton really deprecated or noted as such by mistake?

In [src/objects/Skeleton.d.ts](https://github.com/mrdoob/three.js/blob/441fefadc017ebafd2d10089b6d48e1cd28795c2/src/objects/Skeleton.d.ts#L6) it is mentioned that Skeleton is deprecated and one should use Points instead.
This is even mentioned twice.
Is this by mistake and Skeleton has not been deprecated or it missing in the Upgrade Guide how to migrate to Points.
I fail to see how Points can replace the skeleton as the bone texture is only being built from bones.

Any pointers are highly welcome. I refrained from asking on SO as it feels like it has been marked deprecated by mistake.

##### Three.js version

- [X] all since r101

discordier:
Fixes #18056 - Skeleton has been marked as deprecated by mistake.

### DIFF
--- a/src/objects/Skeleton.d.ts
+++ b/src/objects/Skeleton.d.ts
@@ -2,13 +2,6 @@ import { Bone } from './Bone';
 import { Matrix4 } from './../math/Matrix4';
 import { DataTexture } from './../textures/DataTexture';
 
-/**
- * @deprecated Use {@link Points THREE.Points} instead.
- */
-/**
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #18078, opened 2019-12-05
Object3D types: missing readonly keyword on some properties

### LINKED ISSUE #18076: Object3D type definition: missing readonly keyword for some attributes
The `position`, `rotation`, `quaternion`, `scale`, `modelViewMatrix` and `normalMatrix` attributes of `Object3D` are defined using the [`Object.defineProperties`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties) static method (See [here](https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.js#L70)). This method will set properties as read-only (writable = false) by default.

The problem is that those properties are not defined as read-only in the type definition (See [here](https://github.com/mrdoob/three.js/blob/dev/src/core/Object3D.d.ts#L62)), which results in this code to compile (which I guess should not): 
```typescript
const camera = new THREE.PerspectiveCamera(...);

// This compiles, but throws an error at runtime
camera.position = new THREE.Vector3(1, 1, 1);
```

I can come up with a PR fixing this if that is considered a bug.

martinRenou:
Fix #18076

### DIFF
--- a/src/core/Object3D.d.ts
+++ b/src/core/Object3D.d.ts
@@ -59,26 +59,26 @@ export class Object3D extends EventDispatcher {
 	/**
 	 * Object's local position.
 	 */
-	position: Vector3;
+	readonly position: Vector3;
 
 	/**
 	 * Object's local rotation (Euler angles), in radians.
 	 */
-	rotation: Euler;
+	readonly rotation: Euler;
 
 	/**
 	 * Global rotation.
 	 */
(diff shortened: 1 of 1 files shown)


Mugen87 (approved):

mrdoob:
Thanks!

PR #18089, opened 2019-12-06
GLTFLoader: Clean up UV set messages

zeux:
Almost all models that use occlusionTexture use texCoord 0 - the loader
prints a console message for each material for such models which seems
redundant / excessive.

However, if other textures use texCoord that's not 0, the resulting
texture will not be rendered properly. We now detect this case and issue
a warning.


mrdoob:
Thanks!

PR #18091, opened 2019-12-06
Stop support for Internet Explorer

Mugen87:
I think the upcoming turn of the year is a good opportunity to officially stop support for Internet Explorer. The PR removes any references to IE from the docs and also any IE related workarounds from the core and the examples.

- The `Uint8ClampedArray` hack only affected older builds of IE11. So latest versions don't need it anyway.
- `Math.log2` can be polyfilled by users if necessary.
- Removed all hacks from `CSS3DRenderer`.

### DIFF
--- a/docs/examples/en/loaders/BasisTextureLoader.html
+++ b/docs/examples/en/loaders/BasisTextureLoader.html
@@ -64,10 +64,6 @@
 		<p>
 			Transcoding to PVRTC1 (for iOS) requires square power-of-two textures.
 		</p>
-		<p>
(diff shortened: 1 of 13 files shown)


looeee:
> The Uint8ClampedArray hack only affected older builds of IE11

Specifically builds pre KB2929437 which was released in July 2014. I think it's very reasonable to no longer support that.

Mugen87:
It seems the PR did not get enough support. I hope we can revisit this issue at a later point.

PR #18098, opened 2019-12-07
Examples: remove TerrainShader

WestLangley:
As suggested in https://github.com/mrdoob/three.js/issues/18005#issuecomment-562787567.


mrdoob:
Thanks!

PR #18114, opened 2019-12-10
changed from var to let

ahmedstryout:


donmccurdy:
We don't currently use let/const in the threejs codebase. If you're having a particular issue that would benefit from this change, please open an issue — thanks!

PR #18123, opened 2019-12-11
BasisTextureLoader: Fix severe work distribution imbalance

zeux:
When used from GLTF loader, it's common for BasisTextureLoader promises
to be created in quick succession for a number of textures.

What happens is that the methods wait until the transcoder is loaded,
and then call _getWorker in succession. After each _getWorker call
succeeds, the continuation for the promise is scheduled - however, the
worker task cost is only adjusted when that continuation is ran, and the
next _getWorker call runs immediately after.

As a result, the _taskLoad-based sorting doesn't work, and most textures
are assigned to the last worker, which means that the decode is mostly
single-threaded.

This change fixes this by integrating cost adjustment into _getWorker
(renamed to _allocateWorker for clarity); we also don't need _taskCosts
anymore because we can close over the taskCost local instead.

Also fix the problem where we were creating redundant object URLs for
the transcoder binary because of many calls to _initTranscoder (during
_getWorker) during which the transcoder promise is setup but the binary
hasn't been loaded yet.

### DIFF
--- a/examples/js/loaders/BasisTextureLoader.js
+++ b/examples/js/loaders/BasisTextureLoader.js
@@ -119,7 +119,9 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 		var worker;
 		var taskID;
 
-		var texturePending = this._getWorker()
+		var taskCost = buffer.byteLength;
+
+		var texturePending = this._allocateWorker( taskCost )
 			.then( ( _worker ) => {
 
 				worker = _worker;
@@ -128,8 +130,6 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 				return new Promise( ( resolve, reject ) => {
 
 					worker._callbacks[ taskID ] = { resolve, reject };
-					worker._taskCosts[ taskID ] = buffer.byteLength;
-					worker._taskLoad += worker._taskCosts[ taskID ];
 
 					worker.postMessage( { type: 'transcode', id: taskID, buffer }, [ buffer ] );
 
@@ -181,9 +181,8 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 
 				if ( worker && taskID ) {
 
-					worker._taskLoad -= worker._taskCosts[ taskID ];
+					worker._taskLoad -= taskCost;
 					delete worker._callbacks[ taskID ];
-					delete worker._taskCosts[ taskID ];
 
 				}
 
@@ -195,7 +194,7 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 
 	_initTranscoder: function () {
 
-		if ( ! this.transcoderBinary ) {
+		if ( ! this.transcoderPending ) {
 
 			// Load transcoder wrapper.
 			var jsLoader = new THREE.FileLoader( this.manager );
@@ -239,7 +238,7 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 
 	},
 
-	_getWorker: function () {
+	_allocateWorker: function ( taskCost ) {
 
 		return this._initTranscoder().then( () => {
 
@@ -248,7 +247,6 @@ THREE.BasisTextureLoader.prototype = Object.assign( Object.create( THREE.Loader.
 				var worker = new Worker( this.workerSourceURL );
 
 				worker._callbacks = {};
-				worker._taskCosts = {};
 				worker._taskLoad = 0;
 
 				worker.postMessage( {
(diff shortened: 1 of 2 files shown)


zeux:
I've discovered this when profiling http://zeux.io/tests/rubyrose which uses a 4MB .glb model with 13 Basis encoded textures of varying sizes. Here's the Chrome performance timeline before the fix; you can see that most of the transcode requests execute on just one worker:

image

Here's the timeline after the fix:

image

zeux:
(as an aside I have no idea what the "empty" web workers are about, if anybody knows how to track this down I'd be happy to fix this as well)

donmccurdy (approved):
Nice catch, thanks — I might need to figure out how to unit test this behavior, it's been a bit tricky to get worker management right, both here and in DRACOLoader.

@kaisalmen I'd be curious how feasible you think it would be to have a WorkerLoader-like utility that (1) can support arbitrary types of resources (Texture or BufferGeometry in these cases) and (2) doesn't necessarily put the threejs library itself on the Web Worker? Shared task management would benefit DRACOLoader and BasisTextureLoader, and an isolated utility for that purpose might be easier to test against bugs like this. That said, I probably would not want to add such a dependency until after we've removed the `examples/js` folder and can just rely on imports through `examples/jsm`.

zeux:
> WorkerLoader-like utility

I was thinking the same thing - this really warrants a single generalized helper, that can just schedule arbitrary tasks onto a given worker pool, and handles work balancing, message communication, loading .js/.wasm files and error handling. All of this logic seems like it's not really dependent on the Basis details in the loader, and it's the part that is actually complex, the rest of the Basis decoder is basically streamlined code that is easier to read and validate.

mrdoob:
Thanks!

donmccurdy:
> as an aside I have no idea what the "empty" web workers are about, ...

I'm traveling and my internet connection isn't letting me access the demo, but will have a look when I get back. 👍

kaisalmen:
> I'd be curious how feasible you think it would be to have a WorkerLoader-like utility that (1) can support arbitrary types of resources (Texture or BufferGeometry in these cases) and (2) doesn't necessarily put the threejs library itself on the Web Worker?

I have something like this in my mind for quite some time for a general `AssetLoader`, but never realized it, yet. Having a set of common worker utility functions with bespoke features could ease the use parallel worker tasks. These util functions could evolve from the independent worker util `WorkerSupport` used by `OBJLoader2` V3, but that is of course only one possibility.
Sounds like a nice project to kick off 2020. 😄  Let's discuss this topic further if you like.

donmccurdy:
> as an aside I have no idea what the "empty" web workers are about, ...

@zeux I think the empty web workers are an artifact of the profiling method, as some recorded data persists across a page refresh.

**Method 1:**

1. Start recording.
2. Refresh page.
3. Stop recording.
4. Observe 4 "empty" workers and 4 active workers.

**Method 2:**

1. Set breakpoint early in application.
2. Refresh page and hit breakpoint.
3. Start recording (which unpauses debugger immediately in Chrome)
4. Stop recording.
5. Observe only 4 workers, all active.

donmccurdy:
@kaisalmen apologies if this is one abstraction too far, but would such a utility need to be specific to "loading"? You could also think of it as scheduling tasks to happen off the main thread, with different tasks having different associated costs and priorities. 

For example, the new [RoughnessMipmapper](https://github.com/mrdoob/three.js/pull/18110) is not a loader, but could do its work on a WW.

But for all the cases I'm currently concerned with, yes — a worker-like utility would be just fine. :)

kaisalmen:
> apologies if this is one abstraction too far, but would such a utility need to be specific to "loading"?

No, it should not be constraint to loading. You are right. Scheduling, defined communication patterns and overall management are important requirements, but also getting arbitrary code into the Workers easily will help people adapt such a util. Any idea how ready "ES Modules in workers" are across browsers these days? A worker can de-generate to a set of imports of where the execution functions obeys a communication/invocation contract...

donmccurdy:
> Any idea how ready "ES Modules in workers" are across browsers these days?

Unfortunately I don't know. It might come down more to bundling tools than native ES module support, and either way can get complicated fast. Arguably a general-purpose task management library could be a standalone library, or part of something like [ecsy](https://github.com/MozillaReality/ecsy), rather than part of three... if we were going to invest a _lot_ of time in something really robust, I'd suggest the former. But hopefully we can get a simple version working to at least de-duplicate the code that's already being used by various loaders, within three.

Is there an existing issue where it would make sense to move this conversation, or should we open a new one?

kaisalmen:
> Is there an existing issue where it would make sense to move this conversation, or should we open a new one?

A new one, I guess. I am not aware of an existing one, but I currently don't have a good overview.

donmccurdy:
Opened https://github.com/mrdoob/three.js/issues/18234.

PR #18125, opened 2019-12-11
Handle getting a detached ArrayBuffer from Cache

Oletus:
ArrayBuffers that are stored in the Cache may not remain valid in case they end up in a detached state. Buffers can enter detached state when transferred to a web worker for example.

This fixes feeding the same URL multiple times to FileLoader when using DRACOLoader, that offloads decoding files to a worker.

### DIFF
--- a/src/loaders/Cache.js
+++ b/src/loaders/Cache.js
@@ -26,6 +26,7 @@ var Cache = {
 
 			// This ArrayBuffer may be in a detached state, so it's unusable.
 			// A buffer can enter detached state when it is transferred to a worker for example.
+			console.warn( 'THREE.Cache: could not get a loaded ArrayBuffer from cache since it may have been in detached state. This can result from the ArrayBuffer having been transferred to a worker for processing. Cache key: ' + key );
 			this.remove( key );
 
 		}
--- a/src/loaders/Cache.js
+++ b/src/loaders/Cache.js
@@ -22,6 +22,14 @@ var Cache = {
 
 		if ( this.enabled === false ) return;
 
+		if ( this.files[ key ] !== undefined && this.files[ key ] instanceof ArrayBuffer && this.files[ key ].byteLength === 0 ) {
+
+			// This ArrayBuffer may be in a detached state, so it's unusable.
+			// A buffer can enter detached state when it is transferred to a worker for example.
+			this.remove( key );
+
+		}
+
 		// console.log( 'THREE.Cache', 'Checking key:', key );
 
 		return this.files[ key ];


Oletus:
@Mugen87 @mrdoob Please review this bug fix! An alternative to this would be to store separate copies in the cache, but I decided against suggesting it since it would waste memory. Other alternatives like creating copies of buffers only when they're needed would be significantly more complex and less robust.

Mugen87:
I remember we applied a similar fix in `AudioLoader` in order to avoid detached states errors. 

https://github.com/mrdoob/three.js/blob/efaeea930ad74eb955fa0b6712a499f6c7e89e44/src/loaders/AudioLoader.js#L26-L28

However, audio buffers are usually small. So cloning the buffer is fine in this case. However, mesh data tend to be much larger so cloning could be indeed problematic.

Could we also completely avoid the storage of arraybuffers in `Cache`? If there are not present in the first place, there is no need to remove them.

donmccurdy:
The effect of this PR will be to re-download the detached resource silently — I’m not sure that improves things for users. At minimum, some warning or error should be logged.

But more importantly, what should the application be doing? Copying buffers and re-decoding them seems bad, too. Perhaps the fix should be added to DRACOLoader and BasisTextureLoader (which transfer data to web workers) rather than in FileLoader.

donmccurdy:
Using a WeakMap in DRACOLoader to look up decode results given the same buffer might be a solution.

Oletus:
I think a warning would be a good addition to this. I'd assume most apps that need several copies of the same mesh for example can create the copies from the loaded mesh, instead of loading the same url multiple times.

Adding caching to DRACOLoader and BasisTextureLoader could be a good complement to this, but I don't think it would be sufficient on itself. Apps can also use FileLoader directly.

donmccurdy:
> Adding caching to DRACOLoader and BasisTextureLoader could be a good complement to this, but I don't think it would be sufficient on itself. Apps can also use FileLoader directly.

I think we should fix the issue in DRACOLoader and BasisTextureLoader first. I'm not sure that FileLoader should be checking the state of things it caches, even to evict them from the cache or to log a warning... consider if instead of detaching the ArrayBuffer, application code was directly modifying its values — anything reading the (now modified) entry in the cache will similarly be broken.

IMO, code that reuses objects from a cache needs to be careful about what it does to those objects. In the case of AudioLoader, making a copy is perhaps fine. But as @Mugen87 mentions, the solution for larger objects may be a different one.

I'll try to put together a PR for DRACOLoader and BasisTextureLoader.

donmccurdy:
Opened https://github.com/mrdoob/three.js/pull/18136.

PR #18132, opened 2019-12-12
GLTFLoader: Load materials and geometries in parallel

### LINKED ISSUE #18097: GLTFLoader: Separate asset loading order

Right now when loading a .gltf file, the GLTFLoader will load all of the images first and then the .bin files. I was wondering if there was a reason for this loading order. This can be seen in [this example](https://threejs.org/examples/?q=gltf#webgl_loader_gltf).

One instance I can think of where loading the .bin files earlier would be preferred is with DRACO encoded meshes. If we were to load the .bin files in tandem with the images, it would allow the DRACO meshes to start being decoded earlier in the loading process, potentially before all images are loaded. This would allow certain DRACO gltf files to load faster, with the decoding happening in parallel to image loading.

##### Three.js version

- [x] Dev
- [x] r111

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

zeux:
Instead of waiting for all materials of a mesh before fetching geometry
data, we now push all promises into one large array and wait on all of
this at once.

This means we can load .bin file in parallel with texture data when
texture assets are using separate files, and potentially start decoding
geometry earlier.

Fixes #18097

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -2591,108 +2591,109 @@ THREE.GLTFLoader = ( function () {
 
 		}
 
-		return Promise.all( pending ).then( function ( originalMaterials ) {
+		pending.push( parser.loadGeometries( primitives ) );
 
-			return parser.loadGeometries( primitives ).then( function ( geometries ) {
+		return Promise.all( pending ).then( function ( results ) {
 
-				var meshes = [];
+			var materials = results.slice( 0, results.length - 1 );
+			var geometries = results[ results.length - 1 ];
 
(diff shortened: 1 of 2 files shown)


zeux:
Network loads before this change (threejs.org):

image

Network loads after this change (from localhost, so they are quicker):

image

donmccurdy (approved):
FYI @TyLindberg

TyLindberg:
Nice! Thanks for getting this done @zeux!

mrdoob:
Thanks!

PR #18196, opened 2019-12-18
Pre-compile PMREM

elalish:
I added shader pre-compilation to PMREMGenerator, as I did with RoughnessMipmapper. This can help first load noticeably since it is easy to compile these shaders while the texture download is in flight. There are two prepare functions you can call depending on what type of texture you intend to process. For this to work, I also had to change the programCacheKey, as it was dependent on all the parameters despite RawShaderMaterial not using any of them (and therefore not requiring a recompile). 

I also fixed some mistakes I'd made in the PMREM examples where I was assuming a cubemap where it was really an equirectangular texture.


mrdoob:
Thanks!

PR #17145, opened 2019-08-01
WIP: OrbitControls: added support for zoom to cursor

WestLangley:
As suggested in https://github.com/mrdoob/three.js/pull/16961#issuecomment-516116475.

I think this should be the default for `MapControls`. I'm not sure if it makes as much sense for `OrbitControls`, but it can be toggled on.

This is a WIP. Feedback welcome.

<EDIT - live link removed>

PR #18202, opened 2019-12-19
Zoom over position.

Dubiy:
Zooming over point of interest (under cursor), not over center of canvas.

Available to switch on that option, by setting this.zoomOverPosition = true;
Disabled by default.


Mugen87:
Does your PR aim for the same goal as #17145?

Dubiy:
> Does your PR aim for the same goal as #17145?

Looks similar, but my solution for OrthographicCamera

Mugen87:
The PR of @WestLangley actually supports both orthographic and perspective cameras.

Mugen87:
Closing in favor of #17145. Thanks for your effort though!

PR #18237, opened 2019-12-23
Added an ability to reset current state of Orbit controls

pavelkeyzik:
The problem: 
==
I'm using Three.js and OrbitControls in WebWorker but the problem is that OrbitControls uses DOM. I added `mouseup` event listener to `document` and sent a message to my Worker that I want to unsubscribe from events in my controls to avoid this tiny bug.

ezgif-4-4193cbe78b06

Solution:
==
I added `resetState` method to OrbitControls and this method will reset state to `STATE.NONE` value, but if I have to add something more, please, let me know. I think that I also have to dispatch `change` event, but not sure about it.

### DIFF
--- a/docs/examples/en/controls/OrbitControls.html
+++ b/docs/examples/en/controls/OrbitControls.html
@@ -273,6 +273,11 @@ controls.touches = {
 			Get the current vertical rotation, in radians.
 		</p>
 
+		<h3>[method:null resetState] ()</h3>
+		<p>
+			Reset the current state of the controls.
+		</p>
+
 		<h3>[method:null reset] ()</h3>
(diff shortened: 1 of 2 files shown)


Mugen87:
> I'm using Three.js and OrbitControls in WebWorker

The usage of control classes in workers is actually not supported. Because of their dependencies to the DOM). So the idea is to implement user interaction in the main thread with a dummy camera and transfer the respective transformation to the worker (where it's used for rendering). Hence, I don't think this PR is necessary.

pavelkeyzik:
@Mugen87 Oh, okay. Thank you for your response 😊

PR #18242, opened 2019-12-25
Temporary fix for broken things while the editor is beign moved to modules

sdinesh86:
Many of the current implementations in editor rely on the global editor variable, due to move to modules for editor

### DIFF
--- a/editor/index.html
+++ b/editor/index.html
@@ -67,7 +67,8 @@
 
 			//
 
-			var editor = new Editor();
+            var editor = new Editor();
+            window['editor'] = editor;
 
 			var viewport = new Viewport( editor );
 			document.body.appendChild( viewport.dom );


Mugen87:
Can you please list what things are broken now? I think we should find a way to fix them in a more appropriate way than globally exposing `editor`.

editor/index.html
```diff
@@ -67,7 +67,8 @@
 
 			//
 
-			var editor = new Editor();
+            var editor = new Editor();
+            window['editor'] = editor;
```
Mugen87: Please watch for correct formatting.

sdinesh86: will take care of it

sdinesh86:
> Can you please list what things are broken now? I think we should find a way to fix them in a more appropriate way than globally exposing `editor`.

Drag and drop in the outliner is not working,i didnt get time to look into other things but will be looking into them

Mugen87:
> Drag and drop in the outliner is not working,i didnt get time to look into other things but will be looking into them

Solved via #18243. 

Closing this PR since adding `editor` to `window` is not an appropriate solution. Let's try to fix other issues similar to #18243.

PR #18243, opened 2019-12-25
Editor: Fix drag/drop in Outliner.

Mugen87:
see https://github.com/mrdoob/three.js/pull/18242#issuecomment-568899652


mrdoob:
Thanks!

PR #18256, opened 2019-12-28
Examples: added instancing_modified example

WestLangley:
Credit for the inspiration for this PR goes to @takahirox.  

https://twitter.com/superhoge/status/1176740112345911297

This example is simpler, and just shows how to use code injection to add a custom instanced attribute -- in this case color -- to `InstancedMesh`.

Suggestions welcome.

### DIFF
--- a/examples/webgl_instancing_modified.html
+++ b/examples/webgl_instancing_modified.html
@@ -1,0 +1,196 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - instancing - modified</title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+		<link type="text/css" rel="stylesheet" href="main.css">
+	</head>
+
+	<body>
+		<div id="info">
+			<a href="http://threejs.org" target="_blank" rel="noopener">three.js</a> - webgl instancing - modified
+		</div>
+
+		<script type="module">
+
+			import * as THREE from '../build/three.module.js';
+
+			import Stats from './jsm/libs/stats.module.js';
+
+			var camera, scene, renderer, stats;
+
+			var mesh;
+
+			var dummy = new THREE.Object3D();
+
+			var amount = 8;
+			var count = Math.pow( amount, 3 );
+
+			init();
+			animate();
+
+			function init() {
+
+				renderer = new THREE.WebGLRenderer( { antialias: false } ); // false improves the frame rate
+				renderer.setPixelRatio( window.devicePixelRatio );
+				renderer.setSize( window.innerWidth, window.innerHeight );
+				document.body.appendChild( renderer.domElement );
+
+				renderer.outputEncoding = THREE.sRGBEncoding;
+
+				scene = new THREE.Scene();
+
+				camera = new THREE.PerspectiveCamera( 40, window.innerWidth / window.innerHeight, 1, 100 );
+				camera.position.set( 0, 0, 20 );
+
+				new THREE.BufferGeometryLoader().load( 'models/json/suzanne_buffergeometry.json', function ( geometry ) {
+
+					var instanceColors = [];
+
+					for ( var i = 0; i < count; i ++ ) {
+
+						instanceColors.push( 0.5 + 0.5 * Math.random() );
+						instanceColors.push( 0.5 + 0.5 * Math.random() );
+						instanceColors.push( 0.5 + 0.5 * Math.random() );
+
+					}
+
+					geometry.setAttribute( 'instanceColor', new THREE.InstancedBufferAttribute( new Float32Array( instanceColors ), 3 ) );
+
+					geometry.computeVertexNormals();
+
+					geometry.scale( 0.5, 0.5, 0.5 );
+
+					//console.log( geometry );
+
+					//
+
+					new THREE.TextureLoader().load( 'textures/matcaps/matcap-porcelain-white.jpg', function ( texture ) {
+
+						texture.encoding = THREE.sRGBEncoding;
+
+						var material = new THREE.MeshMatcapMaterial( { color: 0xaaaaff, matcap: texture } );
+
+						var colorParsChunk = [
+							'attribute vec3 instanceColor;',
+							'varying vec3 vInstanceColor;',
+							'#include <common>'
+						].join( '\n' );
+
+						var instanceColorChunk = [
+							'#include <begin_vertex>',
+							'\tvInstanceColor = instanceColor;'
+						].join( '\n' );
+
+						var fragmentParsChunk = [
+							'varying vec3 vInstanceColor;',
+							'#include <common>'
+						].join( '\n' );
+
+						var colorChunk = [
+							'vec4 diffuseColor = vec4( diffuse * vInstanceColor, opacity );'
+						].join( '\n' );
+
+						material.onBeforeCompile = function ( shader ) {
+
+							shader.vertexShader = shader.vertexShader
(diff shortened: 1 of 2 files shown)


WestLangley:
Temporary live link:
https://raw.githack.com/westlangley/three.js/dev_instancing_modified/examples/webgl_instancing_modified.html

examples/webgl_instancing_modified.html
```diff
@@ -0,0 +1,196 @@
...
+
+							//console.log( shader.uniforms );
+							//console.log( shader.vertexShader );
+							//console.log( shader.fragmentShader );
+
+						};
+
+						mesh = new THREE.InstancedMesh( geometry, material, count );
+
+						mesh.instanceMatrix.setUsage( THREE.DynamicDrawUsage ); // will be updated every frame
```
donmccurdy: Oh! We've omitted this in `instancing / scatter` and `instancing / dynamic`. The documentation describes `.instanceMatrix` as "For internal use only," which might discourage users from doing this themselves. Should we (a) document this `.setUsage` suggestion, or (b) just enable it by default with InstancedMesh?

examples/webgl_instancing_modified.html
```diff
@@ -0,0 +1,196 @@
...
+
+						var fragmentParsChunk = [
+							'varying vec3 vInstanceColor;',
+							'#include <common>'
+						].join( '\n' );
+
+						var colorChunk = [
+							'vec4 diffuseColor = vec4( diffuse * vInstanceColor, opacity );'
+						].join( '\n' );
+
```
donmccurdy: How much trouble would it be to make this modification offset the UVs rather than changing the color — aside from the effort of finding a model and texture that look good with some offsets? 🤔

There's an easier way to set per-instance color, which I included in `instancing / scatter`:

https://github.com/mrdoob/three.js/blob/45418089bd5633e856384a8c0beefced87143334/examples/webgl_instancing_scatter.html#L82-L98

^But if for some reason that's not a recommended way to do this, I'm happy to address it there.

donmccurdy: I'm imagining a translation-only UV offset... so e.g. four sub-textures could be packed adjacent in one image...

. 1 – 2 – 3 – 4
◾️◾️◾️◾️

... and the offset would select a different sub-texture for different instances. That would seem like a good example for something like minecraft-style voxels, where there are a finite number of voxel types.

WestLangley:
```js
blossomGeometry.setAttribute( 'color', new THREE.InstancedBufferAttribute( color, 3 ) );

blossomMesh = new THREE.InstancedMesh( blossomGeometry, blossomMaterial, count );
```

Yikes. I can't imagine adding an  `InstancedBufferAttribute` to `BufferGeometry` ever working. I would not recommend we encourage that coding pattern -- even if it does work in this case.

>^But if for some reason that's not a recommended way to do this, I'm happy to address it there.

Yes, I'd recommend you think about changing that.

examples/webgl_instancing_modified.html
WestLangley: I would not have described it as "for internal use only".

I'd document the set usage suggestion.

examples/webgl_instancing_modified.html
donmccurdy: It's `InstancedMesh` + `InstancedBufferGeometry` + `InstancedBufferAttribute` there, which doesn't seem obviously wrong, but ¯\\_(ツ)_/¯

examples/webgl_instancing_modified.html
Mugen87: I vote for option (b).

WestLangley: Remove "internal use only" and document option (a).

I think what I have done in this example is highly appropriate.  I even included a comment for clarification.

Users should set `DynamicDrawUsage` for any attribute that is updated every frame. The default should remain `StaticDrawUsage`.

Instancing is appropriate for static objects such as trees. I would hope that spinning monkeys is not a typical use case.

examples/webgl_instancing_modified.html
WestLangley: It seems to me that `InstancedBufferAttribute` should only be added to `InstancedBufferGeometry`. Adding it to regular `BufferGeometry` is inappropriate -- IMO.

donmccurdy: There is no `BufferGeometry` in any of these examples — I'm passing `InstantedBufferGeometry` to the `InstancedMesh` constructor.

examples/webgl_instancing_modified.html
donmccurdy: Sounds good to me.

examples/webgl_instancing_modified.html
WestLangley: > I'm passing InstantedBufferGeometry to the InstancedMesh constructor.
```js
blossomGeometry.setAttribute( 'color', new THREE.InstancedBufferAttribute( color, 3 ) );
```
Is not `blossomGeometry` is an instance of `BufferGeometry`?

>I'm passing `InstancedBufferGeometry` to the `InstancedMesh` constructor.

The `InstancedMesh` constructor expects a `BufferGeometry`, no?

donmccurdy: Ahh you're right. In my head that was definitely InstancedBufferGeometry. 🤔

I would expect the `InstancedMesh` constructor to accept `InstancedBufferGeometry`, to allow users to keep its API even when using other instanced attributes, but apparently I haven't actually tested that yet!

donmccurdy: ^But I was not expecting BufferGeometry+InstancedBufferAttribute to work. I'll update that example...

mrdoob:
Thanks!

WestLangley:
@mrdoob @donmccurdy Should I change the name to `webgl_instancing_custom_attribute`?

donmccurdy:
> Should I change the name to webgl_instancing_custom_attribute?

I don't have a preference between the current name and this suggestion, but I do like that neither is specifically focusing on color being the customized attribute.

Also... had you realized that this example is also adding InstancedBufferAttribute to BufferGeometry? I didn't earlier, but:

Screen Shot 2019-12-30 at 3 16 53 PM

WestLangley:
@donmccurdy Oh, gosh. I'm doing the same thing that you are doing. :/

OK, I think I will have to give it up and live with it.

Sorry for the unnecessary noise... And thank you for your feedback. :-)

donmccurdy:
I do agree that adding `InstancedBufferAttribute` to `BufferGeometry` feels bad when `InstancedGeometry` exists, and it's obviously an easy mistake to make... that suggests either:

(a) if there are real reasons this is fragile or might fail in the future, we should perhaps log a warning now.

(b) if `InstancedBufferGeometry` has no remaining distinction from `BufferGeometry`, except the `.maxInstancedCount` property which is now redundant with `InstancedMesh.prototype.count`, then maybe we should remove `InstancedBufferGeometry` and just accept that `InstancedBufferAttributes` can be added to `BufferGeometry`, valid only if the geometry is used with `InstancedMesh`.

mrdoob:
How about a visibility attribute?

WestLangley:
I think for now, I'd recommend removing the example until the time at which someone suggests a use case for which customized instancing is required.

mrdoob:
> until the time at which someone suggests a use case for which customized instancing is required.

I just did 😅

WestLangley:
>since InstancedMesh now supports instance colors, webgl_instancing_modified.html needs to be changed to modify some other per-instance attribute.
>>How about a visibility attribute?

Is there a practical use case for which customized instancing is _required_? If so, we could show an example of that.

PR #18263, opened 2019-12-29
BufferGeometryUtils: Replace nbsp with normal one. Fixes #18260

FunFetched:
Pushing changes to the right file this time (js/ instead of jsm/). See #18260 for details.


Mugen87:
I think this should be already solved via e1ea3219510b569755db957950b84e2649a134b6. see https://github.com/mrdoob/three.js/pull/18260#issuecomment-569515050

Mugen87:
Thanks though!

PR #18267, opened 2019-12-30
Update Useful links

packtpartner:
Added a new request


Mugen87:
Sorry, but I don't think the `Useful Links` section is the right place to promote payed content.

Issue #18265, opened 2019-12-30
New MeshStandardMaterial in r112 has banding artifacts on some GPUs


toji:
As of r112 the default MeshStandardMaterial (in this case specifically as it's used when loading glTF files, but probably in other scenarios as well) is exhibiting banding artifacts on some GPUs. Specifically, I have see this issue on a Pixelbook and every generation of Pixel phones. The issue does not occur on the Nvidia desktop GPUs I have tried, nor on an Oculus Go or Oculus Quest.

It should also be noted that I've observed the material's new shader incurred a noticable performance hit relative to r111 for my particular application on various mobile devices, including ones that don't exhibit the rendering artifacts.

The artifacts look like this (Rendered with Three.js r112 on a Pixelbook)

![Screenshot 2019-12-29 at 9 15 19 PM](https://user-images.githubusercontent.com/805273/71569200-4e89f300-2a82-11ea-820a-a07c6542ef09.png)

The expected output looks like this (Rendered with Three.js r111 on the same Pixelbook)

![Screenshot 2019-12-29 at 9 24 31 PM](https://user-images.githubusercontent.com/805273/71569215-78431a00-2a82-11ea-84ee-6231bfa7dfb6.png)

Live link, currently using r112: https://xrdinosaurs.com

All of the dinosaurs on that page exhibit the issue, but ones with large areas of flat or smooth colors (like the TRex's belly) tend to stand out more.

The material in the screenshot (pasted in full below) uses the glTF extension "KHR_materials_pbrSpecularGlossiness", and has a diffuse, normal, and specular/glossiness textures.

```json
    {
      "doubleSided": true,
      "emissiveFactor": [
        0,
        0,
        0
      ],
      "extensions": {
        "KHR_materials_pbrSpecularGlossiness": {
          "diffuseFactor": [
            0.46512957319999998,
            0.46512957319999998,
            0.46512957319999998,
            1
          ],
          "diffuseTexture": {
            "index": 4,
            "texCoord": 0
          },
          "glossinessFactor": 0.27610518290000002,
          "specularFactor": [
            0.92244664629999995,
            0.92244664629999995,
            0.92244664629999995
          ],
          "specularGlossinessTexture": {
            "index": 6,
            "texCoord": 0
          }
        }
      },
      "name": "TRex",
      "normalTexture": {
        "index": 5,
        "scale": 1,
        "texCoord": 0
      }
    }
```

Mugen87:
[...]

PR #18272, opened 2019-12-30
GLTFLoader: Fix specular-glossiness material.

Mugen87:
- Fixed the modified `lights_physical_fragment` shader chunk, see #18265.
- Fixed the broken usage of tangent-space normal maps. The shader material did not define `normalMapType` which means `TANGENTSPACE_NORMALMAP` was never defined.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -622,6 +622,7 @@ THREE.GLTFLoader = ( function () {
 				'bumpMap',
 				'bumpScale',
 				'normalMap',
+				'normalMapType',
 				'displacementMap',
 				'displacementScale',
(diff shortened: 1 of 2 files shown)


Mugen87:
Prod: https://threejs.org/examples/webgl_loader_gltf_extensions.html
PR: https://raw.githack.com/Mugen87/three.js/dev26/examples/webgl_loader_gltf_extensions.html

This is how the boom box looks in `prod` now:

image

It also logs the warning:

> three.module.js:18258 THREE.WebGLProgram: gl.getProgramInfoLog() WARNING: Output of vertex shader 'vTangent' not read by fragment shader

After this change:

image
 
And the warning is gone, too.

mrdoob:
Thanks!

PR #18298, opened 2020-01-03
VRMLLoader: Allow minus in Identifier

tuarrep:
Lot's of files I'm using use minus (`-`) in node identifiers (KiCAD 3D models).
I've haven't found any objections about using dash in VRML specs.

Files can be found here https://kicad.github.io/packages3d/

Correct me if I'm wrong.


Mugen87:
According to http://gun.teipir.gr/VRML-amgem/spec/part1/concepts.html#SyntaxBasics, a minus is not allowed. More details here:

> Field, event, prototype, and node names must not begin with a digit (0x30-0x39) but may otherwise contain any characters except for non-printable ASCII characters (0x0-0x20), double or single quotes (0x22: ", 0x27: '), sharp (0x23: #), plus (0x2b: +), comma (0x2c: ,), minus (0x2d: -), period (0x2e: .), square brackets (0x5b, 0x5d: []), backslash (0x5c: \) or curly braces (0x7b, 0x7d: {}). Characters in names are as specified in ISO 10646, and are encoded using UTF-8. VRML is case-sensitive; "Sphere" is different from "sphere" and "BEGIN" is different from "begin."

I do not vote to change the regex since it corresponds to the standard. It's actually the problem of KiCAD since it export VRML files with invalid identifiers.

tuarrep:
@Mugen87 

Thank for pointing this out. I haven't see this paragraph.
Closing this PR so.

tuarrep:
Sorry for bumping this lately. KiCAD maintainers just replied on the issue I've opened after closing this PR.

For us, it's allowed to have a minus in the node name.

See their comment here https://github.com/KiCad/kicad-packages3D/issues/636#issuecomment-688744765

I'm not comfortable enough with the spec to tell. @Mugen87 could you have a look?

mrdoob:
I see... So dash (—) is allowed, but minus (-) is not?

mrdoob:
Looking at `Battery_CR1225.wrl` in [Battery.3dshapes.7z](https://kicad.github.io/download/packages3d/Battery.3dshapes.7z) I can see:

```
appearance Appearance {material DEF PIN-01 Material {
```

That's a minus `-` and not a dash`—`... 🤔

mrdoob:
Hmm, https://www.web3d.org/documents/specifications/14772/V2.0/part1/concepts.html#4.1.1 seems to be a newer spec.

I think I would just follow KiCAD on this one.

tuarrep:
Indeed, I've mixed up between dash and minus (I'm not an English native).

The original PR was made to allow the minus (-) sign, not the dash.

mrdoob:
@tuarrep seems like you deleted the branch for this PR and I can't reopen it.

Mugen87:
> I think I would just follow KiCAD on this one.

👍

tuarrep:
> @tuarrep seems like you deleted the branch for this PR and I can't reopen it.

Yes. I'll reopen a fresh one rebased on the current master.

tuarrep:
#20321

PR #18302, opened 2020-01-04
Fixed binary compressed color in PCDLoader

sniklaus:
Fixed a small bug in the PCDLoader where the color of binary compressed data was not loaded correctly. A parenthesis was in the wrong spot which erroneously divided the address of the color by 255 instead of dividing the actual color value.

### DIFF
--- a/examples/js/loaders/PCDLoader.js
+++ b/examples/js/loaders/PCDLoader.js
@@ -298,9 +298,9 @@ THREE.PCDLoader.prototype = Object.assign( Object.create( THREE.Loader.prototype
 
 				if ( offset.rgb !== undefined ) {
 
-					color.push( dataview.getUint8( ( PCDheader.points * ( offset.rgb + 2 ) + PCDheader.size[ 3 ] * i ) / 255.0 ) );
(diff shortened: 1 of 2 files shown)


WestLangley:
Thank you for your contribution.

You will need to fix the file in the `js` directory, instead.

The `jsm` file is auto-generated, but it is ok to include both files in your PR.

sniklaus:
Thank you for your prompt response and for letting me know, I went ahead and also updated the file in the `js` directory!

mrdoob:
Thanks!

Mugen87:
Um, it's not possible to load the binary compressed test file anymore:

https://github.com/mrdoob/three.js/blob/dev/examples/models/pcd/binary_compressed/pcl_logo.pcd

The following runtime error is reported:

> Uncaught RangeError: Offset is outside the bounds of the DataView

It seems this happens when the loop variable `i` has a value of `6455`.

@sniklaus Can you please have a look what's going wrong?

Mugen87:
`pcl_logo` looked like so before:

image

sniklaus:
My apologies, I will be looking into this later today and issue another pull request with a fix.

sniklaus:
Looks like not only the parenthesis was at the wrong position, but also the calculation of the color offset was incorrect. Looks like expected now, will be issuing a pull request in a minute.

PR #18317, opened 2020-01-07
Shaders: clarified method limitations

WestLangley:
`inverseTransformDirection()` only works in certain circumstances. The inline comment now makes that clear.

For the record, the method will also work if the matrix is a pure-rotation matrix with uniform-scale, but in our use case, the matrix is `viewMatrix`, which is orthogonal.


mrdoob:
Thanks!

PR #18321, opened 2020-01-07
SVGLoader: Implement units

### LINKED ISSUE #17630: SVGLoader - issue with unit conversion

yomboprime:
Fixes #17630

Implemented the 6 most common units: ```px```, ```mm```, ```in```, ```cm```, ```pt```, ```pc```.

The ```em```, ```ex``` and ```%``` unit specifiers were not possible to implement because there is no text implemented to refer to, nor viewports.

Default DPI are 90, which can be changed in ```svgLoader.defaultDPI```.

Also the ```svgLoader.defaultUnit``` member can be set to one of the 6 unit strings, to set the loader output units (default is ```px```, i.e. user units)

For testing, I added a new SVG  test file (named "Units") in which three rectangles with equal widths are defined in different units.


Mugen87:
For testing: http://raw.githack.com/yomboprime/three.js/svgunits/examples/webgl_loader_svg.html

Mugen87 (approved):

examples/js/loaders/SVGLoader.js
```diff
@@ -8,6 +8,12 @@ THREE.SVGLoader = function ( manager ) {
 
 	THREE.Loader.call( this, manager );
 
+	// Default dots per inch
+	this.defaultDPI = 90;
+
+	// Accepted units: 'mm', 'cm', 'in', 'pt', 'pc', 'px'
+	this.defaultUnit = "px";
```
Mugen87: You might want to add these two new properties to the TS file.

yomboprime:
Done 👍

mrdoob:
Thanks!

PR #18328, opened 2020-01-07
simplified soft shadow shader

### LINKED ISSUE #18315: SoftShadows turn MeshStandardMaterial yellow on Pixel 3

I assume this is a driver bug on the Pixel 3, and it's probably long-standing; it repros on dev, as well as r112 and r110. I haven't tried any more versions. Try this [fiddle](https://jsfiddle.net/elalish/51Lbtuck/9/); normally it's just a white sphere:
image
But on Pixel 3, specifically with this line enabled:
`renderer.shadowMap.type = THREE.PCFSoftShadowMap;`
We get yellow:
image

##### Three.js version

- [x] Dev
- [x] r112
- [ ] ...

##### Browser

- [ ] All of them
- [x] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [x] Android
- [ ] iOS

Pixel 3

elalish:
Fixes #18315 

I tested this in one particular configuration, and given that the original issue appears to be a glsl compiler bug resulting in undefined behavior, I'd appreciate anyone else confirming this is working.

Since the bug seemed to be triggered by shader complexity, I refactored this to make it more efficient and this seems to have fixed the bug as well. I removed the dependence on radius for PCFSoftShadowMap, as anything other than 1 gave a non-smooth shadow anyway. For radius 1, this math should give an identical result to the previous version, but with far fewer divides and texture look-ups. Of course the compilers might have optimized some of those out anyway, but who knows?

### DIFF
--- a/docs/api/en/lights/shadows/LightShadow.html
+++ b/docs/api/en/lights/shadows/LightShadow.html
@@ -74,8 +74,8 @@
 
 			High values will cause unwanted banding effects in the shadows - a greater [page:.mapSize mapSize]
 			will allow for a higher value to be used here before these effects become visible.<br />
-			If [page:WebGLRenderer.shadowMap.type] is set to [page:Renderer PCFSoftShadowMap], it is recommended to
-			leave radius at 1 and increase softness by decreasing [page:.mapSize mapSize] instead.<br /><br />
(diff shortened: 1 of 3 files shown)


mrdoob:
Thanks!

PR #14044, opened 2018-05-12
Make Matrix4.compose more efficient

mqp:
This is a CPU hotspot for rendering if there are a lot of objects in your scene graph. This change makes it moderately more efficient by calculating & writing the resulting matrix values just once, instead of multiple times in the course of applying rotation-scaling-position.

According to my testing, this patch improves overall performance on the `updateMatrixWorld` benchmark in the repository (which just tests calling `updateMatrixWorld` at the root of a large scene graph, like the renderer does) by roughly 10-15% on both Firefox and Chrome.

PR #18344, opened 2020-01-08
Add 'setMatrix' method to Object3D

willcassella:
This method updates the 'position', 'quaternion', and 'scale' properties
from the 'matrix' property, the inverse of 'updateMatrix'.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -145,6 +145,16 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 	},
 
+	setMatrix: function ( m ) {
+
+		this.matrix.copy( m );
+
+		this.matrix.decompose( this.position, this.quaternion, this.scale );
+
+		this.matrixWorldNeedsUpdate = true;
+
+	},
+
 	setRotationFromAxisAngle: function ( axis, angle ) {
 
 		// assumes axis is normalized


DavidPeicho:
Love the shortcut! I am super often doing that manually.
Don't know if the name is super clear about what it does though.

Mugen87:
Just want to note: It was considered to make `Object3D.matrix` private so it's easier to implement performance optimizations in `Object3D. If we introduce this method, it's more unlikely to remove `Object3D.matrix` from the public interface.

DavidPeicho:
I didn't know the existence of  #14138, but now I am voting to migrate `matrix` to be private 😄

Mugen87:
AFAIK, Unity does not expose the local matrix. Not even as a read-only property. The only way to get it is the usage of [Matrix4x4.TRS](https://docs.unity3d.com/ScriptReference/Matrix4x4.TRS.html) which is the equivalent to `Matrix4.compose()`. If I remember correctly, `Babylon.js` does not expose the local matrix, too.

DavidPeicho:
I am definitely not against putting `matrix` private. Though it would be great to have a nice way of manipulating it. Maybe just `fromMatrix4`. `applyMatrix4` can't be used when you wanna start from an identity matrix.

Mugen87:
> Maybe just fromMatrix4.

This can also be done on application level if it's really necessary. I don't think the average user actually needs to work with matrices. It should be an internal thing of the engine.

willcassella:
I decided to refactor it to `setMatrix`, also giving it the responsibility of assigning to the internal matrix. That way if the `matrix` field is ever made private this API will continue to function. This particular change sprung about because the WebXR API mainly hands out matrices instead of the translation/rotation transforms people are usually familiar with, so having a way to set it directly is nice even if it's a bit of a niche use-case. It's also a win for performance, since then you don't end up decomposing the matrix yourself to pass the components to Three.js only to have it recompose them back into a matrix :)

WestLangley:
If you want to update the `matrix` directly, instead of the `position`, `quaternion`, and `scale`, then you should set

```js
object.matrixAutoUpdate = false;
```

Then, as needed,
```js
object.matrix.copy( matrix );
```

That is how three.js was designed to be used. In this case, the `position`, `quaternion`, and `scale` are not used, and the user can set the matrix directly.


If there is an issue with the world matrix not updating, then you may have to set the `object.matrixWorldNeedsUpdate` flag.

Mugen87:
Against the current API status, `setMatrix()` is somewhat contradictory since you can already use the approach @WestLangley mentioned. If you need up-to-date `position`, `rotation` and `scale` properties, you can perform the decompose on application level. It's just a one-liner.

If `Object3D.matrix` is made private, `matrixAutoUpdate` will be removed, too. A method like `setLocalMatrix()` will make more sense then. 

In any event, the current handling of `matrixAutoUpdate` and `Object3D.matrix`  is confusing since the  engine allows object properties  (`position`, `rotation` and `scale`) to have outdated/incorrect values. Another reason for not having the local matrix exposed.

WestLangley:
>If Object3D.matrix is made private, matrixAutoUpdate will be removed, too.

`matrixAutoUpdate` serves two very-useful purposes. I can't imagine it being removed. Can you please provide a bit more detail as to what you mean?

>the current handling of matrixAutoUpdate and Object3D.matrix is confusing since the engine allows object properties (position, rotation and scale) to have outdated/incorrect values.

Can you provide an example (or two if needed) to explain what you are referring to? Why do you say the position be "outdated" or "incorrect"?

Mugen87:
> Can you please provide a bit more detail as to what you mean?

If you introduce a cache a dirty flag mechanism to avoid unnecessary matrix computations, `matrixAutoUpdate` is not necessary anymore. This was all discussed in #14138, #14543 and #15706. We could not implement the approaches because `Object3D.matrix` is public and writable.

> Can you provide an example (or two if needed) to explain what you are referring to? 

I just mean if you set the matrix of a 3D object but not update its `position`, `rotation` and `scale` properties at the same time, the state of the object is wrong. Three properties represent values which are not correct.

WestLangley:
> If Object3D.matrix is made private, matrixAutoUpdate will be removed, too.
> matrixAutoUpdate is not necessary anymore

Actually, that is not true.  Users have to be able to set the matrix directly, as I explained above. Shear matrices, for example, cannot be represented by a `position`, `quaternion`, and `scale`. Users can set the matrix and also set `matrixAutoUpdate` to `false`. The `position`, `quaternion`, and `scale` are not used in that case.

Mugen87:
>  Users have to be able to set the matrix directly, as I explained above.

You can't do this in Unity so I really don't understand why this an important feature. Where do we set shear matrices in the example?

> The position, quaternion, and scale are not used in that case.

They might be not used but they are still there. And might habe outdated/wrong values that might confuse users.

WestLangley:
>You can't do this in Unity so I really don't understand why this an important feature. Where do we set shear matrices in the examples?

This has nothing to do with the examples. It has to do with a feature set users rely on.

If you are interested in learning more about this topic you can read a post I [filed](https://github.com/mrdoob/three.js/issues/3845) in 2013.

Setting the matrix directly is useful when, for example, the matrix data is streamed from another source.

DavidPeicho:
Modifying the matrix is indeed sometimes useful.

What would be the issue if the modification was made through a public access, like `setLocalMatrix()`? If there was a function like that, you could set the dirty flag to `true` there, and then you have the best of both world.

Mugen87:
@DavidPeicho  Keep in mind, that `setLocalMatrix()` would need to perform a decompose in order to keep the `position`, `rotation` and `scale` properties in sync. So it's not even a performance win. Because of https://github.com/mrdoob/three.js/issues/3845#issuecomment-23919272, I would not provide it and leave it up to app-level code. BTW: In this context, I would also remove `Object3D.applyMatrix()`.

I'm still convinced that the engine does not need a `setLocalMatrix()`. You can represent the transformation of a 3D object just with three properties and keep things much simpler in this way. I mean if bigger and more popular engines like Unity follow this approach, we could do the same. 

Anyway, I know how the discussions usually proceed when talking with @WestLangley about math topics. I made my point clear and quit here since we won't come to a common conclusion.

DavidPeicho:
Yes it would need the `decompose()` call, but I guess that's fine, as long as the user knows about it. It's an "advanced" usage IMO. Anyway, I really like the ideas expressed in !14138, would be awesome to have.

EliasHasle:
> You can represent the transformation of a 3D object just with three properties and keep things much simpler in this way. I mean if bigger and more popular engines like Unity follow this approach, we could do the same.

It is an argument that we _could_, not one that we _should_. Unity's popularity may be for other reasons. And being big doesn't mean being right. I personally prefer the three.js ways to Unity' s. With the status quo, three.js is easy to use, yet allows advanced usage. I have been thinking about experimenting with sheer matrices and other out-of-the-box transformations, though not done it yet. For animations or morphing of exotic transforms, modifying the geometry will not compete.

Mugen87:
@EliasHasle Please keep in mind that the matrix multiplication overhead was one reason that triggered this discussion about making `Object3D.matrix` private. The current matrix multiplication overhead in `three.js` does not compete with other 3D engines.

> Unity's popularity may be for other reasons.

It's not only Unity. Other engines do similar things. I've actually not seen a single engine that allows to directly set the local matrix on a transform node.

WestLangley:
>The current resulting overhead of the current implementation in three.js does not compete with other 3D engines.

@Mugen87  Based on what? If you can provide actual data to back up that claim for non-contrived use cases I'd like to see it.

Setting the local matrix directly with  `matrixAutoUpdate = false` removes all overhead. That's the point.

>I would also remove Object3D.applyMatrix().

@Mugen87 Why don't you want users to be able to call `applyMatrix()`, either?

Mugen87:
> Setting the local matrix directly with matrixAutoUpdate = false removes all overhead. That's the point.

Another point is that almost no `three.js` user set `matrixAutoUpdate` to `false`. A better design is to avoid matrix computations automatically like most other engines do. In this way, users don't have to bother with flags. I've never liked `matrixAutoUpdate` or all this `needsUpdate` stuff. They are error prone and not user friendly.

> @Mugen87 Why don't you want users to be able to call applyMatrix(), either?

In order to avoid decompose operations. In my presented approach, it would be necessary to sync `position`, `rotation` and `scale`.

EliasHasle:
@WestLangley Maybe you can win Michael over by making a really cool skew matrix demo. 🥇 

(Disappointingly, my web searches on "skew matrix 3d demo" and "shear matrix  3d demo" returned nothing. 🤔)

Mugen87:
> Maybe you can win Michael over by making a really cool skew matrix demo. 

TBH, I don't get the point of constructing a use case  which seems not relevant to much larger communities of other projects.

sciecode:
I share @WestLangley's opinion on keeping the local public. My argument being that we have a more versatile system than other engines. 

While it is true that there can be significant overhead in automatically updating the matrices, we also offer the possibility of updating it manually for users that wish to do so. In that scenario, it's hard to say three.js will perform any worse than the aforementioned engines.

If anything, I would improve the current API for manually updating matrices, so that beginner users feel less intimidated about controlling the update workflow themselves. Avoiding "unnecessary" API breaking changes.

Mugen87:
https://github.com/mrdoob/three.js/pull/14044

sciecode:
@Mugen87 the numbers are not painting the whole picture. In fact, I think the thread provides further evidence of why we shouldn't go towards this design. 

The linked comment only provides a comparison against `matrixAutoUpdate = true` and in the case where the objects are static, which is a fairly unfair comparison. If we were to compare the same method against `matrixAutoUpdate = false`, the proposed method would perform worse.

Further along the thread, there are comparisons made on dynamic objects and in that scenario `matrixAutoUpdate = true` clearly performs better than any of the options presented. As it is expected.

So, right now we have a versatile system that individually offer the best possible performance out of the recommended use cases. Why would we move towards a fixed design that takes away the ability of application-level optimizations and performs worse on every instance?

You could argue that for a naive implementation the proposed design would on average perform better, which likely a fair statement. But, in my eyes, this is trying to solve a problem by deflecting the real issue. Which is actually educating the user base on the correct usage of three.js. We should better  document the implications of having `matrixAutoUpdate = true/false` and expose feasible examples of the recommended usage.

I won't further extend this discussion, I just wanted to share my view on the issue.

Mugen87:
> But, in my eyes, this is trying to solve a problem by deflecting the real issue.

The real issue is a design flaw in the engine. I think it's wrong to educate user fiddling around with flags if things could be nicely automated.

And yes, I clearly prefer a solution that performs better for the vast majority of the user base.

EliasHasle:
Three.js users range from novices to experts. It has a stated aim of being easy to use, but it must also leave advanced users some freedom/power to keep them interested. An advanced user does not value being held in their hand while their wings are being clipped.

Mugen87:
From the readme file:

> The aim of the project is to create an easy to use, lightweight, 3D library with a default WebGL renderer. 

`three.js` was always focused on making beginners in computer graphics and especially ordinary web devs coding with WebGL easier. The API should always reflect this goal. 

If the library can improve the average performance and API for all these users, I would say we should go for it.

>  but it must also leave advanced users some freedom/power to keep them interested. 

Not sure I follow this argumentation. I would say engines like Unity do attract advanced users, no?

EliasHasle:
> > but it must also leave advanced users some freedom/power to keep them interested.

> Not sure I follow this argumentation. I would say engines like Unity do attract advanced users, no?

There are many reasons Unity does not attract me. One is the relative closed-ness of the source and engine development. One is the heavy weight. One is having to use a particular IDE/Editor. One is the lack of control over global code flow when used in the intended/obvious way ("agent" scripts attached to in-game objects). One is the difficulty with interfacing to other code. For some advanced users, all these properties may be outweighed by good properties, and many may be worked around if one has sufficient resources. But I don't expect this group of advanced users to be the same as the advanced users of three.js (although some users may fit in both groups).

trusktr:
If we get rid of .matrix, perhaps we should at least have more properties that can be useful for various reasons. For some of the below concepts, we currently need the ability to manipulate matrices directly because Object3D doesn't have all the features:

- skewX
- skewY
- rotation
- position
- scale
- origin (for scale, skew, and rotation, preferably separate for each for the most flexibility, without having to make additional tree nodes in cases where the values of these properties are simpler).

See for example skew in web standard CSS transforms (f.e. [skew](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/skew)) and [DOMMatrix](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix#instance_properties).

A few examples of skew:

https://codepen.io/longfurbyqueen/pen/YLdVjw

https://codepen.io/enbee81/pen/yLyrmyg

https://codepen.io/trusktr/pen/ExObxEX

https://codepen.io/agaroff/pen/erjwzv

Mugen87:
Just want to note that my previous opinion has changed. If `Object3D.matrix` would be private, I'm okay with having `Object3D.setMatrix()` as a convenience method.

Mugen87:
There is no consensus yet how to proceed with the update matrix lifecycle so it's not clear if `setMatrix()` is wanted. At the moment, there are no plans to make `Object3D.matrix` private. I understand `setMatrix()` would still be interesting as a convenient method but unfortunately it gained to less interest. Closing for now.

PR #18386, opened 2020-01-13
Build: switch to a custom class transform using Babel

Andarist:
After a short [Twitter discussion](https://twitter.com/mrdoob/status/1216138181189537794) about making Three more tree-shakeable and a private conversation with @mrdoob about it I've prepared this PR.

One of the greatest tree-shaking offenders are:
- top-level scope function calls
- top-level property reads & writes

So continuing to use a custom, hand-written, classes with prototypes manipulation you fall into this trap. The recommendation on this one is to just rewrite your classes to class syntax - you have a "legacy" build that has to support es5 environments though. So given the fact that various tools can produce a sub-optimal output for you (because they operate on a generic case, they often need to handle edge cases etc) and you are not so fond about that, the preferred way to handle this situation is to just write a custom class transform, which can output any code structure you like.

This PR doesn't implement all the stuff that is needed to **really** support transpiling ES classes, but I throw various errors in the transform itself, which should help you catch unsupported cases early. And the logic can always be extended to handle more.

I've also converted 2 classes - `AnimationClip` & `AnimationAction` - as it made my development of this easier (I had more test cases with them, so to speak).

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
   },
   "homepage": "https://threejs.org/",
   "devDependencies": {
+    "@babel/core": "^7.8.0",
     "@types/webgl2": "^0.0.5",
     "@typescript-eslint/eslint-plugin": "^2.15.0",
     "@typescript-eslint/parser": "^2.15.0",
@@ -84,7 +85,7 @@
     "http-server": "^0.12.0",
     "qunit": "^2.9.3",
     "rollup": "^1.28.0",
-    "rollup-plugin-buble": "^0.19.8",
+    "rollup-plugin-babel": "^5.0.0-alpha.1",
     "typescript": "^3.7.4"
   },
   "jspm": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,118 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/core": {
+      "version": "7.8.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.8.0.tgz",
+      "integrity": "sha512-3rqPi/bv/Xfu2YzHvBz4XqMI1fKVwnhntPA1/fjoECrSjrhbOCxlTrbVu5gUtr8zkxW+RpkDOa/HCW93gzS2Dw==",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "^7.8.0",
+        "@babel/generator": "^7.8.0",
+        "@babel/helpers": "^7.8.0",
+        "@babel/parser": "^7.8.0",
+        "@babel/template": "^7.8.0",
+        "@babel/traverse": "^7.8.0",
+        "@babel/types": "^7.8.0",
+        "convert-source-map": "^1.7.0",
+        "debug": "^4.1.0",
+        "gensync": "^1.0.0-beta.1",
+        "json5": "^2.1.0",
+        "lodash": "^4.17.13",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
+      },
+      "dependencies": {
+        "@babel/code-frame": {
+          "version": "7.8.0",
+          "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.0.tgz",
+          "integrity": "sha512-AN2IR/wCUYsM+PdErq6Bp3RFTXl8W0p9Nmymm7zkpsCmh+r/YYcckaCGpU8Q/mEKmST19kkGRaG42A/jxOWwBA==",
+          "dev": true,
+          "requires": {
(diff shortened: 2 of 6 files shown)


Mugen87:
Notice that `BoxGeometry` is already converted to a class for testing purposes. Also note #17425 which converts all geometry generator to classes.

It was not a matter of time why ES6 class migration stopped  last year. One reason was the fact that an inheritance issue become obvious mentioned in https://github.com/mrdoob/three.js/pull/17425#issue-313966874.

donmccurdy:
The issue mentioned in #17425 about ParametricGeometries is not a blocker, correct? It will just take us more work to migrate these examples to more standard ES6 class inheritance?

package.json
```diff
@@ -84,7 +85,7 @@
     "http-server": "^0.12.0",
     "qunit": "^2.9.3",
     "rollup": "^1.28.0",
-    "rollup-plugin-buble": "^0.19.8",
+    "rollup-plugin-babel": "^5.0.0-alpha.1",
```
donmccurdy: Do we need something in the 'alpha' version here, or would a stable version be better?

donmccurdy (commented):
@Andarist thanks for this!

Mugen87:
> The issue mentioned in #17425 about ParametricGeometries is not a blocker, correct?

No, I hope it's not^^. But yes, the impact of this change is bigger than initially assumed since a lot of example code has to be converted, too.

Mugen87:
Anyway, I really like to push this topic forward. I can invest more time in class migration in the next time since I would love to see smaller builds^^.

donmccurdy:
>  lot of example code has to be converted, too.

Do you know roughly how much? Everything that extends a core class, like `Loader`?

If so, the `examples/` migration will probably also break our current `modularize` script (which we want to get away from anyway). Maybe we move things to ES6 classes and ES6 modules at the same time, i.e. removing the versions in the `examples/js/` folder in favor of `examples/jsm/`.

Mugen87:
> If so, the examples/ migration will probably also break our current modularize script (which we want to get away from anyway).

Do you know why? It seems the script can handle the conversion of `ParametricGeometries`.

> Do you know roughly how much? Everything that extends a core class, like Loader?

I would convert everything, similar to the module migration. Yes, this will take some time but I would prefer a consistent code base if possible. However, this is just my personal opinion. I accept whatever @mrdoob prefers here. Same for the `js`/`jsm` topic.

donmccurdy:
I was expecting 'modularize.js' did not support ES6 classes, but haven't tested that.

Andarist:
> Notice that BoxGeometry is already converted to a class for testing purposes.

Yeah, I've seen that - it's just a simple class with just a constructor though. That's why I've converted two separate classes here, as they have methods and static methods.

> One reason was the fact that an inheritance issue become obvious mentioned in #17425 (comment).

Hmm, yes - that's true. Native classes cannot be tricked with `.call` or anything like that. Once you put a native class into an inheritance chain you have to extend with class syntax. Is this a problem for your consumers? I would imagine that if they need to transpile stuff to support older envs, they also have to transpile your code (so your classes) and then the issue goes away because they are not "classes" anymore.

package.json
Andarist: Well, I'm actually maintainer of that plugin 😅 We are in the process of migrating it into the rollup plugins monorepo - so the final v5 should get released this month. There really won't be any breaking changes for you here. This one is preferable over v4 because it's slightly faster and we can use `skipPreflightCheck` option to avoid unnecessary work (which conflicts a little bit with this custom class transformer introduced in this PR).

Mugen87:
>  Is this a problem for your consumers?

Yes, but also for the project itself since there are files in the examples directory which inherit from core files. For example all loaders, geometry generators and some other single files (like `CinematicCamera`).

Andarist:
> But I think this should not block a conversion since we can use the migration guide in order to highlight all breaking changes and explain what migration tasks are necessary.

That would be sweet. Your files can just be converted, but of course, it will take some time. And the migration guide should be enough for your consumers 👍 You'd only have to establish how to proceed with this step by step (the overall migration, not only this PR).

Andarist:
Are there any action points regarding the discussion here? Is there anything i can do to help migrating to es classes? I would love to improve tree-shakeability of three and having those “manual” prototypes manipulation here is the biggest deopt for that

donmccurdy:
Seems like we need to decide the right order to migrate directories. Not being able to extend ES6 classes from prototype-style "classes" unfortunately means we'll likely have to convert things in `examples/js/*` (which don't actually benefit from tree shaking) first before things in the `src/*` folder.

/cc https://github.com/mrdoob/three.js/pull/17425#issuecomment-587551098

Mugen87:
It was lately discussed that the ES6 class conversion will be continued after `examples/js` was removed at the end of the year. Closing for now.

PR #18399, opened 2020-01-16
WebGLRenderer: Procedural uniforms update.

sunag:
Basically this turns `WebGLRenderer` more flexible and this too is a simplification of `uniforms update`.

Result for tests:
https://raw.githack.com/sunag/three.js/dev-uniformsupdate-test/examples/index.html?q=material

Related:
https://github.com/mrdoob/three.js/pull/18162
https://github.com/mrdoob/three.js/pull/18175
https://github.com/mrdoob/three.js/pull/18068


sunag:
All refresh material functions: `refreshUniformsCommon`, `refreshUniformsLambert`, `refreshUniformsToon`, `refreshUniformsPhong`, `refreshUniformsPhysical`... is replaced to `refreshUniforms`:

https://github.com/sunag/three.js/blob/757b052cd34e4a89418200bc8f6f73ba409822f8/src/renderers/WebGLRenderer.js#L1944-L1972


Properties identical in `material` and `uniform` are updated automatically. In case not need the automatic refresh the user can add `.needsRefresh = false` in your `uniform`.

In special cases, like `diffuse` or `envMap` is possible use `.onUpdate` to a customized update.
`Uniform.onUpdate ( uniforms, material, renderer, scene ) `

Note that there has been a **reduction in the amount of code**, that would be even bigger if there were more materials.

`NodeMaterial` update like https://github.com/mrdoob/three.js/pull/18162 became possible.

/ping @mrdoob @WestLangley @bhouston @donmccurdy @Mugen87

src/renderers/WebGLRenderer.js
```diff
@@ -1882,91 +1881,15 @@ function WebGLRenderer( parameters ) {
 
 			}
 
-			// refresh uniforms common to several materials
+			// refresh uniforms common to materials
 
 			if ( fog && material.fog ) {
 
 				refreshUniformsFog( m_uniforms, fog );
```
Mugen87: What about `refreshUniformsFog()`? Could this also be replaced?

src/renderers/WebGLRenderer.js
```diff
@@ -2018,285 +1941,32 @@ function WebGLRenderer( parameters ) {
...
+			var uniform = uniforms[ property ];
 
-			uvScaleMap = material.normalMap;
+			if ( uniform.onUpdate ) {
 
-		} else if ( material.bumpMap ) {
+				uniform.onUpdate( uniforms, material, _this, scene );
 
-			uvScaleMap = material.bumpMap;
+			} else if ( uniform.needsRefresh !== false && material[ property ] !== undefined ) {
```
Mugen87: Where does `needsRefresh` come from?

src/renderers/WebGLRenderer.js
```diff
@@ -2018,285 +1941,32 @@ function WebGLRenderer( parameters ) {
...
 
-		} else if ( material.roughnessMap ) {
+				var value = material[ property ];
 
-			uvScaleMap = material.roughnessMap;
+				if ( value !== null && typeof value === 'object' && !value.isTexture ) {
```
Mugen87: `typeof value === 'object'` is this safe? I think I would explicitly check for the respective classes via `isVector2`, `isVector3`, `isVector4`, `isMatrix3` and `isMatrix4`.

src/renderers/shaders/UniformsLib.js
```diff
@@ -6,16 +7,162 @@ import { Matrix3 } from '../../math/Matrix3.js';
...
-		diffuse: { value: new Color( 0xeeeeee ) },
+		diffuse: { value: new Color( 0xeeeeee ), onUpdate: function( uniforms, material ) {
+
+			if ( material.color ) {
+
+				uniforms.diffuse.value.copy( material.color );
+				
+			}
+
+		} },
```
WestLangley: Nit: suggested formatting pattern:
```js
var UniformsLib = {

	common: {

		diffuse: {

			value: new Color( 0xeeeeee ),

			onUpdate: function ( uniforms, material ) {

				if ( material.color ) {

				}

			}

		},

		opacity: { value: 1.0 },

		map: { value: null },

		uvTransform: {

			value: new Matrix3(),

			onUpdate: function ( uniforms, material ) {

				// uv repeat and offset setting priorities
				// 1. color map
				// 2. specular map
				// 3. normal map
				// 4. bump map
				// 5. alpha map
				// 6. emissive map

				var uvScaleMap;

				if ( material.map ) {

					uvScaleMap = material.map;

				} else if ( material.specularMap ) {

					uvScaleMap = material.specularMap;

				} else if ( material.displacementMap ) {

					uvScaleMap = material.displacementMap;

				} else if ( material.normalMap ) {

					uvScaleMap = material.normalMap;

				} else if ( material.bumpMap ) {

					uvScaleMap = material.bumpMap;

				} else if ( material.roughnessMap ) {

					uvScaleMap = material.roughnessMap;

				} else if ( material.metalnessMap ) {

					uvScaleMap = material.metalnessMap;

				} else if ( material.alphaMap ) {

					uvScaleMap = material.alphaMap;

				} else if ( material.emissiveMap ) {

					uvScaleMap = material.emissiveMap;

				}

				if ( uvScaleMap !== undefined ) {

					// backwards compatibility

					if ( uvScaleMap.isWebGLRenderTarget ) {

						uvScaleMap = uvScaleMap.texture;

					}

					if ( uvScaleMap.matrixAutoUpdate === true ) {

						uvScaleMap.updateMatrix();

					}

					uniforms.uvTransform.value.copy( uvScaleMap.matrix );

				}

			}
		},
```

src/renderers/WebGLRenderer.js
sunag: I think that no for now. `fog` and `lights` come out of the pattern of material properties.

src/renderers/WebGLRenderer.js
sunag: This was to prevent conflict of  `uniform` and `material` property with same name but without interaction each other. But this seems unnecessary, I will remove it for now.

src/renderers/WebGLRenderer.js
sunag: Yes, this was to be temporary. Your suggestion is better.

WestLangley:
Everything appears to be working in my test cases so far...

src/renderers/WebGLRenderer.js
Mugen87: > fog and lights come out of the pattern of material properties.

Um, I'm not sure I understand this. Can you please explain in more detail? 😇

sunag: I was based on the signature of the function, I forgot that I add `scene` parameter in the function. I add this feature 👍

src/renderers/shaders/UniformsLib.js
WestLangley: This can help you pass the formatting test: https://zz85.github.io/mrdoobapproves/

taphos:
@sunag is this ticket abandoned?

sunag:
With `MaterialNode` and new field in materials for nodes like `mateirla.colorNode` I think that we need not work more at this point for now.
https://github.com/mrdoob/three.js/pull/21322

PR #18400, opened 2020-01-16
Example: Initial GUI for webgl2_materials_texture2darray.

linbingquan:
I think this demo should have GUI to control the `depth` value, but I'm not sure this PR is OK,  if it needs to change, please tell me.

### DIFF
--- a/examples/webgl2_materials_texture2darray.html
+++ b/examples/webgl2_materials_texture2darray.html
@@ -62,6 +62,7 @@
 
 			import Stats from './jsm/libs/stats.module.js';
 			import { WEBGL } from './jsm/WebGL.js';
+			import { GUI } from './jsm/libs/dat.gui.module.js';
 
 			if ( WEBGL.isWebGL2Available() === false ) {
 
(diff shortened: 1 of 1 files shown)


Mugen87:
TBH, I would keep the example simple and not add a GUI. Because of the existing animation, the depth already becomes quite obvious.

linbingquan:
> TBH, I would keep the example simple and not add a GUI. Because of the existing animation, the depth already becomes quite obvious.

Yes, I agree with you. I read a CT patent news yesterday, I remembered `three.js` has a similar demo as [materials / texture2darray](https://threejs.org/examples/?q=webgl2_materials_texture2darray#webgl2_materials_texture2darray), I can't pause it or select it what I want to see, so I created this PR in my spare time, if it isn't necessary, I will closed it, how do you think.

linbingquan:
> TBH, I would keep the example simple and not add a GUI. Because of the existing animation, the depth already becomes quite obvious.

Yes, I agree with you. I read a CT patent news yesterday, I remembered `three.js` has a similar demo as [materials / texture2darray](https://threejs.org/examples/?q=webgl2_materials_texture2darray#webgl2_materials_texture2darray), but I can't pause it or select it what I want to see, so I created this PR in my spare time, if it is not necessary, I will closed it, how do you think.

PR #18406, opened 2020-01-16
Allow setting opacity in LineMaterial example

OlivierKamers:
This  PR adds support for setting the opacity of the example LineMaterial.

Code was changed in the `examples/js` directory and compiled to `examples/jsm` using the util script. However, I'm not sure if we're supposed to add this output to a PR.


WestLangley:
Thanks for this, but please see the discussion in #16570.

OlivierKamers:
Thanks, I missed those related tickets. I'll look into the alternatives.

PR #18407, opened 2020-01-16
Added new tone mapping example

WestLangley:
This is a complete re-write, and is focused on tone mapping only. It is based off of the `GLTFLoader` example. (see https://github.com/mrdoob/three.js/pull/18330#issuecomment-573401941)

1. The hdr environment map has an increased dynamic range.
2. The scene background is tone mapped, and is properly-updated when tone mapping is changed.
3. The GUI menu items are adaptive, and inlclude tone mapping parameters only
4. EffectComposer has been removed.
5. The model better-demonstrates the tone mapping subtleties than the previous model (IMO, anyway)

### DIFF
--- a/examples/webgl_tonemapping.html
+++ b/examples/webgl_tonemapping.html
@@ -1,40 +1,38 @@
 <!DOCTYPE html>
 <html lang="en">
 	<head>
-		<title>three.js webgl - inline tone mapping</title>
+		<title>three.js webgl - tone mapping</title>
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
 		<link type="text/css" rel="stylesheet" href="main.css">
 	</head>
-	<body>
 
-		<div id="container"></div>
-		<div id="info"><a href="https://threejs.org" target="_blank" rel="noopener">threejs</a> - Inline Tone Mapping (within a Material's fragment shader) without<br/>using a pre-processing step or float/half buffers by <a href="http://clara.io/" target="_blank" rel="noopener">Ben Houston</a>.</div>
+	<body>
+		<div id="info">
+			<a href="https://threejs.org" target="_blank" rel="noopener">three.js</a> - Tone Mapping<br />
+			Battle Damaged Sci-fi Helmet by
+			<a href="https://sketchfab.com/theblueturtle_" target="_blank" rel="noopener">theblueturtle_</a><br />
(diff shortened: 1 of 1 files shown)


WestLangley:
temporary live link:
https://raw.githack.com/westlangley/three.js/dev_tonemapping_example2/examples/webgl_tonemapping.html

WestLangley:
Hijacking my own thread...

@donmccurdy @elalish In this example, I did not use `RoughnessMipmapper`. It is easy to compare this example with the `GLTFLoader` example, which does use it.

The differences are visible, but not compelling in this case -- at least IMHO.

I'm thinking we should have an example that shows a compelling use case of `RoughnessMipmapper`.

elalish:
@WestLangley RoughnessMipmapper only has an effect when mipmapping would have an effect. Therefore it is primarily noticeable when you zoom out from the model (or the model has very high res textures). I think if you enable zooming in this example like in the GLTFLoader example, you will be able to see the difference when zoomed out, especially as a difference in sparkling as you rotate.

WestLangley:
@elalish  I tried `controls.maxDistance = 20` and `controls.maxDistance = 50` in each example. There are differences, but I would be hard-pressed to argue that one is "better" than the other. They both "sparkle".

I am still thinking that a compelling example is needed.

I was under the impression that `RoughnessMipmapper` helps to prevent normal maps from looking too flat when zooming out --  or is it _roughness maps_ from looking too flat? And it only applies when both are present, of course.

elalish:
@WestLangley It makes the most difference with shiny materials with high-frequency normal maps. It keeps them from becoming mirrors as you zoom out. Agreed that a dedicated test object would be helpful here. The ones I've been using I don't have rights to, unfortunately.

elalish:
@WestLangley A thought I hadn't considered: Ideally it would also be possible to apply RoughnessMipmapper to materials with a normal map but no roughness map, thus transforming a uniform roughness material into one that is affected by its normal map. Would you like to see a PR along those lines?

WestLangley:
@elalish I would like to see anything you are willing to work on that would help provide a compelling reason to use `RoughnessMipmapper` -- and when to use it.

elalish:
@WestLangley ah, I see what you mean now; I forgot that @mrdoob changed the skybox from a sunrise to that room. The sunrise happened to glint off the scratches near the chin in a way that demonstrated this effect, but it's nearly invisible now. I'll see if I can find a model that demonstrates this effect more uniformly.

WestLangley:
@elalish Maybe an example similar to this one: https://kosmonautblog.wordpress.com/2018/09/17/roughness-mip-maps-based-on-normal-maps/.

But even in that example, the author manually sets the LOD -- otherwise, I assume the image is too small to see the detail anyway.

Since users copy the patterns in the three.js examples, I would prefer to add a comment to the `GLTFLoader` example that the use of RoughnessMipmapper is optional -- at least for now.

mrdoob:
Thanks!

PR #18413, opened 2020-01-17
Implement polygon offset possibility for shadows

OndrejSpanel:
Polygon offset is an industry standard technique to help fighting shadow acne and Peter Panning effects. This PR allows per-material control over the polygon offset when rendering shadows. An example is added (derived from current VSM example) which allows to test various values.

Note: the example is using front face to cast shadows, because this way it is easier to demonstrate the polygon offset effect for the ground self-shadowing.

It would be possible to extend the sample to use different settings for ground / objects and to allow selecting front / back shadow casting for ground / objects separately, if desired.

Some notes: it seems polygon offset has some compatibility issues. It works only when I use OpenGL as ANGLE back-end on my Windows x64 Chrome, not when using DirectX 9 (which is default for me). When running on DX 9 ANGLE the example still runs, but the polygon offset settings have no effect at all. I currently do not see a way how could the sample test for this, there is no capability flag, no exception, only the functionality is broken.

### DIFF
--- a/examples/webgl_shadowmap_polygon_offset.html
+++ b/examples/webgl_shadowmap_polygon_offset.html
@@ -1,0 +1,239 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - Shadows with a polygon offset example </title>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
+		<link type="text/css" rel="stylesheet" href="main.css">
+	</head>
+	<body>
+		<div id="info">
+		<a href="https://threejs.org" target="_blank" rel="noopener">three.js</a> - Shadows with a polygon offset example by <a href="https://github.com/OndrejSpanel">Ondřej Španěl</a>
+		</div>
+
+		<script type="module">
+
+			import * as THREE from '../build/three.module.js';
+
+			import Stats from './jsm/libs/stats.module.js';
+			import { GUI } from './jsm/libs/dat.gui.module.js';
+
+			import { OrbitControls } from './jsm/controls/OrbitControls.js';
+
+			var camera, scene, renderer, clock, stats;
+			var dirLight, spotLight;
+			var torusKnot, dirGroup;
+			var groundMaterial, objectMaterial;
+
+			var config = {
+				'Bias': 0,
+				'Offset factor': 4,
+				'Offset units': 3000,
+			};
+
+			init();
+			animate();
+
+			function init() {
+
+				initScene();
+				initMisc();
+				
+				// Init gui
+				var gui = new GUI();
+
+				gui.add( config, 'Bias', -0.001, 0.001, 0.00001 ).min( -0.005 ).max( 0.005 ).onChange( function ( value ) {
+
+					dirLight.shadow.bias = value;
+					spotLight.shadow.bias = value;
+
+				} );
+
+				gui.add( config, 'Offset factor' ).min( -10 ).max( 10 ).onChange( function ( value ) {
+
+					groundMaterial.shadowPolygonOffsetFactor = value;
+					objectMaterial.shadowPolygonOffsetFactor = value;
+
+				} );
+
+				gui.add( config, 'Offset units' ).min( -10000 ).max( 10000 ).onChange( function ( value ) {
+
+					groundMaterial.shadowPolygonOffsetUnits = value;
+					objectMaterial.shadowPolygonOffsetUnits = value;
+
+				} );
+
+				document.body.appendChild( renderer.domElement );
+				window.addEventListener( 'resize', onWindowResize, false );
+
+			}
+
+			function initScene() {
+
+				camera = new THREE.PerspectiveCamera( 45, window.innerWidth / window.innerHeight, 1, 1000 );
+				camera.position.set( 0, 10, 30 );
+
+				scene = new THREE.Scene();
+				scene.fog = new THREE.Fog( 0xCCCCCC, 50, 100 );
+
+				// Lights
+
+				scene.add( new THREE.AmbientLight( 0x444444 ) );
+
+				spotLight = new THREE.SpotLight( 0x888888 );
+				spotLight.name = 'Spot Light';
+				spotLight.angle = Math.PI / 5;
+				spotLight.penumbra = 0.3;
+				spotLight.position.set( 8, 10, 5 );
+				spotLight.castShadow = true;
+				spotLight.shadow.camera.near = 8;
+				spotLight.shadow.camera.far = 200;
+				spotLight.shadow.mapSize.width = 256;
+				spotLight.shadow.mapSize.height = 256;
+				spotLight.shadow.bias = config['Bias'];
+				spotLight.shadow.radius = 2;
+				scene.add( spotLight );
+
+
+				dirLight = new THREE.DirectionalLight( 0xFFFFFF, 1 );
+				dirLight.name = 'Dir. Light';
+				dirLight.position.set( 3, 12, 17 );
+				dirLight.castShadow = true;
+				dirLight.shadow.camera.near = 0.1;
+				dirLight.shadow.camera.far = 500;
+				dirLight.shadow.camera.right = 17;
+				dirLight.shadow.camera.left = - 17;
+				dirLight.shadow.camera.top	= 17;
+				dirLight.shadow.camera.bottom = - 17;
+				dirLight.shadow.mapSize.width = 512;
+				dirLight.shadow.mapSize.height = 512;
+				dirLight.shadow.radius = 2;
+				dirLight.shadow.bias = config['Bias'];
+				scene.add( dirLight );
+
+				dirGroup = new THREE.Group();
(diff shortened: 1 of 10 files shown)


Mugen87:
> Polygon offset is an industry standard technique to help fighting shadow acne and Peter Panning effects.

Can you please explain a use case where the existing [LightShadow.bias](https://threejs.org/docs/index.html#api/en/lights/shadows/LightShadow.bias) can't be used to resolve the mentioned shadow artifacts but polygon offset does?

OndrejSpanel:
Polygon offset was introduced because it has more flexibility regarding handling surfaces which are lit at different angles. It allows to have very small bias when the surface is almost perpendicular to the light, and the bias is getting larger as the surface gets more and more sloped. The problem with a single bias value is that once you set it large enough to prevent shadow acne even for sloped surfaces, you get Peter Panning (disconnected shadows) even in simple situations.

Polygon offset is esp. useful when handling large or dynamic scenes (varying light), it is usually possible to achieve smaller artifacts compared to plain bias.

WestLangley:
This is interesting...

Can you please modify the example to:

1. remove the spinning lights
2. remove the spinning objects
3. use basic shadows map type ( adding a GUI to change it would be nice )
4. cast shadows from back faces, as is typical in three.js, but add a GUI to toggle it
5. add a GIU to toggle `model.receiveShadow` -- it should cast one, of course
6. add a GIU to toggle `floor.castShadow` -- it should receive one, of course

I think only one model and a ground is sufficient. If different kinds of models have different issues, you can have a toggle for models, too.

I am curious if _both_ bias and polygon-offset need to be supported. The example I am suggesting will help @mrdoob and others understand the issue. Maybe polygon-offset is sufficient.

What are the units of your two new parameters?

OndrejSpanel:
> What are the units of your two new parameters?
I think the best is to read [OpenGL documentation](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glPolygonOffset.xml) on this:

> The value of the offset is factor × DZ + r × units , where DZ is a measurement of the change in depth relative to the screen area of the polygon, and r is the smallest value that is guaranteed to produce a resolvable offset for a given implementation.

WestLangley:
>I think the best is to read OpenGL documentation on this:

Yes, I am familiar with the documentation, which has never been clear to me. I am searching for the answer to my question. :-)

OndrejSpanel:
I will try to update an example so that it shows the effects (hopefully in a few days).

OndrejSpanel:
>  When running on DX 9 ANGLE the example still runs, but the polygon offset settings have no effect at all.

The same is true when running on DX 11 ANGLE. I need to run with OpenGL as the backend, but is such configuration Chrome is freezing to me a few times a day, even on pages with no 3D content at all. :( Discouraging. I am not sure how large percent of the user would have the issue, but it does not seem like something which could be ignored. I am thinking about two possible workarouds:

- implement polygon offset in the depth shader instead (using dx/dy this should be possible)
- using Depth Textures for shadows (this is likely to help, but needs testing)

OndrejSpanel:
I have implemented the polygon offset in the depth shader instead of using gl.polygonOffset. This is a minor performance drawback, but it is a solution compatible with ANGLE via Direct3D. The shader is unchanged unless the polygon offset is used.

I am not much sure how should I introduce new uniforms into the depth shader - if it should be done differently, please. advise.

I have tried adjusting the example so that it shows better the differences. There are two presets available, one can clearly see the bias has more visible Peter Panning, while the polygon offset on the other hand shows slight acne on some areas.

OndrejSpanel:
Any comments? Anything else to implement, either in the code or in the example?

OndrejSpanel:
Could someone review this and comment? I would like to implement some more shadow improvements (possibly depth texture based shadows, perhaps even cascades), but before I do I would like to verify my style of work is acceptable and compatible with what is expected here.

What I would especially want is review of how are uniforms passed into the shader - what I did works and I am not aware of any drawback or issues, but I am unsure if this is how it is supposed to be done or if there is perhaps other better / preferred way.

WestLangley:
@OndrejSpanel 

1. Thanks for improving your scene. It is over-bright, however. Please reduce the directional light intensity to a value not exceeding 1.0. You can set `renderer.outputEncoding = THREE.sRGBEncoding` if you want.

2. Your code looks good to me, but I will not be reviewing the implementation details. They will have to be approved by @Mugen87 and @mrdoob.

3. The shadows do not look very good. Not sure if that is intended, but it would be nice if the rendering looked better.

4. I think you are going to have to clearly demonstrate a _problem that is being solved_ by this approach. That means generating a scene using the conventional approach and then having a toggle to demonstrate your approach.

5. I am not sure if you intend this to be a replacement for our current approach, as an alternative to it,  or in addition to it. Perhaps you can clarify.

6. I still do not understand what the units are on your parameters. It would be nice if they could be normalized to take values in [ 0, 1 ], but that may not be possible.

OndrejSpanel:
First I would like to repeat this is industry standard technique for handling shadow artifacts. You can find many articles and papers about it on the internet.

> I am not sure if you intend this to be a replacement for our current approach,

It is an alternative. It could also be used in combination, but it makes little sense, because the "unit" part has more or less the same meaning as the "bias" has. The "factor" part is what makes it different, which allows better handling of surfaces not perpendicular to the light.

> That means generating a scene using the conventional approach

The current example scene has presets which should correspond to the current approach and polygon offset approach. It should be possible to toggle between the two and see the difference. Each of the techniques has its pros and cons and allows different tuning 

> The shadows do not look very good

The scene intentionally uses very coarse resolution of the shadow maps, as this makes the artifacts clearly visible. If you mean anything else, can you please be more specific?

OndrejSpanel:
Bias technique with shadows casted from back sides tends to result in Peter Panning effect: objects look like flying, as the shadow does not start immediately behind them:

image

Using front sides causes very strong acne:

image

If the bias is high enough to avoid the acne we get Peter Panning again (in a different location):

image

No such artifact can be seen with the polygon offset technique, as it allows to use front sides for casting without causing too much acne:

image

OndrejSpanel:
My main motivation for this technique is that when implementing Cascaded Shadow Maps (which I already did before in a different game engine) in my experience is it is very hard to avoid artifacts for the far (coarse) cascades with the bias technique, but I have acceptable results with polygon offsets. This is why the demo shows coarse shadows.

With fine shadows (this can be tested in the updated demo) both techniques look fine, as long as there is no fine detail in the models. Once fine detail would be added into the scene the same artifacts would be visible when zooming close enough.

WestLangley:
Thanks for your reply. I am trying to help you create a compelling case for @mrdoob and @Mugen87 to review.

>... the "unit" part has more or less the same meaning as the "bias" has. The "factor" part is what makes it different, which allows better handling of surfaces not perpendicular to the light.

Good answer. As I said, I would like to see an example where our current `bias` is not adequate, and your two-parameter API is adequate.

>No such artifact can be seen with the polygon offset technique, as it allows to use front sides for casting 

You do know we can set `.shadowSide` with the current API?

>It is an alternative. It could also be used in combination, but it makes little sense, because the "unit" part has more or less the same meaning as the "bias" has.

That is what I thought. I think the API should make it clear that the parameters are not intended to be used together... Or, you can propose the new API as a replacement for the old, I guess...

//

How do your parameters compare with Unity's `bias` and `normalBias`?

Thank you for responding to my comments and questions. :-)

OndrejSpanel:
> shadowSide with the current API?

Yes, and my demo does that. The point is that with a plain bias you will get either acne or Peter Panning no matter what side you choose. With polygon offset front side casted shadows become acceptable.

> How do your parameters compare with Unity's bias and normalBias?

I have no experience with Unity, but after reading [their corresponding docs](https://docs.unity3d.com/Manual/ShadowOverview.html#LightBias) I think this is something different. The `normalBias` seems to enlarge / shrink the model.

OndrejSpanel:
> I would like to see an example where our current bias is not adequate, and your two-parameter API is adequate.

I think the demo scene with the shadow resolution of 256 shows this and I have provided screenshots in my comment above. I did not find a value of bias which would not have either unacceptable acne or Peter Panning. With polygon offset I think this is visibly better.

Mugen87:
Is this PR still relevant after adding `LightShadow.normalBias`? From the [documentation](https://threejs.org/docs/index.html#api/en/lights/shadows/LightShadow.normalBias):

>  Increasing this value can be used to reduce shadow acne especially in large scenes where light shines onto geometry at a shallow angle.

I'm not sure if the engine should provide the user a third option for shadow tweaking (especially if the common shadow rendering issue can be mitigated with the existing API).

mrdoob:
Indeed... I suspect `bias` and `normalBias` should be enough parameters.

Thanks though!

OndrejSpanel:
I experimented with `bias` + `normalBias` for shadows compared to "polygon offset" (also known as "sloped bias") recently. I was extremely prejudiced against normal bias, as polygon offset is what I have always used in my past game developer career, and it seems to me like a more robust and theoretically sound solution, but despite of this prejudice I have to admit `bias` + `normalBias` works very well and I was not able to create a scene where polygon offset would produce significantly better results than `bias` + `normalBias`, even in combination with large scenes and CSM.

PR #18450, opened 2020-01-22
Examples: Add XRControllerModelFactory class.

toji:
This utility makes it super simple to load controller meshes from the [webxr-input-profile registry](https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/index.html), giving users a controller visualization that matches what they're holding in their hand, with individual components animated in response to real-world input.

Effectively, it gives WebXR apps a high-quality default controller with just a couple of lines of code.

*Usage Example:*
```js
let xrControllerModelLoader = new XRControllerModelLoader(gltfLoader);

let controller = renderer.xr.getControllerGrip(0); // See #18362
scene.add(controller);

let model = xrControllerModelLoader.getControllerModel(controller);
model.setEnvironmentMap(envMap); // Optional
controller.add(model);
```

And done! When the XR controller is detected it'll automatically reach out to a list of known controller assets (currently pulling from jsdelivr, though we're working on having a dedicated CDN in the near future) and downloads one that most closely matches the device the user is holding, as reported by the WebXR API.

Given that I'm still fairly new to contributing to Three, I've got a few questions regarding this PR that I'd like clarity on:

 - Naming conventions: The class is currently called a "Loader" but it doesn't behave quite like the other Loaders in the Examples. Is there a more preferable term? Manager? Helper?
 - Similarly, I'm using the term "Model" in the name, but I'm not sure if that's going to cause confusion? Mesh didn't seem appropriate because Three has a well defined concept of a Mesh and this isn't it. It's technically a Group, but "XRControllerGroup" doesn't feel appropriately descriptive.
 - This currently loads both an external JS library and the assets from a CDN, and I'm not sure how Three feels about that. The asset path is easy for apps to reconfigure if they want to point at their own copy of the controller assets, and the `motion-controller.js` library is small enough that it could feasibly be checked into Three as a third party library, but that would make it a bit harder to keep up-to-date.
 - Oh, and the version of those files that are currently pointed to will need to be updated before merging. We're about to publish version 1.0.0. I just wanted to start getting feedback early.

### DIFF
--- a/examples/jsm/webxr/XRControllerModelLoader.d.ts
+++ b/examples/jsm/webxr/XRControllerModelLoader.d.ts
@@ -1,0 +1,24 @@
+import {
+	Group,
+	Object3D,
+	Texture
+} from '../../../src/Three';
+
+import { GLTFLoader } from '../loaders/GLTFLoader';
+
+export class XRControllerModel extends Object3D {
+	constructor( );
+
+	motionController: any;
+
+	setEnvironmentMap( envMap: Texture ): XRControllerModel;
+}
+
+export class XRControllerModelLoader {
+	constructor( gltfLoader?: GLTFLoader );
+	gltfLoader: GLTFLoader | null;
+	path: string;
+
+	setGLTFLoader( gltfLoader: GLTFLoader ): XRControllerModelLoader;
+	getControllerModel( controller: Group ): XRControllerModel;
+}
--- a/examples/jsm/webxr/XRControllerModelLoader.js
+++ b/examples/jsm/webxr/XRControllerModelLoader.js
@@ -1,0 +1,286 @@
+/**
+ * @author Nell Waliczek / https://github.com/NellWaliczek
+ * @author Brandon Jones / https://github.com/toji
+ */
+
+import {
+	Mesh,
+	MeshBasicMaterial,
+	Object3D,
+	Quaternion,
+	SphereGeometry,
+} from "../../../build/three.module.js";
+
+import {
+	Constants as MotionControllerConstants,
+	fetchProfile,
+	MotionController
+} from 'https://cdn.jsdelivr.net/npm/@webxr-input-profiles/motion-controllers@0.1.0/dist/motion-controllers.module.js'
+
+const DEFAULT_PROFILES_PATH = 'https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@0.1.0/dist/profiles'
+
+function XRControllerModel( ) {
+
+	Object3D.call( this );
+
+	this.motionController = null;
+	this.envMap = null;
+
+}
+
+XRControllerModel.prototype = Object.assign( Object.create( Object3D.prototype ), {
+
+	constructor: XRControllerModel,
+
+	setEnvironmentMap: function ( envMap ) {
+
+		if ( this.envMap == envMap ) {
+
+			return this;
+
+		}
+
+		this.envMap = envMap;
+		this.traverse(( child ) => {
+
+			if ( child.isMesh ) {
+
+				child.material.envMap = this.envMap;
+				child.material.needsUpdate = true;
+
+			}
+
+		});
+
+		return this;
+
+	},
+
+	/**
+	 * Polls data from the XRInputSource and updates the model's components to match
+	 * the real world data
+	 */
+	updateMatrixWorld: function ( force ) {
+
+		Object3D.prototype.updateMatrixWorld.call( this, force );
+
+		if ( !this.motionController ) return;
+
+		// Cause the MotionController to poll the Gamepad for data
+		this.motionController.updateFromGamepad();
+
+		// Update the 3D model to reflect the button, thumbstick, and touchpad state
+		Object.values( this.motionController.components ).forEach(( component ) => {
+
+			// Update node data based on the visual responses' current states
+			Object.values( component.visualResponses ).forEach(( visualResponse ) => {
+
+				const { valueNode, minNode, maxNode, value, valueNodeProperty } = visualResponse;
+
+				// Skip if the visual response node is not found. No error is needed,
+				// because it will have been reported at load time.
+				if ( !valueNode ) return;
+
+				// Calculate the new properties based on the weight supplied
+				if ( valueNodeProperty === MotionControllerConstants.VisualResponseProperty.VISIBILITY ) {
+
+					valueNode.visible = value;
+
+				} else if ( valueNodeProperty === MotionControllerConstants.VisualResponseProperty.TRANSFORM ) {
+
+					Quaternion.slerp(
+						minNode.quaternion,
+						maxNode.quaternion,
+						valueNode.quaternion,
+						value
+					);
+
+					valueNode.position.lerpVectors(
+						minNode.position,
+						maxNode.position,
+						value
+					);
+
+				}
+
+			});
+
+		});
+
+	}
+
+} );
+
+/**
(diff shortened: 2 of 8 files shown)


examples/jsm/webxr/XRControllerModelLoader.d.ts
```diff
@@ -0,0 +1,24 @@
...
+export class XRControllerModel extends Object3D {
+	constructor( );
+
+	motionController: any;
+
+	setEnvironmentMap( envMap: Texture ): XRControllerModel;
+}
+
+export class XRControllerModelLoader {
+	constructor( gltfLoader?: GLTFLoader );
```
NellWaliczek: At some point we may want to expand the @webxr-input-profiles/assets schema to support more than just gltf/glb so let's maybe change this to be passed in a generic Loader?

examples/jsm/webxr/XRControllerModelLoader.js
```diff
@@ -0,0 +1,286 @@
...
+	Object3D,
+	Quaternion,
+	SphereGeometry,
+} from "../../../build/three.module.js";
+
+import {
+	Constants as MotionControllerConstants,
+	fetchProfile,
+	MotionController
+} from 'https://cdn.jsdelivr.net/npm/@webxr-input-profiles/motion-controllers@0.1.0/dist/motion-controllers.module.js'
```
NellWaliczek: Are we planning on using this from within @webxr-input-profiles/viewer?  If so, we can't hardcode this path or it won't build against the @webxr-input-profiles/motion-controllers package in the same repo.

NellWaliczek (commented):
A few quick thoughts as I putter with my side project that I'm looking forward to using this in...

Mugen87:
> : The class is currently called a "Loader" but it doesn't behave quite like the other Loaders in the Examples. Is there a more preferable term? Manager? Helper?

Have you considered `XRControllerModelFactory`? AFAICS, `getControllerModel()` returns a new instance of `XRControllerModel`. So technically, it's somewhat a factory method. In that regard, `createControllerModel()` seems even more appropriate.

> Similarly, I'm using the term "Model" in the name, but I'm not sure if that's going to cause confusion?

`XRControllerModel` sounds fine to me.

> This currently loads both an external JS library and the assets from a CDN, and I'm not sure how Three feels about that. 

Usually, the project prefers to copy such files into `jsm/libs`. I think I would do the same here for now.

toji:
> Have you considered `XRControllerModelFactory`? ... In that regard, `createControllerModel()` seems even more appropriate.

Those both sounds fine to me!

> Usually, the project prefers to copy such files into `jsm/libs`. I think I would do the same here for now.

That'll be trivial for `motion-controllers.module.js`, but I question whether that's really what you want for the asset files? It's currently 67Mb worth of *.glb files, and it's expected to grow. As maintainers of that library we'd prefer to have developers point at a central repository when loading controllers to ease the personal burden on their servers, improve caching behavior across sites, and allow for automatic updates as new devices get added to the repository. That said, if you want Three to be completely standalone in this regard we can add them, and we'll just make sure that it's readily apparent that the root path can be changed.

Mugen87:
Okay, I see. Let's see what @mrdoob prefers here.

toji:
Updated the PR based on some of the feedback. Main class is now called `XRControllerModelFactory` with a `createControllerModel()` method. I also copied `motion-controllers.module.js` into `jsm/libs`. CDN usage is pending Doob review.

I also took a moment to clean up a couple of things that I found were bugging me. The factory will now create it's own GLTFLoader if you don't supply one, to make drop-in usage even easier, and I updated three of the samples that used controllers to make use of the library. (Sculpt/Paint were left out because they already have simple controller visualizations.)

To address a couple of comments from @NellWaliczek:

> At some point we may want to expand the @webxr-input-profiles/assets schema to support more than just gltf/glb so let's maybe change this to be passed in a generic Loader?

I think we could easily extend this in the future to allow that (though I think it would need a bit more than just passing in new loaders.) Since we only support glTF right now, and my assumption was that we'd require _at least_ a glTF file be present even if other assets were available, this seemed like a safe bet without over engineering v1. It's worth noting that Three has never been a library to *ahem* shy away from breaking changes, so I don't feel like this is a choice that will be limiting to us long term.

> Are we planning on using this from within @webxr-input-profiles/viewer? If so, we can't hardcode this path or it won't build against the @webxr-input-profiles/motion-controllers package in the same repo.

It would be nice to use it there, yes. (Though I don't view it as an absolute requirement.) I'd like to understand the build issues a bit more, though: In the latest change that library has been updated to link from a copy in `jsm/libs` at @Mugen87's suggestion. That should prevent any outright build problems, such as circular dependencies. We'd definitely like to use our own version of the library when building for the viewer, but I don't know of any way to handle import paths like this dynamically and I'd prefer to not force it to be something that users pass in. It's possible that as part of our build process we could copy `motion-controllers.module.js` into the appropriate three.js folder. Slightly janky, but should do the job? Am I overlooking a larger problem?

toji:
For the record, I discussed @NellWaliczek's feedback with her offline yesterday and it's my understanding that the issues have now been resolved and the current code will allow us to re-use this module as part of our viewer as hoped. I'm going to mark her comments as resolved.

mrdoob:
Great stuff!
I suspect we'll redesign the API but should be good for now.

mrdoob:
Thanks!

mrdoob:
Tested this a little bit more today. Phenomenal work!

@NellWaliczek @toji Thanks for solving this problem! 🙏

mrdoob:
Minor bug report: Oculus Quest controllers seem to be a bit off. I can take screenshots if needed but if you make the controllers touch each other you'll see that there's a bit of offset.

toji:
If you're referring to the position of the controllers (how well they line up with the real-world counterparts) then I fixed it this morning and just haven't published a new version of the package yet. Three should pick it up automatically when I do. If it's anything else (material issue, animation issue, etc.) please file a bug with screenshots! Thanks!

mrdoob:
That's what I meant. Magic!

Mugen87:
The visual feedback of the controllers (what button is pressed etc.) is really amazing!

NellWaliczek:
So glad you all like it!  <3

mrdoob:
@toji I'm still seeing the offset:

83850213_1515301941972680_4226151929824149504_o

I'm making the touch controllers... er..  touch... but in VR they intersect.

toji:
Oh! I forgot to publish the adjusted meshes to npm. I'll do that tonight.

toji:
And published! I don't know exactly how long it'll take for the changes to propagate through the various tubes, but I'd imagine you should start seeing the fixed controller positions within the hour (and maybe less with a cleared cache.)

tbyholm:
Hello! 

I tried this also, and for me it worked well on the Oculus Rift S rig, except that it's using the old Rift controls for some reason. The interesting thing is that if I use the HTC Vive on my own code the screen goes white if i turn the control 90 degrees around my wrist and stays like that, but in the three.js dragging example it does not. Furthermore, on my own solution it shows the correct HTC Vive controls but when i run the threejs dragging example I can see the correct Vive controllers for a split second and then they change into some default controller with a thumb pad. I've been using both Chrome 80 and latest Firefox (using the webxr pollyfill). I haven't had time to try it with the old Rift.  Don't know really if I'm doing something wrong but the only difference between my setup is that I only attach one controller and I don't add it to the scene, but to a dolly object so that it will follow when the character moves.  In general everythings is fine using THREE.VRController but I'm trying to migrate to the XR-standard.


-Thomas

PR #18467, opened 2020-01-24
cubemap background Increase blur level

gameatp:


WestLangley:
Related #18241.

Mugen87:
Closing for now until it's clear in #18241 what API is preferred.

WestLangley:
Reopening. This will not be considered as an option if it is closed.

@gameatp Thank you for your suggestion!

Mugen87:
But the implementation in #18241 is almost identical?

WestLangley:
@Mugen87 As you wish...

PR #18471, opened 2020-01-24
Editor: Add support for background texture.

Mugen87:
It was already possible to define the background's color, now it's possible to define the background via a texture. The new select UI also enable to set the background to `null`.

image

This is actually a nice preparation to support more advanced scene backgrounds in the editor.

BTW: It's not yet possible to safe and reload a scene with a background texture since `ObjectLoader` is unable to serialize/deserialize such a scene. I'll have a look at this when the PR goes in 👍

### DIFF
--- a/editor/js/Editor.js
+++ b/editor/js/Editor.js
@@ -122,7 +122,8 @@ Editor.prototype = {
 		this.scene.uuid = scene.uuid;
 		this.scene.name = scene.name;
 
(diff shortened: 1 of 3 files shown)


Mugen87:
http://raw.githack.com/Mugen87/three.js/dev9/editor/index.html

mrdoob:
Nice!

mrdoob:
Thanks!

PR #18479, opened 2020-01-25
Scene: Add support for serialization/deserialization texture background.

### LINKED ISSUE #15180: Editor: Texture map doesn't render when refresh web page in Chrome, but this problem doesn't reappear in Firefox.

### LINKED ISSUE #16154: Editor: Scene.background and .environment are not set when importing a JSON scene.

When loading a scene, the background of the editor is not set to the background of the scene. This causes an issue when exporting because the scene's background will be overwritten to the Editor's default gray.

##### Three.js version

- [x] Dev
- [x] r103

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

Mugen87:
It's not possible to safe and restores scenes with a background texture.

The PR also ensures that textured objects of loaded (JSON) scenes are not black in the editor anymore. So far, it was always necessary to interact with the viewport to trigger a redraw so the textures were visible. This happens now automatically.

Fixed #15180.
Fixed #16154.

### DIFF
--- a/src/core/Object3D.js
+++ b/src/core/Object3D.js
@@ -688,6 +688,29 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ), 
 
 		}
 
+		if ( this.isScene ) {
+
+			var background = this.background;
+
+			if ( background !== null ) {
+
+				if ( background.isColor ) {
+
+					object.background = background.toJSON();
+
+				} else if ( background.isTexture ) {
+
+					object.background = background.toJSON( meta ).uuid;
+
+				}
+
+			}
+
+			if ( this.environment !== null ) object.environment = this.environment.toJSON( meta );
+			if ( this.fog !== null ) object.fog = this.fog.toJSON();
+
+		}
+
 		//
 
 		function serialize( library, element ) {
--- a/src/scenes/Scene.js
+++ b/src/scenes/Scene.js
@@ -49,18 +49,6 @@ Scene.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 	},
 
-	toJSON: function ( meta ) {
-
-		var data = Object3D.prototype.toJSON.call( this, meta );
-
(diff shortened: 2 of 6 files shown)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev9/editor/index.html

src/scenes/Scene.js
```diff
@@ -49,18 +49,6 @@ Scene.prototype = Object.assign( Object.create( Object3D.prototype ), {
 
 	},
 
-	toJSON: function ( meta ) {
-
-		var data = Object3D.prototype.toJSON.call( this, meta );
-
-		if ( this.background !== null ) data.object.background = this.background.toJSON( meta );
```
Mugen87: We have to move this section to `Object3D.toJSON()` since the `meta` object does not contain `textures` and `images` properties when this overwritten method is called with a scene just containing a background and untextured objects.

src/core/Object3D.js
```diff
@@ -688,6 +688,29 @@ Object3D.prototype = Object.assign( Object.create( EventDispatcher.prototype ),
...
+
+			if ( background !== null ) {
+
+				if ( background.isColor ) {
+
+					object.background = background.toJSON();
+
+				} else if ( background.isTexture ) {
+
+					object.background = background.toJSON( meta ).uuid;
```
Mugen87: We just want to save the UUID to the texture here.

Mugen87:
I think this PR performs too many changes at once. Let me try to split things up a bit.

PR #18485, opened 2020-01-26
Editor: Add frametime info.

### LINKED ISSUE #16007: Editor: add FPS counter

Mugen87:
Fixed #16007.


Mugen87:
http://raw.githack.com/Mugen87/three.js/dev11/editor/index.html

mrdoob:
Thanks!

Issue #13880, opened 2018-04-17
CubeTexture Support for the Editor


atul-mourya:
This is a feature request.

For a long time, threejs editor has been lacking the support of `CubeTexture`. 
Not just for importing but exporting as well. 

I'm assuming that many people would love this feature.

Most of the time, we have to go through threejs editor in order to create a scene with proper textures and materials and use it as an asset for the app development. But the lack of CubeTexture makes the app development slightly tricky. 

I'm aware that it's not a straightforward feature and may require a CubeTexture UI Panel, changes in .toJson and .fromJson in the editor and also the changes in the ObjectLoader and some more local storage related adjustments. 

However, I believe if this feature gets added, the editor will become more powerful then ever.

Let me if this request qualifies to be a feature request in upcoming revision?

Mugen87:
#14235 is a first step in order to enable this feature :blush:

Mugen87:
> I'm aware that it's not a straightforward feature and may require a CubeTexture UI Panel,

Some progress: There is now a `UICubeTexture` element 😊

PR #18494, opened 2020-01-27
Editor: Added CubeMap support for Scene.background.

Mugen87:
This PR introduces `UICubeTexture` which can be used to configure the scene's background with a cube texture.

Similar to a textured background, it's not yet possible to save/reload this setting.

Related: #13880

### DIFF
--- a/editor/js/Sidebar.Scene.js
+++ b/editor/js/Sidebar.Scene.js
@@ -3,7 +3,7 @@
  */
 
 import { UIPanel, UIBreak, UIRow, UIColor, UISelect, UIText, UINumber } from './libs/ui.js';
-import { UIOutliner, UITexture } from './libs/ui.three.js';
+import { UIOutliner, UITexture, UICubeTexture } from './libs/ui.three.js';
 
 var SidebarScene = function ( editor ) {
 
@@ -118,7 +118,8 @@ var SidebarScene = function ( editor ) {
 		signals.sceneBackgroundChanged.dispatch(
 			backgroundType.getValue(),
 			backgroundColor.getHexValue(),
-			backgroundTexture.getValue()
+			backgroundTexture.getValue(),
+			backgroundCubeTexture.getValue()
 		);
 
 	}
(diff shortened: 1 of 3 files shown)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev19/editor/index.html

Mugen87:
@mrdoob Well, implementing `UICubeTexture` was not that easy^^. The current code tries to save as much UI space as possible by using a `select` element in order to select the single sides of the cube texture for configuration.

mrdoob:
> @mrdoob Well, implementing `UICubeTexture` was not that easy^^. The current code tries to save as much UI space as possible by using a `select` element in order to select the single sides of the cube texture for configuration. I'm open for better, more user-friendly approaches 😇 

It's a tricky one! How about this UI?

```
PX: [   ]
PY: [   ]
PZ: [   ]
NX: [   ]
NY: [   ]
NZ: [   ]
```

Mugen87:
```
PX: [   ]
PY: [   ]
PZ: [   ]
NX: [   ]
NY: [   ]
NZ: [   ]
```
I've tried this one first but noticed that a lot of space gets lost since the background configuration is always visible when the `SCENE` tab is active.
```js
P: [   ] [   ] [   ]
N: [   ] [   ] [   ]
```
I've not though of that option so far^^. I like it because we can save space and get rid of the select element. Let me give this a try.

Mugen87:
BTW: My actual goal is to fully support background/ environment maps so we can reproduce examples like the basic glTF (hemlet)  demo in the editor. Against this backdrop, a new `ENVIRONMENT MAPS` section in the `PROJECT` tab similar to `MATERIALS` could be useful. It would allow the users to globally define environment maps and then apply them to the background and to `Material.envMap`. What do you think.

In any case we need `UICubeTexture`. But a little road map could help to clarify the development targets^^.

Mugen87:
Looks much better now:

image

mrdoob:
> BTW: My actual goal is to fully support background/ environment maps so we can reproduce examples like the basic glTF (hemlet)  demo in the editor. Against this backdrop, a new `ENVIRONMENT MAPS` section in the `PROJECT` tab similar to `MATERIALS` could be useful. It would allow the users to globally define environment maps and then apply them to the background and to `Material.envMap`. What do you think? 

Sounds good!

mrdoob:
Thanks!

PR #18499, opened 2020-01-28
Remove unused numXxxx members of WebGLLights

OndrejSpanel:
I think `WebGLLights.state` members `numDirectionalShadows`, `numPointShadows` and `numSpotShadows` are no longer used. Corresponding members of `hash` are used, as well as local variables with the same name.


Mugen87 (approved):
That change looks correct but I suggest to merge the PR after the `R113` release.

Mugen87:
@mrdoob I think this is good to go now^^.

mrdoob:
Thanks!

PR #18510, opened 2020-01-29
Vector3: support the zero vector in projections

WestLangley:
As suggested in https://github.com/mrdoob/three.js/issues/17545#issuecomment-579721218.

The convention followed in this PR is the zero vector is orthogonal to every vector.

Following this convention prevents `NaN` from being returned in the related `Vector3` methods.


mrdoob:
Thanks!

PR #18515, opened 2020-01-30
PCDLoader now ignores NaN positions for points.

f-motu:
PCDLoader now ignores NaN positions for points.  Only for binary_compressed data!

### DIFF
--- a/examples/js/loaders/PCDLoader.js
+++ b/examples/js/loaders/PCDLoader.js
@@ -283,16 +283,27 @@ THREE.PCDLoader.prototype = Object.assign( Object.create( THREE.Loader.prototype
 			var decompressedSize = sizes[ 1 ];
 			var decompressed = decompressLZF( new Uint8Array( data, PCDheader.headerLen + 8, compressedSize ), decompressedSize );
 			var dataview = new DataView( decompressed.buffer );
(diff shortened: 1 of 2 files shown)


Mugen87:
> NaN positions for points

Do you know why PCD files contain such invalid data in the first place? Does this happen because of a buggy exporter?

MorpheusFT:
No I don't know why the invalid data is there. The PCD files are generated from a camera. The NaN values were causing an issue when the parse function calls computeBoundingSphere().

Mugen87:
We normally do not apply such changes to loaders since we expects assets according to the respective standard. PCD files containing invalid numbers are surely not conform to the PCD standard. In such cases, it's actually best to fix the software which produces the problematic file contents.

MorpheusFT:
Alright, makes sense.

MorpheusFT:
How about if I add an optional boolean to enable filtering of NaN values?

Mugen87:
Sorry, I do not vote for this. We are trying to develop the loaders as equal as possible. And in general, they should not perform sanity checks or data cleanup.

Mugen87:
I would advise you to create a copy of `PCDLoader` and then implement your custom checks. I doubt that the loader will change much in the future...

Issue #11315, opened 2017-05-11
What is three.js's normal map handedness convention?


donmccurdy:
It may be helpful to have this documented. @emackey pointed out an issue with GLTF2Loader (https://github.com/donmccurdy/three-gltf-viewer/issues/10) and normal map inversion seems to be responsible. For comparison:

> BabylonJS: x-axis = left, y-axis = down
> OpenGL: x-axis = right, y-axis = up
> DirectX: x-axis = right, y-axis = down

emackey:
Has there been progress on this?

In my own projects I'm [fixing this manually](https://github.com/AnalyticalGraphicsInc/gltf-vscode/blob/c7be56a56b52a6b49bde1ca8696930bcc6568f4d/pages/three.js#L133) by setting:

```
    material.normalScale.x = -1;
```

donmccurdy:
I haven't heard anything. It would be nice to know, but I'm inclined to just flip `normalScale.x` if that works correctly with the sample models.

emackey:
The answers have been worked out by @WestLangley and @donmccurdy here:  https://github.com/mrdoob/three.js/pull/13784#issuecomment-379511456

Hopefully I'm summarizing correctly:

* ThreeJS's default normal map handedness is OpenGL, right-handed, same as glTF.
* ThreeJS's default UV origin is lower-left, opposite of glTF's upper-left UV origin.
* As a result, GLTFLoader must flip the UVs and then flip the normal handedness to compensate.

PR #18522, opened 2020-01-31
Fix and document normal map handedness

Oletus:
A few examples use normal maps that have not been authored with the OpenGL normal map handedness convention. In these cases the normal scale y needs to be negated to get the normal maps to display correctly.

A mention of normal map handedness is also added to documentation - otherwise it can be hard to find out that the handedness differences exist and how to handle them in three.js.

Related earlier discussion in issue #11315

### DIFF
--- a/docs/api/en/materials/MeshMatcapMaterial.html
+++ b/docs/api/en/materials/MeshMatcapMaterial.html
@@ -115,6 +115,8 @@
 		<p>
 			The texture to create a normal map. The RGB values affect the surface normal for each pixel fragment and change
 			the way the color is lit. Normal maps do not change the actual shape of the surface, only the lighting.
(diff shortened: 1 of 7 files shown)


Mugen87:
For comparison:

https://raw.githack.com/higharc/three.js/fix-normal-map-handedness/examples/misc_controls_fly.html
https://threejs.org/examples/misc_controls_fly

http://raw.githack.com/higharc/three.js/fix-normal-map-handedness/examples/webgl_materials_physical_clearcoat.html
https://threejs.org/examples/webgl_materials_physical_clearcoat

mrdoob:
Thanks!

Oletus:
Thanks for the very quick review! :)

PR #18538, opened 2020-02-04
LOD: Add ability to get current level

CreaticDD:
Sometimes it's useful to have the current LOD level. This is awkward to do outside of LOD as it involves duplicating the logic inside of LOD.

One use case is the loading of progressively larger textures when the LOD activates a certain level.

### DIFF
--- a/src/objects/LOD.d.ts
+++ b/src/objects/LOD.d.ts
@@ -12,7 +12,6 @@ export class LOD extends Object3D {
 	levels: { distance: number; object: Object3D }[];
 	autoUpdate: boolean;
 	readonly isLOD: true;
-	readonly _currentLevel: 0;
 
 	addLevel( object: Object3D, distance?: number ): this;
 	getCurrentLevel(): number;
--- a/src/objects/LOD.d.ts
+++ b/src/objects/LOD.d.ts
@@ -12,8 +12,10 @@ export class LOD extends Object3D {
 	levels: { distance: number; object: Object3D }[];
 	autoUpdate: boolean;
 	readonly isLOD: true;
+	readonly _currentLevel: 0;
 
 	addLevel( object: Object3D, distance?: number ): this;
+	getCurrentLevel(): number;
 	getObjectForDistance( distance: number ): Object3D | null;
 	raycast( raycaster: Raycaster, intersects: Intersection[] ): void;
 	update( camera: Camera ): void;
--- a/src/objects/LOD.js
+++ b/src/objects/LOD.js
@@ -33,7 +33,7 @@ LOD.prototype = Object.assign( Object.create( Object3D.prototype ), {
 	constructor: LOD,
 
 	isLOD: true,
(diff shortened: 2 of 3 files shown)


src/objects/LOD.js
```diff
@@ -32,6 +32,8 @@ LOD.prototype = Object.assign( Object.create( Object3D.prototype ), {
 	constructor: LOD,
 
 	isLOD: true,
+	
+	currentLevel: 0,
```
Mugen87: Um, I think  I would favor `LOD.getCurrentLevel()`.  Users could think this property is writable.

donutcoffee: I see your point. Is there a performance concern of a getter over a direct access property, or are we beyond that in three.js?

I.e.
`var _currentLevel`
and
`Object.defineProperties ... get: function() { return _currentLevel; }`

Mugen87: Just make a private variable and then a simple get method. I don't think `defineProperties` is necessary here.

donutcoffee: Ok, done. Thanks for your input

src/objects/LOD.js
```diff
@@ -7,6 +7,7 @@ import { Object3D } from '../core/Object3D.js';
  * @author mrdoob / http://mrdoob.com/
  */
 
+var _currentLevel = 0;
```
Mugen87: I don't think this code works when you define `_currentLevel` as a module scope variable. This will break as soon as you have more than on `LOD` object. `_currentLevel` has to be a real property of `LOD`.

donutcoffee: Oh, absolutely correct... however then I don't understand how you would get away without a getter in the current code ...

Mugen87: I would just write `this. _currentLevel` in the constructor. It's more or less a convention that such properties are intended as private.

donutcoffee: Wilco

Mugen87:
Can you please also enhance the TS declaration file by `getCurrentLevel()`?

donutcoffee:
> 
> 
> Can you please also enhance the TS declaration file by `getCurrentLevel()`?

No problem, thanks for your patience.

src/objects/LOD.d.ts
```diff
@@ -12,8 +12,10 @@ export class LOD extends Object3D {
 	levels: { distance: number; object: Object3D }[];
 	autoUpdate: boolean;
 	readonly isLOD: true;
+	readonly _currentLevel: 0;
```
Mugen87: Can you please remove this line? We usually do not respect private properties in TS files.

donutcoffee: Done

Mugen87 (approved):

Mugen87:
Looks good now! 👍

mrdoob:
Thanks!

PR #18545, opened 2020-02-05
Editor: Fix black textures when loading JSON scenes.

### LINKED ISSUE #15180: Editor: Texture map doesn't render when refresh web page in Chrome, but this problem doesn't reappear in Firefox.

Mugen87:
Fixed #15180.

Ensures the editor only sets a scene/object if textures are actually loaded. This done by moving the related logic in the `onLoad()` callback of `ObjectLoader.parse()`. The editor already use this approach when loading `glTF` assets.

### DIFF
--- a/editor/js/Editor.js
+++ b/editor/js/Editor.js
@@ -644,16 +644,6 @@ Editor.prototype = {
 	fromJSON: function ( json ) {
 
 		var loader = new THREE.ObjectLoader();
-
-		// backwards
-
-		if ( json.scene === undefined ) {
-
-			this.setScene( loader.parse( json ) );
-			return;
-
-		}
-
 		var camera = loader.parse( json.camera );
 
 		this.camera.copy( camera );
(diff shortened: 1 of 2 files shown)


editor/js/Editor.js
```diff
@@ -644,16 +644,6 @@ Editor.prototype = {
...
 		var loader = new THREE.ObjectLoader();
-
-		// backwards
-
-		if ( json.scene === undefined ) {
-
-			this.setScene( loader.parse( json ) );
-			return;
-
-		}
```
Mugen87: This block was introduced here 7431f1a3e07f0090804cbfd60a5b1fd8c2943ceb, five years ago. I think we can safely remove this now.

PR #18562, opened 2020-02-06
Editor: Fix material number input fields.

Mugen87:
This PR fixed a little bug in context of floating point arithmetic.

- Open the editor: https://threejs.org/editor/
- Add a cube
- Go to the MATERIAL tab
- Set roughness value of `0.94` and press enter
- Now set a value of `0.93`

The value is not accepted since `0.94 - 0.93 >= 0.01` is not true. At least on my system, `0.94 - 0.93` is  `0.009999999999999898`. The PR fixes this issue by comparing against an epsilon value.

### DIFF
--- a/editor/js/Sidebar.Material.js
+++ b/editor/js/Sidebar.Material.js
@@ -41,6 +41,8 @@ var SidebarMaterial = function ( editor ) {
 
 	var currentMaterialSlot = 0;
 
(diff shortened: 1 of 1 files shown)


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev41/editor/index.html

mrdoob:
Thanks!

PR #18585, opened 2020-02-09
Fix empty name problem in GLTFLoader

drcmda:
References: 

https://github.com/mrdoob/three.js/issues/18530#issuecomment-583831617

https://spectrum.chat/react-three-fiber/general/help-required-unable-to-load-a-gltf-model~d77788d6-35ff-4e5d-b3e9-23682c661699

GLTF's can be exported with empty names, or missing names. GLTFLoader seems to have problems with the former. For instance, this is the nodes array of the following model: https://sketchfab.com/3d-models/red-city-356c773c4b9a45d8b9d4aa04c60ecb27

```jsx
  ...
  { mesh: 0, name: '' },
  { mesh: 1, name: '' },
  { mesh: 2, name: '' },
  { mesh: 3, name: '' }
```

GLTFLoader will auto generate names, like mesh_0, mesh_1 and so on, this happens in loadMesh:

```jsx
mesh.name = meshDef.name || ( 'mesh_' + meshIndex );
if ( geometries.length > 1 ) mesh.name += '_' + i;
```

Notice how it checks the string itself. So, by now the mesh has an actual name, but a small bug wipes it out again afterwards in the LoadNode function:

```jsx
if ( nodeDef.name !== undefined ) {
  node.userData.name = nodeDef.name;
  node.name = THREE.PropertyBinding.sanitizeNodeName( nodeDef.name );
}
```

Here it checks against undefined, so despite the def name being empty and the auto-generated name being present, it goes in there and overwrites the name with an empty string.

I have fixed all the other places that had this problem.

### DIFF
--- a/examples/js/loaders/GLTFLoader.js
+++ b/examples/js/loaders/GLTFLoader.js
@@ -1826,7 +1826,7 @@ THREE.GLTFLoader = ( function () {
 
 			texture.flipY = false;
 
-			if ( textureDef.name !== undefined ) texture.name = textureDef.name;
+			if ( textureDef.name ) texture.name = textureDef.name;
 
 			// Ignore unknown mime types, like DDS files.
 			if ( source.mimeType in MIME_TYPE_FORMATS ) {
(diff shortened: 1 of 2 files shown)


donmccurdy (approved):
Thanks! Agreed the change.

/cc @lexaknyazev do you think empty string, `"name": ""`, should result in a validation warning? It seems similar to `null` to me, but not a strong preference.

mrdoob:
Thanks!

PR #18594, opened 2020-02-10
DragControls: Avoid unnecessary object creation.

Mugen87:
I've realized today that `DragControls` does not apply the optional target to `Raycaster.intersectObjects()` which means the method always creates an array when raycasting is performed. Since raycasting is performed even when moving the mouse, it's more GC friendly to manage a single array instance in `DragControls` and apply it to the method.


Mugen87:
https://raw.githack.com/Mugen87/three.js/dev26/examples/misc_controls_drag.html

mrdoob:
Thanks!

Issue #18342, opened 2020-01-08
overrideMaterial for nested scenes


nullEuro:
It is possible to add a scene to another scene. However, setting an `overrideMaterial` on a nested scene has no effect. Only the `overrideMaterial` of the root scene is honored (and changes the material of children of nested scenes also as expected).

I expected (and am suggesting now) that it is possible to set an `overrideMaterial` on any scene, and that it overrides the `overrideMaterial` of parent scenes. Though that may be debatable. But the documentation in this regard could be improved in either case.

My use case: I have several Sub-Scenes loaded from different GLTF files in my root scene, and want to highlight specific ones under certain conditions with a special material.

[Live Demo](https://jsfiddle.net/qr4ovwpu/1/)

WestLangley:
I do not recall support for sub-scenes ever being considered in that regard...

In any event, a workaround is to use multiple render passes.

https://jsfiddle.net/d2je03mc/

makc:
@nullEuro would this not have unexpected effects, e g when people just do someObject.add(gltfLoaderResult.scene) and then try to apply effects that set overrideMaterial on the root scene and the entire gltf is still rendered as is because its overrideMaterial is not set?

WestLangley:
Yeah. Never liked that `GLTFLoader` adds a `THREE.Scene` as a child of `THREE.Scene`.

 /ping @donmccurdy

donmccurdy:
Looking at other properties on Scene objects, nesting them would be problematic... `.background`, `.environment`, `.fog`... I don't think we can support those properties, or `.overrideMaterial`, on nested scenes.

@nullEuro I believe you'll need to directly change the `.material` property on meshes you want to effect.

@WestLangley a THREE.Scene is the most obvious representation of a glTF Scene object. It's difficult to say universally whether a glTF file _is_ a scene or is an object added to a scene. The following is valid:

```js
loader.load( 'foo.glb', ( gltf ) => {
  renderer.render( gltf.scenes[0], gltf.cameras[0] );
} );
```

I don't mind changing the return values as shown below, though.

```typescript
// before
{
  scenes: Scene[];
} 

// after
{
  scenes: Object3D[];
} 
```

But I'm not aware that nested scenes have ever been discussed...

Mugen87:
> Looking at other properties on Scene objects, nesting them would be problematic... .background, .environment, .fog... I don't think we can support those properties, or .overrideMaterial, on nested scenes.

[...]

PR #18601, opened 2020-02-11
GLTFLoader: Return Group, not Scene.

donmccurdy:
See https://github.com/mrdoob/three.js/issues/18342#issuecomment-578981172.


mrdoob:
Thanks!

gkjohnson:
Don't forget the typescript definitions :)

donmccurdy:
https://github.com/mrdoob/three.js/pull/18604 👍

PR #18612, opened 2020-02-12
Update OrbitControls.js to be in compliance with default browser behavior

EAS-GUO:
removed event.preventDefault() from onTouchStart(event) to be in compliance with the behavior of "https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent" 

changed document.add/removeEventListener to scope.domElement for consistency in code

changed == to === for js conditions

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -255,8 +255,8 @@ THREE.OrbitControls = function ( object, domElement ) {
 		scope.domElement.removeEventListener( 'touchend', onTouchEnd, false );
 		scope.domElement.removeEventListener( 'touchmove', onTouchMove, false );
 
-		document.removeEventListener( 'mousemove', onMouseMove, false );
-		document.removeEventListener( 'mouseup', onMouseUp, false );
+		scope.domElement.removeEventListener( 'mousemove', onMouseMove, false );
+		scope.domElement.removeEventListener( 'mouseup', onMouseUp, false );
 
 		scope.domElement.removeEventListener( 'keydown', onKeyDown, false );
 
@@ -596,7 +596,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	function handleTouchStartRotate( event ) {
 
-		if ( event.touches.length == 1 ) {
+		if ( event.touches.length === 1 ) {
 
 			rotateStart.set( event.touches[ 0 ].pageX, event.touches[ 0 ].pageY );
 
@@ -613,7 +613,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	function handleTouchStartPan( event ) {
 
-		if ( event.touches.length == 1 ) {
+		if ( event.touches.length === 1 ) {
 
 			panStart.set( event.touches[ 0 ].pageX, event.touches[ 0 ].pageY );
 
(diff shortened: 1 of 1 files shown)


examples/js/controls/OrbitControls.js
```diff
@@ -255,8 +255,8 @@ THREE.OrbitControls = function ( object, domElement ) {
 		scope.domElement.removeEventListener( 'touchend', onTouchEnd, false );
 		scope.domElement.removeEventListener( 'touchmove', onTouchMove, false );
 
-		document.removeEventListener( 'mousemove', onMouseMove, false );
-		document.removeEventListener( 'mouseup', onMouseUp, false );
+		scope.domElement.removeEventListener( 'mousemove', onMouseMove, false );
```
Mugen87: This change will break applications where the canvas is not full screen. As you can see at the following fiddle, it's not possible to transform the camera when you leave the canvas with the mouse.

https://jsfiddle.net/wfc48ukz/

However, it should be like so:

https://jsfiddle.net/wfc48ukz/1/

Mugen87:
> removed event.preventDefault() from onTouchStart(event) to be in compliance with the behavior of "https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent"

Can you please explain in more detail what you mean by that?

EAS-GUO:
I meant to remove the preventDefault() because the eventorder (see the link) would not call the mouse/click events. 
This resulted from us when we wanted to add an onclick eventhandler in our application. 
This is possible on all browsers because they emulate mouse events from touchevents (as you can also see in the documentation).

examples/js/controls/OrbitControls.js
EAS-GUO: On my PC (Win 10) and on the newest Firefox, Chrome and Edge the fiddles work exactly the same?
I'm sorry for the removal of the document usage.

Mugen87:
> This resulted from us when we wanted to add an onclick eventhandler in our application.

Okay, thanks for this bit.

So when removing `preventDefault()` at this place the controls will fire additional mouse events. However, this change might cause unexpected side effects for other users who don't want these events. So this PR might be useful for your use case but it might break other, existing ones.

In the past we have encountered some back and forth in context of `OrbitControls`. Various devs tried to change the class so it better fits to their personal requirements. However, such a change could cause issue in other projects. Hence, changes are only merged if there is a compelling reason. I personally don't see this reason here so I don't vote to merge the PR.

However, if @mrdoob or @WestLangley think the change is appropriate, I won't block it.

EAS-GUO:
Yeah i thought this could be useful because then ThreeJS would have the same (default) behavior as every browser on touch-devices. 
This means it does not fire additional events, it follows the default event-firing order of the current browser running an ThreeJS application.

WestLangley:
I added the `preventDefault()` in #13706, but I do not recall the reason for doing so. Also, there was some push-back in #13706 for adding it. So, I do not have an opinion one way or the other.

/ping @greggman Do you have an opinion on this?

greggman:
You can see where preventDefault is important for touchstart/touchmove in

https://threejs.org/examples/webgl_multiple_elements.html

Without preventDefault when you try to touch an object to orient it the page will scroll under your finger making it difficult to orient the objects. With preventDefault just the objects move, the page does not.

Maybe add a comment what the point is? Or maybe make that sample add it's own event handlers?

Mugen87:
> Maybe add a comment what the point is?

That sounds like a good idea.

PR #18618, opened 2020-02-13
TS: Add optionalTarget in CatmullRomCurve3.getPoint()

marquizzo:
CatmullRomCurve3.getPoint() should accept an [optional Vector3 target](https://github.com/mrdoob/three.js/blob/master/src/extras/curves/CatmullRomCurve3.js#L104) argument. It's missing in the TypeScript declaration file, which leads to compiler errors if users try to use this option.


WestLangley:
Can you target the `dev` branch, please?

PR #18622, opened 2020-02-13
Vector2.angle() performance

gonnavis:
I'm not sure whether the performance is better.
But in glsl, I heard "the less conditional the better".
Test result can check http://glslsandbox.com/e#61015.59 .

### DIFF
--- a/src/math/Vector2.js
+++ b/src/math/Vector2.js
@@ -380,9 +380,7 @@ Object.assign( Vector2.prototype, {
 
 		// computes the angle in radians with respect to the positive x-axis
 
-		var angle = Math.atan2( this.y, this.x );
-
-		if ( angle < 0 ) angle += 2 * Math.PI;
+		var angle = Math.atan2( -this.y, -this.x ) + Math.PI;
 
 		return angle;
 
--- a/src/math/Vector2.js
+++ b/src/math/Vector2.js
@@ -380,7 +380,7 @@ Object.assign( Vector2.prototype, {
 
 		// computes the angle in radians with respect to the positive x-axis
 
-		var angle = Math.atan2( -this.y, -this.x ) + Math.PI;
+		var angle = Math.atan2( - this.y, - this.x ) + Math.PI;
 
 		return angle;
 


src/math/Vector2.js
```diff
@@ -380,9 +380,7 @@ Object.assign( Vector2.prototype, {
 
 		// computes the angle in radians with respect to the positive x-axis
 
-		var angle = Math.atan2( this.y, this.x );
-
-		if ( angle < 0 ) angle += 2 * Math.PI;
+		var angle = Math.atan2( - this.y, - this.x ) + Math.PI;
```
Mugen87: > I'm not sure whether the performance is better.

Please do only make such PRs if you can actually demonstrate a performance improvement. The problem is that the readabilty of `angle()` is now worse than before. So I don't see a fair tradeoff so far.

Mugen87: I think it would be good to implement the unit test for this method:

https://github.com/mrdoob/three.js/blob/d69f9a7d7cac55a4674639f259275a5119f1bc17/test/unit/src/math/Vector2.tests.js#L363-L367

In this way, it's easier to verify if the current and new implementation actually produce the same result.

gonnavis:
OK, I tested in my own chrome console with these code:
First define the two angle functions, **angle** for current, **angle2** for new.
```
function angle(x,y){
    var angle = Math.atan2( y, x );

    if ( angle < 0 ) angle += 2 * Math.PI;

    return angle;
}
function angle2(x,y){
    var angle = Math.atan2( - y, - x ) + Math.PI;

    return angle;
}
```
And then run these two functions in console 1000000000 times seperately.
```
var start=performance.now()
for(let i=0;i<1000000000;i++){
	angle(1,1)
}
console.log(performance.now()-start)
```
```
var start=performance.now()
for(let i=0;i<1000000000;i++){
	angle2(1,1)
}
console.log(performance.now()-start)
```
With **angle**, I got an average time 1330ms.
With **angle2**, I got an average time 560ms.
So I think it's a big performance improvement.

src/math/Vector2.js
gonnavis: I'll try how to use QUnit.

mrdoob:
Thanks!

PR #18663, opened 2020-02-18
EffectComposer: Remove duplicate declaration of Pass class

Itee:
Pass class is already exported by jsm/postprocessing/Pass.js and fixed a circular deps. It is time to remove the old duplicate.


Mugen87:
`EffectComposer` is auto-generated by `utils/modularize.js`. Your change will be undone as soon as somebody runs this script.

We can only remove the internal `Pass` declaration when `exmaples/js` has been removed.

Itee:
I will give a look to modularize (i think i could offer you some experience about threejs parsing ;)

PR #18665, opened 2020-02-18
Make "dispose" documentation up to date

OndrejSpanel:
As WeakMaps are used for geometries since #17242 (which I think was published in rev. 108), the documentation about `dispose` seems no longer accurate.

I have tried updating it to the best of my knowledge. While I am confident geometries are garbage collection now, I am not quite certain how is the situation with materials, textures or render targets. If someone knows they need some special handling, please suggest improvements.


Mugen87:
I don't think we should expose the ramifications of `WeakMap` to the user and promote the usage of `dispose()` as before. Otherwise the guide only produces confusion.

OndrejSpanel:
Fine with me. It is your call. Feel free to close or discard or whatever.

Mugen87:
From https://github.com/mrdoob/three.js/pull/17242#issuecomment-521845876

> In sum - please keep the dispose() methods in Three.js, and encourage developers to call them in order to promptly reclaim the GPU memory associated with objects that are no longer needed.

I think this recommendation makes sense and I would tell the user something else.

PR #18667, opened 2020-02-18
TransformControls: Better error handling and docs.

Mugen87:
`TransformControls` can only be used if its attached 3D object is part of the scene graph. This was not obvious for all users. The PR improves the docs and the error handling so affected users hopefully easier understand the cause of certain runtime errors.

https://discourse.threejs.org/t/how-to-handle-controls-after-removing-all-object-from-scene/8839/10?u=mugen87


mrdoob:
Thanks!

PR #18674, opened 2020-02-19
TS: Fix SpotLight.d.ts.

### LINKED ISSUE #18673: SpotLight typescript decl still has "exponent", should be "penumbra"

The file `SpotLight.d.ts` declares the constructor for `SpotLight` to have an "exponent" argument, but the actual name of that argument (since a long time ago) is "penumbra". It also declares the `exponent` attribute for the object. Doesn't impact behavior, but it's very confusing in an IDE -- made me think I had an old version of three.js somehow. 

Source (.d.ts) file is at https://github.com/mrdoob/three.js/blob/master/src/lights/SpotLight.d.ts

Mugen87:
Fixed #18673.

### DIFF
--- a/src/lights/SpotLight.d.ts
+++ b/src/lights/SpotLight.d.ts
@@ -13,7 +13,7 @@ export class SpotLight extends Light {
 		intensity?: number,
 		distance?: number,
 		angle?: number,
-		exponent?: number,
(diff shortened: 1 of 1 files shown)


mrdoob:
Thanks!

PR #18703, opened 2020-02-22
Fix CSM Example

gkjohnson:
It looks like #18627 wound up breaking the CSM example with changes to the `lights_fragment_begins` shader chunk so I've gone ahead and updated it. I've included the line breaks from the original shader chunk so it's easier to update and compare in the future.

Temporary live link:

https://raw.githack.com/gkjohnson/three.js/fix-csm/examples/webgl_shadowmap_csm.html

@vHawk


Mugen87 (approved):

mrdoob:
Thanks!

PR #18718, opened 2020-02-24
Editor: Make decimal precision on export configurable.

### LINKED ISSUE #6206: Allow for adjusting decimal precision on export in editor
hey it would be great to have an option to reduce the floating point precision when exporting things from the threejs editor, I uploaded a file with two decimal places and after exporting it had 14-15 decimal places. For objects with 1m+ vertices that can add up quickly

Mugen87:
Fixed #6206.


mrdoob:
Thanks!

PR #18723, opened 2020-02-25
CSM: Use cache objects in initCascades

gkjohnson:
The `initCascades` function must be called whenever `fov`, `near`, `far`, or `aspect` fields are changed on a camera so it can be a hot function especially when resizing a window or animating a transition between camera types. I've modified it here to reuse `Frustum` instances rather than create new ones whenever it's called to save on object allocation.

I'll be doing the same thing for `.update` next.

Temporary live link:

https://raw.githack.com/gkjohnson/three.js/csm-cache-objects/examples/webgl_shadowmap_csm.html

@vHawk 

_Edit: Looks like the failure is a diff issue from `css3d_youtube`:_

> ERROR! Diff wrong in 0.079 of pixels in file: css3d_youtube


mrdoob:
Thanks!

PR #18724, opened 2020-02-25
CSM: Use cached objects in .update

gkjohnson:
Follow up to #18723

The `update` function gets fired every frame so I've removed any instantiation of new objects in that loop. I also updated the example page to pass the camera world matrix rather than the local matrix when calculating the shadow camera volumes.

Both `getExtendedBreaks` and `getBreaks` can get called frequently, as well, and create new arrays every call, so I'll address those next.

Temporary live link:
https://raw.githack.com/gkjohnson/three.js/csm-update-cache-objects/examples/webgl_shadowmap_csm.html

@vHawk


mrdoob:
Thanks!

PR #18728, opened 2020-02-25
Utils: Move converters to modules.

Mugen87:
`fbx2three.js` and `obj2three.js` are still using examples files from the `js` directory. This PR ensures that both scripts use modules. Since the ES6 module support in context of node is complicated, the scripts now require the usage of [esm](https://www.npmjs.com/package/esm).

Tested the change with `Samba Dancing.fbx`, `stanford-bunny.fbx` and `emerald.obj`.

Sidenote: I also like to suggest to remove both scripts and refer to the editor instead. The project only provides for two 3D formats such converters which is somewhat inconsistent. Besides, even the current implementation of `fbx2three.js` does not cover all features compared to the browser version and requires some ugly monkey-patching so textures can be handled. Not a proper solution, IMO.

### DIFF
--- a/utils/converters/README.md
+++ b/utils/converters/README.md
@@ -1,11 +1,12 @@
 Utilities for converting model files to the Three.js JSON format.
+It's necessary to install the [esm](https://www.npmjs.com/package/esm) npm package before you can use the converters.
 
 ## obj2three.js
 
 Usage:
 
 ```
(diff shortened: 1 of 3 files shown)


mrdoob:
Nice!

mrdoob:
Thanks!

PR #18743, opened 2020-02-26
Global: Remove references to examples/js.

Mugen87:
This PR removes references to `examples/js` from the code base and docs. Besides, the code examples now assume that example files were imported via modules and thus are not in the `THREE` namespace anymore.

This PR is a preparation to deprecate `examples/js`.

### DIFF
--- a/docs/api/en/animation/KeyframeTrack.html
+++ b/docs/api/en/animation/KeyframeTrack.html
@@ -58,7 +58,7 @@
 
 		<p>
 			Some examples of how to manually create [page:AnimationClip AnimationClips] with different sorts
-			of KeyframeTracks can be found in the [link:https://threejs.org/examples/js/animation/AnimationClipCreator.js AnimationClipCreator]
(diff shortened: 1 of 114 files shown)


mrdoob:
Thanks!

hdtung88:
I think you guys shouldn't deprecate 'examples/js', because:
When I import three.js and its examples functions via modules, then obfuscate the whole code, webpack will obfuscate three.js too, this will makes the output js file size bigger => slower to load the application.
So normally I put all 'three.min.js' & others examples needed code in 'js' not 'jsm' & other libraries into a seperate file called 'libs.js'. and minify it. So webpack will have to obfuscate my main code only. This will reduce the total size of js codes alot.
Maybe there is someone have the same problem like me too ? What do you think ?
Anyway, thanks for your awesome library. You guys are my idols xD

Mugen87:
@hdtung88 Can you please move your post from this PR to #18749?

hdtung88:
ok sorry, do i need to remove it here also?

Mugen87:
I'll mark it as off-topic to avoid confusion.

hdtung88:
ok man

PR #18744, opened 2020-02-26
Box3: use .union() in .expandFromObject()

### LINKED ISSUE #18740: Box3.expandFromObject doesn't work when the given Line doesn't have vertices
When giving object3D with an empty bounding box to expandFromObject, new Box3 is not empty.
Max of new Box3 is (+Inf, +Inf, +Inf) and min of that is (-Inf, -Inf, -Inf).
```
const line = new Line(new Geometry());
line.geometry.computeBoundingBox();

console.log(line.geometry.boundingBox.min);
// Vector3 {x: Infinity, y: Infinity, z: Infinity, constructor: Object}
console.log(new Box3().setFromObject(line).min);
// Vector3 {x: -Infinity, y: -Infinity, z: -Infinity, constructor: Object}

console.log(line.geometry.boundingBox.max);
// Vector3 {x: -Infinity, y: -Infinity, z: -Infinity, constructor: Object}
console.log(new Box3().setFromObject(line).max);
// Vector3 {x: Infinity, y: Infinity, z: Infinity, constructor: Object}
```

This is because expandFromObject use expandFromPoint to set max and min.
To fix this, it should set max and min directly or ignore when max = -Inf and min = +Inf.
https://github.com/mrdoob/three.js/blob/c23b245814ac71999d8303c07434fcfd244288dd/src/math/Box3.js#L265-L266

##### Three.js version

r111 and later

##### Browser

- [x] All of them

##### OS

- [x] All of them

WestLangley:
Fixes #18740

### DIFF
--- a/src/math/Box3.js
+++ b/src/math/Box3.js
@@ -262,8 +262,7 @@ Object.assign( Box3.prototype, {
 			_box.copy( geometry.boundingBox );
 			_box.applyMatrix4( object.matrixWorld );
 
-			this.expandByPoint( _box.min );
-			this.expandByPoint( _box.max );
+			this.union( _box );
 
 		}
 


mrdoob:
Thanks!

PR #18752, opened 2020-02-27
Update WebAudio visualizer example with 3d content

Itee:
All in title.
In case you want keep the current "shadered" version maybe this one could be renamed as webaudio_visualizer_3d.

cheers !


mrdoob:
http://raw.githack.com/Itee/three.js/audio-visualizer/examples/webaudio_visualizer.html

Itee:
The default sound won't work with the previous link, BUT you still be able to use your microphone ;)

About Travis, need it to pre-build something (update image etc...) ?

Itee:
Mmmh it seems that githack does not implement the navigator.mediaDevices api, and disallow CORS request about the media device... So bad to see the fun stuff it is.

Itee:
This one work for me !

https://rawcdn.githack.com/Itee/three.js/bf5062d3b43222b4d1b7f20a9c6d43d21f5eb1da/examples/webaudio_visualizer.html

Mugen87:
The current version of `webaudio_timing` is nice and simple. This PR introduces an unnecessary complexity. I think the forum (category `Showcase`) is the better place to share this code.

Itee:
@Mugen87 

> The current version of webaudio_visualizer is nice and simple.

IMO it does not look like other webgl examples, it's not 3d which was the first goal of this library.

> This PR introduces an unnecessary complexity

Some others examples are much more complex than this one, but i could simplify a bit some parts to focus only on sound.

It display an FFT over the time which is more useful than the current, and it display the waveform too which is not in the current.

If you are talking about the usage of ThreeApplication this one could to contrary simplify hundred of current examples. It allow to end user to only see what the example try to show, and hide the rendering/camera/controls setup.
We had talking about this before... I can't remember where and when but this clearly could be usefull. If this part hurt you, i could refactor it...

Finally this example show a microphone usage with audio library of threejs and allow user to play with/see their voice. Which is one of most interactive example in this library !

munrocket:
In general you can run
```
npm run make-screenshot webaudio_visualizer
```
to update screenhot for Travis.

But right now webaudio [not deterministic](https://github.com/mrdoob/three.js/blob/bb8e4c57e4353df679fc3cac2cedd6cb624c5f75/test/e2e/deterministic-injection.js#L65) enough. Probably we will have a troubles in this case.

munrocket:
I like this PR. Please update screenshot with
```
npm run make-screenshot webaudio_visualizer
```
when you will done.

looeee:
I think this looks great, and is a nice improvement over the previous example. 

It's not any more complex than many other examples. However, if the complexity is a blocker, perhaps we could remove the controls? The initial settings look great in any case, and the music is funky.

makc:
> http://raw.githack.com/Itee/three.js/audio-visualizer/examples/webaudio_visualizer.html

Screen Shot 2020-03-05 at 1 00 27

Itee:
> > http://raw.githack.com/Itee/three.js/audio-visualizer/examples/webaudio_visualizer.html
> 
> Screen Shot 2020-03-05 at 1 00 27

Use this one: https://rawcdn.githack.com/Itee/three.js/bf5062d3b43222b4d1b7f20a9c6d43d21f5eb1da/examples/webaudio_visualizer.html

The audio file is not correctly hosted by raw.gitrack but is using rawcdn.gitrack (need to be in production mode)

Mugen87:
Thanks for the PR but the forum is a better place for this code.

PR #18761, opened 2020-02-27
CSM: Add fade

gkjohnson:
Add fade between shadow cascades for the CSM example and a fade out for the last cascade. The fade capability can be enabled or disabled by setting `csm.fade` to true or false and then calling `csm.updateUniforms`. The fade is achieved by expanding the shadow extents by some margin which is a function of the distance from the camera.

When `fade` is disabled the shader code is exactly the same as before.

It's subtle but here's an image showing the fade between cascades with practical split (look in the middle of the image):

image

It's more noticeable with the logarithmic split:

image

And the fade out behavior:

image

Temporary live link (fade must be enabled):

https://raw.githack.com/gkjohnson/three.js/csm-cascade-blend/examples/webgl_shadowmap_csm.html

/ping @vHawk


examples/jsm/csm/Shader.js
```diff
@@ -78,24 +78,64 @@ IncidentLight directLight;
...
+	#if defined( CSM_FADE )
+	for ( int i = 0; i < NUM_DIR_LIGHTS; i ++ ) {
+
+		directionalLight = directionalLights[ i ];
+		getDirectionalDirectLightIrradiance( directionalLight, geometry, directLight );
+
+		float margin = 0.25 * pow( linearDepth, 2.0 );
+		float csmx = CSM_cascades[ i ].x - margin / 2.0;
+		float csmy = CSM_cascades[ i ].y + margin / 2.0;
+		if( i < NUM_DIR_LIGHT_SHADOWS && linearDepth >= csmx && ( linearDepth < csmy ||  i == CSM_CASCADES - 1 ) ) {
```
gkjohnson: This is the case that prompted #18751.

It would be nice to remove the below `if(i < NUM_DIR_LIGHT_SHADOWS) {` block by using `#if ( UNROLLED_LOOP_INDEX < NUM_DIR_LIGHT_SHADOWS )` . Are there any suggestions for how to apply `#pragma unroll_loop` here?

vHawk:
Looks great!

WestLangley:
Maybe rename to `webgl_shadowmap_cascade.html` ?

vHawk:
@WestLangley The current name was mrdoob's suggestion. And I think it's certainly better than just "cascade". 
 https://github.com/mrdoob/three.js/pull/18481#issuecomment-585956450

vHawk:
If I set `shadow far` to ~300 and then check and uncheck `fade` checkbox shadows render incorrectly.

image

mrdoob:
This is looking great!

examples/jsm/csm/Shader.js
```diff
@@ -78,24 +78,65 @@ IncidentLight directLight;
 	DirectionalLightShadow directionalLightShadow;
 	#endif
 
+	#if defined( CSM_FADE )
+	for ( int i = 0; i < NUM_DIR_LIGHTS; i ++ ) {
+
+		directionalLight = directionalLights[ i ];
+		getDirectionalDirectLightIrradiance( directionalLight, geometry, directLight );
+
+		float margin = 0.25 * pow( linearDepth, 2.0 );
```
vHawk: Could you please explain why `margin` is calculated like this?

gkjohnson:
> If I set shadow far to ~300 and then check and uncheck fade checkbox shadows render incorrectly.

Thanks for playing with the extremes. I found another issue when making the `shadow far` options really small (~100) which causes the background to become brighter when zooming the camera far away. Checking the fade box on and off changes the result, too. A shader field probably isn't being updated somewhere.

image

I'll take a look at both in the next few days. Thanks!

WestLangley:
>@WestLangley The current name was mrdoob's suggestion. And I think it's certainly better than just "cascade".

...Only if you know what "csm" stands for... and if you know to search for "csm" instead of "cascade" in the examples.

`webgl_shadowmap_cascading.html` seems more user-friendly to me.

mrdoob:
> ...Only if you know what "csm" stands for... and if you know to search for "csm" instead of "cascade" in the examples.

I think it's a good name for now. Depending how things evolve it could become the default anyway.

mrdoob:
Thanks!

gkjohnson:
Oh! I was going to fix those two issues before merging this. When is the next release? Maybe I can address them before that otherwise I'll get it done for r115.

mrdoob:
Tomorrow 🤞

gkjohnson:
Ha okay sounds like r115 is more likely then!

PR #18768, opened 2020-02-28
CSM: Miscellaneous Fixes

gkjohnson:
A quick follow up to #18761.

- Adds a `dispose` function to remove CSM from materials.
- Fixes issue which was causing new shaders to be added to the `shaders` array every time a material was recompiled resulting.
- Fix issue mentioned in https://github.com/mrdoob/three.js/pull/18761#issuecomment-592138879 which was caused by stale values being used in `onBeforeCompile`.

I'm still working on the issue I found (https://github.com/mrdoob/three.js/pull/18761#issuecomment-592162657) but the solution wasn't immediately obvious.

Temporary live link:
https://raw.githack.com/gkjohnson/three.js/csm-fixes/examples/webgl_shadowmap_csm.html

@vHawk

### DIFF
--- a/examples/jsm/csm/CSM.js
+++ b/examples/jsm/csm/CSM.js
@@ -386,4 +386,26 @@ export default class CSM {
 
 	}
 
+	dispose() {
+
+		const shaders = this.shaders;
+		shaders.forEach( function ( shader, material ) {
+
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

PR #18773, opened 2020-02-28
Examples: Better searching filter

munrocket:
Draft for #18700 fix.
Actually it's already works, but `tags` filled with `info`.
Same searching performance as before.


mrdoob:
This approach forces adding a tag to all the examples.
My suggestion in #18700 only added them in a case by case as needed.

Not sure what's best.

looeee:
@mrdoob not all the examples have tags here though:

> { 'file': 'webgl_camera_array', 'tags': '' },

If you want to keep the previous style and only add the 'tags' key when there are actual tags, what about this?

``` js
var files = {
	'webgl_animation_cloth',
        { 'file': 'webgl_animation_keyframes', 'tags': 'animation,keyframes' },
        'webgl_animation_skinning_blending',
        ...
```

If the entry is a string, proceed as before, and if it's an object then look for tags key. 

A more complicated alternative would be to use `<meta name="keywords" content="animation,keyframes">` That would require parsing all the HTML files though, or otherwise building an index.

munrocket:
Cons: this commit is wipe all the history, but we can commit it as `anonymous@domain.org`.
Pros: obviously better file structure, that force others to add tags

looeee:
> Cons: this commit is wipe all the history, but we can commit it as anonymous@domain.org.

Not sure I understand what you mean here. What history is being wiped?

makc:
@looeee I guess it means git blame will no longer work

munrocket:
If you are using [GitLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) or similar tool, you can see who edited each line. This file is very rich with different authors.

```js
var files = {
	'webgl_animation_cloth',
        { 'file': 'webgl_animation_keyframes', 'tags': 'animation,keyframes' },
        'webgl_animation_skinning_blending',
        ...
```
This file structure will be deoptimized by V8 and other JS engines, because each line have different shapes. Better to keep similar type in each line.
```js
var files = {
	{ 'file': 'webgl_animation_cloth', 'tags': '' },
        { 'file': 'webgl_animation_keyframes', 'tags': 'animation,keyframes' },
        { 'file': 'webgl_animation_skinning_blending', tags': '' },
        ...
```

makc:
@munrocket here at github you have this little button that fixes it:
Screen Shot 2020-02-29 at 12 23 37

munrocket:
@makc right now it's your commit, but my pull request.


You not solve a problem in general as I asked too, I have a same issue with example `translucency` which I am tried to find with `subsurface scattering`. To solve this problem properly we need to check all examples by hand and add tags not only in `bloom` example.

makc:
@munrocket yes, totally, as commit message says, it fixes #18700 which was about glow specifically, and you could add any number of tags once it is accepted. 

I did not do it because I did not see an easy way to separate good tags from bad tags in your PR, sorry

munrocket:
Ok, as you wish. I am probably will try to add more tags in whatever system will be.

looeee:
> If you are using GitLens or similar tool, you can see who edited each line. This file is very rich with different authors.

Ah, got it. I don't think this is a reliable way of knowing who authored the actual examples though, so that probably doesn't matter. 

> This file structure will be deoptimized by V8 and other JS engines, because each line have different shapes. 

This code is not on a hot path so I doubt it will be optimized in the first place.

PR #18787, opened 2020-03-01
Examples: add contact shadow example

marcofugaro:
Following discussion in #18746

### DIFF
--- a/examples/webgl_shadow_contact.html
+++ b/examples/webgl_shadow_contact.html
@@ -270,6 +270,7 @@
 				scene.background = initialBackground;
 
 				renderer.render( scene, camera );
+				stats.update();
 
 			}
 
--- a/examples/webgl_shadow_contact.html
+++ b/examples/webgl_shadow_contact.html
@@ -56,6 +56,7 @@
 				camera.position.set( 0.5, 1, 2 );
 
 				scene = new THREE.Scene();
+				scene.background = new THREE.Color( 0xffffff );
 
 				stats = new Stats();
 				document.body.appendChild( stats.dom );
@@ -103,7 +104,7 @@
 
 
 				// make a plane and make it face up
-				planeGeometry = new THREE.PlaneGeometry( PLANE_WIDTH, PLANE_HEIGHT ).rotateX( Math.PI / 2 );
+				planeGeometry = new THREE.PlaneBufferGeometry( PLANE_WIDTH, PLANE_HEIGHT ).rotateX( Math.PI / 2 );
 				planeMaterial = new THREE.MeshBasicMaterial( {
 					map: renderTarget.texture,
 					opacity: state.opacity,
@@ -128,20 +129,21 @@
 				cameraHelper = new THREE.CameraHelper( shadowCamera );
 
 				// like MeshDepthMaterial, but goes from black to transparent
-				depthMaterial = new THREE.ShaderMaterial( {
-					uniforms: {
-						darkness: { value: state.darkness },
-					},
-					vertexShader: THREE.ShaderChunk[ 'depth_vert' ],
-					fragmentShader: `
+				depthMaterial = new THREE.MeshDepthMaterial();
+				depthMaterial.userData.darkness = { value: state.darkness };
+				depthMaterial.onBeforeCompile = function ( shader ) {
+
+					shader.uniforms.darkness = depthMaterial.userData.darkness;
+					shader.fragmentShader = `
 						#define DEPTH_PACKING 3200
 						uniform float darkness;
-						${THREE.ShaderChunk[ 'depth_frag' ].replace(
+						${shader.fragmentShader.replace(
 					'gl_FragColor = vec4( vec3( 1.0 - fragCoordZ ), opacity );',
 					'gl_FragColor = vec4( vec3( 0.0 ), ( 1.0 - fragCoordZ ) * darkness );'
 				)}
-					`,
-				} );
+					`;
+
+				};
 				depthMaterial.depthTest = false;
 				depthMaterial.depthWrite = false;
 
@@ -159,7 +161,7 @@
 				gui.add( state, 'blur', 0, 15, 0.1 );
 				gui.add( state, 'darkness', 1, 5, 0.1 ).onChange( function () {
 
-					depthMaterial.uniforms.darkness.value = state.darkness;
+					depthMaterial.userData.darkness.value = state.darkness;
 
 				} );
 				gui.add( state, 'opacity', 0, 1, 0.01 ).onChange( function () {
@@ -183,10 +185,9 @@
 
 				//
 
-				renderer = new THREE.WebGLRenderer( { antialias: true, alpha: true } );
+				renderer = new THREE.WebGLRenderer( { antialias: true } );
 				renderer.setPixelRatio( window.devicePixelRatio );
 				renderer.setSize( window.innerWidth, window.innerHeight );
-				renderer.outputEncoding = THREE.sRGBEncoding;
 				document.body.appendChild( renderer.domElement );
 
 				//
--- a/examples/webgl_shadow_contact.html
+++ b/examples/webgl_shadow_contact.html
@@ -1,0 +1,277 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<title>three.js webgl - contact shadows</title>
+		<meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


looeee:
[Live version](https://raw.githack.com/marcofugaro/three.js/contact-shadow/examples/webgl_shadow_contact.html). 

I'm not seeing any shadows though.

Annotation 2020-03-01 084606

marcofugaro:
Hmm @looeee which browser are you using?

Mugen87:
I see the same like @looeee  on Chrome 80.0.3987.122 and Firefox 73.0.1,  tested on macOS 10.15.3.

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,278 @@
...
+
+			function init() {
+
+				camera = new THREE.PerspectiveCamera( 50, window.innerWidth / window.innerHeight, 0.1, 100 );
+				camera.position.set( 0.5, 1, 2 );
+
+				scene = new THREE.Scene();
+
+				stats = new Stats();
+				document.body.appendChild( stats.dom );
```
Mugen87: You have to update `stats` in the animation loop.

marcofugaro: Done!

marcofugaro:
Alright fixed, not a browser issue, my fork was referring to an outdated version of the Three.js build.

Githack should update soon.

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,279 @@
...
+				renderTarget = new THREE.WebGLRenderTarget( 512, 512 );
+				renderTarget.texture.generateMipmaps = false;
+
+				// the render target that we will use to blur the first render target
+				renderTargetBlur = new THREE.WebGLRenderTarget( 512, 512 );
+				renderTargetBlur.texture.generateMipmaps = false;
+
+
+				// make a plane and make it face up
+				planeGeometry = new THREE.PlaneGeometry( PLANE_WIDTH, PLANE_HEIGHT ).rotateX( Math.PI / 2 );
```
Mugen87: Can you please use `PlaneBufferGeometry` here?

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,279 @@
...
+
+				// the camera to render the depth material from
+				shadowCamera = new THREE.OrthographicCamera( - PLANE_WIDTH / 2, PLANE_WIDTH / 2, PLANE_HEIGHT / 2, - PLANE_HEIGHT / 2, 0, CAMERA_HEIGHT );
+				shadowCamera.rotation.x = Math.PI / 2; // get the camera to look up
+				shadowGroup.add( shadowCamera );
+
+				cameraHelper = new THREE.CameraHelper( shadowCamera );
+
+				// like MeshDepthMaterial, but goes from black to transparent
+				depthMaterial = new THREE.ShaderMaterial( {
```
Mugen87: Could you also use `MeshDepthMaterial` here and perform the changes with `onBeforeCompile()`?

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,279 @@
...
+
+						scene.remove( cameraHelper );
+
+					}
+
+				} );
+
+				//
+
+				renderer = new THREE.WebGLRenderer( { antialias: true, alpha: true } );
```
Mugen87: I don't think it's necessary to set `alpha` to `true`...

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,279 @@
...
+				renderer.setPixelRatio( window.devicePixelRatio );
+				renderer.setSize( window.innerWidth, window.innerHeight );
+				renderer.outputEncoding = THREE.sRGBEncoding;
```
Mugen87: This line can also be removed.

examples/webgl_shadow_contact.html
marcofugaro: Yes, it's to show that the shadows rendered are transparent (i was using them in AR).
The white you see is the white of the body.

Should I use `renderer.setClearColor(0xffffff, 1)` instead?

Mugen87: Try it with `scene.background = new THREE.Color( 0xffffff );`

looeee:
I think this would look better if the ground plane was not transparent. It looks kind of weird when the object cut through their shadows:

Annotation 2020-03-01 212823

Or is there a reason why the ground is transparent?

marcofugaro:
@looeee yeah the objective was to show that they can be rendered against a transparent plane, However I can try putting a slider with the floor opacity

munrocket:
Please, add screenshot in PR with
```
npm run make-screenshot webgl_shadow_contact
```
we need to add automatic suggestion like this inside Travis

marcofugaro:
Thank you @munrocket didn't know that. Also a "Contributing" section in the readme or in the docs would be nice

munrocket:
Ok, will do.

munrocket:
@marcofugaro actually we already have new contribution guide. And without screenshot this probably will not merged, because every new commit will be with broken Travis too. So try to make it, if it possible on your GPU.

mrdoob:
> @marcofugaro actually we already have new contribution guide. And without screenshot this probably will not merged

Actually. Lets revisit that.

I would prefer not to make the contribution process more complicated. I can just run the make-screenshot myself.

marcofugaro:
No problem, just had to get around to doing it 🙂 

@munrocket you're right, I found it in the wiki! I've been using Three for years and didn't know the wiki even existed!

Also I added a plane with the color as suggested by @looeee, ended up making another mesh rather than playing with blendings in the shadow plane. Please review @Mugen87

examples/webgl_shadow_contact.html
```diff
@@ -0,0 +1,312 @@
...
+				verticalBlurMaterial.uniforms.v.value = amount * 1 / 256;
+
+				renderer.setRenderTarget( renderTarget );
+				renderer.render( blurPlane, shadowCamera );
+
+				blurPlane.visible = false;
+
+			}
+
+			function animate( time ) {
```
Mugen87: It seems `time` is defined but never used.

munrocket:
Kek, travis froze on raycasting. Never happened before and here again 😏

munrocket:
Looks like the current screenshot generator work only for me.
```
ERROR! Diff wrong in 0.987 of pixels in file: webgl_shadow_contact
```

munrocket:
I need to ask help from friends with Windows machines and another GPUs.
@marcofugaro please, just add `webgl_shadow_contact` in [exception_list](https://github.com/mrdoob/three.js/blob/dev/test/e2e/puppeteer.js) if it's not hard for you.
Sorry but you was the first who generated a new screenshot by yourself.

marcofugaro:
No problem, I'll leave you guys generate the screenshots as @mrdoob said 😅

Mugen87 (approved):

munrocket:
@Mugen87 @looeee I asked two people to check current regression tests. When software is working _only_ on my machine and in CI this is beyond my force. I literally can't fix it even if I want.

Right now I have two version _one_ in three.js repo, and [_another_](https://github.com/munrocket/puppeteer-webgl-ci) experimental in my repo, the last change that I have is to try to mix it in draft PR with new approach and make a stress test on current codebase. 

Probably some flag from that repo will be suitable for Marcofugaro and Mr.Doob PCs.

makc:
what a weird thing about this demo, it uses two blur shaders (h/v) each of them has a bunch of texture reads. normally this is slow on my old laptop, but this on is very fast. I wish I could understand why.

marcofugaro:
@makc not even that, the demo is doing 2 blur rounds, so in total there are 4 passes each frame!
The blur shader implemented in Three.js is very efficient, you can read more about it [here](http://vodacek.zvb.cz/archiv/251.html) and [here](http://rastergrid.com/blog/2010/09/efficient-gaussian-blur-with-linear-sampling/)

Mugen87:
There is now a merge conflict  because of #18803. Do you mind removing `webgl_shadow_contact.png` from your PR?

marcofugaro:
Sure!

marcofugaro:
@Mugen87 done!

munrocket:
Sorry that I am talking about tests again here, but I am asked people with nvidia 2060 and 1060ti. And all of them able to generate new screenshots. @marcofugaro what videocard/OS do you use, I can't reproduce the bug.

mrdoob:
Thanks!

PR #18789, opened 2020-03-01
CSM: Remove unnecessary logic in "update"

gkjohnson:
With #18777 it's no longer necessary to update the shadow camera bounds every frame because the the xy dimensions are always based on the longest edge of the frustum. This PR moves the shadow bounds update into `updateFrustums` which only gets called when the camera frustum changes in order to reduce the work needed to update every frame. It also removes the need to pass `matrixWorld` into `update` and instead just uses the member cameras `matrixWorld`.

Temporary Live Link:
https://raw.githack.com/gkjohnson/three.js/csm-simplify/examples/webgl_shadowmap_csm.html

@vHawk


mrdoob:
Thanks!

PR #18793, opened 2020-03-02
InstancedMesh: Add note about material sharing.

Mugen87:
see https://github.com/mrdoob/three.js/issues/17701#issuecomment-593164460


mrdoob:
Thanks!

PR #18795, opened 2020-03-02
Raycaster: Stop traversing children when layers test false

Tirzono:
#18694 introduced testing on layers instead of the visibility of the object. However, with that change it also changed the behaviour where it wouldn't stop traversing children if the object would test false. I am using `Group`s and setting the layer/visibility on the group. The renderer picks this up correctly: if the `Group`'s visibility is `false` or if the `Group`'s layer is testing `false`, the children are also not rendered. I would like to see `Raycaster` to match this behaviour.

I am not sure if this was intentional or not, so I am happy to discuss this if the change was intentional.

### DIFF
--- a/src/core/Raycaster.js
+++ b/src/core/Raycaster.js
@@ -46,11 +46,9 @@ function ascSort( a, b ) {
 
 function intersectObject( object, raycaster, intersects, recursive ) {
 
-	if ( object.layers.test( raycaster.layers ) ) {
+	if ( object.layers.test( raycaster.layers ) === false ) return;
 
-		object.raycast( raycaster, intersects );
-
-	}
+	object.raycast( raycaster, intersects );
 
 	if ( recursive === true ) {
 


WestLangley:
My understanding is layers are not inherited.

To me, that means the layer of a non-renderable objects (`Object3D` and `Group`) has no effect.

Mugen87:
> I am not sure if this was intentional or not

Yes, this was done on purpose to provide more flexibility in context of raycasting. You can solve this issue on the application layer by configuring the same layer for all subsequent object of a group.

Tirzono:
Groups are easy to use for me in a way to apply the visibility to it without having to traverse all the objects. Especially because the objects that I add to some of the groups are loaded in an async way: I didn't have to care about applying the right visibility state _at that moment_ because it is inherited from the group. 

Right now I am working towards some helper functions doing this:

```javascript
const setVisible = (object, visible = true) => {
    object.layers.disable(visible ? Constants.LAYERS.INVISIBLE : CanvasConstants.LAYERS.VISIBLE);
    object.layers.enable(visible ? Constants.LAYERS.VISIBLE : CanvasConstants.LAYERS.INVISIBLE);
    object.visible = visible;
};

const visibleLayer = new Layers(Constants.LAYERS.VISIBLE);

const isVisible = (object) => object.layers.test(visibleLayer);
```

It feels a bit clunky as I only use these layers now for visibility and I also have to apply them to all the objects and make sure their state is always correct. On the other hand, I see why the change has been made, because it gives you a bit more flexibility.

Thank you for your quick responses. Feel free to close the PR as the change was intentional. I'll have a look to see what works best for me to work around this change. For now I'll probably port the old Raycaster to be able to upgrade this package without a big rewrite.

PR #18813, opened 2020-03-05
Update types of nullable bounding boxes and spheres

flostellbrink:
Bounding boxes and spheres can be computed on geometry.
These bounding boxes are not computed by default and are instead set to null.
They only become non-null when they are explicitly being computed.

This change reflects that by changing the types of bounding boxes and spheres to be nullable. Otherwise typescript assumes that these values cannot be null.


mrdoob:
Thanks!

PR #18822, opened 2020-03-06
AnimationMixer: Additive Animations

c-morten:
# Additive Animation Overview

Additive animation is a useful feature present in several 3D engines such as Unity and Unreal Engine. It can be used to layer motions on top of each other to build complex performances from smaller re-usable assets. 

By default animation is considered to be ‘override’ - if you play 2 animations on the same rig simultaneously each animation will have 50% influence over the result. The result is always normalized, so the more animations playing at the same time, the smaller amount of influence each individual animation has over the final result. All of the keyframes in override animations are stored relative to the parent. Say for example you have an object with 2 override animations. The first animation has a translation value of [[0, 0, 0]] at frame 0, then it interplates to [[0, 2, 0]] on frame 30, then back to [[0, 0, 0]] on frame 60. The second animation has a translation value of [[0, 0, 0]] at frame 0, interpolates to [[0, 0, 2]] on frame 30, and then back to [[0, 0, 0]] on frame 60. If you played these animations simultaneously at full weight, frame 30 would result in a translation value of [[0, 1, 1]]. Neither the Y or Z axes would ever be able to fully reach a value of 2 with both animations playing.

Additive animation is unique because it does not use that type of normalization logic. You can have N-number of additive animations playing simultaneously and each one will have the exact amount of influence specified. To accomplish this, additive animation values are relative to the object’s value in the previous animation played in a stack, not the parent. So if the animations in the example above were played additively, frame 30 would result in a value of [[0, 2, 2]] because the second animation’s value adds on top of the first.

There are generally 2 aspects to implementing additive animations. The obvious part is runtime playback, different math is used when evaluating animations additively than is used when evaluating override animations. The other part that needs to be addressed is authoring. Because additive animation on an object is relative to its own space, there is no intuitive way to author these animations for objects in hierarchies. Instead a method of conversion needs to be implemented. This involves specifying a frame or pose to ‘subtract’ out of the other keyframes so they are relative to it.

&nbsp;
&nbsp;
## Additive Animation for three.js

The animation system in three.js only supports override animation by default. By making modifications to existing functionality and adding in a conversion method additive animation can be accomplished:

&nbsp;
### math/Quaternion.js

The addition of a **multiplyQuaternionsFlat** method allows for quaternion multiplication to be performed on and applied to flat arrays of numbers rather than Quaternion objects. This allows the evaluation of additive keyframes to happen directly on the PropertyMixer buffers.

&nbsp;
### animation/AnimationUtils.js

The addition of a **makeClipAdditive** method allows existing AnimationClip objects’ keyframes to be converted to an additive format. The reference frame supplied determines the value that resulting keyframes will be relative to. For non-quaternion numeric KeyframeTracks, the value of the reference frame is simply subtracted from every keyframe in the track. For QuaternionKeyframeTracks, the value at the reference frame is normalized and conjugated. It is then pre-multiplied with every keyframe in the track. Non-numeric tracks are left un-altered since additive does not have real meaning for these types.

&nbsp;
### animation/PropertyMixer.js

This is where most of the work is done for additive animation implementation. The buffer was extended to accomodate additive values and cumulative weight values are stored separately for additive animation. Quaternion mixers have one extra buffer region at the end for storing intermediate quaternion multiplication results. A _setIdentity property is bound to the appropriate _setIdentity method based on input type.

An **accumulateAdditive** method has been added to handle accumulation on additive AnimationActions. When this method is called and cumulativeAdditive weight is 0, it will first copy the identity values back into the accumulation region. Regardless of cumulativeAdditive weight, the incoming value will then be mixed with the accumulation region by the incoming weight value.

The **apply** method has been modified to use the additive mix function to add the additive buffer region to the current override cumulative buffer region and store the result in the override buffer region’s next cumulative region just before the buffer is applied to the binding.

The **saveOriginalState** method has been modified to call the _setIdentity method after saving the original state to the override buffer regions.

A **_setAdditiveIdentityNumeric** method has been added to fill the additive buffer region with 0.

A **_setAdditiveIdentityQuaternion** method has been added to fill the additive buffer region with 0, 0, 0, 1.

A **_setAdditiveIdentityOther** method has been added to fill the additive buffer region with the original property values.

A **_slerpAdditive** method has been added to perform mixing for additive quaternions. It multiplies the accumulation and incoming regions, then slerps between the accumulation and result by the weight amount.

A **_lerpAdditive** method has been added to perform mixing on all other additive numeric types. It is similar to _lerp except that it does not normalize the weights of the accumulation and incoming regions before adding them. It just adds the weighted incoming region to the unmodified accumulation region.

&nbsp;
### animation/AnimationMixer.js

The **clipAction** method was modified to add an isAdditive parameter to allow users to specify whether they want the resulting AnimationAction to be evaluated as additive. This does not make any modifications to the contents of the AnimationClip parameter, it is expected that the user will make any clip keyframe conversions they see necessary using AnimationUtils.makeClipAdditive.

&nbsp;
### animation/AnimationAction.js

The **_update** method was modified to use either 'accumulate' or 'accumulateAdditive' on the PropertyMixers, depending on the AnimationAction's isAdditive state.

&nbsp;
### examples/webgl_animation_skinning_additive_blending.html

This example was added to demonstrate additive animation. The panel folder labeled ‘Base Actions’ contains items that can be clicked on to transition between override animations on the character. ‘None’ will transition to the character’s base pose. The panel folder labeled ‘Additive Action Weights’ contains a series of sliders whose values can be manipulated 0 - 1. These determine the amount each additive animation is added on top off the base animations. The model was downloaded from [www.mixamo.com](http://www.mixamo.com/), as were characters from other existing examples.

&nbsp;
### examples/files.js

This was modified to add 'webgl_animation_skinning_additive_blending'

&nbsp;
### examples/models/gltf/Xbot.glb

This is the model and animations used in the 'webgl_animation_skinning_additive_blending' example


marcofugaro:
Really interesting stuff! It's fun playing with it 🙂 
Also the shadows are on the opposite side!

munrocket:
Can I help you with screenshot? What OS and GPU do you use? Looks like it's not fit for travis.

c-morten:
I’m on Windows 10. I had to do a bunch of extra installation steps for puppeteer just to get the screenshot to work at all. And I get a different result depending on if the screenshot file already exists locally, if it does the new screenshot has the UI in it and if it doesn’t then it does not. I’m assuming this is an issue of the timing of when the screenshot is taken, originally I was getting a white image because it was being taken before the glTF asset loaded, I moved the render loop initialization to later in the script and that resolved it locally but it could still be an issue for travis.

munrocket:
This PR is fun! She is sneaking :D

munrocket:
@c-morten good point I am took one screenshot with dat.gui and stat.js too.

WestLangley:
I am not an expert in the use of the three.js animation tools, but I hope those who are can give the OP feedback. I think this is awesome.

src/animation/AnimationAction.js
```diff
@@ -11,11 +11,12 @@ import { WrapAroundEnding, ZeroCurvatureEnding, ZeroSlopeEnding, LoopPingPong, L
  *
  */
 
-function AnimationAction( mixer, clip, localRoot ) {
+function AnimationAction( mixer, clip, localRoot, isAdditive = false ) {
```
mrdoob: How about making `AdditiveAnimationAction` and extend `AnimationAction` instead?

bhouston: I think that this could be a mode rather than a class.  Similar to Pixel Blending modes.  The default is Over ( source * a +  dest * (1 - a) ), and this is Additive ( source * a + dest * 1 ).  I wouldn't be surprised if there are additional animation modes.

bhouston: Unity calls this Blending mode and it is an enum type, rather than a true/false.

- "On each layer, you can specify the mask (the part of the animated model on which the animation would be applied), and the Blending type. Override means information from other layers will be ignored, while Additive means that the animation will be added on top of previous layers."  (Source: https://docs.unity3d.com/Manual/AnimationLayers.html)

In the Porter Duff compositing modes we get:
- SRC - replace.
- SRC_OVER - is override.
- ADD is additive.

c-morten:
I will introduce a new action as mrdoob suggests. That way the action can
manage which type of blending is called, and each new blending mode that
may be introduced would be a new action.

src/animation/AnimationAction.js
c-morten: So I could introduce a new enum called AnimationBlendMode containing Over and Additive values, then change that parameter to be called 'blendMode' and default it to 'Over'. Does that sound about right?

mrdoob: Blending mode sounds good to me. I was just trying to avoid ending in a situation like this:

```js
function AnimationAction( mixer, clip, localRoot, isAdditive = false, isSubtrative = true )
```

Can we reuse already available constants?

```js
function AnimationAction( mixer, clip, localRoot, blending = NoBlending | AdditiveBlending )
```

c-morten:
I could use the existing BlendingMode enum for materials, but do you think it would be confusing that only two of the values in that enum are valid for animation? If not I could use the NormalBlending value for override and AdditiveBlending value for additive.

src/animation/AnimationAction.js
WestLangley: >Can we reuse already available constants?

Please, lets not do that. Instead, define something like:
```js
NormalAnimationBlending = 3300; // or 0
AdditiveAnimationBlending = 3301; // or 1
OverrideAnimationBlending = 3302; // or 2 -- if we support that
```

c-morten:
Perfect, thanks for the quick response. I will get the PR updated with this change first thing Monday.

mrdoob:
This is almost ready to go 👌

@WestLangley `NormalAnimationBlending` or `NormalAnimationBlendMode`?

WestLangley:
>@WestLangley NormalAnimationBlending or NormalAnimationBlendMode?

`BlendMode` sounds better to me.

src/animation/AnimationMixer.js
```diff
@@ -516,7 +517,7 @@ AnimationMixer.prototype = Object.assign( Object.create( EventDispatcher.prototy
 	// return an action for a clip optionally using a custom root target
 	// object (this method allocates a lot of dynamic memory in case a
 	// previously unknown clip/root combination is specified)
-	clipAction: function ( clip, optionalRoot ) {
+	clipAction: function ( clip, optionalRoot, blendMode = OverrideBlendMode ) {
```
donmccurdy, line 520: If an AnimationAction has a blendMode, then should an AnimationClip have that too?

Otherwise, if we were to implement loading additive animation in FBXLoader ([the FBX format supports this](https://help.autodesk.com/view/FBX/2017/ENU/?guid=__files_GUID_29727281_5732_4F89_A06A_BBADD4C8993F_htm)) the user would need to know which blendMode to use when creating actions from the clips returned by the loader. Similarly, the information would be lost when serializing animations to/from JSON.

src/animation/AnimationUtils.js
```diff
@@ -231,6 +233,105 @@ var AnimationUtils = {
 
 		return clip;
 
+	},
+
+	makeClipAdditive: function ( sourceClip, referenceFrame = 0, cloneOriginal = false, clonedName, fps = 30 ) {
```
donmccurdy, line 238: I would suggest omitting the 'cloneOriginal' and 'clonedName' parameters. The user can clone or rename the source clip if they would like to do so.

src/animation/AnimationUtils.d.ts
```diff
@@ -23,4 +23,11 @@ export namespace AnimationUtils {
 		endFrame: number,
 		fps?: number
 	): AnimationClip;
+	export function makeClipAdditive(
+		sourceClip: AnimationClip,
+		referenceFrame?: number,
+		cloneOriginal?: boolean,
+		clonedName?: string,
+		fps?: number
+	): AnimationClip;
```
donmccurdy, line 32: Could you help me understand how `makeClipAdditive` should be used? I am new to animation blending modes, and trying to learn this. :) From reading a bit, it sounds like one common workflow would involve starting with several standalone — not yet additive — animations:

- idle
- walk
- run
- arm wave `(~‾⌣‾ )~`

Now suppose we want to make the "arm wave" animation additive, so it can be played on top of the others. Is it always the case that the additive animation must have a frame at the base pose? Might the artist need to make it additive by subtracting a different animation, or a static pose, from it? 

I guess I am wondering if the `makeClipAdditive` function should perhaps take _two_ clips, `referenceClip` and `targetClip`, and subtract the reference frame of `referenceClip` from each frame of `targetClip`? Like:

```ts
makeClipAdditive(referenceClip, referenceFrame, targetClip, fps = 30)
```

donmccurdy (commented):
This looks very useful, thank you @c-morten!

src/animation/AnimationUtils.d.ts
c-morten: makeClipAdditive would be generally be used for animations on hierarchies, like a biped skeleton. Additive animations add on top of the result of non-additive animations, so if you want the character's 'bones' to remain the same length when you add an additive animation, you'd need to author the animation with all of the joints having local translation values of [0,0,0]. Not very intuitive since you just end up with a blob at the origin. Instead you can author the animation as you normally would and then use makeClipAdditive to subtract a pose out, which would result in zeroing out the translation values if you did not change them during the animation.

I like your idea of including a reference clip as an input, seems like it will make things a bit more flexible. I will definitely add that in, thanks for the feedback!

src/animation/AnimationMixer.js
c-morten: I could add it to AnimationClip as well and rather than default to override blending when creating an animation action, default to the blend mode stored on the clip.

donmccurdy: That would be my preference, but let's confirm with @mrdoob or @bhouston before you spend time on the change. :)

src/animation/AnimationUtils.d.ts
donmccurdy: > ... you'd need to author the animation with all of the joints having local translation values of [0,0,0]. Not very intuitive since you just end up with a blob at the origin. Instead you can author the animation as you normally would ...

I see, thanks — it appears that authoring programs like Maya and Blender do support additive animation. Do you mean that creating the animation that way is just less convenient or less common?

c-morten: Those programs do support additive animation, but it's not a standard feature for exported formats. The only exported file format I know of that supports additive animation is FBX, and it requires that the  animation layers be kept intact. This is not a common export case because most people animate using control rigs and so have to bake the results when exporting. Baking merges everything into the base layer and results in adding the reference pose/animation back in.

src/animation/AnimationClip.js
```diff
@@ -16,11 +17,12 @@ import { MathUtils } from '../math/MathUtils.js';
  * @author David Sarno / http://lighthaus.us/
  */
 
-function AnimationClip( name, duration, tracks ) {
+function AnimationClip( name, duration, tracks, blendMode ) {
 
 	this.name = name;
 	this.tracks = tracks;
 	this.duration = ( duration !== undefined ) ? duration : - 1;
+	this.blendMode = blendMode || NormalAnimationBlendMode;
```
WestLangley, line 25: We avoid this pattern because it assumes `blendMode` is non-zero.

It is best to use the same pattern as is used for `duration`.

src/animation/AnimationUtils.js
```diff
@@ -231,6 +234,116 @@ var AnimationUtils = {
 
 		return clip;
 
+	},
+
+	makeClipAdditive: function ( targetClip, referenceFrame = 0, referenceClip, fps = 30 ) {
+
+		if ( referenceClip === undefined ) {
```
mrdoob, line 241: Can we unify all these and 
```js
makeClipAdditive: function ( targetClip, referenceFrame, referenceClip, fps ) {

	if ( referenceFrame === undefined ) referenceFrame = 0;
	if ( referenceClip === undefined ) referenceClip = targetClip;
	if ( fps === undefined ) fps = 30;
```

src/animation/AnimationUtils.js
```diff
@@ -231,6 +234,116 @@ var AnimationUtils = {
...
+		for ( var i = 0; i < numTracks; ++ i ) {
+
+			var referenceTrack = referenceClip.tracks[ i ];
+			var referenceTrackType = referenceTrack.ValueTypeName;
+
+			// Skip this track if it's non-numeric
+			if ( referenceTrackType === 'bool' || referenceTrackType === 'string' ) continue;
+
+			// Find the track in the target clip whose name and type matches the reference track
+			var targetTrack = targetClip.tracks.find( track => {
```
mrdoob, line 262:
```js
var targetTrack = targetClip.tracks.find( function ( track) {
```

mrdoob:
Thanks!

mrdoob:
@c-morten I think there is something wrong with the normals of the model.

Screen Shot 2020-04-13 at 12 32 03 PM

c-morten:
> @c-morten I think there is something wrong with the normals of the model.
> 
> Screen Shot 2020-04-13 at 12 32 03 PM

@mrdoob Is this a blocking issue that needs to be addressed? I got the model and animations from Mixamo and used Blender to combine everything into one file and convert to .glb. I was able to reproduce the image above and it appears to be an issue with exporting from Blender to .glb. If I export the same model from Maya to .glb the normals look correct. Unfortunately Blender is the only application I've found that supports exporting multiple animations to the same .glb file.

munrocket:
Tests seems ok here.

mrdoob:
@donmccurdy any ideas?

mrdoob:
@c-morten could you share the .blend file?

WestLangley:
Setting only a hemisphere light with red and green colors, it appears to me the normals are rotated 90 degrees.

c-morten:
[xbot.zip](https://github.com/mrdoob/three.js/files/4476334/xbot.zip)

@mrdoob Here's the source file I exported from.

mrdoob:
@c-morten Thanks!

Seems like the Xbot is rotated 90° in the X axis and I think we do not apply the object rotated normal to the skinning.

Screen Shot 2020-05-13 at 7 51 31 AM

What's easier? Removing the rotation on the model? Or trying to improve the renderer? 🤔

c-morten:
@mrdoob funny that you got back to me about this today, I just ran into this same issue with some other fbx assets I brought into Blender and exported back out as glTF. I think it's actually an issue with Blender and importing files from different coordinate systems. Blender is Z up, it appears that when it imports Y up files it is skinning them in the Z up rotation and rotating everything afterwards. It looks ok the first time you import but if you export to glTF and then import that file back into Blender the normals look incorrect and the mesh looks really shiny. I found that what works best for imported rigs is to import just the mesh and freeze all the transforms off, then bring in the rigged version and copy the weights from that onto the clean version of the mesh. I'll submit a PR with a new xbot file using this process.

src/animation/AnimationUtils.js
```diff
@@ -231,6 +234,112 @@ var AnimationUtils = {
...
+
+			var valueSize = referenceTrack.getValueSize();
+			var lastIndex = referenceTrack.times.length - 1;
+			var referenceValue;
+
+			// Find the value to subtract out of the track
+			if ( referenceTime <= referenceTrack.times[ 0 ] ) {
+
+				// Reference frame is earlier than the first keyframe, so just use the first keyframe
+				referenceValue = AnimationUtils.arraySlice( referenceTrack.values, 0, referenceTrack.valueSize );
```
FlorentMasson, line 275: Not that is breaks anything, but while debugging I had `referenceTrack.valueSize` being undefined. I think it should be `referenceTrack.getValueSize()` or `valueSize`

mrdoob: /ping @c-morten

c-morten: @mrdoob I actually fixed this along with some other issues in a new PR: https://github.com/mrdoob/three.js/pull/20021. This PR fixes the issue I reported here: https://github.com/mrdoob/three.js/issues/20025

tasticad:
> I found that what works best for imported rigs is to import just the mesh and freeze all the transforms off, then bring in the rigged version and copy the weights from that onto the clean version of the mesh.

Hi guys, there's another way to do it for Maya users: you can set the up axis to Z in FBX export options → Axis Conversion.

PR #18836, opened 2020-03-07
OrbitControls: Align dolly function names with behavior

sciecode:
Pretty sure the naming convention for `DollyIn` and `DollyOut` is inverted and, to counter-act this, the checks are also inverted. So the behavior is correct, but the functions are doing the exact opposite.  :anguished:

[misc_controls_orbit.html](https://rawcdn.githack.com/sciecode/three.js/c4c4795a013b39ed9b6499a3d5be95cd38f20c7f/examples/misc_controls_orbit.html)

### DIFF
--- a/examples/js/controls/OrbitControls.js
+++ b/examples/js/controls/OrbitControls.js
@@ -414,7 +414,7 @@ THREE.OrbitControls = function ( object, domElement ) {
 
 	}();
 
-	function dollyIn( dollyScale ) {
+	function dollyOut( dollyScale ) {
 
 		if ( scope.object.isPerspectiveCamera ) {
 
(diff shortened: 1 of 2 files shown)


sciecode:
Just expanding a bit on why I believe this is the case:

https://github.com/mrdoob/three.js/blob/3cfe5ed63f964614fd41cf4aec52e4bbb5ca4d12/examples/jsm/controls/OrbitControls.js#L515-L523

`getZoomScale()` is 0.95 by default and is always a number between [0,1].

https://github.com/mrdoob/three.js/blob/3cfe5ed63f964614fd41cf4aec52e4bbb5ca4d12/examples/jsm/controls/OrbitControls.js#L448-L453

https://github.com/mrdoob/three.js/blob/3cfe5ed63f964614fd41cf4aec52e4bbb5ca4d12/examples/jsm/controls/OrbitControls.js#L191

You can see, that `dollyOut` is always decreasing the radius from the camera to the target ( zooming-in ).

Mugen87 (approved):

sciecode:
Think we can merge this one.
Apparently, Mugen confirmed this is indeed the case.
WestLangley also confirmed this privately.

mrdoob:
Thanks!

PR #18846, opened 2020-03-09
Post-processing with WebXR

Neptilo:
I picked up takahirox's work to support post-processing in VR (issue  #15840) to make it up to date with the latest WebXR.


Mugen87:
Please remove all build files from your PR. Also undo the changes to `package-lock.json`. 

It seems #15840 can be closed?

Mugen87:
I guess on reason why #15840 was never merged is the performance aspect of post-processing with WebXR. Has somebody actually verified the performance of `webxr_vr_postprocessing` on a common VR device like Oculus Quest?

Right now, I feel making `EffectComposer` support WebXR is just more a technical showcase but not necessarily useful in practise. I doubt that e.g. `SSAOPass` or `SAOPass` will perform well in context of VR. And stuff like `BokehPass` makes probably no sense at all.

AdaRoseCannon:
At Samsung Post processing in XR is something we are actively working with and currently have to maintain our own versions of THREE to support it, it's more that a technical showcase for us. @wizgrav has been doing the most work with it and has been working with the Oculus quest.

wizgrav:
@Mugen87 here's a demo of the home brew post proc suite @AdaRoseCannon mentioned https://samsunginter.net/three-effects/examples/madrid/

Even though it's coalescing the blending parts of effects in a single ubershader, It is not running at full frame rate on the quest but it has several effects running and complex geometry as well. Check it out and judge for yourself. I agree that ssao doesn't make sense but basic effects like bloom are reasonable

Neptilo:
> Please remove all build files from your PR. Also undo the changes to `package-lock.json`.

Ok, done. Sorry, I assumed the build files had to be up-to-date with the source.

> It seems #15840 can be closed?

If you think my PR is on the right track, yes.

> I guess on reason why #15840 was never merged is the performance aspect of post-processing with WebXR. Has somebody actually verified the performance of webxr_vr_postprocessing on a common VR device like Oculus Quest?

We are also using it professionally, hence the motivation for my changes. :) Yes, Oculus Quest is the device we use the most. Some effects are simple enough to work well in VR, and it is often possible to adjust parameters / disable some passes, based on the device.

Mugen87:
> @Mugen87 here's a demo of the home brew post proc suite @AdaRoseCannon mentioned https://samsunginter.net/three-effects/examples/madrid/

Um, I've tried to test your link with Oculus Quest and Oculus Browser (8.2.1.x). However, when entering XR, most of the screen is black. I can only see the scene in a small part of the viewport. The official `three.js` examples work fine though. Any ideas what's going wrong?

wizgrav:
Hmm haven't tested it since october, probably something changed in the xr spec, I'll take a look this afternoon

Mugen87:
> I also have to point out that effect composer is inefficient and should not be used in production even for higher spec platforms than the quest and even for non vr scenarios as well. 

In this case, I see no reason for adding XR support for `EffectComposer`. It might make more sense to  think about a new solution for post-processing and XR.

> This is the old approach unity had, which it has now abandoned in favor of using an ubershader to blend the results of the various effects prepasses

That sounds interesting. How would something like this look like? I mean a basic setup with e.g. a simple grayscale pass.

wizgrav:
A basic grayscale pass by itself wouldn't benefit much, instead lets break down a common production setup that would include a bloom effect, a vignette effect, color correction and fxaa.

In a preprocess step the blurred texture for the bloom is created, meant to be added on top of the base scene rendering in another pass. With the current effect composer that would take a whole pass on it's own. Then another pass to apply the vignette and another one to do the color correction. But the bloom texture addition + vignette calculation + color correction can all be done in a single pass with an uber shader.

Unity has a single common shader for all it's included effects, and controls what gets done with ifdefs. Three could do the same for all the effects included in the base distribution. If the user would like to add a new effect he should modify the final shader with onBeforeCompile

Neptilo:
> I can only see the scene in a small part of the viewport.

This sounds like the exact kind of problem that would be resolved by my changes.
@wizgrav It should fix some effects. I'm curious how many effects it fixes out-of-the-box. However, I know that some passes have to be modified to be VR-compatible (e.g. those that call `renderer.render` again to generate several buffer inputs).

wizgrav:
@Neptilo the demo was working a couple of months ago and it's not using effectcomposer, it does everything in scene.onAfterRender. I suspect that something changed in the xr spec in between and I see that you have commited changes in the XRManager. Unfortunately I don't have the quest at hand right now so I'll check it out in the afternoon

vanruesc:
For an alternative ubershader approach I'd like to point to [postprocessing](https://github.com/vanruesc/postprocessing#performance) which implements structured effect merging and supports [custom effects](https://github.com/vanruesc/postprocessing/wiki/Custom-Effects).

wizgrav:
@vanruesc that's more or less what I have proposed with the only difference of tapping on Scene.onBeforeRender and Scene.onAfterRender to perform the operations instead of wrapping the renderer in a composer. But even in your approach you will have to tap on at least one of these callbacks because that's the only way to get access to the WebXR cameras which are needed for some effects like god rays

vanruesc:
> tapping on Scene.onBeforeRender and Scene.onAfterRender to perform the operations instead of wrapping the renderer in a composer

Could you elaborate on this or point me to the code where this is done? From my understanding, the `WebGLRenderer` will perform a normal render operation if you call the `render` method. How do you prevent that and render to texture instead without wrapping the render call?

wizgrav:
You attach a callback on Scene.onBeforeRender. Inside that you use renderer.setRenderTarget() to set your renderTarget as the output. Then the renderer will render on this instead of whatever it had previously.

Then in  Scene.onAfterRender you use your renderTarget as input for the post processing and finally output to screen or whatever renderTarget was set before renderer.render() started. You can find that with renderer.getRenderTarget(). The camera argument in both onBeforeRender and onAfterRender will be the camera you passed on renderer.render(), or the internally managed ArrayCamera when running in VR mode.

With this approach post processing becomes a plugin of sorts. The user doesn't have to do anything just attach the callbacks and everything will be taken care of automatically. You can see it in code here: https://github.com/SamsungInternet/three-effects/blob/master/src/fx.js

vanruesc:
Thanks for the detailed explanation! I totally slept on `setRenderTarget` and always thought you'd have to hack `WebGLRenderer.render` to accomplish that.

wizgrav:
@Mugen87 as I suspected it was an XR update that broke it. I selectively updated WebXRManager.setSession() with the one from this PR and the limited viewport got fixed. You can check the demo now https://samsunginter.net/three-effects/examples/madrid/

takahirox:
Sorry I don't have time to follow the discussion and changes in this PR but I wrote https://github.com/mrdoob/three.js/pull/15840#issuecomment-596755413

Mugen87:
> You can check the demo now https://samsunginter.net/three-effects/examples/madrid/

Okay, it works now! Yeah, your modified version of `three.js` seems to perform quite well. It would be actually great to have a FPS counter like [here](https://immersive-web.github.io/webxr-samples/inline-session.html) so its easier to see the performance impact when certain effects are enabled.

BTW: It seems some effects like godrays do not work? Maybe related to the WebXR code update?

examples/js/postprocessing/EffectComposer.js
```diff
@@ -60,6 +60,22 @@ THREE.EffectComposer = function ( renderer, renderTarget ) {
...
+	// for VR
+
+	var rendererSize = new THREE.Vector2();
+	var scope = this;
+
+	function onSessionStateChange() {
+
+		renderer.getDrawingBufferSize( rendererSize );
+		scope.setPixelRatio( renderer.getPixelRatio() );
+		scope.setSize( rendererSize.x, rendererSize.y );
```
WestLangley: I appears you are scaling by pixel ratio twice here.

wizgrav:
@Mugen87 no no no sorry about that. I haven't finished porting all the effects yet. Working are only color(which uses a lut that does nothing), bloom, glitch and filmgrain. Yeah the fps counter is a good suggestion, I'll do that. 

Just for reference the three.js used doesn't actually have anything more than vanilla three performance wise, the demo itself is quite optimized with basis textures and the post proc approach I wrote about. The bloom effect is also custom and probably more optimal than the one in the three examples but I have to benchmark that

examples/js/postprocessing/EffectComposer.js
Neptilo: The addition of `setPixelRatio` after `getDrawingBufferSize` was made in the previous PR (commit f0fdd62) and I didn't change it. But it seems you're right. I'll change `getDrawingBufferSize` to `getBufferSize`.

Mugen87:
To sum up: @wizgrav's approach outlined in https://github.com/mrdoob/three.js/pull/18846#issuecomment-596504066 seems to be the most promising to me right now. This approach does not aim to integrate the current `EffectComposer` into the renderer but head for single pass solution instead (using an ubershader to blend the results of the various effects).

wizgrav:
@Mugen87 thanks for that, still this PR fixes actual issues in WebXRManager, right now post processing in VR is broken and that's a regression actually, so at least some core bits from this PR should get merged somehow

vanruesc:
> I believe that post proc is a very complex piece to be added into core so just giving the developer entry points to implement it the way he wants is a saner approach imo.

I concur.

Btw: does anybody know why `Camera.onBeforeRender` and `Camera.onAfterRender` are never called in `WebGLRenderer`? I think it would make more sense to bind a post processor to specific cameras instead of scenes.

wizgrav:
@vanruesc valid question, I'll answer it since I did the Scene.onAfterRender PR.

The first thing I tried which seems the most intuitive was to put the entry points on the renderer itself. Unfortunately we're always dealing with a single renderer and if this internally does the post proc which uses the renderer that internally does the post proc which uses the renderer that internally does the post proc ... you can see where this is going, stack overflow. 

Then I thought to utilize the cameras which is also intuitive as you point out and is in fact the way unity does it. Unfortunately the cameras are used in an unorthodox way inside the renderer. If we're running in VR mode the cameras are internally replaced by a pair of cameras wrapped in an ArrayCamera. This ArrayCamera is managed internally by the renderer. So that approach which was valid turned out to be complicated by implementation details.

The scene was utilized as it was the only thing left that made sense. It also had the onBeforeRender and onAfterRender entry points already from Object3D, like Camera. It still has it's shortcomings like for instance when you want to reuse the same Scene(fi when doing depth prepass or generating a texture for reflections in onBeforeRender) but way less than using the renderer itself.

So the choice was entirely pragmatic. I can reconsider using the camera, even though that would bring complexity of its own due to the automatic xr management. And to be really honest I'm not even sure if handling WebXR automatically makes sense anymore. Minor spec changes cause regressions that need to be handled by the library itself as we can see in this PR, leaving the developer with little room to work around them sans forking three

vanruesc:
Thank you for the insight.

> I can reconsider using the camera, even though that would bring complexity of its own due to the automatic xr management. And to be really honest I'm not even sure if handling WebXR automatically makes sense anymore.

Would it be possible to implement WebXR as a plugin? Maybe the pain points could be removed first by revising the current implementation. Unfortunately I'm not familiar with the WebXR parts of three and don't know why the ArrayCamera needs to be managed by the renderer. A clean plugin foundation with plugin inter-compatibility in mind could reduce maintenance issues in the long run.

Easier said than done, I guess.

wizgrav:
WebXR basically provides a context to draw into and positions of tracked elements(Headset-Controllers) through polling. It makes sense to handle it externally from the renderer for other reasons as well. Unity for instance is polling for the tracked transforms twice per frame. Once in the beginning to have some valid data to do calculations with(like raycasting and such) but it also polls the camera transform a second time right before actually rendering the scene. It makes for a smoother experience as the viewpoint presented to the user will be much closer to the actual position/rotation of the headset

munrocket:
Sorry I am not read all comments, but looks like you added new example. You can run `npm run make-screenshot webxr_vr_postprocessing` on your machine ( if you not using MacOS ) and add screenshot for Travis with new commit.

fredcallagan:
Hi,

just checking if this is merged or will be merged at any point ?
Thanks a lot

Mugen87:
I personally don't think so. A more promising approach is summed up here: https://github.com/mrdoob/three.js/pull/18846#issuecomment-596999385

Neptilo:
What about [wizgrav's comment](https://github.com/mrdoob/three.js/pull/18846#issuecomment-597002694):
> Still this PR fixes issues in WebXRManager, right now post processing in VR is broken and that's a regression actually, so at least some core bits from this PR should get merged somehow

takahirox:
Combining passes into a few passes sounds good to me. So we need shader combine engine. Can we reuse Node based material system for it? (Sorry, I'm still not familiar with our node based material system yet). Or do we need to create a new one?

AdaRoseCannon:
Would it be worthwhile to look into a stop-gap solution of providing an efficient method for a few popular features like bloom, ssao and colour grading then looking for a more general solution further down the road?

LeviPesin:
> To sum up: @wizgrav's approach outlined in https://github.com/mrdoob/three.js/pull/18846#issuecomment-596504066 seems to be the most promising to me right now. This approach does not aim to integrate the current EffectComposer into the renderer but head for single pass solution instead (using an ubershader to blend the results of the various effects).

Is there any progress on that approach?

julapy:
as a user coming to use XR in threejs for the first time, my expectation has been that my code should work the same across XR and non-XR

the glitch effect (from examples) performs well on my android using EffectComposer and from what I’ve seen, performance on mobile in and out of XR is comparable.

imo, if code is not performant, it should be up to the user to decide if to use it or not or to find an alternative or choose which devices to support etc.

i imagine others like myself that are familiarising themselves but following the three.js examples and would expect for the EffectComposer to work.

donmccurdy:
> Combining passes into a few passes for the performance sounds good to me. So we need a shader combine engine...

It's worth noting that https://github.com/pmndrs/postprocessing/ has already done this. I've implemented a few custom post-processing effects with that library, and I really appreciate how @vanruesc has set things up, my effects got merged with others fine. I'm not sure how many effects are WebXR-compatible yet, but it can be done:

https://twitter.com/Cody_J_Bennett/status/1482585611781480448

I'm worried we may be doing users a disservice by exclusively promoting the internal post-processing implementation, while so much good development is happening in the community at this point. Maybe we could add 1-2 examples using the pmndrs/postprocessing library, with an "external" tag?

CodyJasonBennett:
I decided to complete the fix in https://github.com/mrdoob/three.js/pull/26160 to champion this issue. I've planned to do the same for [pmndrs/postprocessing](https://github.com/pmndrs/postprocessing), but only now just got to getting it to work since the introduction of WebXR layers now years ago.

My tweet Don mentioned gives me full control for eye-dependent effects which we can do more effectively since WebXRManager exposes XR internals. I'm personally very happy about that since I don't need to eject from three.js to curate advanced behavior.

image

danrossi:
I need this to add lens distortion to the WebXR stereo camera as a polyfill for Iphone. I tried some experiments but only one side renders. Using this project. https://github.com/ycw/three-lens-distortion

danrossi:
I need this integration for a lens distortion effect for webxr. For Iphone due to the webxr polyfill being completely broken with three. I ran tests with StereoEffect and only get one side rendering. 


Using this project

danrossi:
My last comment got flagged as spam for some reason. Confirming it's working. But the lens distortion project doesn't yet. I'll try the other processing project also. 
Screenshot 2023-07-09 053912

Mugen87:
Implemented via #26160.

Mugen87:
We have merged #26160 instead but had to revert for performance reasons, see https://github.com/mrdoob/three.js/pull/26160#issuecomment-1751012842.

PR #18854, opened 2020-03-10
Line: Added support for morph targets.

### LINKED ISSUE #2875: THREE.Line missing morphTargets
I took the [morphTargets](http://mrdoob.github.com/three.js/examples/webgl_morphtargets.html) example and tried to swap a `THREE.Mesh` with a `THREE.Line`. I got the below errors.

Screen Shot 2013-01-02 at 2 00 36 PM

An example of this can be found [here](http://dl.dropbox.com/u/1595444/experiments/Three.js/THREE-Line-morphTarget.html). I expected it to work because previously I had used `THREE.Line` and `THREE.Mesh` interchangeably. I went looking through the source and `THREE.Line` doesn't look like it has any of the initialization logic for morphTargets that `THREE.Mesh` has.

So I think it's missing and I'm requesting that it be added. Is it as simple as adding these properties to `THREE.Line`?

```
this.morphTargetBase = -1;
this.morphTargetForcedOrder = [];
this.morphTargetInfluences = [];
this.morphTargetDictionary = {};
```

Finally, I'm running Chrome stable on OS X 10.8.2

Mugen87:
Fixed #2875.

### DIFF
--- a/docs/api/en/materials/LineBasicMaterial.html
+++ b/docs/api/en/materials/LineBasicMaterial.html
@@ -88,6 +88,8 @@
 			property and it is ignored by the [page:WebGLRenderer WebGL] renderer.
 		</p>
 
+		<h3>[property:Boolean morphTargets]</h3>
+		<p>Define whether the material uses morphTargets. Default is false.</p>
 
 		<h2>Methods</h2>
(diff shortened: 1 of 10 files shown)


mrdoob:
Thanks!

PR #18863, opened 2020-03-11
Math: move to es6 classes

DefinitelyMaybe:
Hi all.
This pull relates to #11552 #17425

I've been working on sorting the maths scripts into ES6 classes. It's close but needs some help.
My local dev website runs all of the code just fine however:

- "geometry / terrain /raycast" example has had a noticeable performance hit
- "camera / array" example fails silently...
- various WebGLProgram warnings (i.e. fakepath() ... pow(f, e) ... negative f values)

I also tried to work my way through some of the unit tests but le brain died at that point.

Going to have a look at the other folders as there may be some easier candidates for ES6 transformation.

update:
[mrdoob's comment on let/const](https://github.com/mrdoob/three.js/pull/17276#issuecomment-527687857)
[Blocking dependencies](https://github.com/mrdoob/three.js/issues/6419#issuecomment-599160231)

### DIFF
--- a/examples/js/math/ColorConverter.js
+++ b/examples/js/math/ColorConverter.js
@@ -3,15 +3,17 @@
  * @author zz85 / http://github.com/zz85
  */
 
-THREE.ColorConverter = {
+import { MathUtils } from "../../../build/three.js";
+
+var ColorConverter = {
 
 	setHSV: function ( color, h, s, v ) {
 
 		// https://gist.github.com/xpansive/1337890#file-index-js
 
-		h = THREE.MathUtils.euclideanModulo( h, 1 );
-		s = THREE.MathUtils.clamp( s, 0, 1 );
-		v = THREE.MathUtils.clamp( v, 0, 1 );
+		h = MathUtils.euclideanModulo( h, 1 );
+		s = MathUtils.clamp( s, 0, 1 );
+		v = MathUtils.clamp( v, 0, 1 );
 
 		return color.setHSL( h, ( s * v ) / ( ( h = ( 2 - s ) * v ) < 1 ? h : ( 2 - h ) ), h * 0.5 );
 
@@ -84,5 +86,6 @@ THREE.ColorConverter = {
 
 	}
 
-
 };
+
+export { ColorConverter };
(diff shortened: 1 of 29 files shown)


Mugen87:
> Going to have a look at the other folders as there may be some easier candidates for ES6 transformation.

FYI: The class migration has been paused for now. So it might be better to wait until it's clear when this task moves on.

Mugen87:
BTW: 23 units test fail with your change (which is definitely an issue).

DefinitelyMaybe:
most definitely. I had a quick look at the tests and there were things like 'x not found on undefined'. not sure how I got there by changing the syntax. I made the PR in case someone wanted to look into before I get back to it. My thinking atm is that one of the other folders might be easier... if at first you don't succeed, right?

DefinitelyMaybe:
> FYI: The class migration has been paused for now. 

Hey, quick question, which discussion(s) are you referencing?

donmccurdy:
See https://github.com/mrdoob/three.js/issues/6419#issuecomment-599152135. We cannot migrate any classes that are extended by files in `examples/js`, yet, at least. I'm not sure if that means _all_ conversions are paused, or if there are other considerations here.

looeee:
@Mugen87 there's no reason to pause _all_ class conversions. Why don't we move ahead with classes that are not extended in the examples? The box geometries have been converted to classes for a few months now without issue.

Mugen87:
I'm not sure it's just about the examples directory. Any class could be dervied by user level code and I don't know how @mrdoob feels about this topic in context of backwards compatibility.

Sidenote: Introducing classes does not mean introducing other ES6 features like `let`/`const`. According to https://github.com/mrdoob/three.js/pull/17276#issuecomment-527687857, the migration should only focused on classes. #18912 does honor this right now. To make things more stable, I think it's more safe to do things step by step.

Mugen87:
Closing due to failing unit tests, merge conflicts and wrong example code (the PR modularized files in the `examples/js` directory).

DefinitelyMaybe:
good. your turn @Mugen87

PR #18864, opened 2020-03-11
Another round of lgtm fixes

makc:
So I went through [a list of unused stuff at lgtm](https://lgtm.com/projects/g/mrdoob/three.js/?mode=list&id=js%2Fsuperfluous-trailing-arguments%2Cjs%2Ftrivial-conditional%2Cjs%2Funneeded-defensive-code%2Cjs%2Funreachable-statement%2Cjs%2Funused-local-variable%2Cjs%2Fuseless-assignment-to-local) and patched the files until I ran out of steam. None of this is tested, use at your own risk!

CC #18845

### DIFF
--- a/examples/js/loaders/LDrawLoader.js
+++ b/examples/js/loaders/LDrawLoader.js
@@ -1174,8 +1174,6 @@ THREE.LDrawLoader = ( function () {
 
 									type = lp.getToken();
 
-									if ( ! parsingEmbeddedFiles ) {
-
 										currentParseScope.triangles = [];
 										currentParseScope.lineSegments = [];
 										currentParseScope.conditionalSegments = [];
@@ -1206,8 +1204,6 @@ THREE.LDrawLoader = ( function () {
 										triangles = currentParseScope.triangles;
 										lineSegments = currentParseScope.lineSegments;
 										conditionalSegments = currentParseScope.conditionalSegments;
-
-									}
 
 									break;
 
--- a/editor/js/LoaderUtils.js
+++ b/editor/js/LoaderUtils.js
@@ -54,7 +54,7 @@ var LoaderUtils = {
 
 					}
 
-					onEntryHandled( entry );
+					onEntryHandled();
 
 				} );
 
--- a/editor/js/libs/codemirror/codemirror.js
+++ b/editor/js/libs/codemirror/codemirror.js
@@ -1525,7 +1525,7 @@
 
         // Try to detect the user choosing select-all
         if (te.selectionStart != null) {
-          if (!ie || (ie && ie_version < 9)) prepareSelectAllHack();
+          if (!ie || (ie_version < 9)) prepareSelectAllHack();
           var i = 0, poll = function() {
             if (display.selForContextMenu == cm.doc.sel && te.selectionStart == 0 &&
                 te.selectionEnd > 0 && input.prevInput == "\u200b")
@@ -2222,7 +2222,7 @@
   // Verify that the selection does not partially select any atomic
   // marked ranges.
   function reCheckSelection(doc) {
-    setSelectionInner(doc, skipAtomicInSelection(doc, doc.sel, null, false), sel_dontScroll);
+    setSelectionInner(doc, skipAtomicInSelection(doc, doc.sel, null, false));
   }
 
   // Return a selection that does not partially select any atomic
@@ -4040,7 +4040,7 @@
       // actually possible. Otherwise, it causes vertical scroll
       // jitter on OSX trackpads when deltaX is small and deltaY
       // is large (issue #3579)
-      if (!dy || (dy && canScrollY))
+      if (!dy || canScrollY)
         e_preventDefault(e);
       display.wheelStartX = null; // Abort measurement, if in progress
       return;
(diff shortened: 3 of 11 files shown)


Mugen87:
> None of this is tested, use at your own risk!

Sorry, this is not the way how PRs are done! Just blindly changes source files is something that can do anybody. It's also important to verify if everything works after the changes.

I refuse to review and merge the PR with such an attitude.

makc:
> something that can do anybody

I dont see anybody doing it, though

> to verify if everything works after the changes

says "All checks have passed" right there 🙄

lol

mrdoob:
@Mugen87 I've looked at the changes and the fixes seem correct?

In general lgtm recomendations tend to be 99% right. I only remember it fails with binary parsing:
https://github.com/mrdoob/three.js/blob/6d403a5bab72f67741b58a43f7fabf1b3d625b07/examples/js/loaders/FBXLoader.js#L3297-L3298

Mugen87:
Well yes, at least the changes to the examples look good.

To be clear: I more upset about the way the OP thinks PRs work. I though we would no accept untested PRs...

mrdoob:
It's definitely risky. But he is not changing `src/` which is where things must really be tested.

Mugen87:
Well, it is a matter of principle to me. Other devs invest a considerable amount of time to test their contributions. But if someone opens a PR with the comment "None of this is tested, use at your own risk!", then I think something goes significantly wrong.

examples/webgl_loader_gcode.html
```diff
@@ -48,7 +48,7 @@
 				renderer.setSize( window.innerWidth, window.innerHeight );
 				container.appendChild( renderer.domElement );
 
-				var controls = new OrbitControls( camera, renderer.domElement );
+				new OrbitControls( camera, renderer.domElement );
```
sciecode: I'm not sure if we want to set this kind of example for the users.
I know LGTM complains about it, but in reality having access to the controls variable offers many possibilities of customization and having an empty constructor also feels wrong to me.

WestLangley: Agreed. Instead, maybe add `controls.min/maxDistance` -- or `controls.min/maxZoom` in the case of orthographic. Or `controls.enableZoom = false`.

That could be done in a separate PR.

mrdoob:
> But if someone opens a PR with the comment "None of this is tested, use at your own risk!", then I think something goes significantly wrong.

Yes, @makc doesn't tend to have the best wording. But he has good intentions and he's trying to help. Maybe he could make things easier for us, but he's helping nonetheless.

examples/webgl_loader_gcode.html
mrdoob: I agree that it's weird. But I think we have used this pattern before?

makc:
It is fine by me if you reject - I do not mind. After all, you did not ask me to do it. But I mean, this is free stuff I am sending your way, and you are basically saying I did not send you enough, I should be spending more time for you to be happy :) And then you reject and - what? End up doing the same commit yourself? Feel free, but like I said, you havent done so for years so there must be a reason for that, and that reason says you probably will not do this.

> makc doesn't tend to have the best wording

yes, but the wording is not part of the commit, you do not have to accept the wording here. The wording will sink to the bottom of the closed issue tickets valley either way.

> Instead, maybe add controls.min/maxDistance -- or controls.min/maxZoom in the case of orthographic. Or controls.enableZoom = false.

+1 for @WestLangley idea

examples/webgl_loader_gcode.html
sciecode: I think I’m more worried about what new users might infer from that, other then the fact that it is completely safe to do so. 

I like WestLangley idea as well.

WestLangley: We have certainly used this pattern:
```js
new FooLoader()
	.load( url, ... )
```

I do not have a problem using the same pattern with `OrbitControls`, but I do have a problem allowing the user to zoom without limitation.

makc: > could be done in a separate PR

if you want I can remove controls commit from PR, I mean it is separate since you know... premonition.

makc: there - gone.

mrdoob:
Thanks!

examples/js/loaders/LDrawLoader.js
```diff
@@ -1174,8 +1174,6 @@ THREE.LDrawLoader = ( function () {
 
 									type = lp.getToken();
 
-									if ( ! parsingEmbeddedFiles ) {
```
mrdoob: @makc I'm not sure about this one, I think `parsingEmbeddedFiles` could be `true`... 🤔

https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/LDrawLoader.js#L1254

makc: @mrdoob there is if ( parsingEmbeddedFiles ) { ...continue; } block above, so no.

makc:
@WestLangley actually, I think Im -1 with the controls issue again. See, if we have to replace
```
var controls = new OrbitControls( camera, renderer.domElement );
```
with, for example,
```
var controls = new OrbitControls( camera, renderer.domElement );
controls.minDistance = 0.3 * camera.position.length();
controls.maxDistance = 3.0 * camera.position.length();
```
in 100 examples, to stfu the linter, it raises the question of why cant the OrbitControl themselves set these default values in constructor instead of 0/Infinity. Same with
```
var controls = new OrbitControls( camera, renderer.domElement );
controls.enableZoom = false;
```
it only makes one wonder why enableZoom is not false by default.

WestLangley:
To determine reasonable constraints on camera movement, one must take into consideration the position and scale of scene elements. That is why the constraints are set at the application level -- not inside the control constructor.

makc:
@WestLangley and how does setting them in the constructor changes the possibility of setting them at the application level, exactly? They would be still exposed and available.

examples/js/loaders/LDrawLoader.js
mrdoob: Oh I see!

PR #18865, opened 2020-03-11
Skeleton: Add dispose().

### LINKED ISSUE #18844: skeleton boneTexture won't dispose?
when support floatVertexTextures and enable skinning ,it will create DataTexture in the webglrender,but i dont find any place to release it

cjsjy123:
Fixed #18844.


src/objects/Skeleton.d.ts
```diff
@@ -21,4 +21,6 @@ export class Skeleton {
 	clone(): Skeleton;
 	getBoneByName( name: string ): undefined | Bone;
 
+	dispose();
```
Mugen87: This is better:
```ts
dispose(): void;
```

src/objects/Skeleton.js
```diff
@@ -168,6 +168,26 @@ Object.assign( Skeleton.prototype, {
...
+
+			this.boneMatrices = undefined;
+
+			this.boneTextureSize = undefined;
+
+		}
+
+		// developer need to clean skeleton reference
+
+		this.disposed = true;
```
Mugen87: It would be good to have a solution with such flags.

cjsjy123: Inherited from EventDispatcher, And add an assignto function to trigger a bind/unbind callback, then record it in the refmap, finally automatically clear the reference when dispose?

src/objects/Skeleton.d.ts
cjsjy123: ok,i will update it

src/objects/Skeleton.js
cjsjy123: But should i reduce the dependency on the es6 objects (Set,Map) when creating the refmap ?

Mugen87: The dispose method should look like so:
```js
dispose: function () {

    if ( this.boneTexture ) {

        this.boneTexture.dispose();
        this.boneTexture = undefined;

    }

}
```
It's not necessary to reset `boneMatrices` and `boneTextureSize`.  Besides, calling `dispose()` should not throw a runtime error at any state of the skeleton object. Besides, there renderer must be able to work with a skeleton after `dispose()` was called. In this case, it will recreate the bone texture like when encountering the skeleton for the first time.

cjsjy123: in fact, the destroyed object should not continue to be rendered. The flag is added to block its rendering, otherwise I need to continue to execute the message for the datatexture that he automatically created later. In fact, if the reference can be maintained automatically By leaving the object's skeleton empty, you don't need to make any changes to webglrender. And now it's been auto-cleaned and ready to commit it

cjsjy123: Now my webglrenderer changes has been rolled back.

Mugen87: > in fact, the destroyed object should not continue to be rendered.

Sorry, but this is not the logic of how `dispose()` works in the engine. `Sekelton.dispose()` has to work exactly like the existing methods.

cjsjy123: Got it

Mugen87 (approved):

mrdoob:
Thanks!

PR #18877, opened 2020-03-12
New controls IndoorControls

wobushijiujiu:
Hello, I'm a new programmer and this is my first time using GitHub to provide source code for an open source project.
I added an indoor controller that can be rotated from a first-person perspective, moved horizontally by clicking on the ground or using the direction keys.
Thanks for reference code pointerLockControls.js and reference web site 'https://showroom.littleworkshop.fr'. The code still has a lot of work to do.
I would be honored if I could be a contributor to three.js, and thank you very much for checking my code!


Mugen87:
https://raw.githack.com/wobushijiujiu/three.js/dev/examples/misc_controls_indoor.html

Um not sure but the example does no load for me. I only see the skybox.

arpu:
Bildschirmfoto von 2020-03-12 13-44-06

looks like you need to wait some more time to load all model images

wobushijiujiu:
> Um not sure but the example does no load for me. I only see the skybox.

I will try to use simple geometry in the example instead of the GLTF model to speed up the loading.
But what is the cause of this error?Thank you for your help.
微信图片_20200313103555
https://travis-ci.org/github/mrdoob/three.js/jobs/661469893?utm_medium=notification&utm_source=github_status

Mugen87:
You add a new example which is currently problematic for the automatic regression tests. This can be fixed later, too.

Before I check the PR in more detail, please try to find a model under 10 MB. 66 MB is definitely to big.

munrocket:
Looks interesting! I wonder if travis can handle so huge scene. Can you make new screenshot for your example with `npm run make-screenshot misc_controls_indoor`, add new commit and check what Travis say?

munrocket:
Also if travis fails again we can change [this variable](https://github.com/mrdoob/three.js/blob/aeb0fcf8f58586fbca2a41dacb10e29afc659e07/test/e2e/puppeteer.js#L37) in current test suite.

wobushijiujiu:
> Looks interesting! I wonder if travis can handle so huge scene. Can you make new screenshot for your example with `npm run make-screenshot misc_controls_indoor`, add new commit and check what Travis say?

npm ERR! missing script: make-screenshot，I wonder if there is something wrong with my execution path?

munrocket:
Looks like you had outdated branch when you make your changes or try `npm i`.

munrocket:
@wobushijiujiu sorry, you can add this example in exception list, I will add screenshot later by myself. >.<

munrocket:
Less pull request is better or what? <.<
Everybody asked about Automatic Regression Rests, so we have a little troubles now.
But I hope that I will fix almost all issues with this tests later.

wobushijiujiu:
Don't be nervous. I just started using github. I didn't know that changing my branch would get a new commit. I will change my code and redo the pull request next week.

wobushijiujiu:
If you have time, you can tell me the detailed solution to this problem about add screenshot. My English is not very good, so some places can be screenshot, my email is wobushi.jiujiu@gmail.com，welcome to email me

WestLangley:
@wobushijiujiu I would not be concerned about the screenshot for now.

Just create a simpler scene -- for example, you can modify the scene in `misc_controls_pointerlock.html`.

In your new controller, avoid calling `new` and `clone()` in methods that are called every frame.

munrocket:
You can add here a line for now, and tests will not block you PR anymore https://github.com/mrdoob/three.js/blob/aeb0fcf8f58586fbca2a41dacb10e29afc659e07/test/e2e/puppeteer.js#L19
I guess I need to increase `maxPageSizeTax` to 10, in order to cover your example.

mrdoob:
I think we should try not to increase the repo that much.
70MB for a new example seems too much to me 🤔

wobushijiujiu:
You can see new models at pull #18908

PR #18879, opened 2020-03-12
Only compute direction once in sky shader and disable depthWrite

Oletus:
depthTest is kept enabled since it may be desirable to render the sky after other geometry is already rendered.

### DIFF
--- a/examples/js/objects/Sky.js
+++ b/examples/js/objects/Sky.js
@@ -22,7 +22,8 @@ THREE.Sky = function () {
 		fragmentShader: shader.fragmentShader,
 		vertexShader: shader.vertexShader,
 		uniforms: THREE.UniformsUtils.clone( shader.uniforms ),
-		side: THREE.BackSide
+		side: THREE.BackSide,
+		depthWrite: false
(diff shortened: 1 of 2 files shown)


Mugen87:
https://raw.githack.com/higharc/three.js/sky-shader-cleanup2/examples/webgl_shaders_sky.html

Mugen87 (approved):

mrdoob:
Thanks!

makc:
@Mugen87 @Oletus starting at luminance = 1.1897 and above you get this at https://raw.githack.com/higharc/three.js/sky-shader-cleanup2/examples/webgl_shaders_sky.html:



Is it supposed to do that?

Mugen87:
@makc I see the same behavior on prod so I guess it's the intended behavior.

Oletus:
@mrdoob Thanks for merging!

@makc That's a pre-existing issue not related to this patch specifically. Probably numerical precision runs out somewhere, so I doubt it's exactly intended. But not sure how much need there is to go in and try to analyze it.

PR #18894, opened 2020-03-13
Water2: Fix encoding issue.

### LINKED ISSUE #18893: Horrible render performance when using Water + MeshStandardMaterial + sRGBEncoding

I'm experiencing horrible render performance when using the combination of the Water.js shader, a plane with a MeshStandardMaterial and the renderer.output using sRGBEncoding.

Recreated [water example](https://threejs.org/examples/#webgl_water) in jsfiddle and added `renderer.outputEncoding = THREE.sRGBEncoding` (line 119) in the fiddle: https://jsfiddle.net/4esfwtkh/

The performance does not dip:
- By not setting render.output to sRGBEncoding and using MeshStandardMaterial it will run ok (comment line 117)
- By setting render.output to sRGBEncoding but using MeshBasicMaterial it will run ok (comment line 55 and uncomment line 56)
- When the water is not visible in the camera frustum (i.e. by looking from below)

##### Three.js version

- [x] Dev
- [x] r114
- [x] r113

Mugen87:
Fixed #18893.

### DIFF
--- a/examples/js/objects/Water2.js
+++ b/examples/js/objects/Water2.js
@@ -26,6 +26,7 @@ THREE.Water = function ( geometry, options ) {
 	var reflectivity = options.reflectivity || 0.02;
 	var scale = options.scale || 1;
 	var shader = options.shader || THREE.Water.WaterShader;
+	var encoding = options.encoding !== undefined ? options.encoding : THREE.LinearEncoding;
 
 	var textureLoader = new THREE.TextureLoader();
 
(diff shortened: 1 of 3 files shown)


WestLangley:
This does not actually "fix" #18893. It is a work-around.

For related discussions, see:

https://github.com/mrdoob/three.js/pull/18709#issuecomment-591701987

https://github.com/mrdoob/three.js/issues/18699#issuecomment-590089847

Mugen87:
> This does not actually "fix" #18893. It is a work-around.

I've already mentioned this here: https://github.com/mrdoob/three.js/issues/18893#issuecomment-598672140

PR #18908, opened 2020-03-16
New controls IndoorControls

wobushijiujiu:
An indoor controller that can be rotated from a first-person perspective, moved horizontally by clicking on the ground or using the direction keys.

Fixed some issues in the IndoorControls that might affect performance.

The model used in the example is taken from the website https://sketchfab.com. I downloaded it for free and simplified the images. So this model is only for testing.
model asset: 
      "author": "mozillareality (https://sketchfab.com/mozillareality)",
      "license": "CC-BY-4.0 (http://creativecommons.org/licenses/by/4.0/)",
      "source": "https://sketchfab.com/models/b69e2a5dd7de4d7a8e091768d54a63e6",
      "title": "Conference Room1"


marcofugaro:
Live version:
https://raw.githack.com/wobushijiujiu/three.js/dev/examples/misc_controls_indoor.html

The model looks really nice!

I personally would show the mouseHelper only on the ground, like the [original example](https://showroom.littleworkshop.fr) you posted. You can do that by checking that the `intersect.face.normal` is pointing up, and setting `mouseHelper.visible = false` when it's now, what do you think?

wobushijiujiu:
Yes, that's a good idea. My original idea was that the user could assign a value to `indoorcontrols.ground` for the model of the floor, rather than checking the whole room as in my example.
Now, of course, models can't do that. Therefore, it is perfectly fine to add these checks in the example to improve the effect, which will not affect the indoorcontrols. js, thank you!

daiping2008:
谢谢




------------------&nbsp;原始邮件&nbsp;------------------
发件人:&nbsp;"LinZH"<notifications@github.com&gt;;
发送时间:&nbsp;2020年3月25日(星期三) 中午11:38
收件人:&nbsp;"mrdoob/three.js"<three.js@noreply.github.com&gt;;
抄送:&nbsp;"Subscribed"<subscribed@noreply.github.com&gt;;
主题:&nbsp;Re: [mrdoob/three.js] New controls IndoorControls (#18908)





 
Reopened #18908.

Mugen87:
Since the controls are quite application specific, I vote to promote `IndoorControls` in the forum rather than adding it to the library.

WestLangley:
>Since the controls are quite application specific, I vote to promote IndoorControls in the forum rather than adding it to the library.

I think the controls are useful, if properly-implemented.

For example, see https://my.matterport.com/show/?m=B2iwWtDwCjd.

Also, Google Street View uses a similar API.

This PR, temporary live link:
https://raw.githack.com/wobushijiujiu/three.js/dev/examples/misc_controls_indoor.html

DefinitelyMaybe:
Looks pretty cool! 👍 

Whichever way is decided, it may also be an idea to make a new PR once finished making changes. There are a lot of commits here.
An extra thought of mine would be to have a quick look around at the other models already within three.js to check if some might be usable for the example as oppose to adding more.

munrocket:
Agree with @DefinitelyMaybe

Mugen87:
Closing in favor of #19199.

PR #18912, opened 2020-03-16
A few ES6 class conversions

DefinitelyMaybe:
Unsure as to whether the following would be the preference of converting
`ArrayCamera.prototype = Object.assign( Object.create( PerspectiveCamera.prototype ), {
...
	isArrayCamera: true`
to
`constructor() { ... 
Object.defineProperty( this, 'isArrayCamera', {
			value: true
		} );
}`

a simple `this.isArrayCamera = true` does the above simply enough but should we worry about someone then calling `___.isArrayCamera = false`?

### DIFF
--- a/src/cameras/ArrayCamera.js
+++ b/src/cameras/ArrayCamera.js
@@ -4,21 +4,21 @@
 
 import { PerspectiveCamera } from './PerspectiveCamera.js';
 
-function ArrayCamera( array ) {
+class ArrayCamera extends PerspectiveCamera {
 
-	PerspectiveCamera.call( this );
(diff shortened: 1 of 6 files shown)


Mugen87:
https://github.com/mrdoob/three.js/pull/18863#issuecomment-597517929

DefinitelyMaybe:
Ah! thank you again. This PR had changed a couple of var's to let's. I've undone that change. Progress might be slow but yes, one thing at a time is best.

DefinitelyMaybe:
oops. missed up the merge commit. Will close request and make a new one.

PR #18913, opened 2020-03-17
GLTFLoader: Implement EXT_mesh_gpu_instancing.

donmccurdy:
A proposed optimization extension for glTF, allowing exporters to prepare data for instancing offline. See https://github.com/KhronosGroup/glTF/pull/1691.


examples/js/loaders/GLTFLoader.js
```diff
@@ -1323,6 +1432,19 @@ THREE.GLTFLoader = ( function () {
 
 	}
 
+	function quaternionFromBufferAttribute( quaternion, attribute, index ) {
+
+		return quaternion.set(
```
Mugen87: We might consider to add `Quaternion.fromBufferAttribute()`.

zeux:
Would it be possible to support signed normalized 16-bit quaternions as an input? The extension specification currently doesn't specify the formats but I'm hoping that we're going to go for the same format that animations use for consistency.

zeux:
I have an implementation for this extension mostly ready in gltfpack. However I'm running into an issue where three.js doesn't seem to support interleaved buffers as part of instancing extension. Here's a simple scene, the non-noq variant uses quantized geometry and as a result has to create an interleaved buffer attribute, which means the file doesn't load.

(I also have other rendering issues even with noq variant on a large test scene but I'm not sure yet if this is a bug in gltfpack or the loader)

[instancing.zip](https://github.com/mrdoob/three.js/files/4613987/instancing.zip)

zeux:
I've worked around the interleaved attribute issues like this for now: https://github.com/zeux/meshoptimizer/pull/142/commits/4bd7a7d90ee8b8f8b82028e3c4cddf6383eb6a29 - there's no need to clone the geometry if there are no custom attributes. It's not the best fix but it's probably good to do this regardless.

zeux:
One other issue is that it doesn't seem sufficient to disable frustum culling for the InstancedMesh (which is done in its constructor); on the attached level I had to disable frustum culling in Object3D constructor for the chairs inside the stadium to render. I'm wondering if frustum culling needs to be disabled on the node hierarchy as well since the bounding box can't be computed properly?

[stadium.glb.zip](https://github.com/mrdoob/three.js/files/4619662/stadium.glb.zip)

mrdoob:
> I'm wondering if frustum culling needs to be disabled on the node hierarchy as well since the bounding box can't be computed properly?

Hmm, that sounds like a bug indeed.

donmccurdy:
> Would it be possible to support signed normalized 16-bit quaternions as an input?

No objection!

> ... there's no need to clone the geometry if there are no custom attributes.

I'm not sure if that's true, I thought I remembered seeing errors when reusing the same geometry for both instanced and non-instanced meshes... in any case, I think we should try to improve BufferGeometry to be capable of cloning itself even when it has interleaved attributes.

zeux:
> I thought I remembered seeing errors when reusing the same geometry for both instanced and non-instanced meshes...

I was under the impression that the cloning is done because the cloned geometry is modified in presence of custom attributes. Maybe that's not true. Regardless I agree cloning should work, but it seems that it's not always necessary anyway.

zlf2002171:
unsubscribe

2020-05-18 

zlf2002171 



发件人：Arseny Kapoulkine <notifications@github.com>
发送时间：2020-05-18 11:25
主题：Re: [mrdoob/three.js] GLTFLoader: Implement EXT_mesh_instancing. (#18913)

I thought I remembered seeing errors when reusing the same geometry for both instanced and non-instanced meshes...
I was under the impression that the cloning is done because the cloned geometry is modified in presence of custom attributes. Maybe that's not true. Regardless I agree cloning should work, but it seems that it's not always necessary anyway.

lokiiarora:
@zeux just curious when is this landing on gltfpack?
Looking forward to it and also this PR landing on three, are there any blockers?

zeux:
@lokiiarora gltfpack supports this in master via `-mi` option. OTOH this PR needs:
a) fix cloning issues with interleaved buffers
b) add support for quantized rotation storage (should be a few lines of code)
c) resolve the culling issues somehow where disabling frustum culling on geometry that is done in the loader is insufficient (unsure what the right fix here is, might touch core)

donmccurdy:
>  c) resolve the culling issues somehow where disabling frustum culling on geometry that is done in the loader is insufficient 

You mean that the bounding box on the geometry won't describe the whole transformed instance group, right? I think that's OK for now. The `InstancedMesh` class has `frustumCulled` disabled by default. I hope we can work toward a way to turn that back on, but it's out of the scope of GLTFLoader to solve.

zeux:
@donmccurdy GLTFLoader setting `frustumCulled` to false was not sufficient in my tests. Initially I assumed that the issue is some sort of hierarchical visibility check in the core scene traversal, but I debugged this just now and the problem is simpler:

```
				// Copy mesh properties first, then node properties. These may be the same object.
				THREE.Object3D.prototype.copy.call( instancedMesh, mesh );
				THREE.Object3D.prototype.copy.call( instancedMesh, node );
```

resets `frustumCulled` back to `true` since the source mesh has it set. Adding

```
				// Reset frustumCulled back to false, since it was set to true by .copy above
				instancedMesh.frustumCulled = false;
```

fixes the problem.

donmccurdy:
Hm that's fun, thanks for tracking it down! 

These all seem like relatively quick fixes, yes.

donmccurdy:
I guess one larger change is that I should try to implement this on top of the new extension plugin system, though...

donmccurdy:
Remaining:

- [ ] Refactor to use GLTFLoader's extension API
- [ ] Test with interleaved and quantized cases
- [ ] Improve or remove example

donmccurdy:
Closing; see https://twitter.com/superhoge/status/1376655750294999043. /cc @takahirox

zeux:
It would still be great to see first-class support for this without external plugins.

donmccurdy:
Agreed; I think @takahirox's implementation is farther along than this branch, so that'd be a better PR at this point. :)

takahirox:
https://twitter.com/superhoge/status/1376665123360698373

> As I wrote in the readme, I still have some works for the GPU plugin. If I resolve them and it looks great match to Three.js core API I would think of moving it into the official glTF loader as built-in plugin. Until then I would keep it in my repo.

PR #18916, opened 2020-03-17
Docs: Improve Uniform page.

### LINKED ISSUE #10759: Better Documentation for Structured Uniforms

By structured uniform, I mean a uniform of the form:

    struct parameters {
        float scale, height, width, offset, baseline;
    };
    uniform parameters transition_model;

Currently, the [uniform documentation](https://threejs.org/docs/?q=unifor#Reference/Core/Uniform) only states that

> Uniform structures and arrays are also supported.

but doesn't go into further detail on how to set such uniforms using three.js. I had to figure out the proper syntax by stepping through the source code. For future reference, the syntax is:

    material.uniforms = {
        transition_model: { value: {
            scale: pScale,
            height: pHeight,
            width: pWidth,
            offset: pOffset,
            baseline: pBaseline
        }
    }

[This old issue thread](https://github.com/mrdoob/three.js/issues/2720) is the only reference I could find on structured uniforms in three.js. Moreover, said issue doesn't conclude on a single syntax (nor is the correct syntax even mentioned).

It would be nice to have this topic covered in the official documentation. Just like there already is a nice example of regular (non-structured) uniform usage.

##### Three.js version

- [ ] Dev
- [x] r84
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [x] All of them
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

Mugen87:
Fixed #10759.

### DIFF
--- a/docs/api/en/core/Uniform.html
+++ b/docs/api/en/core/Uniform.html
@@ -20,8 +20,8 @@
 		<code>
 		uniforms: {
 			time: { value: 1.0 },
-			resolution: new THREE.Uniform(new THREE.Vector2())
-		}
+			resolution: new Uniform( new Vector2() )
+		};
 		</code>
 
 		<h2>Uniform types</h2>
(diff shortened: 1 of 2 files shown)


mrdoob:
Thanks!

PR #18927, opened 2020-03-19
Clean up comments.

### LINKED ISSUE #18925: //!\ DECLARE ALIAS AFTER assign prototype ! are put in our .LICENSE files
When using Terser, which is practically modernized UglifyJs, to minimize Javascript and remove all comments in the go with a Webpack build, only the license-comments inside the code will be kept, but put in a seperate .LICENSE-file. 
Comments after `//` and between `/* */` will be removed to never show up in the distributables nor the .LICENSE files. But comments starting with `/*!` do (and need to) to allow us to insert license-information with these comments in our code.

This is working with and respected by all libs and frameworks I use and know of, but not with Threejs. Because Threejs has some 'license styled'-comments inside the code. Just regular comments, but not starting with `//` or use `/* */`, but use `//!\` instead, so start with `//!` and will therefore rightfully be seen as license-comments and therefore put inside our .LICENSE files by the minifier. 

Now we see comments like `DECLARE ALIAS AFTER assign prototype !` in our license files, and that's obviously now we want.

Comments found in the threejs lib like this in threejs r114:
three/build/three.js: lines 33920 and 43349,
three/build/three.module.js: lines 33912 and 43341,
three/src/animation/PropertyBinding.js: line 705,
three/src/math/Interpolant.js: line 247:
`//!\ DECLARE ALIAS AFTER assign prototype !`

Could you guys please change these all comments into normal js comments by replacing the `//!\` with `//` or using `/* */` so we don't get `DECLARE ALIAS AFTER assign prototype !` all over the place in our .LICENSE files. That should be a small effort to you guys, but would help us a lot, as we obviously don't like to have these comments in our .LICENSE files.

Thanks a lot in advance!

Mugen87:
Fixed #18925.

### DIFF
--- a/src/animation/PropertyBinding.js
+++ b/src/animation/PropertyBinding.js
@@ -702,7 +702,7 @@ Object.assign( PropertyBinding.prototype, { // prototype, continued
 
 } );
 
-//!\ DECLARE ALIAS AFTER assign prototype !
+// DECLARE ALIAS AFTER assign prototype
 Object.assign( PropertyBinding.prototype, {
 
 	// initial state of these methods that calls 'bind'
--- a/src/math/Interpolant.js
+++ b/src/math/Interpolant.js
@@ -244,7 +244,7 @@ Object.assign( Interpolant.prototype, {
 
 } );
 
-//!\ DECLARE ALIAS AFTER assign prototype !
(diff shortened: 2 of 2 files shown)


mrdoob:
Thanks!

PR #18928, opened 2020-03-19
Tests: travis -> github CI

munrocket:
Sometimes Travis fails without any reason #18843 #18787.
This probably happens, because Travis increase calculation power when you look at it.

This is an attempt to switch CI. Maybe after migration tests will be more robust.
Also in with Github Actions we can ignore `docs` path, but config file is bigger.

Owner of repository can turn on Github Actions somehow, because it works for me just with one commit. Probably here:

After this I can play Github Actions, also better to off notifications, because it is noisy without tuning.

https://github.com/settings/notifications

Still existing problems with tests:
- Mugen suggestion about new screenshot
- Attempt with beginFrame API can be successful


looeee:
Is there anything from Travis that we can't do using Github actions?

munrocket:
With Github Actions we can't make small configs (semantics is different and new) and it is not obvious how to make `allow_fails` feature.

The problem that looks like Travis under pressure gives another results and sometime fails.
Also results is differ in [another repo](https://github.com/munrocket/puppeteer-webgl-ci), which is basically a fork.

I am completely ok if this PR will be refused, but we will never understand how Github CI is good. Maybe this CI will work equal in another repos, god bless that it is easy to turn on and try it right here.

mrdoob:
Lets try this in the next dev cycle 👌

mrdoob:
@munrocket For now I've enabled local actions.

Screen Shot 2020-04-03 at 4 36 47 PM

mrdoob:
Lets see how this goes...

mrdoob:
Thanks!

mrdoob:
Screen Shot 2020-04-03 at 5 28 07 PM

Hmm, okay... I'll try allowing third-party.

munrocket:
🤞

mrdoob:
Progress!
https://github.com/mrdoob/three.js/actions/runs/69709625

Mugen87:
Um, it seems it's not possible to force-push anymore:

image

munrocket:
Looking like regress, all is failed.

Mugen87:
I can't even push to a new branch 😞 . It's not making fun anymore....

munrocket:
I am expected to play with this commit before merge. This regression tests not fun at all.
But for some reason Travis status is broken too. We can revert and check Travis manually.

Mugen87:
I do not modify `.github/workflows/ci.yml` so it makes no sense why the error message pops up.

mrdoob:
Reverted.

munrocket:
@mrdoob somehow you fix Travis.
It shows status now, problem solved.

munrocket:
Button still exists. If we don't need it we can off Actions.

mrdoob:
So we should cancel the transition?

munrocket:
Every new build (#19024 #19000) with failed screenshots. But we not touched screenshots 2 month and not touched sources in tests 1 month.

munrocket:
How many bugs we catched with this tests? I know only one. Is it worth our effort?

Mugen87:
> Every new build (#19024 #19000) with failed screenshots. But we not touched screenshots 2 month and not touched sources in tests 1 month.

Could it be that one of the related dev dependencies has a new version which leads a slightly different screenshot output? I've seen that some dependencies use a fix version like e.g. `pixelmatch`. `puppeteer` however uses `^2.1.1`. That means you will get newer minor releases. Could something like this affect the regression tests?

munrocket:
Still errors with fixed versions #19026

Mugen87:
Well, it was worth a try 👍

munrocket:
Going to create an issue in Travis forum.

mrdoob:
> Still errors with fixed versions #19026

Doesn't the PR need to be merged first?

Mugen87:
AFAIC Travis should always pick the `package.json` and `travis.yml` from the PR's branch.

munrocket:
>Doesn't the PR need to be merged first?

We have `package-lock.json` file and node with latest version usually just works.

Seems that we changed something in core because two CI failing on the same screenshots.
We can see a progress in #19028, +12MB in github, but problem localized.

mrdoob:
Do you want to try this again?

munrocket:
@mrdoob yep, it's possible.
I will make it.

munrocket:
I made new PR #19118 
This migration is a thing.

PR #18943, opened 2020-03-21
Implement support for BPTC (BC7) in BasisU

zeux:
This change allows transcoding new UASTC .basis files to BPTC which
results in much higher quality compared to UASTC -> BC3 transcoding.

The change needs to introduce BPTC support to three.js core for this to
work; the format is introduced in EXT_texture_compression_bptc which is
available in Chrome Canary on Windows (and possibly elsewhere?)

The update to Basis transcoder just pulls it directly from Binomial
repository; the transcoder is 20% larger now, unfortunately, because
it needs to support UASTC.

Note that we use mode 5 unconditionally to make sure RGB and RGBA
textures work.

BPTC (BC7) and BC3 are the same size in memory so this shouldn't
negatively impact VRAM consumption.


zeux:
ETC1S => BC3:

image

ETC1S => BC7:

image

UASTC => BC3:

image

UASTC => BC7:

image

mrdoob:
Thanks!

PR #18945, opened 2020-03-22
GLTFLoader: Use self instead of window

thomas-oxbotica:
Replacing `window` with `self` has no negative effects and GLTFLoader then works in web workers.

`window.URL` --> `self.URL`

[URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) is available in Web Workers.

Window API: [self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self)

(to use` GLTFLoader` in worker, also using `manager.addHandler` to add an edited `TextureLoader` that uses `ImageBitmapLoader` instead of `ImageLoader`)


donmccurdy (approved):

mrdoob:
Thanks!

PR #18947, opened 2020-03-23
Add quaternion.fromBufferAttribute

donmccurdy:
From https://github.com/mrdoob/three.js/pull/18913#discussion_r395012628.


mrdoob:
Thanks!

PR #18958, opened 2020-03-23
ColladaExporter: Honor normal maps.

### LINKED ISSUE #18950: ColladaExporter ignores normal maps
When exporting mesh that uses normal map with ColladaExporter, only diffuse texture is exported, normal is left out.

Mugen87:
Fixed #18950.


mrdoob:
Thanks!

PR #18964, opened 2020-03-24
Helpers: move to class syntax

DefinitelyMaybe:
Relates to: #18863

Helpers that were left out of the update
SkeletonHelper.js
ArrowHelper.js
because they are both used within the examples/js folder


Mugen87:
Closing see https://github.com/mrdoob/three.js/pull/17425#issuecomment-636168218.

PR #18970, opened 2020-03-25
Allow TrackballControls to work in a static scene

chabb:
The `staticMoving` flag suggests that you can use the trackball control without an animation loop, which is not the case. This change allows people to use trackball control without an animation loop, by calling `update` directly after a mouseMove event.

I'm aware that it will cause two calls to update if you set `staticMoving` to true, and use an animation loop that updates the control. Let me know if there is a better way to do it

### DIFF
--- a/docs/examples/en/controls/TrackballControls.html
+++ b/docs/examples/en/controls/TrackballControls.html
@@ -16,10 +16,12 @@
 			<p>
 				[name] is similar to [page:OrbitControls]. However, it does not maintain a constant camera [page:Object3D.up up] vector.
 				That means if the camera orbits over the “north” and “south” poles, it does not flip to stay "right side up".
+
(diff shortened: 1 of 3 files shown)


Mugen87:
Seems this PR is a duplicate of #17816.

Mugen87:
I think we agreed not to change `TrackballControls` for now until #18496 is resolved.

chabb:
I just realized that I can achieve this in my code for now
```    this.controls.addEventListener('change', () => {
        this.dispatch(this.camera.position, this.camera.quaternion, this.camera.zoom);
        this.renderScene();
      });
      this.controls.addEventListener('start', () => {
        this.controls.update();
        document.addEventListener( 'mousemove', this.mouseTrackballUpdate, false );
      });
      this.controls.addEventListener('end', () => {
        this.controls.update();
        document.removeEventListener('mousemove', this.mouseTrackballUpdate, false);
      });```

Issue #18975, opened 2020-03-25
Matrix added to instance mesh on subsequent frame doesn't work


TheBunyip:
I found that if I try to add another instance to an InstancedMesh on a frame(?) after the one on which the InstancedMesh is created, the second instance it doesn't appear.

http://jsfiddle.net/vm4zpdx1/2/

Mugen87:
Updated fiddle: http://jsfiddle.net/d4oq5ahx/

If you modify instanced data, you also have to set `InstancedMesh.instanceMatrix` to `true`.

Please use the [forum](https://discourse.threejs.org/) or stackoverflow if you need more help.

TheBunyip:
Ah OK. The documentation says this..

.instanceMatrix : BufferAttribute
Represents the local transformation of all instances. For internal use only.

Mugen87:
Yes, the documentation should be better in this case. How about:

> Represents the local transformation of all instances. You have to set its *needsUpdate*  flag to *true* if you modify instanced data via [page:.setMatrixAt()].

Mugen87:
Let's give this a try.

PR #18977, opened 2020-03-25
Docs: Improve InstancedMesh page.

Mugen87:
see https://github.com/mrdoob/three.js/issues/18975#issuecomment-604041820


mrdoob:
Thanks!

PR #18987, opened 2020-03-28
Helpers: Properly specify the type property

WestLangley:
... in the` src/` and `/examples/jsm` files.

The type must be specified after the super is called, otherwise it is overwritten.

### DIFF
--- a/examples/jsm/helpers/FaceNormalsHelper.js
+++ b/examples/jsm/helpers/FaceNormalsHelper.js
@@ -54,6 +54,8 @@ function FaceNormalsHelper( object, size, hex, linewidth ) {
 
 	LineSegments.call( this, geometry, new LineBasicMaterial( { color: color, linewidth: width } ) );
 
+	this.type = 'FaceNormalsHelper';
+
 	//
 
 	this.matrixAutoUpdate = false;
--- a/examples/jsm/helpers/LightProbeHelper.js
+++ b/examples/jsm/helpers/LightProbeHelper.js
@@ -106,6 +106,8 @@ function LightProbeHelper( lightProbe, size ) {
 
 	Mesh.call( this, geometry, material );
 
+	this.type = 'LightProbeHelper';
+
 	this.onBeforeRender();
 
 }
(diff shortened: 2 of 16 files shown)


Mugen87:
The `type` property is mainly used for serialization/deserialization. But even if `ObjectLoader` does not support helpers yet, I think it's good for consistency reasons to specify it properly (see #11233).

Mugen87 (approved):

mrdoob:
Consistency is definitely good 👌

After merging this PR, I propose we move all these `type` properties to a `toJSON()` method instead.

Right now there are two ways of checking what kind of object we're dealing with:

```js
if ( object.isMesh ) {}
if ( object.type === 'Mesh' ) {}
```

`type` is definitely good for serialising and then deserialising with `ObjectLoader`, but having it as a property too seems confusing to me.

This will result on having to add `toJSON` to a bunch of objects. For instance:

```js
toJSON: function ( meta ) {

	var data = Object3D.prototype.toJSON.call( this, meta );
	data.type = 'Mesh';
	return data;

}
```

But it will also allow us to clean up `Object3D.toJSON()` a bit. For example, this code should go in `InstancedMesh.toJSON()`:

https://github.com/mrdoob/three.js/blob/ea3eb81a03701e12fbbcc23c9ff2f99c8dc9610d/src/core/Object3D.js#L683-L689

What do you guys think?

WestLangley:
>The type property is mainly used for serialization/deserialization.

For materials, `type` is used in `renderer.info.programs` and is assigned to program `name`.

Screen Shot 2020-03-28 at 7 09 06 AM

Let's at least retain that information, somehow.

Mugen87:
> What do you guys think?

I like the idea to hide this stuff in `toJSON()` 👍 .

> For materials, type is used in renderer.info.programs and is assigned to program name.

Um, `material.constructor.name` could be an alternative although it won't work for cases like `BackgroundCubeMaterial` and when the library is minified: https://jsfiddle.net/tghrLkvo/1/

WestLangley:
@mrdoob said

>After merging this PR...

Can we do that?

mrdoob:
Thanks!

PR #18988, opened 2020-03-28
OrbitControls bugfix

harshmoorjani:


Mugen87:
`ROTATE`, `DOLLY`, `PAN` has replaced `LEFT`, `MIDDLE`, `RIGHT` so there is not need for a continuous enumeration.

harshmoorjani:
Ok, my bad... and it still did not fix my issue... seems like event is dispatched but there are no handlers for event type = 'start' ... basically nothing happens when clicking on canvas.

PR #18997, opened 2020-03-30
MeshBasicMaterial: Add dithering

gkjohnson:
See forum discussion for context:

https://discourse.threejs.org/t/srgb-texture-encoding-gives-color-banding-issue/13855/17

This PR adds dithering to MeshBasicMaterial which is useful when linear->gamma color conversion is causing banding due to a texture even though the material does not support lighting .


mrdoob:
Thanks!

Issue #18992, opened 2020-03-29
Material: Feature request to use world coordinates of vertex as UV


finnbear:
## Description of the problem

There is an abundance of textures that can be repeated without showing seams. However, if they are applied to anything but the simplest of shapes (squares in a perfect grid), seams will re-emerge. One important feature of seamless textures is that the specific UVs are irrelevant. What matters is that the UVs are consistent between multiple shapes.

In the image, a "seamless" texture has been applied to a grid of hexagons (one of the 3 possible regular polygons that can form a grid). Circled are the intersecting hexagon vertices and arrows point to the edge. Seams have been reintroduced due to the fact that the texture was designed to be seamless over a square, not a hexagon. Some textures, especially ones with higher contrast, appear much worse with these seams.
![Screenshot from 2020-03-29 13-59-20](https://user-images.githubusercontent.com/20015102/77860767-c7a1ff00-71c5-11ea-8446-cbce8b138a01.png)

Using debug texture to highlight that, while specific UVs work for grids of squares, they don't for grids of hexagons (or triangles for that matter):
![Screenshot from 2020-03-29 14-12-02](https://user-images.githubusercontent.com/20015102/77861033-772ba100-71c7-11ea-90f2-221016813ec4.png)

## Describe the bug or feature request in detail

One way to achieve this consistency is the use a global coordinate system to drive the UVs, such as scene/world coordinates. There is a slight complication that the axis that translates world to uv coordinates is relevant to the process. Therefore, I propose that the axis be configurable. The lack of such an axis would indicate that the geometry's specific, local UVs should be used.

Potential API: `material.uvsAxis = quaternion | plane | vector (normal to plane)`

Alternate API: `material.uvsAxis = THREE.XZ | THREE.XY | THREE.YZ` (only support 3 axes)

Alternate API: `geometry.uvsAxis = (see above); material.useUvsAxis = true` (allow each geometry to have its own UV axis.

Both would solve my problem, with the first being highly flexible. **To be clear, the implementation would project the world coordinate onto the axis to determine the resulting UV coordinate.**

Not a solution: User having to manually set the UVs of each vertex of each mesh at all times. This approach doesn't support InstancedMesh, where each instance has the same UVs. It is also needlessly difficult.

## Three.js version

As long as I can remember

##### Browser

- [x] All of them

##### OS

- [x] All of them

donmccurdy:
You can accomplish this with custom shaders, or with the node material system in the examples:

```js
[...]

PR #19010, opened 2020-04-01
Texture: Add WorldMapping to use world coordinates for texturing (#18992)

finnbear:
This PR adds `THREE.WorldMapping`, an alternative to `THREE.UVMapping`. `WorldMapping` works by using two (user defined) axes of the vertex's world coordinates as the UV coordinates for texture mapping. It is configurable via the API:
```javascript
mesh.worldMappingAxes = 'xz'; // the default
mesh.material.map.mapping = THREE.WorldMapping;
```
This is crucial for applying "seamless" textures to grids of non-square objects without manually setting the UVs of each one. As a result, `InstancedMesh` is supported.

The changes don't break any of the examples I tried running.

A working example is available: `webgl_materials_texture_worldmapping`
image

Please see #18992 for more details.

**Warning:** This is my first PR on this repository and the biggest PR I have ever written for any repository. It may break things. Please merge with caution. Suggestions for changes/fixes welcome.


WestLangley:
Ephemeral live link:
https://raw.githack.com/finnbear/three.js/dev/examples/webgl_materials_texture_worldmapping.html

finnbear:
Thanks for adding a live link. I'm in the process of adding my build from this PR to my application to see if it actually solves my problem :wink:. There seems like there might be an issue so hold off on merging for now. Comments are welcome though.

finnbear:
Okay, after some fixes, there are no remaining issues with my app. This is mergable AFAIK :smile:

Mugen87:
see https://github.com/mrdoob/three.js/issues/18992#issuecomment-607127577

finnbear:
see https://github.com/mrdoob/three.js/issues/18992#issuecomment-607285048, in response to previous comment

gkjohnson:
I agree that this is application specific and doesn't belong in core -- I think node materials or ShaderMaterials serve this use case well.

finnbear:
Okay, let's close this and finalize discussion in the issue.

WestLangley:
Removed temporary live link.

Issue #13025, opened 2018-01-03
Documentation: Filter by Method Names


loldrup:
It seems nothing shows up when I filter on 'triangulateShape' on https://threejs.org/docs/
Is that intended behavior? Can you reproduce?

looeee:
TriangulateShape is in the [ShapeUtils page](https://threejs.org/docs/#api/extras/ShapeUtils). 

Currently the search only works for pages, not internal hrefs within pages. 
It would be great if it also searched within pages, so we could return results for things like:

`threejs.org/docs/#api/extras/ShapeUtils.triangulateShape`

Mugen87:
I'm not sure if it's worth to implement something like a full-text search. I prefer to keep things simple even if we have to live with some limitations. The code base of the docs is already quite big (and maybe too complex).

BTW: I can find the documentation page when i just type into google : _triangulateShape_. Even three.js is not required^^.

looeee:
@Mugen87 I agree that a full text search is not necessary - however, each class's parameters and methods are included as internal hyperlinks and it would be useful to be able to search these.

For example, it would be great to easily see which classes override an inherited method such as [Object3D.clone](https://threejs.org/docs/#api/core/Object3D.clone). There is currently no easy way to check this. 
But if you could just search the docs for "clone" and have every class with a clone method show up then checking this would be simple.

Mugen87:
I've renamed the issue so it describes the actual feature request 😇

haroldiedema:
Reference PR #13027

Mugen87:
Merging into #19342.

PR #19038, opened 2020-04-04
[documentation] Add a DocSearch

s-pace:
👋 team,

TL;DR: Add a learn-as-you-type search experience to the documentation

I'm working at Algolia on the a project called [DocSearch](https://community.algolia.com/docsearch/) which goal is to enhance documentation websites with exhaustive, fast and relevant search. You might have seen DocSearch live already on websites like [react](https://reactjs.org/), [Bootstrap](https://getbootstrap.com/), [Brew](https://brew.sh/) or [jQuery](https://jquery.com/).

I have created [a preview of this PR](https://threejs-docsearch.surge.sh/docs/) and what DocSearch on the threejs website could will look like here. Feel free to try it and let us know what you think. Please note the learn-as-you-type experience and the typo tolerance:

The way DocSearch works is by crawling your content, pushing the results into an Algolia index, and then requesting this index directly from the website front-end through JavaScript.

We'll take care of crawling your website and populating the Algolia index with the latest changes every 24h for you. You don't need to change anything to your deployment process. The only thing you need to add are the following CSS and JS snippets that will bind the dropdown to your searchbox.

We built DocSearch with the idea of giving back to the Open-Source community. This is why your [crawling configuration](https://github.com/algolia/docsearch-configs/blob/master/configs/threejs.json) is available on GitHub if you want to change it. We also have our own [documentation](https://community.algolia.com/docsearch/documentation/) to help you tweak the dropdown to your needs. You'll also have access to *analytics on the most searched terms* or those with *no results*. All of this is of course *entirely free*.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -379,7 +379,7 @@
 						hitsPerPage: 10, 'facetFilters': ["lang:" + languageSelect.value]
 					}
 				})
-				if ('URLSearchParams' in window) {
+				if (window.location.hash.includes('q=')) {
 					const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
 					const searchQuery = urlParams.get('q')
 					document.getElementById('filter').value = searchQuery
--- a/docs/index.html
+++ b/docs/index.html
@@ -382,6 +382,7 @@
 				if (window.location.hash.includes('q=')) {
 					const urlParams = new URLSearchParams(window.location.hash.split('?')[1]);
 					const searchQuery = urlParams.get('q')
+					window.innerWidth < 640 && document.getElementById('expandButton').click();
 					document.getElementById('filter').value = searchQuery
 					document.getElementById('filter').focus()
 					docSearch.autocomplete.autocomplete.setVal(searchQuery)
--- a/docs/index.html
+++ b/docs/index.html
@@ -369,9 +369,7 @@
 					apiKey: 'ec69071880b355699ad27cb720becc23',
 					indexName: 'threejs',
 					inputSelector: '#filter',
-					algoliaOptions: {
-						hitsPerPage: 20, 'facetFilters': ["lang:" + languageSelect.value]
-					}
+					algoliaOptions: { 'facetFilters': ["lang:" + languageSelect.value] }
 				})
 			}
 		</script>
--- a/docs/index.html
+++ b/docs/index.html
@@ -6,6 +6,7 @@
 		<meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0">
 		<link rel="shortcut icon" href="../files/favicon.ico" />
 		<link rel="stylesheet" type="text/css" href="../files/main.css">
+		<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
 		<!-- console sandbox -->
 		<script src="../build/three.min.js" async defer></script>
 	</head>
@@ -84,7 +85,6 @@
 				language = value;
 
 				createNavigation();
-				updateFilter();
 				autoChangeUrlLanguage( language );
 
 			}
@@ -139,14 +139,12 @@
 			exitSearchButton.onclick = function( event ) {
 
 				filterInput.value = '';
-				updateFilter();
 				panel.classList.remove('searchFocused');
 
 			}
 
 			filterInput.oninput = function ( event ) {
 
-				updateFilter();
 
 			};
 
@@ -282,117 +280,6 @@
 
 			}
 
-
-			// Filtering
-
-			function updateFilter() {
-
-				// (uncomment the following line and the "Query strings" section, if you want query strings):
-				// updateQueryString();
-
-				var regExp = new RegExp( filterInput.value, 'gi' );
-
-				for ( var pageName in pageProperties ) {
-
-					var linkElement = pageProperties[ pageName ].linkElement;
-					var categoryClassList = linkElement.parentElement.classList;
-					var filterResults = pageName.match( regExp );
-
-					if ( filterResults && filterResults.length > 0 ) {
-
-						// Accentuate matching characters
-
-						for ( var i = 0; i < filterResults.length; i++ ) {
-
-							var result = filterResults[ i ];
-
-							if ( result !== '' ) {
-								pageName = pageName.replace( result, '<b>' + result + '</b>' );
-							}
-
-						}
-
-						categoryClassList.remove( 'hidden' );
-						linkElement.innerHTML = pageName;
-
-					} else {
-
-						// Hide all non-matching page names
-
-						categoryClassList.add( 'hidden' );
-
-					}
-
-				}
-
-				displayFilteredPanel();
-
-			}
-
-			function displayFilteredPanel() {
-
-				// Show/hide categories depending on their content
-				// First check if at least one page in this category is not hidden
-
-				categoryElements.forEach( function ( category ) {
-
-					var pages = category.children;
-					var pagesLength = pages.length;
-					var sectionClassList = category.parentElement.classList;
-
-					var hideCategory = true;
-
-					for ( var i = 0; i < pagesLength; i ++ ) {
-
-						var pageClassList = pages[ i ].classList;
-
-						if ( ! pageClassList.contains( 'hidden' ) ) {
-
-							hideCategory = false;
-
-						}
-
-					}
-
-					// If and only if all page names are hidden, hide the whole category
-
-					if ( hideCategory ) {
-
-						sectionClassList.add( 'hidden' );
-
-					} else {
-
-						sectionClassList.remove( 'hidden' );
-
-					}
-
-				} );
-
-			}
-
-
-			// Routing
-
-			function setUrlFragment( pageName ) {
-
-				// Handle navigation from the subpages (iframes):
-				// First separate the memeber (if existing) from the page name,
-				// then identify the subpage's URL and set it as URL fragment (re-adding the member)
-
-				var splitPageName = decomposePageName( pageName, '.', '.' );
-
-				var currentProperties = pageProperties[ splitPageName[ 0 ] ];
-
-				if ( currentProperties ) {
-
-					window.location.hash = currentProperties.pageURL + splitPageName[ 1 ];
-
-					createNewIframe();
-
-				}
-
-			}
-
 			function createNewIframe() {
 
 				// Change the content displayed in the iframe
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -474,6 +361,21 @@ <h1><a href="http://threejs.org">three.js</a></h1>
...
+		<script>
+			function addSearch() {
+				var languageSelect = document.getElementById('language');
+	
+				docsearch({
+					apiKey: 'ec69071880b355699ad27cb720becc23',
+					indexName: 'threejs',
+					inputSelector: '#filter',
+					algoliaOptions: { 'facetFilters': ["lang:" + languageSelect.value] },
+					debug: true // Set debug to true if you want to inspect the dropdown
```
Shipow, line 373: set debug to false to fix the blur behavior

looeee:
This is nice. Some searches give better results than the old system. But it seems to limit to only 5 results:

alg

Compare that to the old one: 

sea

Also, there's still no way to share searches. 

This means docs search works differently than examples. I can share a link to all geometry examples. 

https://threejs.org/examples/?q=buffergeometry

I can't do that for docs. There are multiple places in my book where I have to say something like: search the docs for "material". I'd rather say: here's a link to material pages in the docs.

Is there any way to configure this to: 

1. return all the search results
2. save the search in the URL like examples?

looeee:
That's better - does it return all results now? However, there's no scroll so when there's a lot of results it goes below the bottom of the screen and I can't see them. 

Maybe you could reduce the vertical space between results?

s-pace:
Should be good now =)

looeee:
Yeah, that's better. I still don't think it's returning all the results though:

se

One thing that's a little annoying - when I click anywhere outside the results box or the window loses focus, the results disappear. I think it would be better if they stay visible until the search term is cleared or a result is selected.

s-pace:
Thanks @looeee for your feedback.

We are displaying 10 results. Giving the relevancy, I do think it is better this way. We also close the dropdown when not focus to avoid disturbing the reading of the documentation. I do think these are the best settings.

looeee:
> Giving the relevancy, I do think it is better this way.

So,  when a user searches the docs for "material", they will no longer be able to find all the material pages? That seems odd to me. When I use a doc search function, I expect to get _all_ results, not just the ones that some AI algorithm decides are relevant. It's fine if some results are not on the first page or something. But all results should be available. 

> We also close the dropdown when not focus to avoid disturbing the reading of the documentation. I do think these are the best settings.

Well, I don't 😁 I like the new search UI, but this is the one thing that stands out as annoying to me. It's not a big deal though, let's see what other people have to say. 

Otherwise, would it be possible to have the results come up in the menu (like the current search results)?

s-pace:
Let's wait for other feedback but we can definitely increase the number of results or/and let the dropdown open.

> Otherwise, would it be possible to have the results come up in the menu (like the current search results)?

This feature will need to a custom UI and it will not be out of the box as DocSearch is.

looeee:
Playing with it a bit more, what's bugging me is that that the search results vanish when the window loses focus, or when I switch tabs. However, the results closing when I interact with another part of the docs page is fine.

s-pace:
We have seen with more than 3 millions of users a month that 5 hits are enough given the conversational search where people can change their query with real-time results. Happy to change it if another contributor thinks it is better.  We display more textual content to give better context of a result so I do not want to charge too much the interactions.

Happy to change it if needed :) 🔍 🚀

mrdoob:
Does this require internet connection to work?

Shipow:
@mrdoob yes it requires an internet connection to connect to the nearest proxy (the API is distributed all around the world to offer the fastest response time possible).
It's possible to implement it as a progressive enhancement where you could keep the current search as a fallback if there are no internet connection.

looeee:
> We have seen with more than 3 millions of users a month that 5 hits are enough given the conversational search where people can change their query with real-time results.

For experienced users of three.js, you may be right. However, I'm thinking about this from the point of view of a beginner who is using the search function to discover the scope of the library. Questions they might reasonably expect the search function to answer:

1. What are all the material classes? (i.e. search "materials")
2. What are all the geometry classes? (i.e. search "geometry")

... and so on. If we switch to a search function that only returns 10 results (or any number less than _all_ results), we are making the docs less beginner-friendly. 

For example, using your search function, "material" doesn't bring up "MeshStandardMaterial", "MeshPhongMaterial", "MeshPhongMaterial", "LineBasicMaterial", or "PointsMaterial". Instead, it mostly brings up a bunch of advanced pages - "ShadowMaterial", "MeshDistanceMaterial", and so on. 

I guess that's something you hope would improve over time once the algorithm has worked with the docs for a while. But I don't think we should leave that up to an algorithm, and in any case, the current search function already handles this fine. 

@s-pace thank you for your responsiveness to my feedback. Your search function is great, but in my opinion, it's not a good fit for the three.js docs unless it returns all results. However, I'll step back now and let other people give their opinions.

Shipow:
@looeee Thanks for your feedback. I totally get your point, and also think it would be very useful for the links in the book. I believe we could have the best of both approach (relevance + exhaustivity) with a 'show more' link.

marcofugaro:
Thanks for your contribution.

From a design standpoint, this looks like patchwork and does not look very integrated with the current design.

It is clipped and the user can't scroll on 13'' screens:


Screenshot 2020-04-09 at 11 26 03

And it overflows the whole page on mobile:




It also comes with a lot of visual noise and can often confuse the user. A solution like this would fit best with the current website:
.


Hope you appreciate this feedback.

Mugen87:
The PR would fix #13025 though.

If it is decided not to merge the PR or work at a similar solution, I suggest to close the above issue.

s-pace:
Thanks @Mugen87 

Can you elaborate why you do not want to use this solution please?

Mugen87:
> Can you elaborate why you do not want to use this solution please?

Oh, I did not want to say this. I just wanted to link the issue here and suggest to close it if the search won't be changed.

s-pace:
Alright thank you @Mugen87 

Just thought it would be sad to not benefit the typo-tolerance feature, high speed and the `search with no results` analytic that can help you to continue to build this great documentation.

s-pace:
Thanks @marcofugaro for the feedback.

I will make sure the integration is better on all device. On it :)

s-pace:
I have deployed the new version here https://threejs-docsearch.surge.sh/docs/#manual/en/introduction/Creating-a-scene

marcofugaro:
@s-pace thanks but it is still broken (on desktop you can't scroll anymore in the sidebar and on mobile it still overflows to the right).

However that is not my point.

My point was on the design and the UX. It doesn't fit with the current design.
It doesn't feel integrated like on https://reactjs.org/ for example.

As a user, I would expect the search to look and behave exactly like it does on the [example page](https://threejs.org/examples/), since they're exactly the same UI element. If it doesn't, I would feel confused and the search would be less straightforward. [This is a basic UX principle, and is called consistency](https://uxdesign.cc/design-principle-consistency-6b0cf7e7339f).

Let me elaborate on the solution that in my opinion could work.

-----------------------

Let's say I am a new Three.js user. I just learned about gamma correction, and want to learn how Three.js handles it from the docs.

Here is with your solution.

Screenshot 2020-04-16 at 23 23 44


This contains a lot of visual noise, for example I don't want to know right away what parameters it accepts and the typings.

Here is how I would prefer it to look instead.




That's much more straightforward and doesn't get in the way of the user.

-------------------

However that is just my opinion, and I would like to hear other contributors' feelings about this.

DefinitelyMaybe:
I agree with @marcofugaro. If the search pops out as another ui element ontop, it feels more like a pop-up than my search results.

> We have seen with more than 3 millions of users a month that 5 hits are enough given the conversational search where people can change their query with real-time results

could you please expand on this a little bit? Is it something like -> Of the 3 million users a month that use your search API receives, 90+% selected one of the first 5 results?

[Converstional search](https://resources.algolia.com/voice/what-is-conversational-search-2) seems like a pretty cool new feature. Unless I'm mistaken, this would be the kind of feature that helps users get to the parts of an exhaustive search that they're looking for, however I'm wondering about how that user experience goes.
- Putting voice input to one side, would someone be simply adding more terms into the search box?
- If previous searches are used as context how would the user update them when they realize it wasn't what they were looking for (in some sense, how could the user escape their own filter bubble).
- Would we need to tell people how to use the feature?

s-pace:
> @s-pace thanks but it is still broken (on desktop you can't scroll anymore in the sidebar and on mobile it still overflows to the right).

This is solved now. https://threejs-docsearch.surge.sh/docs/#manual/en/introduction/Creating-a-scene

> My point was on the design and the UX. It doesn't fit with the current design.
It doesn't feel integrated like on https://reactjs.org/ for example.

My idea is that once you validate the search on the documentation, we will add it to the example section too. It will make the experience consistent.

> This contains a lot of visual noise, for example I don't want to know right away what parameters it accepts and the typings.

Definitely agree. I have made sure we remove the superfluous content from the crawl. What we do now have is not noise anymore but context which will empower your user to solve their issue.

> could you please expand on this a little bit? Is it something like -> Of the 3 million users a month that use your search API receives, 90+% selected one of the first 5 results?

This is not my statement ^^

`We have seen with more than 3 millions of users a month that 5 hits are enough given the conversational search where people can change their query with real-time results. Happy to change it if another contributor thinks it is better. We display more textual content to give better context of a result so I do not want to charge too much the interactions.`

We have seen that displaying 5 records is enough. I have make sure this UI display 10 instead to be sure previous feedback are taking into consideration.

> Putting voice input to one side, would someone be simply adding more terms into the search box?

Voice search is something we are working on. It is not integrated yet but it could be.

> If previous searches are used as context how would the user update them when they realize it wasn't what they were looking for (in some sense, how could the user escape their own filter bubble).

We do not use previous searches as context in this experience

> Would we need to tell people how to use the feature?

It is pretty intuitive. None DocSearch user tells people how to use the search yet.

# OLD vs NEW

1st the search is now shareable. It was not the case in the past

`change`
image

`gamma`
image

`locolly`
image

`gama` (mind the typo)
image

I do think the search will improve the DX of this great project 🚀 

Feel free to share any other question you may have. Happy to help on this.

mrdoob:
@marcofugaro 

> Here is how I would prefer it to look instead.
>
> 
>
> That's much more straightforward and doesn't get in the way of the user.
> However that is just my opinion, and I would like to hear other contributors' feelings about this.

Thanks for doing that mockup! I was going to suggest exactly the same 🙌

It's a shame that proposed solution ignores the current UX.

Also, the fact that it requires internet connection is a big red flag, unfortunately.

mrdoob:
As a side note... I find it "fishy" that I had never seen before the names of the users adding hearts and thumb up reactions to @s-pace posts 🤔

Mugen87:
@s-pace Thanks for your work on this. But since the PR has some conceptual issues, it's not possible to merge it. Closing for now.

PR #19052, opened 2020-04-06
Examples: Improve OrbitControls settings.

Mugen87:
See https://github.com/mrdoob/three.js/pull/18864#discussion_r391314316.

### DIFF
--- a/examples/css2d_label.html
+++ b/examples/css2d_label.html
@@ -39,7 +39,7 @@
 				var EARTH_RADIUS = 1;
 				var MOON_RADIUS = 0.27;
 
-				camera = new THREE.PerspectiveCamera( 45, window.innerWidth / window.innerHeight, 1, 1000 );
+				camera = new THREE.PerspectiveCamera( 45, window.innerWidth / window.innerHeight, 0.1, 200 );
 				camera.position.set( 10, 5, 20 );
 
 				scene = new THREE.Scene();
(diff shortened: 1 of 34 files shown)


mrdoob:
Thanks!

examples/css2d_label.html
```diff
@@ -39,7 +39,7 @@
 				var EARTH_RADIUS = 1;
 				var MOON_RADIUS = 0.27;
 
-				camera = new THREE.PerspectiveCamera( 45, window.innerWidth / window.innerHeight, 1, 1000 );
+				camera = new THREE.PerspectiveCamera( 45, window.innerWidth / window.innerHeight, 0.1, 200 );
```
WestLangley, line 42: I would prefer to avoid small values of the near plane in the examples.

When the far plane is 200, I would expect a near plane of 1 would be fine.

examples/webgl_loader_vrml.html
```diff
@@ -38,7 +38,7 @@
 
 			function init() {
 
-				camera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 0.01, 1e10 );
+				camera = new THREE.PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 0.1, 1e10 );
```
WestLangley, line 41: ditto.

examples/css2d_label.html
Mugen87: Yes, these values can still be improved. I think it make sense to look over the examples in general and optimize the near and far planes.

PR #18839, opened 2020-03-08
Make examples searchable by screenshots.

munrocket:
Trying to solve problem with searching page, since we don't have any progress in #18714

[Live page on GitHack](https://rawcdn.githack.com/munrocket/three.js/98454737436eda21bd96a8a2a277d93047f28062/examples/index.html)

2do list:
- [x] ui button that works on mobile phones ( in own PR #18841 )
- [x] spiral sphere geometry that easily to extend with new examples
- [x] MVP with octahedral mapping
- [x] uv maker with stripe mapping
- [x] automatic spiral resize that keep 800x600 ratio
- [x] analytic raycasting
- [x] image picker that works on phones
- [x] fix redirect & uv maker
- [x] highlight tiles
- [x] uv maker with cpu
- [x] highlight names
- [x] preload,  progressive loading

PR #18714, opened 2020-02-24
New version of index_thumbs.html that displays a list of examples and thumbnails

callumprentice:
Re-added the code and made a new pull request after something bad happened to original repository. 

This change re-adds the index_thumbs.html and associated CSS file -this page displays the list of examples dynamically and now references the official screenshots folder for thumbnails.

Also re-added the node.js script I wrote to capture thumbnails for all examples.

PR #19075, opened 2020-04-08
Tests: e2e coverage

munrocket:
Mugen suggestion: making it easy to add new examples.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,12 @@ jobs:
       env: FORCE_COLOR=0 CI=2
     - <<: *e2e
       env: FORCE_COLOR=0 CI=3
+
+    - &e2e-cov
+      stage: e2e-cov
+      name: e2e-cov
+      script: npm run test-e2e-cov
+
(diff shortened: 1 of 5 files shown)


Mugen87:
Can you please explain a bit what the PR does?

munrocket:
With this PR new examples not require anymore screenshots for passing tests successfully. But I am created a new stage called `e2e-cov` witch check integrity and:

* suggest to add screenshots for new examples
* suggest to remove obsolete screenshots
* suggest to add description in file.js for example
* suggest to remove description in file.js for example

This stage is in `allow_failures` mode and visible only in travis report (or github actions).
Also It can be useful for checking integrity before you releasing a new revision or for thumbnails navigation (#18839 or #18714). And I am suggest a workflow where we add new screenshots before each release.

I am added little bit of refactoring and freezed puppeteer version for future, so this PR can look little bit messy. But I am checked how it works before force-push, all seems ok.

mrdoob:
Thanks!

PR #19070, opened 2020-04-07
EXRLoader: add RGBE - UnsignedByteType support

sciecode:
Based on https://github.com/mrdoob/three.js/issues/19053#issuecomment-610101599

Should I log warning when the input EXR file is RGBA format and output type is `UnsignedByteType`?
To remind that alpha channel information is lost.

[webgl_materials_envmaps_exr.html](https://rawcdn.githack.com/sciecode/three.js/819b64a2eb5fe6d2949bca672d226245af318cfa/examples/webgl_materials_envmaps_exr.html)

PR #19081, opened 2020-04-08
Tests: webgl_materials_envmaps_exr screenshot.

munrocket:
Don't know what is happening here, this is an update or bug in #19070.

PR #19083, opened 2020-04-08
EXRLoader: fix RGBE data encoding

sciecode:
Follow up to #19070

Fixed mirrored mapping on RGBE encoded texture.

[webgl_materials_envmaps_exr.html](https://rawcdn.githack.com/sciecode/three.js/d4ad0e7ca24af61308f6a083a10976b55c376f16/examples/webgl_materials_envmaps_exr.html)

### DIFF
--- a/examples/js/loaders/EXRLoader.js
+++ b/examples/js/loaders/EXRLoader.js
@@ -2151,32 +2151,39 @@ THREE.EXRLoader.prototype = Object.assign( Object.create( THREE.DataTextureLoade
 
 		if ( this.type === THREE.UnsignedByteType ) {
 
-			let v;
+			let v, i, j;
 			const size = byteArray.length;
 			const RGBEArray = new Uint8Array( size );
 
-			for ( let i = 0; i < size; i += 4 ) {
+			for ( let h = 0; h < height; ++ h ) {
 
-				const red = byteArray[ size - ( i + 4 ) ];
-				const green = byteArray[ size - ( i + 3 ) ];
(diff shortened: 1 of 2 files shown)


munrocket:
>... was indeed a bug, filed a fix for it on

Yep, we should revert #19081 after merge.

WestLangley:
@sciecode It seems that `EXRLoader` loads an `UnsignedByteType` texture shifted vertically -- as compared to float and half-float.

sciecode:
Was using a shifted indexing, I believe last commit should solve it.

WestLangley:
@sciecode Can you confirm with the `EXRLoader` example?

sciecode:
I did make sure the were the same.

[UnsignedByte Example](https://rawcdn.githack.com/sciecode/three.js/7149e853d0ce4813c7ed276daf1876a4f1a5d1d8/examples/webgl_loader_texture_exr.html)

[Float Example](https://rawcdn.githack.com/sciecode/three.js/f12968c3721fecda86ad14c9591c12ff463ad9f3/examples/webgl_loader_texture_exr.html)

WestLangley:
>Visual differences are due to encoding precision

Actually, I expect it is due to the "nearest" filtering. The unsigned-byte case does not look good. Using nearest with the float-case looks equally-bad.

I do not like RGBE for that reason.

WestLangley:
@mrdoob This is ready to be merged.

mrdoob:
Thanks!

PR #19087, opened 2020-04-08
PMREMGenerator: set internal/output render targets as RGBE.

### LINKED ISSUE #19053: EXR Envmap not working on iOS
Following the example on: https://threejs.org/examples/webgl_materials_envmaps_exr.html
and loading a EXR file as environment map with:

```js
var pmremGenerator = new THREE.PMREMGenerator( this.renderer );
pmremGenerator.compileEquirectangularShader();
var exrCubeRenderTarget = pmremGenerator.fromEquirectangular( texture );
this.hdr = exrCubeRenderTarget.texture;
```

This works perfectly on all desktop browsers but on iOS the HDR is completely black and no lighting is applied to the model. This also applies to the example link on iOS: https://threejs.org/examples/webgl_materials_envmaps_exr.html

##### Three.js version

- [ ] Dev
- [x] r115
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [ ] Windows
- [ ] macOS
- [ ] Linux
- [x] Android
- [x] iOS

sciecode:
Fixes #19053 

Based on https://github.com/mrdoob/three.js/issues/19053#issuecomment-610569210 and https://github.com/mrdoob/three.js/issues/19053#issuecomment-610969027

~This PR introduces the possibility for the output texture generated by PMREM to have different type than the inputted texture. This is important for mobile support, which often lacks the ability to render to floating point textures.~

See https://github.com/mrdoob/three.js/pull/19087#issuecomment-612161233 and https://github.com/mrdoob/three.js/pull/19087#issuecomment-612234820

### DIFF
--- a/src/extras/PMREMGenerator.js
+++ b/src/extras/PMREMGenerator.js
@@ -230,9 +230,9 @@ PMREMGenerator.prototype = {
 			magFilter: NearestFilter,
 			minFilter: NearestFilter,
 			generateMipmaps: false,
-			type: equirectangular ? equirectangular.type : UnsignedByteType,
-			format: equirectangular ? equirectangular.format : RGBEFormat,
-			encoding: equirectangular ? equirectangular.encoding : RGBEEncoding,
+			type: UnsignedByteType,
+			format: RGBEFormat,
+			encoding: _isLDR( equirectangular ) ? equirectangular.encoding : RGBEEncoding,
 			depthBuffer: false,
 			stencilBuffer: false
 		};
@@ -358,7 +358,7 @@ PMREMGenerator.prototype = {
 		}
 
 		uniforms[ 'inputEncoding' ].value = ENCODINGS[ texture.encoding ];
-		uniforms[ 'outputEncoding' ].value = ENCODINGS[ texture.encoding ];
+		uniforms[ 'outputEncoding' ].value = ENCODINGS[ cubeUVRenderTarget.texture.encoding ];
 
 		_setViewport( cubeUVRenderTarget, 0, 0, 3 * SIZE_MAX, 2 * SIZE_MAX );
 
@@ -504,6 +504,14 @@ PMREMGenerator.prototype = {
 	}
 
 };
+
+function _isLDR( texture ) {
+
+	if ( texture === undefined || texture.type !== UnsignedByteType ) return false;
+
+	return texture.encoding === LinearEncoding || texture.encoding === sRGBEncoding || texture.encoding === GammaEncoding;
+
+}
 
 function _createPlanes() {
 


src/extras/PMREMGenerator.js
```diff
@@ -96,6 +98,10 @@ function PMREMGenerator( renderer ) {
 	this._equirectShader = null;
 	this._cubemapShader = null;
 
+	this._type = undefined;
+	this._format = undefined;
+	this._encoding = undefined;
```
WestLangley, line 103: instead,
```js
this.setDataType( THREE.UnsignedByteType );
```

src/extras/PMREMGenerator.js
```diff
@@ -104,6 +110,32 @@ PMREMGenerator.prototype = {
...
+				this._type = type;
+				this._format = RGBEFormat;
+				this._encoding = RGBEEncoding;
+				break;
+
+			case FloatType:
+			case HalfFloatType:
+				this._type = type;
+				this._format = undefined;
+				this._encoding = undefined;
```
WestLangley, line 130: Instead,
```js
this._format = RGBAFormat;
this._encoding = LinearEncoding;
```

src/extras/PMREMGenerator.js
```diff
@@ -230,9 +262,9 @@ PMREMGenerator.prototype = {
 			magFilter: NearestFilter,
 			minFilter: NearestFilter,
 			generateMipmaps: false,
-			type: equirectangular ? equirectangular.type : UnsignedByteType,
-			format: equirectangular ? equirectangular.format : RGBEFormat,
-			encoding: equirectangular ? equirectangular.encoding : RGBEEncoding,
+			type: ( this._type !== undefined ) ? this._type : ( equirectangular ) ? equirectangular.type : UnsignedByteType,
+			format: ( this._format !== undefined ) ? this._format : ( equirectangular ) ? equirectangular.format : RGBEFormat,
+			encoding: ( this._encoding !== undefined ) ? this._encoding : ( equirectangular ) ? equirectangular.encoding : RGBEEncoding,
```
WestLangley, line 267: ...and remove this.

That is what I think, anyway...

sciecode: These sound reasonable, was trying to think if there was a use-case not covered. 
But I don't think so, and we can always adapt if necessary.

sciecode:
If we decide to merge, it is important to notify on changelogs about the default output type changing.

WestLangley:
I think `PMREMGenerator` could be smart about setting the filtering. We can avoid using `NearestFilter` with `Float` and `HalfFloat` data types.

sciecode:
> I think `PMREMGenerator` could be smart about setting the filtering. We can avoid using `NearestFilter` with `Float` and `HalfFloat` data types.

I'm ok with that, @elalish would that cause any problems that we are not considering?

polarathene:
> HalfFloatType - > PMREMGenerator -> UnsignedByteType Example
> FloatType - > PMREMGenerator -> UnsignedByteType Example

Tested on my Android phone Huawei Mate 10 Pro, purchased in 2018 for around 600 USD? So not the best but not budget either.

Firefox and Chrome browsers fail both examples. UI renders, but the 3D content appears black.
Firefox Focus(which uses a different web engine from the main Firefox product afaik) can display the 1st example, but also fails the 2nd.

elalish:
@sciecode @WestLangley It's true you could avoid `NearestFilter` with float types, but the only reason to do so is if you change the shaders to not do custom interpolation. Of course this will mean another branch in the bilinearCubeUV shader fragment based on type (perhaps you could use a #define instead). I'm also curious how much performance improvement this will be; it saves the RGBE decodes and lets the compiler take care of interpolation, but needs to push more memory through the cache. 

Also, why do you expect iphones to fail your halfFloat test? It seems like we should test this on quite a few devices before making this change. 

Oh, and what exactly is the change to the default output type?

src/extras/PMREMGenerator.js
```diff
@@ -104,6 +109,32 @@ PMREMGenerator.prototype = {
...
+	 * Defines output and intermediary render target texture types.
+	 */
+	setDataType( type ) {
+
+		switch ( type ) {
+
+			case UnsignedByteType:
+				this._type = type;
+				this._format = RGBEFormat;
+				this._encoding = RGBEEncoding;
```
elalish, line 122: I don't think this is enough granularity. Before it worked so that a JPG turns into `RGBFormat` with `sRGBEncoding`, while an HDR turns into `RGBEFormat` with `RGBEEncoding`. Not that JPG is the highest quality for PMREM, but it's fast, so long as you're not storing it as RGBE. Or is the idea that all JPGs should be decoded as Float? I suppose another option would be to have a data type = auto, which retains the old behavior.

src/extras/PMREMGenerator.js
elalish: Perhaps my comment is related to this. @WestLangley what do you think about the `sRGBEncoding` case?

src/extras/PMREMGenerator.js
WestLangley: PMREM should output to linear color space. Right? Why sRGB?

WestLangley:
@bhouston [said](https://github.com/mrdoob/three.js/issues/6383#issuecomment-97844411):

>But I will warn you, it turns out that doing live HDR decodings from 8-bit per channel RGBA textures is highly problematic. I should have just converted textures with HDR encodings to true half (fp16) textures on load, rather than doing live decoding.

src/extras/PMREMGenerator.js
elalish: Well, it should be linear if it's float, but definitely non-linear if it's 8-bit. RGBE and sRGB are both nonlinear formats. If the input is sRGB, the output should match, unless it's upgraded to float.

WestLangley: We need to separate encoding from color space.

>If the input is sRGB, the output should match

I see no reason for that.

I do not think PMREM should output to sRGB color space.

It is better to convert to linear once on the CPU rather than per pixel, per frame on the GPU.

elalish: Yes, I agree, the problem is that floating point renderTargets aren't widely supported. In leu of that, it seems like converting sRGB is less onerous than converting RGBE, but maybe I'm wrong?

sciecode: That is exactly what I made initially.
If `setDataType( UnsignedByteType )` we modified to `RGBEEncoding`, otherwise we kept the input texture parameters.

sciecode:
> Also, why do you expect iphones to fail your halfFloat test?.

iPhones don't seem to be able to decode `HalfFloatType` from `Uint16Array` data.
I'm know it happens up until 8, not sure about newer models.

image

> Oh, and what exactly is the change to the default output type?

Previously the default behavior was to copy input texture parameters, in the current state it defaults to `UnsignedByteType` `RGBEFormat` `RGBEEncoding`, unless explicitly changed with `setDataType`

elalish:
@sciecode Wait, hold on, I didn't grok this the first time through. If you set data type to float, you're changing all the render targets to float type, which doesn't work on most mobile devices. I thought the idea here was to just accept float input and still use the same 8-bit internal formats for compatibility.

src/extras/PMREMGenerator.js
WestLangley: PMREM should always output to linear color space. It should also do its calculations in linear color space.

PMREM should output to RGBE if the output data type is `UnsignedByteType` and the input is .hdr or .exr. (maybe that is the point of confusion?).

RGBE is assumed to be in linear color space. We do not have an RGBE-sRGB encoding.

sciecode:
Well, the idea was to increase support. Let me explain.

With dev version, PMREM only works on mobile if we input an RGBE 8-bit texture that outputs to RGBE as well.
With this version, PMREM should work on every inputted texture format ( as long as device can read form floating point texture ), but we still need to output to RGBE. 

In other words, we are increasing the support of inputted texture formats, not outputted. If we want to output as `HalfFloatType` or `FloatType`it still won't work, not much we can do about that.

src/extras/PMREMGenerator.js
```diff
@@ -230,9 +261,9 @@ PMREMGenerator.prototype = {
 			magFilter: NearestFilter,
 			minFilter: NearestFilter,
 			generateMipmaps: false,
-			type: equirectangular ? equirectangular.type : UnsignedByteType,
-			format: equirectangular ? equirectangular.format : RGBEFormat,
-			encoding: equirectangular ? equirectangular.encoding : RGBEEncoding,
+			type: this._type,
```
elalish, line 264: This is the format of the renderTargets PMREM will write to and output. If this is `FloatType` above, it will fail on any device that can't write to float framebuffers, which I believe is why we've already had a report of black renders with this change. See related comment above.

src/extras/PMREMGenerator.js
```diff
@@ -104,6 +109,32 @@ PMREMGenerator.prototype = {
...
+
+			case UnsignedByteType:
+				this._type = type;
+				this._format = RGBEFormat;
+				this._encoding = RGBEEncoding;
+				break;
+
+			case FloatType:
+			case HalfFloatType:
+				this._type = type;
```
elalish, line 127: Here the `FloatType` input becomes also the output format, I believe?

sciecode: Nope, this is only when the user manually call `setDataType( FloatType / HalfFloatType )`.

src/extras/PMREMGenerator.js
sciecode: Yes, but the question is whether we want to kill the ability to render to float framebuffers or not. 
The way we're trying to set-up, we let the application decide that.

elalish: Ah, okay, I finally understand the problem. I was confused because this PR changes several things at once. I believe that allowing the application to choose the data type of the output is a foot-gun because anything other than 8-bit will fail on many mobile devices. Also, in order to get any benefit from it, you would also need to change the shaders to handle this case without custom interpolation. 

I agree that you don't want to blindly follow the input data type for the output data type, so that change is good. The only reason it did that originally was to support both RGBE and sRGB output. But it would be a very simple change to just use RGBE always.

src/extras/PMREMGenerator.js
sciecode: Oh, I think I understand why the confusion. `setDataType` is meant to set the framebuffers type ( and associated format/encoding ), not the input texture type.

src/extras/PMREMGenerator.js
sciecode: > Also, in order to get any benefit from it, you would also need to change the shaders to handle this case without custom interpolation.

I think we can try that at a different PR, I could take a look at it. 
But the main benefit right now, in the case of EXR at least, is that we don't need to convert the data to RGBE on the CPU before inputting to PMREM. EXR doesn't have internal data as RGBE, it's always 16/32 float.

elalish: Agreed; personally I'd remove the `setDataType` from this PR and put it in the follow-on if/when that happens. I'm still concerned about it being a foot-gun. The minimal change would be to use RGBE for any input encoding that's not sRGB. The other way is to make the internal format RGBE always. I'll defer to @WestLangley on that one. Just make sure to test a few JPG environments.

WestLangley:
@elalish 

How do you feel about this?

1. Support the following image formats as input:
	(a) Radiance HDR
	(b) OpenEXR
	(c) JPG/PNG

2. Internally, render targets will be `UnsignedByteType `.

3. Output to a texture having user-specified optional data type:
	(a) `UnsignedByteType`, the default, with `RGBEEncoding`, `NearestFilter`
	(b) `Float` or `HalfFloat`, with `LinearEncoding`, `LinearFilter`


In the case of 3(a),  `RGBEEncoding` is only required for HDR, otherwise LinearEncoding -- not sRGB.

elalish:
> How do you feel about this?
> 
> 1. Support the following image formats as input:
>    (a) Radiance HDR
>    (b) OpenEXR
>    (c) JPG/PNG
> 2. Internally, render targets will be `UnsignedByteType `.
> 3. Output to a texture having user-specified optional data type:
>    (a) `UnsignedByteType`, the default, with `RGBEEncoding`, `NearestFilter`
>    (b) `Float` or `HalfFloat`, with `LinearEncoding`, `LinearFilter`
> 
> In the case of 3(a), `RGBEEncoding` is only required for HDR, otherwise LinearEncoding -- not sRGB.
> 
> I discussed this offline with @sciecode, and I think we agree that this makes sense.
> 
> Any "gotchas" we are missing?

@WestLangley 1 and 2 sound good. Number 3 has the same gotchas as number 2; in order to output a float texture you have to be able to render to a float texture, which mobiles can't reliably do. And in order to make use of linear filtering you'll need to change the cube_uv shader fragment to remove the custom interpolation. Hence my suggestion to save those for later and always output RGBE for now.

WestLangley:
> in order to output a float texture you have to be able to render to a float texture

@elalish Why not create a float data texture and populate it from the RGBE Uint internal render target?

elalish:
> @elalish Why not create a float data texture and populate it from the RGBE Uint internal render target?

@WestLangley Do you mean by going through the CPU? Or is there another way to convert from RGBE to float texture without writing to a renderbuffer? I may just be unaware.

sciecode:
> And in order to make use of linear filtering you'll need to change the cube_uv shader fragment to remove the custom interpolation. Hence my suggestion to save those for later and always go with 3a for now.

Ok, I understand now.  `cube_uv_reflection_fragment` was refactored to use the custom bilerp for mipmapped envmap, `LinearFilter` will cause incorrect sampling of the envMap even after the PMREM pipeline. I thought the problem was only about the internal RTT pipeline. 

Now your comment about using hardware lerp vs shader lerp makes more sense to me.

In that case, I say let's go with your proposal (1), (2) and (3a). I can study the impact of using `LinearFilter` later on, but I don't expect too much performance difference to be worth trouble of doing it.

WestLangley:
@elalish @sciecode Yes, let's do the things we agree on. :-)

sciecode:
I think this one is good to go, aye? 👍

mrdoob:
@elalish @WestLangley 👌?

WestLangley:
As I proposed in https://github.com/mrdoob/three.js/pull/19087#issuecomment-612161233, I would only output `RGBE` if the input texture is HDR. Otherwise, the output should be `Linear`.

sciecode:
How do you suggest we test for input HDR texture that?
`type == UnsignedByteType` and `encoding !== RGBE, RGBM16, RGBM7, LogLUV` ?

sciecode:
@WestLangley please check if current solution fits with your idea. 

`RGBEFormat == RGBAFormat` by the way, so we don't need to change that based on LDR/HDR

WestLangley:
twitter

elalish:
> As I proposed in [#19087 (comment)](https://github.com/mrdoob/three.js/pull/19087#issuecomment-612161233), I would only output `RGBE` if the input texture is HDR. Otherwise, the output should be `Linear`.

@WestLangley @sciecode Sorry, but I again have to disagree here; I would approve this without the last commit this comment added. If we have input in an 8-bit nonlinear encoding (sRGB or Gamma), and we encode it as 8-bit linear, we get significant quantization loss. I know it would be nice to not convert back and forth all the time, but without writing floats, there's just not a good way around it.

sciecode:
Yeah, there is a great deal of quantization loss actually.

RGBE
RGBE

Linear
Linear

sciecode:
Let's just shake on `RGBE` for everything for now, just so we have support for multiple input textures, and later on we can continue discussing improvements we can do. @WestLangley 👍 ?

elalish:
> Yeah, there is a great deal of quantization loss actually.

@sciecode Is there a chance you pasted the same image twice?

sciecode:
Nope, I can see clearly the "salt and pepper" difference in the first one. I do have a very high contrast on my monitor, try adjusting a bit in your and I guarantee you're gonna see it.

bhouston:
A couple years ago I quantified the loss when going from gamma to linear in a 256 byte wide color space:
https://docs.google.com/spreadsheets/d/1sEhW8QeilNtsjcgd0fX1dcoO3L1p4Ow-U9v96I0aYiM/edit?usp=drive_web&ouid=117151377272934601196

You start with 256 unique intensities and end up with just 184.  See above spreadsheet.

The trick is to not re-encode these from gamma to linear in 8-bit per channel textures, but rather do inline decoding.

sciecode:
Another one for the non-believers :grin:

RGBE
RGBE

Linear
Linear

sciecode:
@bhouston thanks for the input.

When you say inline decoding, you mean directly on the shader when reading the target texture correct?

bhouston:
> When you say inline decoding, you mean directly on the shader when reading the target texture correct?

Yes, I mean that if you have an sRGB texture, leave it in sRGB format until you are in the shader itself, then decode it directly to floating point values.  If you rencode an 8-bit sRGB texture into a 8-bit linear texture, you've just lost 25% of your possible color values, thus you lost you lost accuracy.  The same goes for any conversion between sRGB, gamma and linear - never do this unnecessarily between 8-bit per channel textures.

WestLangley:
Thank you everybody.

@elalish Please correct if it is still not what you support.

Revised specification:

1. Support the following image formats as input:
   (a) Radiance HDR
   (b) OpenEXR
   (c) JPG/PNG

2. Internal render targets are `UnsignedByteType `.

3. Internal processing is in linear color space for HDR/EXR, and in sRGB color space for JPG/PNG.

4. PMREM output is:
   (a) for HDR/EXR: `UnsignedByteType`, `RGBEEncoding`, `NearestFilter`
   (b) for JPG/PNG: `UnsignedByteType`, `sRGBEncoding`, `NearestFilter`

elalish:
@WestLangley Yes, agreed.

WestLangley:
@mrdoob We can merge this if @elalish approves.

elalish (approved):
Thanks for bearing with me!

WestLangley:
Thank you @bhouston and @elalish for your helpful feedback!

sciecode:
Thank you guys ❤️ 
It's better to have long discussions and reach the correct solution, than having to patch previous PRs over and over.

WestLangley:
@mrdoob We have consensus! :-)

bhouston:
And we will be not support the RGBM-encoded PNG workflows going forward right?   Like this old example supports: https://threejs.org/examples/#webgl_materials_envmaps_hdr

I am okay with this personally.

sciecode:
Well, depends what you mean by support, PMREMGenerator can correctly read it and output a readable render target, the only difference is that the output will be in RGBE instead of RGBM.

bhouston:
> Well, depends what you mean by support, PMREMGenerator can correctly read RGBM and output a readable render target, the only difference is that the output will be in RGBE instead of RGBM.

Good enough for me.  I support this consensus.  This is a greatly simplification and imposition of smart best practice.

WestLangley:
@mrdoob /bump

sciecode:
Anything in specific holding this PR back, @mrdoob?

I know there's been a lot discussion, but I can try to summarize what's been discussed in order to make it easier to review. 

Current `PMREMGenerator`, on dev, tries to match the input texture settings to the internal and output render targets, this works fine for devices that are able to render to floating point framebuffers, however it fails on devices that aren't if the incoming texture is in floating point. ( AKA mobiles ).

In order to allow broader use of input textures on all devices, we now enforce `UnsignedByteType` for the internal/output render targets. We use `RGBEEncoding` for HDR input textures and we keep the same encoding for the case of LDR input textures. This greatly improves the overall support and allows `PMREMGenerator` to work properly in most workflows.

Hopefully this helps.

mrdoob:
Thanks everyone! 🙏

PR #19092, opened 2020-04-09
Bugfix at matrix calculation at FBX import

crest01:
During FBX import the transformation matrix calculation uses the Matrix4.multiply() - function wrong, which causes objects to have a wrong local rotation.


Mugen87:
Can you please modify `examples/js/loaders/FBXLoader.js` instead? You can then generate the `jsm` file via `node utils/modularize.js`.

crest01:
seems like i accidentally created this pull request from the wrong branch. Gonna close this one and create a new one from the dev branch

Issue #15097, opened 2018-10-21
FBXLoader rotated Objects


p1337:
##### Description of the problem

We export FBX Files from Revit (CAD Building Software). These files show rooms, that we want to load in three.js + ar.js .

after using the latest FBXLoader, some of the elements are rotated and some come through in the right orientation.

Images

- [Revit image](http://p1337.synology.me/vrview/images/errors/Revit.png)
- [FBXLoader image](http://p1337.synology.me/vrview/images/errors/FBXLoader.png)
- [Autodesk FBX Viewer image](http://p1337.synology.me/vrview/images/errors/adsk_FBX.png)

Testing this FBX File with other online Viewers - show that the generated FBX File ist quite ok.

FBX File:
- [FBX](http://p1337.synology.me/vrview/images/errors/3D-EXPORT.fbx)

##### Three.js version

- [ ] Dev
- [x] r97
- [ ] ...

##### Browser

- [x] All of them
- [ ] Chrome
- [ ] Firefox
- [ ] Internet Explorer

##### OS

- [ ] All of them
- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] Android
- [ ] iOS

p1337:
Maybe solved:

replaced in FBXLoader.js  ~ row 3976:

`var array = transformData.rotation.map(THREE.Math.degToRad);`

with 

`    if (transformData.rotation.map(THREE.Math.degToRad)[2] < 0) {

                var bereinigt = transformData.rotation.map(THREE.Math.degToRad)[2] * -1.5;
                var array = new Array;
                array.push(transformData.rotation.map(THREE.Math.degToRad)[0]);
                array.push(transformData.rotation.map(THREE.Math.degToRad)[1]);
                array.push(bereinigt);
            }
            else if (transformData.rotation.map(THREE.Math.degToRad)[2] > 0) {

                var bereinigt = transformData.rotation.map(THREE.Math.degToRad)[2] * -3.5;
                var array = new Array;
                array.push(transformData.rotation.map(THREE.Math.degToRad)[0]);
                array.push(transformData.rotation.map(THREE.Math.degToRad)[1]);
                array.push(bereinigt);

            } else if (transformData.rotation.map(THREE.Math.degToRad)[2] == 0 ){

                var bereinigt = transformData.rotation.map(THREE.Math.degToRad)[2] * 1.5;
                var array = new Array;
                array.push(transformData.rotation.map(THREE.Math.degToRad)[0]);
[...]

PR #19093, opened 2020-04-09
Bugfix at matrix calculation at FBX import

crest01:
During FBX import the transformation matrix calculation uses the Matrix4.multiply() - function wrong, which causes objects to have a wrong local rotation.

New Pull request, since #19092 was created from the wrong branch.

### DIFF
--- a/examples/js/loaders/FBXLoader.js
+++ b/examples/js/loaders/FBXLoader.js
@@ -3986,43 +3986,49 @@ THREE.FBXLoader = ( function () {
 
 		// Global Shear*Scaling
 		var lParentTM = new THREE.Matrix4();
-		var lLSM;
-		var lParentGSM;
-		var lParentGRSM;
+		var lLSM = new Matrix4();
+		var lParentGSM = new Matrix4();
+		var lParentGRSM = new Matrix4();
+		
+		var lParentTM_inv = new Matrix4().getInverse( lParentTM );
 
 		lParentTM.copyPosition( lParentGX );
-		lParentGRSM = lParentTM.getInverse( lParentTM ).multiply( lParentGX );
-		lParentGSM = lParentGRM.getInverse( lParentGRM ).multiply( lParentGRSM );
+		lParentGRSM.multiply( lParentTM_inv ).multiply( lParentGX );
+		
+		var lParentGRM_inv = new Matrix4().getInverse( lParentGRM );
+		lParentGSM.multiply( lParentGRM_inv ).multiply( lParentGRSM );
 		lLSM = lScalingM;
 
-		var lGlobalRS;
+		var lGlobalRS = new Matrix4();
 		if ( inheritType === 0 ) {
 
-			lGlobalRS = lParentGRM.multiply( lLRM ).multiply( lParentGSM ).multiply( lLSM );
+			lGlobalRS.multiply( lParentGRM ).multiply( lLRM ).multiply( lParentGSM ).multiply( lLSM );
 
 		} else if ( inheritType === 1 ) {
 
-			lGlobalRS = lParentGRM.multiply( lParentGSM ).multiply( lLRM ).multiply( lLSM );
+			lGlobalRS.multiply( lParentGRM ).multiply( lParentGSM ).multiply( lLRM ).multiply( lLSM );
 
 		} else {
 
 			var lParentLSM = new THREE.Matrix4().copy( lScalingM );
 
-			var lParentGSM_noLocal = lParentGSM.multiply( lParentLSM.getInverse( lParentLSM ) );
+			var lParentLSM_inv = new Matrix4().getInverse( lParentLSM );
+			var lParentGSM_noLocal = new Matrix4().multiply( lParentGSM ).multiply( lParentLSM_inv );
 
-			lGlobalRS = lParentGRM.multiply( lLRM ).multiply( lParentGSM_noLocal ).multiply( lLSM );
+			lGlobalRS.multiply( lParentGRM ).multiply( lLRM ).multiply( lParentGSM_noLocal ).multiply( lLSM );
 
 		}
 
 		// Calculate the local transform matrix
-		var lTransform = lTranslationM.multiply( lRotationOffsetM ).multiply( lRotationPivotM ).multiply( lPreRotationM ).multiply( lRotationM ).multiply( lPostRotationM ).multiply( lRotationPivotM.getInverse( lRotationPivotM ) ).multiply( lScalingOffsetM ).multiply( lScalingPivotM ).multiply( lScalingM ).multiply( lScalingPivotM.getInverse( lScalingPivotM ) );
+		var lTransform = new Matrix4();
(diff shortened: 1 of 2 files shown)


Mugen87:
Is it somehow possible to demonstrate the issue with an asset? I mean is it actually visible when loading a FBX file or is more an internal thing?

mrdoob:
@looeee any chance you could test this?

looeee:
> @looeee any chance you could test this?

Sure, but I need more info. 

@crest01 can you explain in detail why this fix is needed? And can you share any FBX files you are using to test it?

crest01:
matrix1.multiply( matrix2 ) modifies matrix1 in place, which i guess works as intended. In the FBXLoader this is now used as matrix3 = matrix1.multiply( matrix2 ), which becomes a problem, since matrix1 is used in a later calculation during the quite complicated fbx pose calculation process.

However, this only becomes apparent if the objects in a fbx have a local rotation.

Issues #15745 and #15097 might be related.

I have a simple test file of two goats staring at a tree:
[goats_test.zip](https://github.com/mrdoob/three.js/files/4462591/goats_test.zip)

Blender 
image

Three.JS viewer
image

crest01:
Also i only dabble as a js developer, and my solution feels somewhat inelegant. I wouldn't mind some hints on how to refactor that...

WestLangley:
>matrix3 = matrix1.multiply( matrix2 )

Yikes. I would avoid using that pattern. It is very confusing.

One alternative is:

```js
matrix3.copy( matrix1 ).multiply( matrix2 );
```

The matrix methods return `this` to support chaining.

examples/js/loaders/FBXLoader.js
```diff
@@ -3986,43 +3986,51 @@ THREE.FBXLoader = ( function () {
...
 		// Global Shear*Scaling
 		var lParentTM = new THREE.Matrix4();
-		var lLSM;
-		var lParentGSM;
-		var lParentGRSM;
+		var lLSM = new Matrix4();
+		var lParentGSM = new Matrix4();
+		var lParentGRSM = new Matrix4();
+		
+		var lParentTM_inv = new Matrix4().getInverse( lParentTM );
```
looeee, line 3993: Just `new Matrix4()` is probably OK here since `getInverse` is called on the identity matrix.

examples/js/loaders/FBXLoader.js
```diff
@@ -3986,43 +3986,51 @@ THREE.FBXLoader = ( function () {
...
+		var lScalingPivotM_inv = new Matrix4().getInverse( lScalingPivotM );
 		// Calculate the local transform matrix
-		var lTransform = lTranslationM.multiply( lRotationOffsetM ).multiply( lRotationPivotM ).multiply( lPreRotationM ).multiply( lRotationM ).multiply( lPostRotationM ).multiply( lRotationPivotM.getInverse( lRotationPivotM ) ).multiply( lScalingOffsetM ).multiply( lScalingPivotM ).multiply( lScalingM ).multiply( lScalingPivotM.getInverse( lScalingPivotM ) );
+		var lTransform = new Matrix4();
+		lTransform.multiply( lTranslationM ).multiply( lRotationOffsetM ).multiply( lRotationPivotM ).multiply( lPreRotationM ).multiply( lRotationM ).multiply( lPostRotationM ).multiply( lRotationPivotM_inv ).multiply( lScalingOffsetM ).multiply( lScalingPivotM ).multiply( lScalingM ).multiply( lScalingPivotM_inv );
 
 		var lLocalTWithAllPivotAndOffsetInfo = new THREE.Matrix4().copyPosition( lTransform );
 
-		var lGlobalTranslation = lParentGX.multiply( lLocalTWithAllPivotAndOffsetInfo );
+		var lGlobalTranslation = new Matrix4().multiply(lParentGX).multiply( lLocalTWithAllPivotAndOffsetInfo );
```
looeee, line 4030: Formatting: needs spaces in parens `multiply( lParentGX )`

examples/js/loaders/FBXLoader.js
```diff
@@ -3986,43 +3986,51 @@ THREE.FBXLoader = ( function () {
...
 		lGlobalT.copyPosition( lGlobalTranslation );
 
-		lTransform = lGlobalT.multiply( lGlobalRS );
+		lTransform = new Matrix4().multiply(lGlobalT).multiply( lGlobalRS );
```
looeee, line 4033: Spaces in parens.

looeee:
Works great, good job!

> Issues #15745 and #15097 might be related.

This does fix #15745. Unfortunately, files on #15097 were shared on some site that doesn't work now, but I think we can assume this fixes that too. 

> matrix3 = matrix1.multiply( matrix2 )

> Yikes. I would avoid using that pattern. It is very confusing.

Yeah, I think I did it that way to avoid an extra copy. That was definitely premature optimization though. I've done some timing and this code is only adding a couple of ms to the total load time.

mrdoob:
I'll clean up the formatting post merge.

examples/js/loaders/FBXLoader.js
mrdoob: Oh, I've just noticed this comment. Well, we can do further clean up PRs 🙂

WestLangley:
@looeee I prefer what I suggested, but I think you should do whatever you prefer. :-)

mrdoob:
@WestLangley I think the code you're talking about is the previous code and not the new code.

looeee:
OK, then I think this PR is good to go. We can look at improving the readability of this function another time.

examples/js/loaders/FBXLoader.js
```diff
@@ -3986,43 +3986,51 @@ THREE.FBXLoader = ( function () {
...
+		var lParentGRM_inv = new Matrix4().getInverse( lParentGRM );
+		lParentGSM.multiply( lParentGRM_inv ).multiply( lParentGRSM );
 		lLSM = lScalingM;
 
-		var lGlobalRS;
+		var lGlobalRS = new Matrix4();
 		if ( inheritType === 0 ) {
 
-			lGlobalRS = lParentGRM.multiply( lLRM ).multiply( lParentGSM ).multiply( lLSM );
+			lGlobalRS.multiply( lParentGRM ).multiply( lLRM ).multiply( lParentGSM ).multiply( lLSM );
```
WestLangley, line 4005:
```js
lGlobalRS.copy( lParentGRM ).multiply( lLRM ).multiply( lParentGSM ).multiply( lLSM );
```
multiple places...

looeee: > lGlobalRS.copy( lParentGRM )

> lGlobalRS.multiply( lParentGRM )

These two statements are equivalent for `lGlobalRS = I`, so what would be the benefit of using `.copy` over `.multiply`?

WestLangley: For the same reason we would use `.copy()` instead of
```js
matrix1.identity().multiply( matrix2 );
```
It is clearer.

mrdoob: https://github.com/mrdoob/three.js/commit/809fa4f7f9925f4421acb4eb200250f781c3b0dd

mrdoob:
@crest01 do you have any ideas about https://github.com/mrdoob/three.js/issues/15287?

pikilipita:
Just wanted to say, this bugfix works great: no more issues with fbx generated from Cinema 4D, and no regression on other models.
That was a long standing issue, thanks a lot!

PR #19099, opened 2020-04-10
WebXRManager: Introduce WebXRController.

Mugen87:
This PR tries to refactor `WebXRManager` by introducing a new class called `WebXRController` which encapsulated the controller/input source related logic. This refactoring performs a similar modularization that was applied to `WebGLRenderer` in the past. 

It's a pure internal change and does not affect the user API. I've verified this change with Oculus Quest and the examples.


mrdoob:
Very nice!

mrdoob:
Thanks!

